PR #270, opened 2014-11-15
AY: added IPv6 option (bnc#633310)

mchf:
fate#316938


mvidner:
I have tested that it works (but only after a reboot, since is is implemented by blacklisting the module; that should be IMHO mentioned in the release notes).
LGTM.

PR #278, opened 2015-01-08
ipoib mode selection

mchf:
backported implementation of fate#315501 to fulfill fate#315162 requirements.

Tested on top of SLE-11-SP3.


src/modules/LanItems.ycp
```diff
@@ -152,6 +152,12 @@ global boolean tunnel_set_persistent=true;
 global string tunnel_set_owner="";
 global string tunnel_set_group="";
 
+// infiniband options
+global string ipoib_mode = "";
+global map<string, string> ipoib_modes = $[
+    "connected": _("connected"),
```
mvidner: Translator context missing: "připojený počítač" or "spojovaný režim"? 
> Translators: connected mode of IPoIB operation

src/modules/LanItems.ycp
```diff
@@ -152,6 +152,12 @@ global boolean tunnel_set_persistent=true;
...
+    "datagram": _("datagram")
```
mvidner: > Translators: datagram mode of IPoIB operation

src/routines/widgets.ycp
```diff
@@ -186,6 +187,59 @@ _("<p><b><big>Device Activation</big></b></p>
...
+        "items" : maplist(string k, string v, LanItems::ipoib_modes, { return [k, v]; }),
+        "label" : _("IPoIB device mode"),
+        "opt"   : [`hstretch],
+        "init"  : init_ipoib_mode_widget,
+        "store" : store_ipoib_mode_widget
+    ];
+}
+
+const list<list> COMMON_MTU_ITEMS = [
+    ["1500", "1500 (Ethernet, DSL broadband)"],
```
mvidner: Translation markers missing.

mvidner:
It turns out this PR are two backports mixed into one. The first one is related to the device type refactoring in yast/yast-yast2#315. It should probably get a separate PR, and definitely a separate entry in the changelog, because:

- It adapts to an incompatible API change. (I thought that it could still be avoided by adding a compatibility alias in yast2, but )
- It adapts the tests to the changed implementation in yast2

Otherwise it will be much easier to install, or release as a MU, only one of the matching RPMs and mess up.

mvidner:
- And it must have the same target branch: SLE11-SP3

mchf:
ok, I'll split  f69a315 & co into PR for SLE-11-SP3

mchf:
So, device renaming related code was moved into https://github.com/yast/yast-network/pull/279

mchf:
Rebased, updated and re-tested.

src/routines/widgets.ycp
```diff
@@ -186,6 +187,61 @@ _("<p><b><big>Device Activation</big></b></p>
...
+// device type
+const list<list> COMMON_MTU_ITEMS = [
+    ["1500", "1500 (Ethernet, DSL broadband)"],
+    ["1492", "1492 (PPPoE broadband)"],
+    ["576", "576 (dial-up)"]
+];
+
+const list<list> IPOIB_MTU_ITEMS = [
+    ["65520", "65520 (IPoIB in connected mode)"],
+    ["2044", "2044 (IPoIB in datagram mode)"]
```
mvidner: Translation markers missing (5x).

package/yast2-network.changes
```diff
@@ -6,6 +6,13 @@ Wed Jan 14 11:26:43 UTC 2015 - mfilka@suse.com
 - fixed problem with not matching udev rule (bnc#551310)
 - bnc#695262
     - fixed write sequence to enable device renaming.
+- 2.17.202
+
+-------------------------------------------------------------------
+Thu Jan  8 21:07:28 UTC 2015 - mfilka@suse.com
+
+- fate#315162
+  - IPoIB mode configuration in YaST
```
mvidner: It is better to add changelog entries at the top, isn't it?

mchf: hugh. I was happy that I've made the merge ... obviously not well enough ;-)

src/routines/widgets.ycp
mchf: the generic one at line 221 is not enough? Do you prefer one-per-line comments?

jreidinger: @mchf - `_()` missing, not comment for translators.

mvidner: Um,
> \# This is a translator COMMENT
_("This is a translation MARKER")

mchf: ups.

package/yast2-network.changes
```diff
@@ -1,6 +1,13 @@
 -------------------------------------------------------------------
 Wed Jan 14 11:26:43 UTC 2015 - mfilka@suse.com
 
+- fate#315162
+  - IPoIB mode configuration in YaST
+- 2.17.202
+
+-------------------------------------------------------------------
+Thu Jan  8 21:07:28 UTC 2015 - mfilka@suse.com
```
mvidner: Changing the timestamp of an existing entry would result in a failed check "changelog is not monotonic", but here we are lucky because the previous change is not accepted yet :)

mvidner:
LGTM, thanks!

package/yast2-network.changes
mchf: yep, I took the opportunity ;-)

mchf:
thanks for reviews.

PR #281, opened 2015-01-19
Configure vlan at install time

mchf:
fate#316140

It builds. Currently testing if it works.


mvidner:
LGTM.

I was also wondering how this would be ported to SLE12/master but then realized that the situation is completely different there, with ifcfg files being used much earlier.

mchf:
tested using build635

PR #284, opened 2015-01-29
Fixed changing configuration name when adding new device

mchf:
bnc#912904

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -1885,24 +1885,24 @@ module Yast
     # @return true if success
     def Commit
       if @operation != :add && @operation != :edit
-        Builtins.y2error("Unknown operation: %1", @operation)
+        log.error("Unknown operation: %1", @operation)
         raise ArgumentError, "Unknown operation: #{@operation}"
       end
 
       newdev = {}
 
       # #104494 - always write IPADDR+NETMASK, even empty
-      Ops.set(newdev, "IPADDR", @ipaddr)
-      if Ops.greater_than(Builtins.size(@prefix), 0)
-        Ops.set(newdev, "PREFIXLEN", @prefix)
+      newdev["IPADDR"] = @ipaddr
+      if !@prefix.empty?
+        newdev["PREFIXLEN"] = @prefix
       else
-        Ops.set(newdev, "NETMASK", @netmask)
+        newdev["NETMASK"] = @netmask
       end
       # #50955 omit computable fields
-      Ops.set(newdev, "BROADCAST", "")
-      Ops.set(newdev, "NETWORK", "")
+      newdev["BROADCAST"] = ""
+      newdev["NETWORK"] = ""
 
-      Ops.set(newdev, "REMOTE_IPADDR", @remoteip)
+      newdev["REMOTE_IPADDR"] = @remoteip
 
       # set LLADDR to sysconfig only for device on layer2 and only these which needs it
       # do not write incorrect LLADDR.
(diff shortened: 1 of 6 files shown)


src/modules/LanItems.rb
```diff
@@ -1932,45 +1934,43 @@ def Commit
...
+          newdev["BONDING_SLAVE#{i}"] = slave
+          i += 1
         end
 
         #assign nil to rest BONDING_SLAVEn to remove them
-        while Ops.less_than(i, @MAX_BOND_SLAVE)
-          Ops.set(newdev, Builtins.sformat("BONDING_SLAVE%1", i), nil)
-          i = Ops.add(i, 1)
+        while i < @MAX_BOND_SLAVE
+          newdev["BONDING_SLAVE#{i}"]
```
jreidinger: ugh, how this should work?
```
irb(main):001:0> h = {}
=> {}
irb(main):002:0> h["binding0"] = true
=> true
irb(main):003:0> h
=> {"binding0"=>true}
irb(main):004:0> h["binding1"]
=> nil
irb(main):005:0> h
=> {"binding0"=>true}
```

src/modules/LanItems.rb
```diff
@@ -1932,45 +1934,43 @@ def Commit
...
-        Ops.set(newdev, "WIRELESS_KEY", "") # TODO: delete the varlable
-        Ops.set(newdev, "WIRELESS_KEY_0", Ops.get(@wl_key, 0, ""))
-        Ops.set(newdev, "WIRELESS_KEY_1", Ops.get(@wl_key, 1, ""))
-        Ops.set(newdev, "WIRELESS_KEY_2", Ops.get(@wl_key, 2, ""))
-        Ops.set(newdev, "WIRELESS_KEY_3", Ops.get(@wl_key, 3, ""))
+        newdev["WIRELESS_KEY"] = "" # TODO: delete the varlable
+        newdev["WIRELESS_KEY_0"] = Ops.get(@wl_key, 0, "")
+        newdev["WIRELESS_KEY_1"] = Ops.get(@wl_key, 1, "")
+        newdev["WIRELESS_KEY_2"] = Ops.get(@wl_key, 2, "")
+        newdev["WIRELESS_KEY_3"] = Ops.get(@wl_key, 3, "")
```
jreidinger: NP: I think here is better to use merge approach:
```
newdev.merge! ({
  WIRELESS_MODE" => @wl_mode,
  ....
)}
```

jreidinger:
one new bug, one nitpick and no tests. Please fix bug and also add some tests that proof that you do not break it ;)

mchf:
there are tests already which covers bugfixed code.

jreidinger:
LGTM

src/modules/LanItems.rb
```diff
@@ -540,8 +540,10 @@ def ReplaceItemUdev(replace_key, new_key, new_val)
     # Returns new name
     def SetItemName( itemId, name)
       if name && !name.empty?
-        updated_rule = update_udev_rule_key(GetItemUdevRule(itemId), "NAME", name)
-        @Items[itemId]["udev"]["net"] = updated_rule
+        if @Items[itemId]["udev"]
```
mvidner: Shouldn't we instead ensure that `@Items[itemId]` has a `udev` key whose value is a hash?

src/modules/LanItems.rb
```diff
@@ -1932,45 +1934,43 @@ def Commit
       case @type
       when "bond"
         i = 0
-        Builtins.foreach(@bond_slaves) do |slave|
-          Ops.set(newdev, Builtins.sformat("BONDING_SLAVE%1", i), slave)
-          i = Ops.add(i, 1)
+        @bond_slaves.each do |slave|
```
mvidner: Use `Enumerable#each_with_index`

src/modules/LanItems.rb
```diff
@@ -1932,45 +1934,43 @@ def Commit
...
+          newdev["BONDING_SLAVE#{i}"] = slave
+          i += 1
         end
 
         #assign nil to rest BONDING_SLAVEn to remove them
-        while Ops.less_than(i, @MAX_BOND_SLAVE)
-          Ops.set(newdev, Builtins.sformat("BONDING_SLAVE%1", i), nil)
-          i = Ops.add(i, 1)
+        while i < @MAX_BOND_SLAVE
```
mvidner: Ah, never mind, `i` is still needed here.

jreidinger:
OK, LGTM, I think this part of code deserve bigger refactoring, it is quite messy

mchf:
@jreidinger: 
Yes, this part of yast-network wasn't touched that much ... i know that this is small step for you ... but huge step for the code (hopefully) ;-)

Thanks for reviews

PR #286, opened 2015-02-19
Apply udev rules provided by AY profile

mchf:
bnc#905738
L3. Working path confirmed by both customers.


kobliha:
@mchf Travis build has failed, any idea why?

kobliha:
Finished in 0.00352 seconds
3 examples, 0 failures
PASS: udev_test.rb
/bin/bash: line 9: ./udev_auto_test.rb: Permission denied
FAIL: udev_auto_test.rb
......

mchf:
execution bit not set :-/

mchf:
locally it works ... so travis runs tests differently?

jreidinger:
@mchf travis use make check which use Makefile.am and it need executable bit

src/clients/lan_auto.rb
```diff
@@ -159,7 +160,13 @@ def main
         @ret = deep_copy(@autoyast)
       elsif @func == "Write"
         @progress_orig = Progress.set(false)
-        @ret = Lan.WriteOnly
+
+        @ret = LanUdevAuto.Write
+        Builtins.y2error("Writing udev rules failed") if !@ret
+
+        @ret = Lan.WriteOnly && @ret
```
jreidinger: so in this usage it means try write Lan even if LanUdevAuto failed? does it make sense? I think at least comment why it is possible should be here.

src/include/network/routines.rb
```diff
@@ -947,6 +947,18 @@ def SetAllLinksUp
       ret
     end
 
+    def SetAllLinksDown
+      interfaces = GetAllInterfaces()
+      ret = !interfaces.empty?
```
jreidinger: maybe better and readable to have `return false if interfaces.empty?`

src/include/network/routines.rb
```diff
@@ -947,6 +947,18 @@ def SetAllLinksUp
...
+
+      interfaces.each do |ifc|
+        Builtins.y2milestone("Setting link down for interface %1", ifc)
+        ret = SetLinkDown(ifc) && ret
```
jreidinger: so again logic says, that you try to down it all and if any failed, then continue with others.

mchf: yes. I've chosen best effort approach instead of e.g. trying to rollback

src/clients/lan_auto.rb
mchf: yes. best effort approach. LanUdevAuto only writes udev rules according AY profile. Lan.Write* writes configuration for devices, dns, routing, ...

src/include/network/routines.rb
mchf: copy-paste-edit of similar function for ycp times. I'll give it some touches ...

src/modules/LanUdevAuto.rb
```diff
@@ -32,11 +32,15 @@
 
 module Yast
   class LanUdevAutoClass < Module
+
+    include Logger
```
jreidinger: I think it is better to be more explicit and include Yast::Logger as Logger is well known ruby class so to avoid confusion

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
     end
 
     def Write
-      comment = [
-        "# Generated by autoyast",
-        "# program run by the persistent-net-generator.rules rules file.",
-        "#",
-        "# You can modify it, as long as you keep each rule on a single line."
-      ]
+      template = "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", %s==\"%s\", NAME=\"%s\""
```
jreidinger: this looks like constant. And I also think that with so many replacement it is better to use `%{name}` syntax, so it will be easier to read it

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
...
-          rules,
-          Builtins.sformat(
-            "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", %1==\"%2\", NAME=\"%3\"",
-            Ops.get_string(rule, "rule", ""),
-            Ops.get_string(rule, "value", ""),
-            Ops.get_string(rule, "name", "")
-          )
-        )
+
+      @udev_rules.each do |rule|
```
jreidinger: it should be `rules = @udev_rules.map do |rule|`

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
...
-      if Ops.greater_than(Builtins.size(rules), 0)
-        Builtins.y2milestone(
-          "Writing AY udev rules for network (will replace original rules from 1st stage)"
-        )
-        if AllowUdevModify()
-          SCR.Write(path(".udev_persistent.rules_comment"), comment)
-          SCR.Write(path(".udev_persistent.rules"), rules)
-        end
+
+      if (rules.size > 0) && AllowUdevModify()
```
jreidinger: `if !rules.empty? && AllowUdevModify()`

test/udev_auto_test.rb
```diff
@@ -0,0 +1,44 @@
...
+    allow(Yast::LanUdevAuto)
+      .to receive(:AllowUdevModify)
+      .and_return true
+
+    expect(Yast::SCR)
+      .to receive(:Write)
+      .with(path(".udev_persistent.rules"), ay_rules)
+    allow(Yast::SCR)
+      .to receive(:Write)
+      .and_return 0
```
jreidinger: this is not nice indentation as it looks like you call `.and_return` on result of `to`, but it is not true

src/clients/lan_auto.rb
```diff
@@ -141,6 +141,7 @@ def main
         end
         @new = FromAY(@param)
         Lan.Import(@new)
+        LanUdevAuto.Import(@new)
```
mvidner: Interesting. I wonder how this was supposed to work before:
 https://github.com/yast/yast-autoinstallation/blob/3e3cc69b8dd7129f2efa7d31f805ce7980120818/src/clients/inst_autosetup.rb#L354
 see also https://bugzilla.novell.com/show_bug.cgi?id=303916

mchf: well, who can say if it ever worked. Device renaming was broken in sle-11-sp3 for a very long time. In fact it was fixed before one or two months.

ancorgs:
If possible would like to see a test ensuring that all the rules (specially the last one in a given file) ends with the newline character. I'm not sure if that requirement applies to all versions of udev, but at least with openSUSE 13.2 we had a bug report and, since it's impossible for a human eye to distinguish a file with/without a newline character at the end, we got some fun.

test/udev_auto_test.rb
mchf: well. matter of taste ;-) I assume this one is NP.

test/udev_auto_test.rb
```diff
@@ -0,0 +1,44 @@
...
+
+require_relative "test_helper"
+
+require "yast"
+
+include Yast
+
+Yast.import "LanUdevAuto"
+
+describe "LanUdevAuto#Write" do
```
mvidner:
```
$ ls -1 test/*udev*test.rb
test/lan_udev_auto_test.rb
test/udev_auto_test.rb
test/udev_test.rb
```

Also tests should be well organized, so I think this file should be folded into lan_udev_auto_test.rb

src/include/network/lan/udev.rb
```diff
@@ -45,6 +45,28 @@ def update_udev_rule_key(rule, key, value)
...
+      SCR.Execute(path(".target.bash"), "udevadm control --reload")
+
+      # When configuring a new s390 card, we neglect to fill
+      # its Items[i, "udev", "net"], causing jumbled names (bnc#721520)
+      # The udev trigger will make udev write the persistent names
+      # (which it already has done, but we have overwritten them now).
+      SCR.Execute(
+        path(".target.bash"),
+        "udevadm trigger --subsystem-match=net --action=add"
+      ) == 0
```
mvidner: Compared to the old code there is an added `== 0` test. But the behavior is not documented in the comment for the new(!) method. It should say `# @return void` as it is actually used in the new code or `# @return [Boolean] we succeeded in tickling the daemon or whatever` as it is implemented.

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
...
+        log.info("Writing AY udev rules for network")
+
+        comment = [
+          "# Generated by autoyast",
+          "# program run by the persistent-net-generator.rules rules file.",
+          "#",
+          "# You can modify it, as long as you keep each rule on a single line."
+        ]
+
+        SCR.Write(path(".udev_persistent.rules_comment"), comment)
```
mvidner: This line IMHO belongs to `write_update_udevd(rules, comment = "")`

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
...
-        Builtins.y2milestone(
-          "Writing AY udev rules for network (will replace original rules from 1st stage)"
-        )
-        if AllowUdevModify()
-          SCR.Write(path(".udev_persistent.rules_comment"), comment)
-          SCR.Write(path(".udev_persistent.rules"), rules)
-        end
+
+      if !rules.empty? && AllowUdevModify()
+        SetAllLinksDown()
```
mvidner: This is new, introduced in 8561a6f08971e236dbec1a5d2eaceb01486d7d68 named "Wait until udev settles down when updating udev rules."
It seems it is a great opportunity to break things. Why is it needed?

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,43 @@ def Import(settings)
...
+        SCR.Write(path(".udev_persistent.rules_comment"), comment)
+
+        write_update_udevd(rules)
+
+        SCR.Execute(path(".target.bash"), "udevadm settle")
       else
-        Builtins.y2milestone(
-          "No udev rules defined by AY, nothing to write (keep original file)"
-        )
+        log.info("No udev rules defined by AY")
```
mvidner: No, there may be rules which are blocked by `AllowUdevModify`. The log message was only truthful before the refactoring.

src/include/network/routines.rb
```diff
@@ -928,23 +928,29 @@ def GetAllInterfaces
...
-      ret
+      interfaces.all? { |i| SetLinkUp(i) }
+    end
+
+    def SetAllLinksDown
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
+
+      interfaces.all? { |i| SetLinkDown(i) }
```
mvidner: Again, the code explicitly tries to return a boolean value which is

1. not documented
2. not used by any caller

I hate this!

In this particular case, I think the return value does not even make sense.

mvidner:
A .changes entry is missing

src/modules/LanUdevAuto.rb
mchf: well, I was thinking about removing this call when the rest was moved into ```write_update_udevd```. I don't think that the comment brings an additional value to a user. Moreover it is not true - the rules are definitely not generated by ```persistent-net-generator.rules```. What is your opinion?

src/modules/LanUdevAuto.rb
mchf: Bcs udev doesn't apply new rules when device is up.

src/modules/LanUdevAuto.rb
mchf: Then it wasn't true even before. ```AllowUdevModify``` is not newly added. 

I'll rephrase the message.

test/udev_auto_test.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    expect(Yast::SCR)
+      .to receive(:Write)
+      .with(path(".udev_persistent.rules"), ay_rules)
+    allow(Yast::SCR)
+      .to receive(:Write)
+      .and_return 0
+
+    LanUdevAuto.Import({ "net-udev" => udev_rules })
+    LanUdevAuto.Write
```
mvidner: This is a functional test involving `LanUdevAuto` calls only. Do we have an integration test?

src/modules/LanUdevAuto.rb
mvidner: Ah, good point, let's remove the comment.

mchf:
Added change log and some changes according comments.

mchf:
Added comments where requested in the review.

mvidner:
LGTM, thank you.

mchf:
Thanks for all reviews.

PR #287, opened 2015-02-26
Provide proper device description in bond slave candidates list

mchf:
bnc#918356

Not to merge now. Provided for early review.

### DIFF
--- a/src/include/network/widgets.rb
+++ b/src/include/network/widgets.rb
@@ -559,47 +559,48 @@ module Yast
     # @param [Array<Fixnum>] itemIds           list of indexes into LanItems::Items
     # @param [Array<String>] enslavedIfaces    list of device names of already enslaved devices
     def CreateSlaveItems(itemIds, enslavedIfaces)
+      raise ArgumentError if itemIds.nil?
+
       items = []
 
-      Builtins.foreach(itemIds) do |itemId|
-        description = ""
+      itemIds.each do |itemId|
         dev_name = LanItems.GetDeviceName(itemId)
         ifcfg = LanItems.GetDeviceMap(itemId)
 
-        next if IsEmpty(dev_name)
+        next if dev_name.nil? || dev_name.empty?
 
         ifcfg = { "dev_name" => dev_name } if ifcfg == nil
         dev_type = LanItems.GetDeviceType(itemId)
 
-        if Builtins.contains(["tun", "tap"], dev_type)
+        if ["tun", "tap"].include? dev_type
           description = NetworkInterfaces.GetDevTypeDescription(dev_type, true)
         else
           description = BuildDescription(
             dev_type,
             "",
             ifcfg,
-            [Ops.get_map(LanItems.GetLanItem(itemId), "hwinfo", {})]
+            [LanItems.GetLanItem(itemId)["hwinfo"] || {}]
           )
 
           # this conditions origin from bridge configuration
           # if enslaving a configured device then its configuration is rewritten
           # to "0.0.0.0/32"
-          if Ops.get_string(ifcfg, "IPADDR", "") != "0.0.0.0"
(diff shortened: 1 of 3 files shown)


ancorgs:
Looks good, just 2 nitpicking comments

src/include/network/widgets.rb
```diff
@@ -559,46 +559,45 @@ def handleDevice(items, selected)
...
+        if ["tun", "tap"].include? dev_type
           description = NetworkInterfaces.GetDevTypeDescription(dev_type, true)
         else
           description = BuildDescription(
-            "",
+            dev_type,
             "",
             ifcfg,
-            [Ops.get_map(LanItems.GetLanItem(itemId), "hwinfo", {})]
+            [LanItems.GetLanItem(itemId)["hwinfo"] || {}]
```
jreidinger: API looks quite funny as you pass array with single element that is hash.

src/include/network/widgets.rb
```diff
@@ -559,46 +559,45 @@ def handleDevice(items, selected)
...
 
           # this conditions origin from bridge configuration
           # if enslaving a configured device then its configuration is rewritten
           # to "0.0.0.0/32"
-          if Ops.get_string(ifcfg, "IPADDR", "") != "0.0.0.0"
-            description = Builtins.sformat("%1 (%2)", description, "configured")
-          end
+          #
+          # translators: a note that listed device is already configured
+          description << " " << _("configured") if ifcfg["IPADDR"] != "0.0.0.0"
```
jreidinger: are you sure that description returned by by `BuildDescription` is not frozen (e.g. as result of `_` call)? Otherwise it can raise exception.

src/include/network/widgets.rb
```diff
@@ -559,46 +559,45 @@ def handleDevice(items, selected)
...
-          Item(
-            Id(dev_name),
-            Builtins.sformat("%1 - %2", dev_name, description),
-            selected
-          )
+        selected = enslavedIfaces.include?(dev_name) if enslavedIfaces
+        items << Item(
+          Id(dev_name),
+          "#{dev_name} - #{description}",
+          selected
```
jreidinger: I think it will be better to state selected as false if `enslavedIfaces` is nil, I think our goal should be reduce number of nil that we passing around. So better to have selected as true or false and do not allow `nil` there.

jreidinger:
BTW is modified code covered by tests? if not, then I think it is good opportunity to write some.

ancorgs:
I promise I added 2 comments on the code. They seem to be gone :-/

src/include/network/widgets.rb
mchf: What can I say. Yes it is. However, API is given and this patch  do not change this particular thing.

mchf:
@ancor: I've rebased due to a syntax error. Never mind, I have your comments in mail ;-)

mchf:
Added some fixes according comments. Testing the patch is WIP

mchf:
Btw building device description is already covered by tests.

src/include/network/widgets.rb
```diff
@@ -559,46 +559,48 @@ def handleDevice(items, selected)
...
-        items = Builtins.add(
-          items,
-          Item(
-            Id(dev_name),
-            Builtins.sformat("%1 - %2", dev_name, description),
-            selected
-          )
+
+        selected = false
+        selected = enslavedIfaces.include?(dev_name) if enslavedIfaces
```
jreidinger: just idea: maybe more readable to use  ternary operator: `selected = enslavedIfaces ? enslavedIfaces.include?(dev_name) : false`

jreidinger:
in general it looks good for me now

mchf:
Updated changelog. Going to close the pull request bcs of invalid merge target :-/

PR #289, opened 2015-02-28
https://github.com/yast/yast-network/pull/287 #3

mchf:
supersedes https://github.com/yast/yast-network/pull/287

This time hopefully with correct merge target finally.


mchf:
The patch was LGTMed in https://github.com/yast/yast-network/pull/287 so merging. 

Thanks all for reviews.

PR #293, opened 2015-04-21
Write --noproxy to .curlrc

mvidner:
After a PTF has been provided, this is for SLE11 SP4
https://bugzilla.suse.com/show_bug.cgi?id=923788
https://trello.com/c/dUtdKuNg/86-3-11-sp3-l3-923788-curlrc-is-missing-noproxy-setting


src/modules/Proxy.ycp
```diff
@@ -99,6 +99,68 @@ global define boolean Read() {
     return true;
 }
 
+void WriteSysconfig() {
+    SCR::Write(.sysconfig.proxy.PROXY_ENABLED, enabled?"yes":"no");
```
jreidinger: please spaces arond `?:`

src/modules/Proxy.ycp
```diff
@@ -99,6 +99,68 @@ global define boolean Read() {
...
+        proxyuser = user;
+        if (pass != "") proxyuser = user + ":" + pass;
+    }
+
+    map<string, string> options = $[
+        "--proxy-user": proxyuser,
+        // bnc#305163
+        "--proxy":      http,
+        // bsc#923788
+        "--noproxy":    no
```
jreidinger: not done in this pr, but variable with name `no` with file scope is quite horrible

jreidinger:
few improvements but in general LGTM

mchf:
Thanks for spliting Write* methods

src/modules/Proxy.ycp
mvidner: Yes, I will rename it to `the`. Just kidding!

PR #294, opened 2015-04-27
Several patches from SLE-12 branch

mchf:
- bnc#912904
  - do not crash when adding new device and proposed configuration
    name is changed to new one
- bnc#914833
  - do not crash when renaming virtual device via hardware tab
- bnc#910337
  - initialize configuration name in Hardware tab correctly. It
    means according device type. Also uses custom name if user
    used nonstandard one.
- bnc#905738
  - udev rules provided by AutoYaST profile gets applied
- bnc918356
  - populate bond slave candidates list with proper device names


package/yast2-network.changes
```diff
@@ -1,4 +1,22 @@
...
+    name is changed to new one
+- bnc#914833
+  - do not crash when renaming virtual device via hardware tab
+- bnc#910337
+  - initialize configuration name in Hardware tab correctly. It
+    means according device type. Also uses custom name if user
+    used nonstandard one.
+- bnc#905738
+  - udev rules provided by AutoYaST profile gets applied
+- bnc918356
```
mvidner: hash sign missing

package/yast2-network.changes
```diff
@@ -1,4 +1,22 @@
...
+  - populate bond slave candidates list with proper device names
+- 3.1.114
```
mvidner: Note that 3.1.110 - 3.1.112 exist both in master and SLE-12-GA with different contents :( Please remember to use 3.1.112.1 for the next SLE-12-GA submission.

mvidner:
I haven't reviewed the contents, just checked that it is a correct merge of the branch. It is.
Only add the missing '#', then it LGTM.

src/include/network/lan/udev.rb
```diff
@@ -45,6 +45,30 @@ def update_udev_rule_key(rule, key, value)
...
+      SCR.Execute(path(".target.bash"), "udevadm control --reload")
+
+      # When configuring a new s390 card, we neglect to fill
+      # its Items[i, "udev", "net"], causing jumbled names (bnc#721520)
+      # The udev trigger will make udev write the persistent names
+      # (which it already has done, but we have overwritten them now).
+      SCR.Execute(
+        path(".target.bash"),
+        "udevadm trigger --subsystem-match=net --action=add"
+      ) == 0
```
jreidinger: this is not much obvious, can you please use `retcode = SCR.Execute...` and then `retcode == 0` this way it can be easy to overlook

src/include/network/routines.rb
```diff
@@ -928,23 +928,35 @@ def GetAllInterfaces
...
+    # Tries to set all available interfaces down
+    #
+    # @return [boolean] false if some of interfaces cannot be set down
+    def SetAllLinksDown
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
+
+      interfaces.all? { |i| SetLinkDown(i) }
     end
```
jreidinger: this looks duplicate. Can you make new private method to set all links like:
```
def SetAllLinksUp
  SetAllLinks(:up)
end

def SetAllLinksUp
  SetAllLinks(:down)
end

def SetAllLinks(action)

  interfaces.all? do |i|
     case action
     when :up      then SetLinkUp(i)
     when :down then SetLinkDown(i)
     else raise "Unknown action '#{action}'"
     end
  end
end
```

src/include/network/widgets.rb
```diff
@@ -559,46 +559,48 @@ def handleDevice(items, selected)
     # @param [Array<Fixnum>] itemIds           list of indexes into LanItems::Items
     # @param [Array<String>] enslavedIfaces    list of device names of already enslaved devices
     def CreateSlaveItems(itemIds, enslavedIfaces)
-      itemIds = deep_copy(itemIds)
-      enslavedIfaces = deep_copy(enslavedIfaces)
+      raise ArgumentError if itemIds.nil?
```
jreidinger: please specify message for argument error, this way it is quite useless, especially if you use more arguments constrains.

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,34 @@ def Import(settings)
     end
 
     def Write
-      comment = [
-        "# Generated by autoyast",
-        "# program run by the persistent-net-generator.rules rules file.",
-        "#",
-        "# You can modify it, as long as you keep each rule on a single line."
-      ]
+      template = "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", %s==\"%s\", NAME=\"%s\""
```
jreidinger: for longer template I really recomment to use named replacement points. It really helps with applying and also with reading code like:
```
template = "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", %{rule}==\"%{rule_value}\", NAME=\"%{device_name}\""

# and then usage
rules << template % [ rule: opt, rule_value: value.downcase, device_name: devname]
```

test/lan_udev_auto_test.rb
```diff
@@ -4,9 +4,12 @@
 
 require "yast"
 
-module Yast
+include Yast
```
jreidinger: ah, please do not do this. It populates whole global namespace and can cause troubles to other tests.

test/lan_udev_auto_test.rb
```diff
@@ -4,9 +4,12 @@
...
 
-  Yast.import "UI"
+Yast.import "UI"
+Yast.import "LanUdevAuto"
+
+describe '#getDeviceName' do
```
jreidinger: This is wrong, you should always specify whole method call like `LanUdevAuto.getDeviceName` or if it is part of include then e.g. describe "Yast::Udev#getDeviceName"

mchf:
Ok I'm going to update the patch a bit ... switching from "maintenance mode" to "new code" mode ;-)

test/lan_udev_auto_test.rb
```diff
@@ -19,61 +22,93 @@ module Yast
...
-            "mac" => "00:11:22:33:44:FF",
-            "busid" => "0000:00:19.0"
-          }
-        ]
-      }
-    end
+  # general mocking stuff is placed here
+  before( :each) do
+    # mock devices configuration
+    LanUdevAuto.stub( :ReadHardware) {
```
jreidinger: space after `(` I think we should really setup rubocop for network :)

jreidinger: and please do not use stub, it is RSpec2 which do not work in new opensuse.

jreidinger: correct way is `expect(...).to receive(...)`

mchf: regarding these two comments ... this is a pretty old testcase :-/

test/lan_udev_auto_test.rb
```diff
@@ -19,61 +22,93 @@ module Yast
...
+    udev_rules = [
+      {
+        "rule"  => ATTR,
+        "value" => VALUE,
+        "name"  => NAME
+      }
+    ]
+
+    ay_rules = [
+      "SUBSYSTEM==\"net\", ACTION==\"add\", DRIVERS==\"?*\", %s==\"%s\", NAME=\"%s\"" %
```
jreidinger: so you basically copy template from code, this is wrong from test perspective as you should check it against real expected output and do not depend on same template.

test/lan_udev_auto_test.rb
jreidinger: fine for me, it is just warning that it won't work on TW and SP1 probably.

test/lan_udev_auto_test.rb
mchf: well in this case this template overwrites everything in the target file, so it is expected output.

src/clients/lan_auto.rb
```diff
@@ -159,7 +160,13 @@ def main
         @ret = deep_copy(@autoyast)
       elsif @func == "Write"
         @progress_orig = Progress.set(false)
-        @ret = Lan.WriteOnly
+
+        @ret = LanUdevAuto.Write
+        Builtins.y2error("Writing udev rules failed") if !@ret
+
+        @ret = Lan.WriteOnly && @ret
+        Builtins.y2error("Writing lan config failed") if !@ret
```
jreidinger: so it writes `"Writing lan config failed"` even if only UdevAuto failed, is it correct?

src/modules/LanItems.rb
```diff
@@ -539,17 +539,21 @@ def ReplaceItemUdev(replace_key, new_key, new_val)
...
-        @Items[itemId].delete("udev")
+        lan_items[itemId].delete("udev")
       end
 
-      if @Items[ itemId].has_key?( "ifcfg")
-        NetworkInterfaces.Delete2( @Items[ itemId][ "ifcfg"])
-        @Items[ itemId][ "ifcfg"] = name.to_s
+      if lan_items[ itemId].has_key?( "ifcfg")
+        NetworkInterfaces.Delete2(lan_items[ itemId][ "ifcfg"])
+        lan_items[ itemId][ "ifcfg"] = name.to_s
```
jreidinger: we really need to set rubocop for network, too much spaces after opening brackets :)

src/modules/LanItems.rb
```diff
@@ -1886,24 +1878,24 @@ def setup_basic_device_options(devmap)
     # @return true if success
     def Commit
       if @operation != :add && @operation != :edit
-        Builtins.y2error("Unknown operation: %1", @operation)
+        log.error("Unknown operation: %1", @operation)
```
jreidinger: BUM! :bomb:  crash. `log.error` accept only one argument. Testsuite should catch it ;)

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,34 @@ def Import(settings)
...
-          )
-        )
+
+      @udev_rules.each do |rule|
+        opt = rule["rule"] || ""
+        value = rule["value"] || ""
+        devname = rule["name"] || ""
+
+        rules << template % [opt, value.downcase, devname]
       end
```
jreidinger: in fact instead of each you should use map ( so `roles = []` is not needed )
like:
```
     rules = @udev_rules.map do |rule|
        opt = rule["rule"] || ""
        value = rule["value"] || ""
        devname = rule["name"] || ""

        template % [opt, value.downcase, devname]
     end
```

src/modules/LanUdevAuto.rb
```diff
@@ -194,39 +198,34 @@ def Import(settings)
...
+        log.info("Writing AY udev rules for network")
+
+        write_update_udevd(rules)
+
+        SCR.Execute(path(".target.bash"), "udevadm settle")
       else
-        Builtins.y2milestone(
-          "No udev rules defined by AY, nothing to write (keep original file)"
-        )
+        log.info("No udev rules created by AY")
```
jreidinger: Nitpicking: for log we often skip parens like `log.info "No udev rules created by AY"`

test/bridge_test.rb
```diff
@@ -62,9 +62,9 @@ module Yast
     ]
 
     before(:each) do
-      NetworkInterfaces.stub(:FilterDevices).with("netcard") { NETCONFIG_ITEMS }
+      allow(NetworkInterfaces).to receive(:FilterDevices).with("netcard") { NETCONFIG_ITEMS }
 
-      LanItems.stub(:ReadHardware) { HWINFO_ITEMS }
+      allow(LanItems).to receive(:ReadHardware) { HWINFO_ITEMS }
```
jreidinger: :+1:

test/lan_udev_auto_test.rb
```diff
@@ -4,9 +4,11 @@
 
 require "yast"
 
-module Yast
+include Yast::UIShortcuts
```
jreidinger: not needed if you use `require "yast/rspec"` helpers already include UIShortcuts and add it only in rspec scope. See https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/rspec/shortcuts.rb

src/modules/LanItems.rb
mvidner: So this need fixing in the SLE branch too.

jreidinger:
LGTM

mchf:
Thanks for reviews.

PR #295, opened 2015-05-11
Rubocop

jreidinger:
you can ignore auto corrections commit, as it is done by rubocop itself without my intervention.

I would like to pin up real bug fix found by rubocop which report unused variable with typo see:


src/lib/network/install_inf_convertor.rb
```diff
@@ -318,31 +314,31 @@ def create_wlan_ifcfg
...
+      when "wep_open", "wep_restricted"
+        if wlan_key_type == "password"
+          type = "h:"
+        elsif wlan_key_type == "ascii"
+          type = "s:"
+        else
+          type = wlan_key_type[0] + ":"
+        end
+
+        wlan_auth_mode = wlan_auth == "wep-open" ? "open" : "sharedkey"
```
jreidinger: this is actually bug fix, as original auth_mode is not used and below wlan_auth_mode crash.

mchf:
one test didn't survive changes (lan_udev_auto)

jreidinger:
I will fix test, I try to finish it before left to NUE, but it failed, so I will fix it tomorrow

mchf:
I've learned something, tests pass and I have no more objections so LGTM

jreidinger:
great

PR #299, opened 2015-05-18
iSCSI - Firewall proposal in 1st stage

kobliha:
- When iSCSI is in use, we need to adjust the firewall proposal accordingly
- bsc#916376


jreidinger:
LGTM

PR #300, opened 2015-05-18
iSCSI-related firewall proposal in 1st stage - also in master

kobliha:
- Port from SLE 12 GA to master -> https://github.com/yast/yast-network/pull/299
- bsc#916376
- Related also to https://github.com/yast/yast-yast2/pull/356

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -7,7 +7,7 @@ Mon May 18 12:59:27 CEST 2015 - locilka@suse.com
 - 3.1.115
 
 -------------------------------------------------------------------
-Fri Mar  6 13:48:24 UTC 2015 - mfilka@suse.com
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon May 18 12:59:27 CEST 2015 - locilka@suse.com
+
+- bsc#916376
+  - proposing iSCSI-related configuration in firewall proposal
+    in first stage if iSCSI is in use
+- 3.1.115
```
jreidinger: missing change in spec file

jreidinger:
LGTM

package/yast2-network.changes
kobliha: Ah, true, that's because version have diverged and this is just a port from SLE 12. I forgot to raise the spec file version as well. Thanks!

PR #301, opened 2015-05-18
Versions between SLE 12 and master have diverged long time ago

kobliha:
- This is a simple "fix" for the divergence so we do not keep this problem on the plate anymore
- 3.1.113 is a "new" SLE 12 version, created just today and not yet submitted to SLE 12 BS


mvidner:
LGTM

PR #302, opened 2015-05-20
Adjusted help text. bnc#931478

mchf:


jreidinger:
@mchf Problem is that I forgot to add rubocop to .travis.yml. Can you please add it and check if it passes? Thanks

mvidner:
The text looks good. Have you omitted version + changelog intentionally?

mchf:
@mvidner: yes, I think it is non critical and simple update which can be postponed.

jreidinger:
LGTM

PR #303, opened 2015-05-21
Always do network card detection in the first stage (bnc#915180)

lslezak:
Always detect network cards when saving the network config before reboot in the first stage to include the correct card name in the `ifcfg-*` file.

- 2.17.207

- Tested in the latest SP4 build, after patching the installer there is no confirmation popup at the end of the first stage and the saved `/etc/systconfig/network/ifcfg-eth0` file contains the correct card name (no generic name).
- In the second stage the correct name is displayed in the proposal.


src/routines/routines.ycp
```diff
@@ -640,7 +641,10 @@ define list<map> ReadHardware(string hwtype) {
     /* Confirmation: label text (detecting hardware: xxx) */
     string hwstring = _("All Network Devices");
     if(haskey(hwstrings, hwtype)) hwstring = hwstrings[hwtype]:"";
-    if(!Confirm::Detection(hwstring, "yast-lan")) return [];
+
+    // don't ask in the 1st stage, if we need to save a netcard config then network was already
+    // used for installation and asking is useless
```
mchf: the comment is misleading. saving network configuration is called unconditionally.

lslezak:
Note: the config is saved in the first stage only when installing from network, when installing from CD network config is not saved (and the detection is not called at all).

src/routines/routines.ycp
lslezak: Really? When I installed from DVD I did not see the popup before reboot... (Or did I overlook it?)

mchf: when you start DVD installation normally then detection is done at the beginning of installation and the popup is not raised.

Try to start installer with manual=1 (this is case for s390) and will see it as well

lslezak: I just tried the latest DVD SP4 build on x86_64 with `manual=1` and the only confirmation displayed in the first stage was the USB driver load confirmation and no network config was saved at the end of the first stage.

I guess the s390s are usually installed from network and therefore there is a network detection and the config is then saved.

lslezak: https://github.com/yast/yast-installation/blob/Code-11-SP4/src/clients/network_finish.ycp#L55

mchf: ok I've played with that a bit and only way how to use network for e.g. driver update when in manual mode. So, no more objections against the comment ;-)

LGTM

lslezak: OK, thanks! :wink:

PR #305, opened 2015-05-25
Fixed duplicated keys in hash literals.

mvidner:
Seen on the console during Tumbleweed installation:

  warning: duplicated key at line 999 ignored: "foo"

Ruby would use the latter value, but I used the value that made more sense:
ralink-firmware is dropped; the former help text was the proofread one.

See also
https://bugs.ruby-lang.org/issues/10315


mchf:
Nice. LGTM

PR #307, opened 2015-06-02
Coveralls support

mchf:


lslezak:
Um, this is incomplete, you need to update the `test_helper.rb` to run the code coverage, see e.g. https://github.com/yast/yast-kdump/blob/master/test/test_helper.rb or https://github.com/yast/yast-kdump/pull/29/files for the complete support in kdump package.

lslezak:
You need to activate the code coverage by `COVERAGE=1` env setting, usually done via `COVERAGE=1 rake test:unit`.

This package partly uses the old testsuite so you will need to add that command to the `.travis.yml` file.

(The code coverage will not be reported for the old tests, in theory it is possible to merge several tests to one report, but that's not trivial and it would be better to convert the old tests to RSpec anyway...)

You can run `COVERAGE=1 rake test:unit` locally to see if it works correctly. The coveralls report should print something like this https://travis-ci.org/yast/yast-registration#L989 at the Travis output.

mvidner:
```
Formatter Coveralls::SimpleCov::Formatter failed with TypeError: Term is not a module (/var/lib/gems/2.1.0/gems/term-ansicolor-1.3.0/lib/term/ansicolor.rb:1:in `<top (required)>')
```
I have addressed this already in kdump: https://github.com/yast/yast-kdump/commit/cad8561b67166670e8a1748a9db37b7fff85c490 and the bad news is, `include Yast` is much used here. Maybe we need a different fix.

mvidner:
```
/home/travis/build/yast/yast-network/test/s390_helpers_test.rb:11:in `<top (required)>': NetworkLanS390Include is not a class (TypeError)
```
this is easy:
```
test/s390_helpers_test.rb:11:class NetworkLanS390Include < Yast::Client
src/include/network/lan/s390.rb:10:  module NetworkLanS390Include
```

jreidinger:
in general all `include Yast` in global namespace is very bad as it can cause easy collisions, as it e.g. overwrite ::String with Yast::String which some code ( even outside of YaST ) do not expect.

mvidner:
Apparently it should work to enclose the tests in `module Yast` instead.

jreidinger:
@mvidner well, also yast/rspec provide a lot of helpers for easier call of methods, which should be prefered way

mchf:
It seems, there are some collisions in testsuites, I'm going to investigate it further. However, I need to move the code into own branch -> closing this PR.

PR #313, opened 2015-06-11
Moved "Network Settings" into "System" group in YCC

mchf:
fate#318804


mchf:
Tested in ncurses and works as expected

src/desktop/network.desktop
```diff
@@ -1,22 +0,0 @@
...
-X-SuSE-YaST-AutoInst=write
-X-SuSE-YaST-Geometry=
-X-SuSE-YaST-SortKey=
-X-SuSE-YaST-AutoInstResource=network
-
-Icon=yast-network
-Exec=xdg-su -c "/sbin/yast2 network"
-
-Name=Expert Network Configuration
-GenericName=Configure network
```
mvidner: This is strange, I don't see this item on a Tmbleweed installation already.

mchf: me too ... I'm checking that in SLES12 .. once it gets installed.

mvidner: Ah, it is removed in the spec file. Please clean it up there.

src/desktop/lan.desktop
```diff
@@ -6,12 +6,12 @@ X-KDE-ModuleType=Library
...
 X-SuSE-YaST-Call=lan
 
-X-SuSE-YaST-Group=Network
+X-SuSE-YaST-Group=System
 X-SuSE-YaST-Argument=
 X-SuSE-YaST-RootOnly=true
 X-SuSE-YaST-AutoInst=all
 X-SuSE-YaST-Geometry=
-X-SuSE-YaST-SortKey=
+X-SuSE-YaST-SortKey=!first
```
mvidner: What is the intent of this? In the qt control center it does not seem to have any effect.

mvidner: It does affect the sorting in the curses menu, I see. But then won't the Kernel Settings module want to become "!!first!!!1!" ?

mvidner: I think it would make sense for it to be the first only if it got to the Network Services group, as their "prerequisite". But in the System group, it should get no special treatment.

mchf: have it first was required in fate#318804. I personally don't care at all. However, if ```X-SuSE-YaST-SortKey``` works only in ncurses then it make sense to have it sorted alphabetically to have same appearance in both ncurses even qt.

mvidner:
LGTM

mchf:
Thanks for review

PR #314, opened 2015-06-15
Fix keep install network

imobach:
@mchf: I've getting an error when using the `keep_install_network` parameter on AutoYaST:

<pre>Details: undefined method `map' for nil:NilClass
Caller: /usr/share/YaST2/modules/LanUdevAuto.rb:199:in `Write'</pre>

 Taking a look at the code, maybe there's a typo.  I've tried this fix and at least installation ends up properly. If that's not the problem, just feel free to ditch this PR away :)


mchf:
LGTM

PR #315, opened 2015-06-17
Filling default domain using the one defined in Hostname module

mchf:


mvidner:
The yast2-network package is already submitted, and this change is not necessary for the fix to work. It is just a code clean-up and it is OK to submit it only later when a material change appears.

LGTM.

mchf:
Thanks for review

PR #316, opened 2015-06-18
Fixed openQA issue

mchf:
bnc#930448

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -301,6 +301,8 @@ module Yast
     end
 
     # Returns configuration of item (see LanItems::Items) with given id.
+    #
+    # @param itemId [Integer] a key for {#Items}
     def GetLanItem(itemId)
       Ops.get_map(@Items, itemId, {})
     end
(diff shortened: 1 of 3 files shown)


src/modules/LanItems.rb
```diff
@@ -1229,7 +1229,7 @@ def BuildBondIndex
       deep_copy(index)
     end
 
-    def startmode_overview
+    def startmode_overview(item_id)
```
mvidner: \# @param item_id [Integer] a key for {#Items}

I keep forgetting which API uses numbers and which uses the kernel names.

mvidner: NTH: Consider adding this to the many other methods which are missing an explanation for item_id / itemId.

src/modules/LanItems.rb
```diff
@@ -1257,7 +1257,8 @@ def startmode_overview
         )
       }
 
-      startmode_descr = startmode_descrs[NetworkInterfaces.Current["STARTMODE"].to_s] || _("Started manually")
+      ifcfg = GetDeviceMap(item_id) || []
```
mvidner:
```
$ ruby -e '[]["foo"]'
-e:1:in `[]': no implicit conversion of String into Integer (TypeError)
        from -e:1:in `<main>'
```

mvidner:
LGTM

mchf:
Thanks for reviews.

PR #317, opened 2015-06-25
Fixed proposing domain

mchf:


src/modules/DNS.rb
```diff
@@ -246,24 +246,14 @@ def default_dhcp_hostname
...
-      @hostname = split[0] || ""
-      @domain = split[1] || ""
-
-      # last resort
-      if @hostname == ""
-        @hostname = "linux"
-        @domain = "site"
-      end
+      # reads setup from /etc/HOSTNAME, returns a default if nothing found
+      @hostname, @domain = *Hostname.SplitFQ(Hostname.CurrentFQ) if fqhostname.nil? || fqhostname.empty?
```
mvidner: So if `fqhostname` is not empty... then it is not used?

mchf: ups. thanks

src/modules/DNS.rb
```diff
@@ -246,24 +246,16 @@ def default_dhcp_hostname
...
-      fqhostname = ""
       # In installation (standard, or AutoYaST one), prefer /etc/install.inf
       # (because HOSTNAME comes with netcfg.rpm already, #144687)
       if (Mode.installation || Mode.autoinst) && FileUtils.Exists("/etc/install.inf")
         fqhostname = read_hostname_from_install_inf
       end
 
-      fqhostname = read_hostname_from_etc if fqhostname.empty?
+      # reads setup from /etc/HOSTNAME, returns a default if nothing found
+      fqhostname = Hostname.CurrentFQ if fqhostname.nil? || fqhostname.empty?
```
mvidner: The docs for `Hostname.CurrentFQ` say 
> Retrieve currently set fully qualified hostname
> (uses hostname --fqdn)

It does that, but in fact it falls back to /etc/HOSTNAME and to linux.$tld. Notably it does not use /etc/hostname like the deleted method. So there may be a different bug hidden in there.
https://github.com/yast/yast-yast2/blob/02ec574edeaba563f0e52a9b3d5cf58e0fe13d59/library/types/src/modules/Hostname.rb#L154

mchf: I know. Updating yast2 is a next step. It has to be updated to fulfill bnc#923990. My primary target here was deduplicating the proposal.

src/modules/DNS.rb
```diff
@@ -246,24 +246,16 @@ def default_dhcp_hostname
...
-      split = Hostname.SplitFQ(fqhostname)
-      @hostname = split[0] || ""
-      @domain = split[1] || ""
-
-      # last resort
-      if @hostname == ""
-        @hostname = "linux"
-        @domain = "site"
-      end
+      @hostname, @domain = *Hostname.SplitFQ(fqhostname)
```
mvidner: SplitFQ may return `[foo, bar]` or `[foo]`. Are you sure you want a nil in `@domain`?

mchf: I'll fix ```SplitFQ``` to return [foo, ""] in case when no domain is defined. To be consistent with ```CurrentDomain``` API.

mchf:
Should be merged after: https://github.com/yast/yast-yast2/pull/376

package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jun 25 14:15:06 UTC 2015 - mfilka@suse.com
+
+- bnc#922765
```
mvidner: This is the SP4 bug, while yast/yast-yast2#376 mentions [bnc#923990](https://bugzilla.suse.com/show_bug.cgi?id=923990), an openSUSE bug. I think both PRs should mention both bugs.

mvidner:
Just add the other bug reference, otherwise it LGTM.

PR #318, opened 2015-06-30
Fixed duplicated keys in hash literals (.changes only)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=930870
was filed for a fix that hadn't had a bugzilla entry:
bac49d9681e73e6c8326849630b6cccdfb609b6d  https://github.com/yast/yast-network/pull/305


mchf:
LGTM

PR #319, opened 2015-07-01
Fixed internal error when adding an alias

mchf:
bnc#864264

### DIFF
--- a/test/alias_dialog_test.rb
+++ b/test/alias_dialog_test.rb
@@ -36,6 +36,9 @@ describe "VirtualEditDialog" do
     allow(Yast::LanItems).to receive(:device).and_return("")
 
     expect(Yast::UI).to receive(:UserInput).once
-    expect(AliasDialog.new.VirtualEditDialog(0, term(:empty), nil)).not_to be nil
+
+    new_id = 0
+    existing_item = term(:empty)
+    expect(AliasDialog.new.VirtualEditDialog(new_id, existing_item)).not_to be nil
   end
 end
--- a/test/alias_dialog_test.rb
+++ b/test/alias_dialog_test.rb
@@ -1,0 +1,41 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+
+require "yast"
+
+class AliasDialog
+  include Yast::I18n
+  include Yast::UIShortcuts
+
+  def initialize
+    Yast.include self, "network/lan/virtual.rb"
+  end
+end
+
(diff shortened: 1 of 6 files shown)


jreidinger:
LGTM

mvidner:
Actually this does not fix the originally reported bug. It fixes a bug that we introduced only 2 months ago in https://github.com/yast/yast-network/commit/4efb006a22ee0dd8489ffdc023c099670d5a09da#diff-4b8e1b8810d82526b81cb37dcf76f811R244
```ruby
-      while true
+      loop do
         host = nil
         ret = UI.UserInput
```

(and that should warn us to be careful when "fixing" Rubocop problems)

mvidner:
Moving a `CloseDialog` into a loop looks suspicious, and indeed it will break if the user presses Cancel. IMHO a better fix is to declare `ret = nil` before the `loop`.

Why don't we add unit tests for the dialog? It should be possible now thanks to Ancor's work in journald.

mvidner:
The code looks good now, but we should add tests. Let me first try resurrecting the coverage PR #307 so that we can see an improvement in the metric ;-)

test/alias_dialog_test.rb
```diff
@@ -0,0 +1,41 @@
...
+      .with(Id(:ipaddr), :Value)
+      .and_return("1.1.1.1")
+    allow(Yast::UI)
+      .to receive(:QueryWidget)
+      .with(Id(:netmask), :Value)
+      .and_return("255.0.0.0")
+    allow(Yast::LanItems).to receive(:device).and_return("")
+
+    expect(Yast::UI).to receive(:UserInput).once
+    expect(AliasDialog.new.VirtualEditDialog(0, term(:empty), nil)).not_to be nil
```
mvidner: `0, term(:empty), nil` ? Even if this is only a smoke test, I have absolutely no idea what the arguments mean. I'll write an improvement later today.

mchf: that's ... bad design ... I can tell you what is what ... but ... you'll see ;)

mvidner:
OK, in the end I just removed the disused argument and documented the method API. Please add https://github.com/mvidner/yast-network/commit/3cecd09f723f66fb36b3cacf32041b5f28a6042b from https://github.com/mvidner/yast-network/commits/master , then it LGTM.

mchf:
updated according instructions ;-)

Going to merge once travis is done. Thanks for reviews.

PR #320, opened 2015-07-01
Coverage report: local and coveralls

mvidner:
aka Return of the #307

### DIFF
--- a/src/clients/firewall_stage1_proposal.rb
+++ b/src/clients/firewall_stage1_proposal.rb
@@ -76,7 +76,9 @@ module Yast
         # Don't override users settings
         SuSEFirewall4Network.prepare_proposal unless SuSEFirewallProposal.GetChangedByUser
 
-        adjust_configuration
+        # this method is not easily mockable in rspec and currently is out of scope
+        # for testing in firewall_stage1_proposal_test.rb
+        adjust_configuration if !Mode.test
 
         script_return = {
           "preformatted_proposal" => preformatted_proposal,
--- a/test/netcard_test.rb
+++ b/test/netcard_test.rb
@@ -69,9 +69,6 @@ MOCKED_ITEMS = {
 
 require "yast"
 
-include Yast::UIShortcuts
-include Yast::I18n
-
 Yast.import "LanItems"
 
 describe "When querying netcard device name" do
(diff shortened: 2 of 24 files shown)


mvidner:
If you try to reproduce this locally, `test/routing_test.rb` needs a **disabled SuSEFirewall**.

mchf:
I've updated original branch - several other tests fixed: https://github.com/mchf/yast-network/tree/coverage

The branch passes through travis, however several more tests have to be fixed.

mchf:
it works, so we can merge it ... or?

src/clients/firewall_stage1_proposal.rb
```diff
@@ -76,7 +76,9 @@ def main
         # Don't override users settings
         SuSEFirewall4Network.prepare_proposal unless SuSEFirewallProposal.GetChangedByUser
 
-        adjust_configuration
+        # this method is not easily mockable in rspec and currently is out of scope
+        # for testing in firewall_stage1_proposal_test.rb
+        adjust_configuration if !Mode.test
```
jreidinger: Ugh? Mode.test is not set in rspec, so how it can affect it? Also I do not belive that it is hard to mock in rspec. Can you be more specific why you cannot simple mock it out?

test/netcard_test.rb
```diff
@@ -69,9 +69,8 @@
 
 require "yast"
 
-include Yast
-include UIShortcuts
-include I18n
+include Yast::UIShortcuts
```
jreidinger: UI shortcuts is by default now part of rspec helper so not needed - https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/rspec/shortcuts.rb#L8

test/netcard_test.rb
```diff
@@ -235,16 +234,17 @@
...
   # this test covers bnc#914833
   it "doesn't try to update udev rules when none exists for the item" do
-    allow(LanItems)
+    allow(subject)
       .to receive(:Items)
       .and_return(MOCKED_ITEMS)
 
-    item_id = LanItems.Items.find { |_k, v| !v.key?("udev") }.first
-    expect(LanItems.SetItemName(item_id, new_name)).to eql new_name
+    item_id = subject.Items.find { |_k, v| !v.key?("udev") }.first
```
jreidinger: if not found it crashes, but probably not so big problem in tests

test/remote_test.rb
```diff
@@ -11,7 +11,7 @@ module Yast
   import "Package"
   import "Packages"
 
-  RSpec.configure do |c|
+  ::RSpec.configure do |c|
```
jreidinger: I see this more times, maybe add it to test_helper?

test/routing_test.rb
```diff
@@ -4,17 +4,12 @@
...
-include I18n
-
+include Yast::I18n
 Yast.import "Routing"
 
-describe Routing do
-  SYSCTL_IPV4_PATH = path(RoutingClass::SYSCTL_IPV4_PATH)
-  SYSCTL_IPV6_PATH = path(RoutingClass::SYSCTL_IPV6_PATH)
+describe Yast::Routing do
+  SYSCTL_IPV4_PATH = Yast::Path.new(Yast::RoutingClass::SYSCTL_IPV4_PATH)
```
jreidinger: path should work, as rspec helper contain it.

test/routing_test.rb
```diff
@@ -23,60 +18,60 @@
...
+      allow(Yast::SCR).to receive(:Execute) { nil }
     end
 
     def fw_independent_write_expects
-      expect(SCR)
+      expect(Yast::SCR)
         .to receive(:Execute)
         .with(
-          path(".target.bash"),
+          Yast::Path.new(".target.bash"),
```
jreidinger: same here and also below

test/widgets_test.rb
```diff
@@ -4,18 +4,24 @@
 
 require "yast"
 
-include Yast
-include UIShortcuts
-include I18n
+include Yast::UIShortcuts
```
jreidinger: same as above

src/clients/firewall_stage1_proposal.rb
mvidner: That is done by a specific ` allow(Mode).to receive(:test).and_return(true)` for that test ;-)
I guess the mocking is problematic because this is a client.

jreidinger: yep, it can be reason as clients are loaded and evaluated in same time. OK, so fine for me. Just maybe change "this method" to "client class" to be precise and not confusing

test/netcard_test.rb
mvidner: Indeed. But I18n isn't.

jreidinger: I am talking only about UIShortcuts. I18n is not needed by majority of tests, so there is no pressure to add it. But UI shortcuts are useful for any UI testing.

test/remote_test.rb
mvidner: Right, will do

test/routing_test.rb
mvidner: Indeed, but only if I wrap the constants in a `before(:all)`.

jreidinger: ah, so it is instance variables and not class ones. OK

jreidinger: but others below should work as expected

PR #321, opened 2015-07-02
Enabled coveralls

mchf:
Follows https://github.com/yast/yast-network/pull/307 and https://github.com/yast/yast-network/pull/320


mchf:
still an issue here ... coveralls do not send reports

mvidner:
Yes. That is what I've been fixing yesterday. I think I will take your fix of firewall_stage1_proposal_test.rb and make #320 working.

mchf:
May be we should merge our branches to avoid work duplication.

mvidner:
Used 320 in the end. Thank you!

PR #322, opened 2015-07-08
Added tests of the CLI (bsc#935937).

mvidner:
WIP, not ready for merging as is, but comments are welcome.

**Warning**, they will reconfigure your system (actually the current ones
try to clean up after themselves, but they won't stay this way), so use
them in a scratch VM.


test/cli/t/add-del.t
```diff
@@ -0,0 +1,49 @@
...
+# this will tell TAP "not ok" IFF foo has failed
+tapfail() {
+  echo -n "not "
+}
+
+YAST=/usr/sbin/yast
+
+echo "# assume and check that eth0 exists, as a base for the vlan"
+ip addr show dev eth0 || tapfail
+echo "ok 1 eth0: exists already"
```
jreidinger: if I get your code in case of fail it returns `not ok 1 eth0: exists already` which is a bit confusing for me.

Another note I have is that I think we should use different directory for such kind of tests as in "test" are unit tests which is run automatic and is non-destructive.

mvidner: Yes, `not ok 1 eth0: exists already` is according to http://testanything.org/ . I hope it will be a little less confusing when I insert a dash after the test number.

mvidner:
As for the directory, yes, these tests should be separate from the unit tests. Putting them to a sub-dir of those is a borderline case, and I have no problem with moving them elsewhere. Maybe `/test-live`, to signify that they operate on the live system?

ancorgs:
Some notes:

1) I would remove the unknown-command file. It does not belong to yast2-network and the same functionality is already tested by RSpec tests in yast2.

2) Looks like the rollback functionality in openQA is not 100% stable and we should not abuse it. So we have two options. Either we use pennyworth instead (or inside) openQA, either we make sure that most tests clean up their own mess, so we only need rollback in a few specific cases.

3) I agree we should move the tests out of `/test`. Actually the convention with TAP is to simply have the tests under a directory called `/t`. People familiar with TAP (like Perl developers) will be tempted to run `prove` as soon as they see a `/t` directory. But that's what README files are for :wink:

ancorgs:
Oops, I didn't intent to add that pull request. I was targeting my fork. Sorry

ancorgs:
If we leave this as it is, we could use https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/614 to run the CLI tests in openQA.

About the rollback. If the test fails, openQA will take care of performing the rollback. So as long as the tests leave everything in the original state (as this one does), we are safe.

If some CLI test in the future does not cleanup its own mess, then we can introduce the rollback explicitly in the openQA test. But it's not needed now.

t/add-del.t
```diff
@@ -0,0 +1,49 @@
...
+ip addr show dev vlan50 || tapfail
+echo "ok 3 vlan50: interface exists"
+
+ls /etc/sysconfig/network/ifcfg-vlan50 || tapfail
+echo "ok 4 vlan50: ifcfg exists"
+
+echo "# find it"
+# the embarassing part: no way to identify by "vlan50"
+ID=$($YAST lan list |& grep Virtual.LAN | cut -f1)
+echo "ok 5 vlan50: yast specific ID found"
```
jreidinger: ??? where is test?

t/add-del.t
```diff
@@ -0,0 +1,49 @@
...
+
+echo "# delete it"
+$YAST lan delete id=$ID verbose
```
jreidinger: why no tapfail here?

t/add-del.t
ancorgs: Yes. This should be clearly merged with test 6

t/add-del.t
ancorgs: Good question. I think we are using tapfail in a quite arbitrary way. How should we use it in my opinion?

`prove` interrupts the execution of the whole script if some line returns non-zero. So in my opinion we should never use tapfail when the failure is critical and would break the following steps. So IMO, 1 and 2 should not use tapfail because if they fail we really want to abort (it will be reported as an error by prove even without the `echo not`.

The rest of the tests should use tapfail because we can still continue trying after them.

jreidinger:
LGTM

jreidinger:
btw for future some tutorial how to write this kind of tests would be nice ( like when use tapfail ), how format should look like etc. ( no need to write it from scratch, but some collection of links would be nice )

PR #323, opened 2015-07-23
Enables IPv6 forwarding

mchf:
bnc#916013


src/modules/Routing.rb
```diff
@@ -137,54 +137,64 @@ def RemoveDefaultGw
...
         SCR.Write(
           path(SYSCTL_IPV4_PATH),
-          forward_ipv4
-        )
-        SCR.Write(
-          path(SYSCTL_IPV6_PATH),
-          forward_ipv6
+          sysctl_val
         )
         SCR.Write(path(".etc.sysctl_conf"), nil)
```
jreidinger: this is not necessary as you write it with second call.

testsuite/tests/Network_YaPI.out
```diff
@@ -295,10 +295,9 @@ Read	.etc.sysctl_conf."net.ipv6.conf.all.forwarding" nil
 Execute	.target.bash_output " LANG=C TERM=dumb COLUMNS=1024 systemctl --no-legend --no-pager --no-ask-password show SuSEfirewall2.service  --property=Id  --property=MainPID  --property=Description  --property=LoadState  --property=ActiveState  --property=SubState  --property=UnitFileState  --property=FragmentPath " $["exit":0, "stderr":"", "stdout":"laptop.suse.cz"]
 Execute	.target.bash_output " LANG=C TERM=dumb COLUMNS=1024 systemctl --no-legend --no-pager --no-ask-password is-enabled SuSEfirewall2.service " $["exit":0, "stderr":"", "stdout":"laptop.suse.cz"]
 Write	.etc.sysctl_conf."net.ipv4.ip_forward" "0" true
+Write	.etc.sysctl_conf nil true
```
jreidinger: as I said, this is not neccessary, onle one nil is enough at the end

src/modules/Routing.rb
```diff
@@ -137,54 +137,64 @@ def RemoveDefaultGw
...
       end
 
-      SCR.Execute(
-        path(".target.bash"),
-        "echo #{forward_ipv4} > /proc/sys/net/ipv4/ip_forward"
-      )
-      SCR.Execute(
-        path(".target.bash"),
-        "echo #{forward_ipv6} > /proc/sys/net/ipv6/conf/all/forwarding"
```
mvidner: So why do we no longer do this? "Dropped useless code" does not help me understand the reason.
AFAIK the /etc/sysctl_conf config is the *permanent* configuration of the settings and this /proc/sys/net was the *immediate* configuration. Does that mean the settings will not be effective until a reboot?

package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 23 16:00:07 UTC 2015 - mfilka@suse.com
+
+- bnc#916013
+  - enables IPv6 forwarding
```
mvidner: I guess that is not the actual fix, as the IPv6 forwarding checkbox existed even before this change. Can you please describe what is being fixed from the user's point of view?

src/modules/Routing.rb
mchf: yes ```/proc``` was for immediate action. But I've replace that using ```sysctl -w``` call. I like it more bcs it hides dependency on /proc and is more consistent with updating persistent sysctl config few lines above.

mchf: I've reworded the commit message.

src/modules/Routing.rb
```diff
@@ -50,8 +50,9 @@ class RoutingClass < Module
     # "routes" file location
     ROUTES_FILE = "/etc/sysconfig/network/routes"
 
-    SYSCTL_IPV4_PATH = ".etc.sysctl_conf.\"net.ipv4.ip_forward\""
-    SYSCTL_IPV6_PATH = ".etc.sysctl_conf.\"net.ipv6.conf.all.forwarding\""
+    SYSCTL_AGENT_PATH = ".etc.sysctl_conf"
+    SYSCTL_IPV4_PATH = "net.ipv4.ip_forward"
+    SYSCTL_IPV6_PATH = "net.ipv6.conf.all.forwarding"
```
mvidner: I think this is VERY confusing. How should I know which of these to use?
```rb
SCR.Read(path(SYSCTL_IPV6_PATH))
SCR.Read(path("#{SYSCTL_AGENT_PATH}.#{SYSCTL_IPV6_PATH}"))
SCR.Read(path("#{SYSCTL_AGENT_PATH}.\"#{SYSCTL_IPV6_PATH}\""))
```

Thinking of it, the best seems
```rb
SYSCTL_AGENT_PATH = path(".etc.sysctl_conf")
SYSCTL_IPV4_PATH = SYSCTL_AGENT_PATH + "net.ipv4.ip_forward"  # add as a single path component
SYSCTL_IPV6_PATH = SYSCTL_AGENT_PATH + "net.ipv6.conf.all.forwarding"
...
SCR.Read(SYSCTL_IPV6_PATH)
```

(There is a difference between `path(".foo") + ".bar.baz"` and `path(".foo") + path(".bar.baz")` which I will need to document properly, [`Yast.Path#+`](http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast%2FPath%3A%2B) does not mention it)

mvidner: Oh, I see, `SYSCTL_IPV9_PATH` are used as an argument to `sysctl -w`. Then I'd use

```rb
# sysctl keys, used as *single* SCR path components below
IPV4_SYSCTL = "net.ipv4.ip_forward"
IPV6_SYSCTL = "net.ipv6.conf.all.forwarding"
# SCR paths
SYSCTL_AGENT_PATH = path(".etc.sysctl_conf")
SYSCTL_IPV4_PATH = SYSCTL_AGENT_PATH + IPV4_SYSCTL
SYSCTL_IPV6_PATH = SYSCTL_AGENT_PATH + IPV6_SYSCTL
```

testsuite/tests/Network_YaPI.out
mvidner: The small additional cost is IMHO outweighed by more regular code, so not a problem.

test/routing_test.rb
```diff
@@ -9,8 +9,8 @@
 
 describe Yast::Routing do
   before(:all) do
-    SYSCTL_IPV4_PATH = path(Yast::RoutingClass::SYSCTL_IPV4_PATH)
-    SYSCTL_IPV6_PATH = path(Yast::RoutingClass::SYSCTL_IPV6_PATH)
+    SYSCTL_IPV4_PATH = path("#{Yast::RoutingClass::SYSCTL_AGENT_PATH}.\"#{Yast::RoutingClass::SYSCTL_IPV4_PATH}\"")
+    SYSCTL_IPV6_PATH = path("#{Yast::RoutingClass::SYSCTL_AGENT_PATH}.\"#{Yast::RoutingClass::SYSCTL_IPV6_PATH}\"")
```
mvidner: So this should become
```rb
SYSCTL_IPV4_PATH = Yast::RoutingClass::SYSCTL_IPV4_PATH
SYSCTL_IPV6_PATH = Yast::RoutingClass::SYSCTL_IPV6_PATH
```

mvidner:
LGTM, thank you!

mchf:
rebased, merging once travis is done. Thanks for reviews.

PR #324, opened 2015-07-31
Fix crash when using CLI to delete an interface

ancorgs:


mvidner:
Tested with https://github.com/yast/yast-network/pull/322 and https://github.com/os-autoinst/os-autoinst-distri-opensuse/pull/614
LGTM

PR #328, opened 2015-08-12
Bnc940892 lanudevauto internal error

mchf:


src/modules/LanUdevAuto.rb
```diff
@@ -184,7 +184,7 @@ def Import(settings)
           createUdevFromIfaceName(Ops.get_list(settings, "interfaces", []))
         )
       else
-        @udev_rules = Ops.get_list(settings, "net-udev", [])
+        @udev_rules = settings["net-udev"] || {}
```
ancorgs: shouldn't this be ``settings["net-udev"] || []`. Looking at the rest of the code, it seems to be an array of hashes, not a hash.

mchf: you're right. thx

ancorgs:
lgtm

kobliha:
Hmm, sounds like we actually need a test case?

mchf:
thanks for review.

mchf:
hups, wrong button :-)

kobliha:
OK, thanks, feel free to merge

test/lan_udev_auto_test.rb
```diff
@@ -105,10 +105,29 @@
...
+    udev_rules = [
+      {
+        "rule"  => ATTR,
+        "value" => VALUE,
+        "name"  => NAME
+      }
+    ]
+
+    expect(LanUdevAuto.Import({ "net-udev" => udev_rules }))
+      .to be_true
```
ancorgs: better use `.to eq true`

test/lan_udev_auto_test.rb
```diff
@@ -105,10 +105,29 @@
...
-      LanUdevAuto.Import({ "net-udev" => udev_rules })
-      LanUdevAuto.Write
-    end
+    LanUdevAuto.Import({ "net-udev" => udev_rules })
+    LanUdevAuto.Write
+  end
+end
+
+describe "LanUdevAuto#Import" do
+  it "imports defined udev rules correctly" do
```
ancorgs: This does not test that there are imported correctly. It only tests it does not crash.

test/lan_udev_auto_test.rb
```diff
@@ -77,11 +77,11 @@
 
 end
 
-describe "LanUdevAuto#Write" do
+ATTR = "ATTR{address}"
+VALUE = "aa:BB:cc:DD:ee:FF"
+NAME = "custom-name"
```
ancorgs: Better use `let` instead of polluting the global namespace with constants

ancorgs:
Yes, I know it's already merged, but I couldn't resist to point some things in the test.

mchf:
Thanks for notes. I'll fix it when merging into master

test/lan_udev_auto_test.rb
mchf: yes it was targeted mainly against rewriting from Builtins into ruby, so a smoke test.

PR #329, opened 2015-08-12
Mege SLE-12 into master branch

mchf:


test/lan_udev_auto_test.rb
```diff
@@ -108,3 +109,26 @@
...
+    udev_rules = [
+      {
+        "rule"  => attr_addr,
+        "value" => value,
+        "name"  => name
+      }
+    ]
+
+    expect(Yast::LanUdevAuto.Import({ "net-udev" => udev_rules }))
+      .to be true
```
ancorgs: better `eq` than `be`

test/lan_udev_auto_test.rb
```diff
@@ -6,105 +6,135 @@
...
-      expect(Yast::LanUdevAuto.getDeviceName(NEW_STYLE_NAME)).to be_equal NEW_STYLE_NAME
+  describe ".getDeviceName" do
+    subject(:getDeviceName) { Yast::LanUdevAuto.getDeviceName(oldname) }
+
+    let(:new_style_name) { "spec0" }
+    let(:mac) { "00:11:22:33:44:FF" }
+    let(:busid) { "0000:00:19.0" }
+
+    # general mocking stuff is placed here
+    before(:each) do
```
jreidinger: `before(:each)` can be replaced by simple `before`

src/modules/LanUdevAuto.rb
```diff
@@ -180,7 +180,7 @@ def Import(settings)
           createUdevFromIfaceName(Ops.get_list(settings, "interfaces", []))
         )
       else
-        @udev_rules = Ops.get_list(settings, "net-udev", [])
+        @udev_rules = settings["net-udev"] || []
```
jreidinger: I do not get how it fixes internal error. For me it looks equivalent or I overlook something?

mchf: actual fix is in the other file in the commit. I was also checking if an expected default value is set properly everywhere, so I turned it from converted code into ruby here.

mchf: note that the fix was already present in the master, so here it disappears in the merge.

test/lan_udev_auto_test.rb
mchf: can be, but I prefer to be explicit. Is it blocker?

src/modules/LanUdevAuto.rb
jreidinger: I see only change here and in test suite, spec and changes, nothing else.

jreidinger: ahm having it in master, exaplains it

test/lan_udev_auto_test.rb
jreidinger: not blocker just `nth`, as before(:all) is almost not used, so each is good and intuitive default

kobliha: I use it quite often and nobody ever objected :)

jreidinger:
LGTM

mchf:
Thanks for reviews

PR #332, opened 2015-08-27
Bridge slaves selection stays persistent when switching tabs

mchf:
bnc#846201

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -446,6 +446,24 @@ module Yast
       nil
     end
 
+    # Initializes widget (BRIDGE_PORTS) which contains list of devices available
+    # for enslaving in a brige.
+    #
+    # @param [String] key	id of the widget
+    def InitBridge(key)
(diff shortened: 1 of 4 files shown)


src/include/network/lan/address.rb
```diff
@@ -446,6 +446,24 @@ def StoreAddrWidget(key, event)
       nil
     end
 
+    # Initializes widget (BRIDGE_PORTS) which contains list of devices available
+    # for enslaving in a brige.
+    #
+    # @param [String] key	id of the widget
+    def InitBridge(key)
+      br_ports = (@settings["BRIDGE_PORTS"] || "").split
+      br_ports ||= (NetworkInterfaces.Current["BRIDGE_PORTS"] || "").split
```
jreidinger: this never happen as `"".split` is `[]` so or is always `true`

mchf: ok, i'll improve the copy paste fix ;-)

jreidinger:
LGTM

mchf:
Thanks for review

PR #333, opened 2015-09-02
Fixed handling of "no info returned from GetDeviceMap(...)"

kobliha:
- bsc#911571
- Similar solution proposed and tested by Adam Spiers
- I've tested this only using a unit test


mchf:
Well, this issue appears in several forms over the time. The obvious fix as present in this PR is not favorite one for me. Bcs it simply hides the real problem. So, do we need hotfix that urgently?

kobliha:
OK, we don't want to hide the real problem, but we need a solution.

PR #334, opened 2015-09-15
Update openSUSE package to version 3.1.114

imobach:
This merge brings version 3.1.114 of yast2-network to openSUSE 13.2. At this point, the tarball generated for openSUSE 13.2 have the almost the same contents than [revision 633](https://build.opensuse.org/package/show/YaST:Head/yast2-network?rev=633) in OBS (which corresponds to 3.1.114 for YaST:Head). The difference is the Rakefile which is ok from my POV:

```diff
 Yast::Tasks.configuration do |conf|
+  conf.obs_api = "https://api.opensuse.org"
+  conf.obs_target = "openSUSE_13.2"
+  conf.obs_sr_project = "openSUSE:13.2:Update"
+  conf.obs_project = "YaST:openSUSE:13.2"
   #lets ignore license check for now
   conf.skip_license_check << /.*/
```

I've tested this package manually and fixes [bsc#911571](https://bugzilla.suse.com/show_bug.cgi?id=911571). Moreover it contains many fixes from SLE-12.


package/yast2-network.changes
```diff
@@ -1,4 +1,41 @@
...
+  - do not crash when renaming virtual device via hardware tab
+- bnc#910337
+  - initialize configuration name in Hardware tab correctly. It
+    means according device type. Also uses custom name if user
+    used nonstandard one.
+- bnc#905738
+  - udev rules provided by AutoYaST profile gets applied
+- bnc#918356
+  - populate bond slave candidates list with proper device names
+- 3.1.114
```
kobliha: This is a problem, we already have this version (had this version in the past and it was different I guess). Maintenance branches must use X.Y.Z.A versioning instead of X.Y.Z (as used in development).

kobliha:
I don't have many comments. In general the code doesn't look bad as it's just a merge. Changes file also looks OK. The only problem I see might be the final version - if it actually differs.

package/yast2-network.changes
imobach: @kobliha Yes, you're right, we had this version in the past but the code is exactly the same (the only difference is the Rakefile). Anyway, as latest yast2-network version in this branch is 3.1.110, we could tag this version as 3.1.110.1 or something like that.

imobach:
@kobilha I've updated version to 3.1.110.1, so now it matches maintenance versions numbering.

kobliha:
@imobach Thanks, then LGTM from my POV, but we need also a LAN expert view on that. If something is likely to break because of XYZ, I would not see it.

imobach:
@kobliha Thanks. Let's wait for @mchf POV then.

mchf:
LGTM

imobach:
@mchf @kobliha Thanks for your reviews!

PR #336, opened 2015-09-16
Fix for cmdline in Routing module

mchf:
bnc#945947

### DIFF
--- a/src/modules/Routing.rb
+++ b/src/modules/Routing.rb
@@ -72,6 +72,7 @@ module Yast
       Yast.import "Map"
       Yast.import "Mode"
       Yast.import "SuSEFirewall"
+      Yast.import "FileUtils"
 
       Yast.include self, "network/runtime.rb"
       Yast.include self, "network/routines.rb"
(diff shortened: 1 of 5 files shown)


src/modules/Routing.rb
```diff
@@ -289,6 +290,24 @@ def Write
...
+      # work around bnc#19476
+      if device == ANY_DEVICE
+        filename = ROUTES_FILE
+
+        return SCR.Write(path(".target.string"), filename, "")
+      else
+        filename = "#{ROUTES_DIR}/ifroute-#{device}"
+
+        return SCR.Execute(path(".target.remove"), filename) if FileUtils.Exists(filename)
+        return true
```
jreidinger: more logic way for me is to quick exit for non exist file like:
```
return true unless FileUtils.Exists(filename)

SCR.Execute(path(".target.remove"), filename)
```

test/ifroute_test.rb
```diff
@@ -81,6 +81,8 @@
 end
 
 describe "Routing#write_routes" do
+  Yast.import "FileUtils"
```
jreidinger: this should not be needed as it is already imported in Routing or I overlook something?

mvidner:
LGTM, thanks!

mchf:
Thanks for reviews

PR #338, opened 2015-09-23
Fixed AY import to keep caches consistent when doing so several times in a row

mchf:
bnc#866742


mchf:
incorrect target branch

PR #339, opened 2015-09-23
Fixed AY import to keep caches consistent when doing so several times in a row

mchf:
bnc#866742

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -668,16 +668,7 @@ module Yast
     def Import(settings)
       settings = {} if settings.nil?
 
-      NetworkInterfaces.Import("netcard", settings["devices"] || {})
-      NetworkInterfaces.List("netcard").each do |device|
-        LanItems.AddNew
-        LanItems.Items[LanItems.current] = { "ifcfg" => device }
-      end
-
-      LanItems.autoinstall_settings["start_immediately"] = settings["start_immediately"] || false
-      LanItems.autoinstall_settings["strict_IP_check_timeout"] = settings["strict_IP_check_timeout"] || -1
-      LanItems.autoinstall_settings["keep_install_network"] = settings["keep_install_network"] || false
-
+      LanItems.Import(settings)
       NetworkConfig.Import(settings["config"] || {})
       DNS.Import(settings["dns"] || {})
       Routing.Import(settings["routing"] || {})
@@ -695,7 +686,6 @@ module Yast
 
       @ipv6 = settings.fetch("ipv6", true)
 
-      LanItems.SetModified
       true
     end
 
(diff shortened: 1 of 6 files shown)


src/modules/Lan.rb
```diff
@@ -666,34 +666,14 @@ def WriteOnly
     # @param [Hash] settings settings to be imported
     # @return true on success
     def Import(settings)
-      settings = deep_copy(settings)
-      NetworkInterfaces.Import("netcard", Ops.get_map(settings, "devices", {}))
-      Builtins.foreach(NetworkInterfaces.List("netcard")) do |device|
-        LanItems.AddNew
-        Ops.set(LanItems.Items, LanItems.current, { "ifcfg" => device })
-      end
+      settings = {} if settings.nil?
```
jreidinger: `settings ||= {}`

src/modules/Lan.rb
```diff
@@ -703,11 +683,9 @@ def Import(settings)
       else
         NetworkService.use_wicked
       end
-      if Builtins.haskey(settings, "ipv6")
-        @ipv6 = Ops.get_boolean(settings, "ipv6", true)
-      end
 
-      LanItems.modified = true
+      @ipv6 = settings["ipv6"] || true
```
jreidinger: it is a bit confusing to have spread over whole file settings default. What about set it at beginning or maybe even better use merging? like `settings = DEFAULT_SETTINGS.merge(settings)`

mchf: well, it is. And not only in one file. But I'm not able to save them all now, so I'm only touching those directly affected by a PR.

src/modules/LanItems.rb
```diff
@@ -1088,6 +1081,45 @@ def Read
...
+    def reset_cache
+      LanItems.Items = {}
+
+      @modified = false
+    end
+
+    # Imports data from AY profile
+    #
+    # As network related configuration is spread over the whole AY profile
+    # the function requires hash map with whole AY profile.
```
mvidner: Not quite. It is not the whole profile but rather its `<profile>/<networking>` element, AFTER it has been processed by [FromAY](https://github.com/yast/yast-network/blob/6ca4d37d55fd1517751cef86b696e9692dedbd53/src/clients/lan_auto.rb#L169).

src/modules/Lan.rb
mvidner: `@ipv6 = settings["ipv6"] || true` NOOO!

That is the classical case where we must not use the `|| default` for supplying the default value. What if the user explicitly had `settings["ipv6"] = false`? That's why we must use `settings.fetch("ipv6", true)`.

src/modules/LanItems.rb
mchf: right. Bad wording. I'm going to rephrase it.

src/modules/Lan.rb
mchf: thanks

mvidner:
Just that one `fetch` thing, otherwise it looks good.

mvidner:
LGTM

mchf:
Thanks for reviews

PR #340, opened 2015-09-23
Merge SLE-12 into master

mchf:


src/modules/LanItems.rb
```diff
@@ -1108,6 +1103,46 @@ def Read
...
+    def Import(settings)
+      reset_cache
+
+      NetworkInterfaces.Import("netcard", settings["devices"] || {})
+      NetworkInterfaces.List("netcard").each do |device|
+        AddNew()
+        LanItems.Items[current] = { "ifcfg" => device }
+      end
+
+      autoinstall_settings["start_immediately"] = settings["start_immediately"] || false
```
jreidinger: well `|| false` looks strange, better use fetch.

mvidner:
LGTM

mchf:
Thanks for reviews

PR #341, opened 2015-10-01
Moved "Network Settings" into "System" group in y2cc-gnome

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=867809


mchf:
LGTM

mvidner:
Karl Eichwalder mentioned earlier in Bugzilla that it may be too late to fix the docs, so waiting for his feedback before merging this.

mvidner:
Got a *go* from him.

PR #342, opened 2015-10-02
Relax-NG ... Some defines for ipv6_forward needs the combine attribute

mvidner:
L3: https://bugzilla.suse.com/show_bug.cgi?id=948206
That bug also needs https://github.com/yast/yast-auth-client/pull/44


mvidner:
The correct target for this is SLE-12-GA so I need to open a new PR.

mvidner:
Superseded by #343

PR #346, opened 2015-10-11
Cleanup: Put some defaults into Yaml

mchf:
Done out of SRUM, to be merged after RC2.

These defaults directly corresponds to sysconfig defaults or distribution defaults. No reason to have them in sources.


src/modules/LanItems.rb
```diff
@@ -198,108 +198,24 @@ def main
...
-        # default options for bonding (bnc#404449)
-        "BONDING_MODULE_OPTS"          => "mode=active-backup miimon=100",
-        # defaults for tun/tap devices
-        "TUNNEL_SET_OWNER"             => "",
-        "TUNNEL_SET_GROUP"             => "",
-        # Infiniband
-        # default mode for IPoIB devices suggested in fate#315501
-        "IPOIB_MODE"                   => "connected"
-      }
+      yaml_dir = File.expand_path("../../data", __FILE__) + "/network"
```
jreidinger: Ehm, no this is not way to go, it kills y2update and various loading location. Correct way is https://github.com/yast/yast-yast2/blob/master/library/general/src/modules/Directory.rb#L125

problem is also when you want experimentaly change only one yaml file in y2update so it should looke like

`@s390_defaults = YAML.load_file(Directory.find_data_path("network/s390_defaults.yml")) if Arch.s390`
and similar to others.

jreidinger: if you find it too long. I suggest to create helper method load_data that do this yaml loading with network prefix.

src/modules/LanItems.rb
```diff
@@ -1529,24 +1445,24 @@ def SelectHWMap(hardware)
       # Wireless Card Features
       @wl_auth_modes = Builtins.prepend(
         Convert.convert(
-          Ops.get(hardware, "wl_auth_modes", @nilliststring),
+          Ops.get(hardware, "wl_auth_modes", nil),
```
jreidinger: nil is default for get so useless and also convert is useless unless so want to use nil for invalid value. so in general it can look like
`@wl_auth_modes = hardware["wl_auth_modes"] ? ["no-encryption", *hardware["wl_auth_modes"]] : nil`
see explicit nil at the end as from previous code it is not much obvious that it will be nil if wl_auth_modes is nil. BTW can it happen at all? that looks wrong logic for me. as no-encryption should be always supported even if hardware do not specify own modes.

src/modules/LanItems.rb
```diff
@@ -1529,24 +1445,24 @@ def SelectHWMap(hardware)
...
       )
       @wl_channels = Convert.convert(
-        Ops.get(hardware, "wl_channels", @nilliststring),
+        Ops.get(hardware, "wl_channels", nil),
         from: "any",
         to:   "list <string>"
       )
       @wl_bitrates = Convert.convert(
-        Ops.get(hardware, "wl_bitrates", @nilliststring),
+        Ops.get(hardware, "wl_bitrates", nil),
```
jreidinger: same for all of them

jreidinger:
otherwise nice improvement :+1:

src/modules/LanItems.rb
mchf: yep, this was fix by s/../../g targeted against the `@nilstring` variable, so I didn't check all replacements. 

Wireless configuration is wild field which will need lot of love. I works, at least somehow, now (most probably thanks to NetworkManager :-/)

I'll tweak it a bit more regarding dropping some more builtins.

src/modules/LanItems.rb
mchf: ok

src/modules/LanItems.rb
mchf: I was so proud that I was able to do it using my skills ;-) Ok, I'll change it.

jreidinger:
LGTM, of course it can be improve even more with introducing new classes that encapsulate some logic, but in general great improvement

mvidner:
Nice!
I thought that it would be useful to have also a low level test for YAML syntax validity, so I wrote this:

cat test/yaml_parse_test.rb

```ruby
#! /usr/bin/env rspec

require_relative "test_helper"
require "yast"

require "yaml"

describe "YAML data files" do
  globs = Yast.y2paths.map { |p| "#{p}/data/**/*.yml" }
  yml_filenames = Dir.glob(globs)
  yml_filenames.each do |f|
    it "parse without error: #{f}" do
      expect { YAML.load_file(f) }.to_not raise_error
    end
  end
end
```

It works for me, passing for the files you committed and failing when I introduce an error.

mchf:
> I thought that it would be useful to have also a low level test for YAML syntax validity, so I wrote this:

Great. More tests is better, I'll add that. Thanks

mchf:
I've added the test as suggested by @mvidner.

Thanks for reviews.

If there are no more comments I'll merge it after RC release.

mchf:
RC2 is out, merging.

kobliha:
Well, Michal, but you should have branched SLE 12 SP1 first

PR #347, opened 2015-10-13
Fixed "AutoYaST fails to configure network" (bsc#949193).

mvidner:
[L3 bug 949193](https://bugzilla.suse.com/show_bug.cgi?id=949193)

The fix for bsc#874259 erroneously removed the "devices" part of FromAY
return value. As a result no interfaces were configured, always keeping
what was used for the installation.

(still testing whether the fix actually helps)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Oct 19 12:17:07 UTC 2015 - mvidner@suse.com
+
+- Fixed a regression where AutoYaST flag keep_install_network=false
+  was ignored (bsc#949193, introduced in 3.1.112.2).
+- 3.1.112.10
+
+-------------------------------------------------------------------
 Fri Oct  2 13:44:00 UTC 2015 - mvidner@suse.com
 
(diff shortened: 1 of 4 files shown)


mvidner:
BTW, regarding `FromAY` and `ToAY`, as I was trying to understand the bug I made a refactoring of `ToAY` to see what it actually does: https://github.com/mvidner/yast-network/commit/7ef06f300580c3325cc51c269f5073a6844e8317 in [branch understand_toay](https://github.com/mvidner/yast-network/tree/understand_toay). Of course we would be crazy to merge it as is, without tests. So JFYI.

mchf:
Note (mainly for myself ;-):
I want to do some testing before submitting the fix for maintenance.

mvidner:
@mchf Please give this a priority, because 3 customers reported the L3.

mchf:
For record:
the bug was in incorrectly tested refactoring. The issue was introduced here cd9a2f948ac23814958218f53b70b1ed430cb317

LGTM and thanks for finding it.

mvidner:
I haven't actually tested this myself (testing installation with updates is trickier than I thought) but the customers in the bug did test and reported success.

PR #348, opened 2015-10-20
Updated Rakefile

mchf:


Rakefile
```diff
@@ -1,6 +1,10 @@
 require "yast/rake"
 
 Yast::Tasks.configuration do |conf|
+  conf.obs_api = "https://api.suse.de/"
+  conf.obs_target = "SLE-12-SP1"
+  conf.obs_sr_project = "SUSE:SLE-12-SP1:GA"
+  conf.obs_project = "Devel:YaST:Head"
```
lslezak: I think we should create a `Devel:YaST:SLE-12-SP1` subproject, just like we have [`Devel:YaST:SLE-12`](https://build.suse.de/project/show/Devel:YaST:SLE-12)...

mchf: it is based on https://github.com/yast/yast-country/blob/SLE-12-SP1/Rakefile as pointed by @kobliha

jreidinger: well, but it is not correct to overuse Head, which is master and will be SP2. It will hit us in future. Also currently there is under review imo change to introduce shortcut with `submit_to` - https://github.com/yast/yast-rake/pull/21

lslezak:
Looks OK, but shouldn't we change all packages in the SP1 branch globally at once? As a part of https://trello.com/c/LqgsBfb7 ?

@jreidinger ?

lslezak:
Um, sorry, still not OK... but I think we should do it globally when doing the SP1 branching...

mchf:
yes, it was discussed with @jreidinger a few days ago. So, the PR didn't get updated. I'll close it not to confuse others with that.

PR #349, opened 2015-10-20
Dropped / moved some of old testsuites

mchf:
providing it for review before it gets to big. However, dropping other old testsuites is WIP.


src/include/network/routines.rb
```diff
@@ -461,37 +461,27 @@ def DistinguishedName(name, hwdevice)
...
 
-      vendor = Ops.get_string(hwdevice, "sub_vendor", "")
-      dev = Ops.get_string(hwdevice, "sub_device", "")
+      vendor = hwdevice["sub_vendor"] || ""
+      dev = hwdevice["sub_device"] || ""
 
-      if vendor == "" || dev == ""
-        vendor = Ops.get_string(hwdevice, "vendor", "")
-        dev = Ops.get_string(hwdevice, "device", "")
+      if vendor.empty? || dev.empty?
```
jreidinger: it would be nice to have comment why sub_vendor have precedence before vendor

test/bond_test.rb
```diff
@@ -0,0 +1,84 @@
...
+        expect(
+          LanItems
+            .GetBondableInterfaces(LanItems.GetCurrentName)
+            .map { |i| LanItems.GetDeviceName(i) }
+        ).to match_array expected_bondable
+      end
+    end
+
+    describe "#GetBondSlaves" do
+      it "returns list of slaves if bond device has some" do
```
jreidinger: maybe better to use "when" instead of "if"

test/bond_test.rb
```diff
@@ -0,0 +1,84 @@
...
+        expect(LanItems.GetBondSlaves("bond0")).to match_array ["eth1", "eth2"]
+      end
+
+      it "returns empty list if bond device doesn't have slaves assigned" do
```
jreidinger: same here

test/routines_test.rb
```diff
@@ -39,3 +39,45 @@ def initialize
...
+    expect(routines.ValidNicName("eth:0")).to be true
+  end
+
+  it "fails in case of long name" do
+    expect(routines.ValidNicName("0123456789012345")). to be false
+  end
+
+  it "fails when it contains invalid character" do
+    expect(routines.ValidNicName("eth0?")).to be false
+  end
```
jreidinger: missing test case that it fails also for empty device.

test/routines_test.rb
```diff
@@ -39,3 +39,45 @@ def initialize
...
+    expect(routines.DeviceName({})).to be_empty
+  end
+
+  it "returns description containing model name when known" do
+    expect(routines.DeviceName(hwinfo.merge("model" => "hw_model"))).to eql "hw_model"
+  end
+
+  it "returns description build from vendor and device details when model name is uknown" do
+    expect(routines.DeviceName(hwinfo)).to eql "hw_vendor hw_device"
+  end
```
jreidinger: missing test case when only one of sub_ is defined, so vendor and device is used.

mchf: I've somehow expected comments like this. But I had to try to pass with only converting old testsuites ;-) ok, ok will extend it too ;-)

src/include/network/routines.rb
```diff
@@ -461,37 +299,27 @@ def DistinguishedName(name, hwdevice)
...
+      "#{vendor} #{dev}".strip
     end
 
     # Validates given name for use as a nic name in sysconfig. See bnc#784952
     def ValidNicName(name)
       # 16 is the kernel limit on interface name size (IFNAMSIZ)
-      return false if !Builtins.regexpmatch(name, "^[[:alnum:]._:-]{1,15}$")
-
-      true
+      !(name =~ /^[[:alnum:]._:-]{1,15}$/).nil?
```
mvidner: `^` and `$` match LINES, not the entire string, in [Ruby regexes](http://ruby-doc.org/core-2.0.0/Regexp.html#class-Regexp-label-Anchors), so `"eth0\n" `would match. Use `\A` and `\z`.

kobliha: This is ideally covered by a unit test

mvidner:
Thanks! Just one thing about Regexp, otherwise LGTM so far.

src/include/network/routines.rb
mchf: Thanks. I've fixed the regexp and added testsuite for this case. These differences in regexp will never stop biting me.

mchf:
Thanks for review. If travis succeeds I'll merge it. 

Subsequent conversions will follow later - this one is getting pretty big.

jreidinger:
I propose to not merge it till we have SP1 branch. Reason is simple. Broken testsuite -> no submit request and it can block some P1 or shit stopper. Common problem I see is inappropriate arch mocking, which result e.g. in s390 build failing, so even jenkins do not discover it

mchf:
well, yast2-networ has sp1 branch. however ok, i'll postpone merging. nothing urgent here

mchf:
ok, branching is done - thanks to @ancorgs. Lets test it and merge ... ;-)

PR #350, opened 2015-11-04
Fix AutoYaST schema

imobach:
Add missing firewall element to AutoYaST schema. Related to [bsc#805275](https://bugzilla.suse.com/show_bug.cgi?id=805275).


src/config/networking.rnc
```diff
@@ -69,7 +69,9 @@ interface =
     wireless? &
     wifi_settings? &
 
-    dhclient_set_down_link?
+    dhclient_set_down_link? &
+
+    element firewall { text }?
```
mvidner: IIUC the contents should be `"yes" | "no"`. I could only find it in ifcfg-lo so I suspect it is obsolete anyway.

imobach:
Yes, you're right. It looks like obsolete and it's limited to `yes` and `no`. I could find it in `/etc/sysconfig/network/config` in SLE11.

mvidner:
LGTM

PR #351, opened 2015-11-04
Adapt Rakefile to submit_to

ancorgs:


jreidinger:
LGTM...to be honest, I think you can do mass change in all repos, as it is easier that creating so many reviews.

PR #352, opened 2015-11-06
Clear hostname from /etc/hosts when deleting nic config

mchf:
Bnc#951330


src/modules/LanItems.ycp
```diff
@@ -1831,28 +1832,49 @@ global map FindMatchingDevice(map interface) {
...
- }
- SetModified();
+    string ifcfg = Items[current, "ifcfg"]:"";
+    string hwcfg = Items[current, "hwcfg"]:"";
+
+    if (IsNotEmpty(ifcfg))
+    {
+        NetworkInterfaces::Delete(ifcfg);
+
+        // this depends on above NetworkInterfaces::Delete
```
mvidner: I don't understand this. Wouldn't it be much better to first `NetworkInterfaces::GetValue(ifcfg, "IPADDR")` and only then `NetworkInterfaces::Delete(ifcfg)`? Like, accessing memory *before* freeing it?

mchf: ```NetworkInterfaces::Delete``` sets ```NetworkInterfaces::current``` internally, which is used in ```GetValue```. Moreover, you want to delete ifcfg even when no IP is present (e.g. dhcp config) so it doesn't hurt.

mvidner: > NetworkInterfaces::Delete sets NetworkInterfaces::current internally, which is used in GetValue

While both statements are true, they don't imply that `GetValue` needs `Delete`. [`GetValue`](https://github.com/yast/yast-yast2/blob/7b9adc08fe956778c9f112a0c095a089c6478194/library/network/src/NetworkInterfaces.ycp#L1567) has ifcfg as an argument, and does call `Select` on it.

> Moreover, you want to delete ifcfg even when no IP is present (e.g. dhcp config) so it doesn't hurt.

Sure, I mean like this:
```cpp
+        // no comment needed here
+        string ip = NetworkInterfaces::GetValue(ifcfg, "IPADDR");
+        if(IsNotEmpty(ip))
+        {
+            y2milestone("Deleting hostnames assigned to %1 from /etc/hosts", ip);
+            Host::SetNames(ip, []);
+        }
+        NetworkInterfaces::Delete(ifcfg);
```

mchf: hm, I've overlooked the ```Select``` in ```GetValue```, so the comment is useless.

No problem to add another level of ```if``` is no problem, so I'll adapt it.

mvidner: Again I don't understand. I thought we would only move the Delete to a later point, not introducing any new `if`.

mchf: all this is closed in ```if(ifcfg)```. I'm just checking if it is secure to drop this one (seems so). I'm a bit paranoid due to unexpected side effects. That's why i talk about second if above.

mvidner:
LGTM

mchf:
Thanks for review

PR #354, opened 2015-11-06
Merge #347 to SP1

mvidner:
#347 is 'Fixed "AutoYaST fails to configure network" [bsc#949193](https://bugzilla.suse.com/show_bug.cgi?id=949193)'

NOTE that due to changes done in the development branch (master AKA SLE-12-SP1) in the meantime, the original bug bsc#949193 is not present, as evidenced by this new test:
yast/autoyast-integration-test#42

It also seemed that a different bug ([bsc#874259#c157](https://bugzilla.suse.com/show_bug.cgi?id=874259#c157)) was present but my testing did not confirm that.

Therefore no version+changelog entry, just merging forward.

### DIFF
--- a/src/clients/lan_auto.rb
+++ b/src/clients/lan_auto.rb
@@ -88,13 +88,13 @@ module Yast
       elsif @func == "Change"
         @ret = LanAutoSequence("")
       elsif @func == "Import"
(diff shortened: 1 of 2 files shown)


jreidinger:
LGTM

PR #355, opened 2015-11-09
Fix AutoYaST schema for SLE11 SP4

imobach:
This is a port of https://github.com/yast/yast-network/pull/350/files for SLE11 SP4 which is also affected by [bsc#805275](https://bugzilla.suse.com/show_bug.cgi?id=805275).


mvidner:
The code looks good, but I think that it would be better to merge SP3 into SP4. This was last done in May. @mchf or are there reasons against it?

mchf:
@mvidner: AFAIK no

imobach:
@mvidner I've merged SP3 into SP4 branch in https://github.com/yast/yast-network/pull/356. I'm closing the current PR.

PR #359, opened 2015-11-09
Do not set links down during ssh (auto)installation

mchf:
bnc#944349

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Sun Nov  8 20:55:53 UTC 2015 - mfilka@suse.com
+
+- bnc#944349
+  - do not set links down when running AY installation over ssh.
+    SSH installation do not freeze.
+- 3.1.134
+
+-------------------------------------------------------------------
 Thu Oct  8 12:02:44 UTC 2015 - mvidner@suse.com
 
 - Fixed "Relax-NG parser error : Some defines for ipv6_forward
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Sun Nov  8 20:55:53 UTC 2015 - mfilka@suse.com
+
+- bnc#944349
+  - do not set links down when installing over network
```
mvidner: The code looks good, but we should improve the explanation:

- "when installing over network" is too broad. HTTP? FTP? VNC?
- This describes *what* is done. But as we are changing a notoriously complex thing, we should say *why*.

src/modules/LanUdevAuto.rb
```diff
@@ -205,7 +207,7 @@ def Write
       end
 
       if !rules.empty? && AllowUdevModify()
-        SetAllLinksDown()
+        SetAllLinksDown() if !Linuxrc.usessh
```
jreidinger: so vnc is not affected?

mchf: it is. But this is a quick fix which breaks another use case so I've limited that only on reported case.  Another improvement in this patch is a must.

mvidner: I *guess* it is also affected but I would not mention it if we haven't *tested* it.

mvidner:
LGTM

src/modules/LanUdevAuto.rb
jreidinger: one part is to not mention it, second one is let here bug

jreidinger: but if release managers are fine with it, then at least please add TODO comment capturing your findings

PR #360, opened 2015-11-09
Another drops from old testsuite

mchf:


src/include/network/lan/udev.rb
```diff
@@ -71,10 +72,9 @@ def update_udevd
 
     # Removes (key,operator,value) tripplet from given udev rule.
     def RemoveKeyFromUdevRule(rule, key)
-      rule = deep_copy(rule)
-      pattern = Builtins.sformat("%1={1,2}[^[:space:]]*", key)
+      pattern = /#{key}={1,2}[^[:space:]]*/
```
jreidinger: usually safer is to use `Regexp.escape(key)`

jreidinger: also this part `[^[:space:]]` doesn't look like ruby regexp. What it actually should catch? something like `\S` ?

src/include/network/lan/udev.rb
```diff
@@ -82,13 +82,11 @@ def RemoveKeyFromUdevRule(rule, key)
     # Tripplet is given as a string in form KEY="VALUE" or
     # MATCHKEY=="MATCHVALUE"
     def AddToUdevRule(rule, tripplet)
-      rule = deep_copy(rule)
-      if !Builtins.regexpmatch(tripplet, ".+={1,2}\".*\"")
-        return deep_copy(rule)
-      end
+      return rule unless tripplet =~ /.+={1,2}\".*\"/
+
       rule = [] if rule.nil?
```
jreidinger: better `rule ||= []`

src/include/network/lan/udev.rb
```diff
@@ -82,13 +82,11 @@ def RemoveKeyFromUdevRule(rule, key)
...
 
-      Builtins.add(rule, tripplet)
+      rule << tripplet
```
jreidinger: be aware that here you return reference to rule passed as argument, so it slightly change behavior. So just ensure it do not break anything. Problem can be e.g. something like this

```
RULE="c=d"

# old behavior
res = AddToUdevRule(RULE, "a=b")
res = AddToUdevRule(RULE, "a=b")
# res will be "c=d c=d a=b" and RULE will be "c=d"
# in new behavior res and RULE will be "c=d c=d a=b a=b"
```

jreidinger:
in general nice improvement, only few minor notes and one possible cave eat :+1:

src/include/network/lan/udev.rb
mchf: thanks for pointing that.I was unsure of this change. I've already checked all occurrences in the network code and it should work fine. However as I think of it now, I'll return to old behavior as it is probably more useful.

jreidinger: old behavior is `rule + [tripplet]`

mchf: I know, I'll update it. Thanks

jreidinger:
in general LGTM

PR #361, opened 2015-11-13
Fix AutoYaST schema in SLE 12 GA

imobach:
I'm adding the `FIREWALL` element because I found it in the `ifcfg-lo` in a SLE 12 machine.

Fixes [bsc#954412](https://bugzilla.suse.com/show_bug.cgi?id=954412) for this module.


schubi2:
LGTM

PR #363, opened 2015-11-16
Device renaming during ssh|vnc AY installation

mchf:
bnc#944349

the most important part of the patch is moving device renaming from 2nd into 1st stage

Do not merge yet, still under testing.

### DIFF
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -299,10 +299,10 @@ module Yast
         key = rule["value"].downcase
         # currently we're interrested only on those interfaces which are already
         # configured - such interfaces cannot be restarted during second stage
-        matching_item = LanItems.Items.find { |_, i| i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key }
+        _, matching_item = LanItems.Items.find { |_, i| i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key }
         next if !matching_item
 
-        name_from = matching_item[1]["ifcfg"]
+        name_from = matching_item["ifcfg"]
 
         log.info("- renaming <#{name_from}> -> <#{name_to}>")
 
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -49,6 +49,7 @@ module Yast
       Yast.import "Storage"
       Yast.import "LanItems"
       Yast.import "Profile"
+      Yast.import "Linuxrc"
 
       Yast.include self, "network/routines.rb"
       Yast.include self, "network/complex.rb"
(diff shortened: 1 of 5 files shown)


src/clients/save_network.rb
```diff
@@ -189,14 +190,14 @@ def copy_udev_rules
         log.info("File #{udev_rules_destdir} exists")
       end
 
-      if !FileUtils.Exists(net_destfile)
+      if !Mode.update
```
kobliha: This deserves a comment, why you copy this file only during Update?

kobliha:
RuboCop complains

src/clients/save_network.rb
mchf: I do not copy it during update to avoid overwriting user changes

mchf:
I've finished last of "basic tests" and everything seems to be fine.

src/clients/save_network.rb
```diff
@@ -263,6 +266,63 @@ def configure_target
...
+    #
+    # Intended mainly for steps which cannot be done in AY's second stage
+    #
+    # FIXME: Currently used only for applying udev rules during network
+    # installations (ssh, vnc, ...). It was introduced as a quick fix for
+    # bnc#944349, so it is currently limited only on {ssh|vnc} installations.
+    # Once properly analyzed and tested then starting of whole network second
+    # stage can be moved here.
+    def ay_mode_configuration
+      # TODO: run only when in {ssh|vnc} mode of AY
```
jreidinger: why is there todo, when code below do it?

src/clients/save_network.rb
```diff
@@ -263,6 +266,63 @@ def configure_target
...
+
+      LanItems.Read
+
+      udev_rules.each do |rule|
+        name_to = rule["name"]
+        attr = rule["rule"]
+        key = rule["value"].downcase
+        # currently we're interrested only on those interfaces which are already
+        # configured - such interfaces cannot be restarted during second stage
+        matching_item = LanItems.Items.find { |_, i| i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key }
```
jreidinger: hint is to use `index, matching_item = ...` this way below you do not need to use nasty `matching_item[1]`

src/clients/save_network.rb
mchf: forgot to drop it ;-)

jreidinger:
few NP, in general looks fine for me.

src/modules/LanItems.rb
```diff
@@ -645,7 +645,7 @@ def WriteUdevItemsRules
           # - remove it completely
           # removing is less error prone when tracking name changes, so it was chosen.
           item_udev_net = RemoveKeyFromUdevRule(item_udev_net, "KERNEL")
-          SetLinkDown(dev_name)
+          SetLinkDown(dev_name) if !Mode.autoinst
```
schubi2: I would put here also an comment, that AutoYaST would have here a problem if it used via SSH installation.

kobliha:
Merging, we need it NOW

PR #366, opened 2015-11-23
Initialize Routing when not done before Exporting. 

mchf:
bnc#956012

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Nov 23 09:54:34 UTC 2015 - mfilka@suse.com
+
+- bnc#956012
+  - When exporting AutoYaST profile, it contains route definitions
+    even when these routes were configured using linuxrc.
+- 3.1.137
+
+-------------------------------------------------------------------
 Mon Nov 16 21:33:18 UTC 2015 - mfilka@suse.com
 
 - bnc#955217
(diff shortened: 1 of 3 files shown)


src/modules/Routing.rb
```diff
@@ -391,6 +394,12 @@ def Import(settings)
     def Export
       exproute = {}
 
+      # It should be case only for installer (1st stage). When routing
+      # was configured via linuxrc, yast needn't to be aware of it. bnc#956012
+      Routing.Read if !@initialized
```
jreidinger: simple `Read() if !@initialized` is enough as it is same module

jreidinger:
just NP, LGTM

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Nov 23 09:54:34 UTC 2015 - mfilka@suse.com
+
+- bnc#956012
+  - routing configuration exported at the end of first stage even
+    when configured via linuxrc
```
mvidner: It is unclear whether this describes the bug or the fix :) Also, can you please mention the user visible symptoms?

kobliha:
@mchf @schubi2 PLS add this use-case to AutoYast autotest

mvidner:
LGTM

mchf:
Thanks for reviews

PR #367, opened 2015-11-25
Clear hostname from /etc/hosts when deleting nic config

mchf:
Ported patch https://github.com/yast/yast-network/pull/1
Ported patch for bnc#951330

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -66,7 +66,7 @@ module Yast
     # Remove all entries from the host table.
     def clear
       @hosts = {}
-      modified = true
+      @modified = true
     end
 
     # @return [hash] address->list of names
@@ -82,7 +82,7 @@ module Yast
     # Give address a new list of names.
     def set_names(address, names)
       @hosts[address] = names
-      modified = true
+      @modified = true
     end
 
     # Add another name to the list for address (which may be empty so far)
@@ -90,7 +90,7 @@ module Yast
       @hosts[address] = [] if @hosts[address].nil?
       @hosts[address] << name
 
-      modified = true
+      @modified = true
     end
 
     def NeedDummyIP
@@ -371,6 +371,7 @@ module Yast
     def GetModified
       @modified
     end
+
     # Function sets internal variable, which indicates, that any
     # settings were modified, to "true"
     def SetModified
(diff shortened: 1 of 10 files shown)


src/modules/Host.rb
```diff
@@ -57,12 +57,39 @@ def main
...
+
+    # Give address a new list of names.
+    def set_names(address, names)
+      @hosts[address] = names
+      modified = true
+    end
+
+    # Add another name to the list for address (which may be empty so far)
+    def add_name(address, name)
+      @hosts[address] = [] if @hosts[address].nil?
```
jreidinger: `@hosts[address] ||= []`

src/modules/Host.rb
```diff
@@ -349,9 +376,7 @@ def SetModified
       nil
     end
 
-    publish :variable => :hosts, :type => "map <string, list <string>>"
     publish :variable => :modified, :type => "boolean"
-    publish :variable => :write_only, :type => "boolean"
```
jreidinger: this removes also accessor, do you really check all SLE12 modules? It is API incompatible change, which is not nice in update branch

src/modules/Host.rb
```diff
@@ -57,12 +57,39 @@ def main
...
-      # Only write configuration
-      @write_only = false
-
       @initialized = false
     end
 
+    # Remove all entries from the host table.
+    def clear
+      @hosts = {}
+      modified = true
```
mvidner: If we replace all `modified = true` with `@modified = true` then we can remove the only remaining`publish :variable`.
It is that way in the original patch. Thanks for porting it, it looks like I missed to merge it back then :-/

src/modules/Host.rb
mvidner: I have double checked that these variables are indeed not used in other packages.

src/modules/Host.rb
jreidinger: btw this is wrong. it have to be `self.modified = true` otherwise it is assigned to local variable and lost.

mchf: I've obviously missed the modification mentioned by @mvidner above. I'll update the PR

mvidner:
LGTM

mchf:
Thanks for reviews

PR #368, opened 2015-11-25
Write udev rules in non-{ssh/vnc} AY installation

mchf:
bnc#956605


jreidinger:
LGTM. just do not forget to really do something with that fixme ;)

mchf:
Thanks.

The fixme already has its PBI and I expect it will be planed for next sprint.

schubi2:
LGTM too. I will start the testing for the mentioned bugs too.

schubi2:
I do not see Micha in the IRC. So I am merging it for him and submitting the package.

PR #365, opened 2015-11-20
Fix AutoYaST schema in SLE 12 SP1

imobach:
Fixes [bsc#954412](https://bugzilla.suse.com/show_bug.cgi?id=954412) for SLE 12 SP1. BTW, `SLE-12-GA` branch is merged into `sle-12-sp1-after_release`.

PR #369, opened 2015-11-30
Merged SLE-12 branch

mchf:


mvidner:
So how does this one relate to #365 which is in an "after_release" branch?

src/include/network/lan/complex.rb
```diff
@@ -443,26 +443,24 @@ def handleOverview(_key, event)
...
-            "STARTMODE"
-            ) == "nfsroot"
-            if !Popup.YesNoHeadline(
-              Label.WarningMsg,
-              _("Device you select has STARTMODE=nfsroot. Really delete?")
-              )
-              # y2r: cannot break from middle of switch
-              # but in this function return will do
-              return nil
+            return :edit
```
jreidinger: indentation here looks wrong, also in whole file

mchf:
Well, I'm not author of "after release" branch, so this one duplicates the content (I'll ask @imobach to check it). As there is update project already open in IBS I don't see a reason for "after release branch"

src/include/network/lan/complex.rb
mchf: yes it is. Indentation is fixed in SLE-12 branch but not in SP1. So it looks weird. I expect rubocop will complain, so I'll fix it with separate commit.

imobach:
It's fine with me. I've created `sle-12-sp1-after_release` branches in some repositories following the agreement on ML (`SLE12-SP1: Post-RC2 Behavior`). We should take this into account when merging that branches to avoid duplications in changes file.

mvidner:
LGTM

mchf:
thanks for reviews

PR #370, opened 2015-12-02
Dropped / moved some of old testsuites

mchf:


test/Makefile.am
```diff
@@ -28,6 +28,7 @@ TESTS = \
   suse_firewall_4_network_test.rb \
   udev_test.rb \
   widgets_test.rb \
+  host_test.rb \
```
mvidner: NP: keep the list in alphabetical order

src/modules/Host.rb
```diff
@@ -207,14 +207,13 @@ def Import(settings)
...
     def Export
-      if Ops.greater_than(Builtins.size(@hosts), 0)
-        # Filter out IPs with empty hostname (so that valid autoyast
-        # profile is created)(#335120)
-        @hosts = Builtins.filter(@hosts) { |_ip, names| names != [] }
-        return { "hosts" => Builtins.eval(@hosts) }
-      else
-        return {}
-      end
+      return {} if hosts.empty?
```
mvidner: Please add a test case for this too.

src/modules/Host.rb
```diff
@@ -207,14 +207,13 @@ def Import(settings)
...
+
+      # Filter out IPs with empty hostname (so that valid autoyast
+      # profile is created)(#335120)
+      hosts.keep_if { |_, names| !names.empty? }
```
mvidner:
```rb
    it "removes empty name lists" do
      allow(host).to receive(:hosts).and_return({"127.0.0.1" => ["localhost"], "10.0.0.1"  => []})
      expect(host.Export).to eql("hosts" => {"127.0.0.1" => ["localhost"]})
    end
```

jreidinger: this creates quite nasty side-effect which result that Export modify hosts element, which looks wrong from my POV.

mchf: yes, it is ugly, but it was so even before. See:
```
@hosts = Builtins.filter(@hosts) { |_ip, names| names != [] }
```

jreidinger: hmm, so please add at least fixme to remove such side effect.

mvidner:
LGTM.

src/modules/Host.rb
```diff
@@ -207,14 +207,15 @@ def Import(settings)
...
-        return { "hosts" => Builtins.eval(@hosts) }
-      else
-        return {}
-      end
+      return {} if hosts.empty?
+
+      # Filter out IPs with empty hostname (so that valid autoyast
+      # profile is created)(#335120)
+      # FIXME: this also removes records with empty names from @hosts. Such
+      # side effect is unexpected and should be removed.
```
mvidner: A link to the original bug, in which  alternative approaches are discussed: https://bugzilla.suse.com/show_bug.cgi?id=335120

mchf: thanks for the reference

mchf:
Rebased.
Thanks for reviews. I'm going to merge it once travis is done.

PR #372, opened 2015-12-14
Export DNS config at the end of installation even when provided via linuxrc

mchf:
[bsc#957377](https://bugzilla.suse.com/show_bug.cgi?id=957377)

### DIFF
--- a/src/modules/DNS.rb
+++ b/src/modules/DNS.rb
@@ -516,6 +516,11 @@ module Yast
     # @return autoinstallation settings
     def Export
       expdns = {}
+
+      # It should be case only for installer (1st stage). When resolver / hostname
+      # was configured via linuxrc, yast needn't to be aware of it. bnc#957377
+      Read() if !@initialized
+
       if Ops.greater_than(Builtins.size(@hostname), 0)
         Ops.set(expdns, "hostname", @hostname)
       end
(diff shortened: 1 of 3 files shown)


src/modules/DNS.rb
```diff
@@ -516,6 +516,11 @@ def Import(settings)
     # @return autoinstallation settings
     def Export
       expdns = {}
+
+      # It should be case only for installer (1st stage). When routing
+      # was configured via linuxrc, yast needn't to be aware of it. bnc#956012
```
mvidner: It seems this confusing comment was copied unchanged from #366.

src/modules/DNS.rb
```diff
@@ -516,6 +516,11 @@ def Import(settings)
...
+      Read() if !@initialized
+
```
mvidner: BTW! Since `expdns` is not `nil` and the indices are literals, those `Ops.set` below can be autocorrected. I only need to tell Rubocop-YaST to consider `Ops.set`. Yay automation!

mchf: good to know. This time I was too lazy to do that by hand ;-)

mvidner:
LGTM

mchf:
Thanks for the review

PR #374, opened 2015-12-15
Merged SLE-12-SP1

mchf:


Rakefile
```diff
@@ -1,5 +1,7 @@
 require "yast/rake"
 
+Yast::Tasks.submit_to :sle12sp2
```
jreidinger: this is wrong. there should not be anything, otherwise public jenkins also send it to SP2

mchf: Thanks. Obviously I live in SLE world only (master == SLE12SP2) :-/

mvidner:
Travis failed because in the two branches we have refactored the code and the tests independently and Git is not smart enough to merge that.
But I see the last commit corrects that, good.

jreidinger:
LGTM

mvidner:
LGTM

mchf:
Thanks for the reviews.

PR #376, opened 2015-12-22
fix exporting hosts (found by openQA)

jreidinger:


imobach:
LGTM

ancorgs:
LGTM. That was a nice example on how too much mocking is evil

jreidinger:
@ancorgs it is nice example why we should use mocking only real calls like instance double

mvidner:
@jreidinger @imobach @ancorgs sorry, this does not look good to me.
Cc @mchf 

I came across this because I suspect we have divergent code in SP1 and master, yet labelled with the same version numbers *and* change files. While most of the differences are in the test suite, this looks like a substantial fix.
But how can I know?? There is no Bugzilla reference, not even an openQA reference. No .changes entry? Not even a PR description? What is the user impact of the bug? Is a maintenance update needed?

You might have done a good job on the coding front, but the bookkeeping aspect is absolutely awful.

mvidner:
OK. Whatever information was there in openQA has been already rotated away, the last logs I see are from Jan 10, 2 weeks old.
This seems to be a hotfix after #374, a big merge from SLE12-SP1 to master, so it is unlikely it even got as far as Tumbleweed users.

PR #377, opened 2016-01-12
Cleanup in old testsuite

mchf:
dropped unused installation client.


src/modules/InternetDevices.rb
```diff
@@ -1,116 +0,0 @@
-# encoding: utf-8
```
mvidner: But what about this use? https://github.com/yast/yast-live-installer/blob/4a62460f1a8abaea00be50e671c719711c430a57/src/clients/inst_live_prepare_net_test.rb#L56

mvidner: It appears unused in the control files so we can probably drop that too and cross reference the PR here.

mvidner:
Also drop https://github.com/mchf/yast-network/blob/old_testsuite_drops/src/installation/manual-tests/test-net-test.rb

src/modules/InternetDevices.rb
jreidinger: yes, it should be also dropped

mchf: I didn't find any control file (except one testing) which mention the client. I've checked mainly SLES and OpenSUSE ones. But I feel I don't have a good overview of all our products. So any ideas what should be checked are welcome.

mchf:
@mvidner thanks for pointing manual tests. I always forget of them :-( I've only grepped for function names not client :-/

Manual test was dropped too.

src/modules/InternetDevices.rb
```diff
@@ -1,116 +0,0 @@
...
-      nil
-    end
-
-    # Find the fastest connection to the Internet
-    # @return true if a "good" connection was found
-    def FindFastest
-      Reset()
-
-      NetworkInterfaces.Read
-      Internet.device = NetworkInterfaces.Fastest
```
teclator: It seems that NetworkInterfaces.Fastest method is not using anymore so maybe could be also droped?

mchf: Short answer is:
If it is not used anymore it can. According my quick search it should be safe.

A bit longer one:
There is a problem how to find if it is not used anymore. ```NetworkInterfaces.Fastest``` is part of public API of ```NetworkInterfaces``` module which is part of ```yast2.rpm```. So, it can be used by any of more than 100 yast modules or e.g. in a script in OES and so on. 

In general, cleaning of the network code is very needed and I'm very happy when an artifact is removed. But you never know where is a skeleton hidden. I was already punched couple of times.

So conclusion is: I'd be happy if you remove the function but you have to double or better triple check you can do it.

mchf:
@teclator See above ;-)

src/modules/InternetDevices.rb
teclator: It is one question i was asking to ancorgs, how you could be sure that anyone is using the published interface, and is safe to remove something.

mchf: well, you check as many things as you can (it means modules under our control), try to imagine what you have seen so far (experience, e.g. OES is often source of troubles in my case) ...

... and last but not least you have to have a luck. We usually do such change in early stage of development - like in present situation when we start SLE-12-SP2 development in master branch right now.

mchf:
Also see https://github.com/yast/yast-live-installer/pull/5

mvidner:
Please make a PR (an unfinished one is OK) for live installer to drop that similar client and reference it here before merging, so that we don't forget and end up referencing removed API.

Otherwise LGTM.

mchf:
Thanks for reviews.

@mvidner thanks for pointing the live-installer one. I wouldn't find it otherwise.

PR #378, opened 2016-01-14
Replace old testsuite for NetworkStorage by rspec ones.

teclator:
- Implemented the old testsuite tests on rspec and added some corner cases.
- Rewrited the getDevices method using SCR agents and removing dependency on grep.


src/modules/NetworkStorage.rb
```diff
@@ -39,21 +42,23 @@ def main
     # Ask /proc/mounts what device a mount point is using.
     # @return e.g. /dev/sda2 (or just "nfs")
     def getDevice(mount_point)
-      cmd = Builtins.sformat(
```
jreidinger: in general if you write tests and do refactoring it is good to split it into two commits, so it is easily verifiable that tests passed even before refactoring

jreidinger: just NP

src/modules/NetworkStorage.rb
```diff
@@ -39,21 +42,23 @@ def main
...
-               else
-                 Ops.get(fields, 0, "")
-               end
-      Builtins.y2milestone("%1 is on device: %2", mount_point, device)
-      device
+      log.info "The mount_point is #{mount_point}"
+      out = SCR.Read(path(".proc.mounts")).find do |m| 
+        m["file"] == mount_point && m["vfstype"] != "rootfs" 
+      end 
+      if out
```
jreidinger: better use quick exit, it is trivial like here:
```
return "" unless out

...
```

test/network_storage_test.rb
```diff
@@ -0,0 +1,35 @@
...
+Yast.import "NetworkStorage"
+
+
+describe Yast::NetworkStorage do
+  describe ".getDevices" do
+    let(:data_dir) { File.join(File.dirname(__FILE__), "data") }
+    
+    around do |example| 
+      change_scr_root(File.join(data_dir, "scr_root"), &example)
+    end
```
jreidinger: nice :+1:

ancorgs:
You offended Rubocop

mchf:
test is much more pretty now ;-)
LGTM

PR #379, opened 2016-01-22
Added rspect tests for YaPI::NETWORK

teclator:
I added unit tests for Write and Read that was covered by the old testsuite tests.

To cover the 100% of the old tests others unit tests should be added to the modules involved.


test/SCRStub.rb
```diff
@@ -58,3 +58,25 @@ def stub_scr_read(path_name)
     allow(Yast::SCR).to receive(:Read).with(path).and_return info
   end
 end
+
+module NetworkInterfacesStub
```
mchf: this is a bit misleading name. NetworkInterfaces is a module ... and you stub not only this module inside

mchf:
I'll continue with review but at first view it seems that you converted ```Network_YaPI``` but deleted ```YaPI_bond``` only. Is the conversion complete or it is work in progress?

test/SCRStub.rb
teclator: You are right, firstly it was only for Network Interfaces but then i also stub other parts so i will change it ;)

teclator:
Before delete Network_YAPI in Read method, interfaces parsing is sill missing. I will add it before merge and in that moment we could delete the old one and autotools files.

mvidner:
Travis fails because we have a dependency problem which we still haven't been able to fix (see yast/yast-yast2#420); but a local workaround you may want to try is cherry-picking https://github.com/yast/yast-network/commit/43dd425ec11764eff10bb6c0d01a2dc57ba25f6c .

teclator:
Thnx @mvidner i saw it, and i'm following your conversation with @lslezak trying to figure out what's happening. Thnx!

test/SCRStub.rb
```diff
@@ -58,3 +58,25 @@ def stub_scr_read(path_name)
     allow(Yast::SCR).to receive(:Read).with(path).and_return info
   end
 end
+
+module YaPINetworkStub
```
mchf: still unchanged

btw stubbing of NetworkInterfaces is so huge that it makes sense to to replace whole class by a nil object

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+
+describe Yast::YaPI::NETWORK do
+  include YaPINetworkStub
+
+  describe ".Write" do
+
+    context "with routes" do
+      # Mock the system to do not break anything
+      before do
+        allow(Yast::Routing).to receive(:Read)
```
mchf: may be that using named subjects (```subject(:routing) { Yast::Routing }```) would save you a lot of typing ;-)

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+      context "with empty gateway in the default  route" do
+        let(:route) { { "default" => { "via" => "" } } }
+
+        it "returns success" do
+          expect(subject).to eq success
+        end
+
+        it "empties the routes" do
+          expect(Yast::Routing).to receive(:Routes=).with []
+          subject
```
mchf: receiving ```Routes=``` with ```[]``` is implementation dependent. What about 
```
expect(Yast::Routing.Routes).to be empty
```

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+        end
+      end
+
+      context "with a gateway in the default route" do
+        let(:route) { { "default" => { "via" => "TheIP" } } }
+        before do
+          expect(Yast::IP).to receive(:Check4).with("TheIP").and_return valid
```
mchf: I didn't get this mocking. Why to mock such simple function as ```Check4``` for validating obviously invalid input? You can use any ip instead of the string ```TheIP```. Moreover there are predefined IPs in a RFC (if think it is RFC5737) which can be used in documentation / examples. Such IP's must not be used in real word.

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+            expect(Yast::Routing).to receive(:Routes=).with(
+              [
+                {
+                  "destination" => "default",
+                  "gateway"     => "TheIP",
+                  "netmask"     => "-",
+                  "device"      => "-"
+                }
+              ]
+            )
```
mchf: again this test is implementation dependent.

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+          end
+        end
+
+        context "if it's not a valid IP4" do
+          let(:valid) { false }
+
+          it "returns failure" do
+            res = subject
+            expect(res["exit"]).to eq "-1"
+            expect(res["error"]).to_not be_empty
```
mchf: NP: ```not_to``` is better readable.

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+          end
+
+          it "doesnt't modify the routes" do
+            expect(Yast::Routing).to_not receive(:Routes=)
```
mchf: same as above.

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+      subject { Yast::YaPI::NETWORK.Write("interface" => interface) }
+
+      before do
+        interface.keys.map do |k|
+          stub_clean_cache(k)
+        end
+      end
+
+      context("setting bootproto and startmode") do
+        let(:interface) { { "eth0" => { "bootproto" => "dhcp6", "startmode" => "onboot" } } }
```
mchf: not used ?

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+            "BOOTPROTO" => "dhcp6",
+            "STARTMODE" => "onboot"
+          )
+          stub_write_interfaces
+          subject
+        end
+      end
+
+      context("setting IP address and netmask") do
+        let(:interface) { { "eth0" => { "ipaddr" => "TheIP/#{netmask}" } } }
```
mchf: not used?

test/yapi_network_test.rb
```diff
@@ -0,0 +1,374 @@
...
+
+        context("with valid netmask in dot notation") do
```
mchf: I'd like to see also a test for invalid input.

May be I'm to demanding, but network module is in bad shape and you'll see that sometimes it crashes even on obviously invalid user's input.

mchf:
Good job. Looks promising ;-)

In general I don't like those "TheIP" mocks and also "<whatever>=".

test/SCRStub.rb
teclator: I change it from NetworkInterfacesStub to YaPINetworkStub, so it contains some methods to stub models called by YaPINetwork.

Yeah, it should better to replace it.

test/yapi_network_test.rb
teclator: Sure :P.

imobach: In this case, is `Yast::Routing` the subject? I though it was `Yast::YaPI::NETWORK`.

test/yapi_network_test.rb
teclator: Apart of the return we are testing the posible calls to others methods from the current implementation. So we are stubbing a call for <code>Routing->Routes( \@route ) with @route = []</code>

For me it seems more implementation dependent on <code>Yast::Routing.Routes</code> the one proposed than the current one.

test/yapi_network_test.rb
mchf: no. Here ```subject``` is ```Yast::YaPI::NETWORK``` as you write. But you can use named subjects to have multiple "instances"

test/yapi_network_test.rb
mchf: sure. But testing for ```=``` assumes that modification is done only by assignment. What I tried to suggest is to find a solution which should be more resistant to further refactoring, when updating ```Routes```can be done differently. Of course we must not rely on ```@Routes``` but relying on ```Routes``` is acceptable from my POV bcs it is a part of public API, a getter which can stay even in a future at least as an alias.

test/yapi_network_test.rb
imobach: Why not `let`?

test/yapi_network_test.rb
teclator: We tried to stub <code>Yast::Routing::Routes.with([])</code> but we was not able to, being called from perl code and published api.

test/yapi_network_test.rb
teclator: Same as above.

test/yapi_network_test.rb
teclator: Taking note ;)

test/yapi_network_test.rb
teclator: We are just reusing TheIP definition in <code>let(:route)</code> and is should be any valid one or a invalid one, and as you said is only a stub for Check4.

I could change it but couldn't reuse the let and expectation and should be declared twice. But it is more legible, of course.

test/yapi_network_test.rb
mchf: yes, ```let``` is better

test/yapi_network_test.rb
teclator: I just added what already was tested in old ones. At final it only parse the arguments and prepare it to <code>NetworkInterfaces->Current</code> so what i did was tested the correct parsing of them. 

It actually returns ever the same, so the output does not change even with wrong config. Just an example:
<code>
irb(main):028:0> Yast::YaPI::NETWORK.Write({"interface" => { "em1" => { "ipaddr" => "192.168.0.1" } } })
=> {"error"=>"", "exit"=>"0"}
irb(main):033:0> Yast.import 'SCR'
=> nil
irb(main):034:0> Yast::SCR.Read(Yast::Path.new('.network.value.em1.IPADDR'))
=> "192.168.0.1/192.168.0.1"
irb(main):035:0> Yast::YaPI::NETWORK.Write({"interface" => { "em1" => { "ipaddr" => "192.168.0.1.31" } } })
=> {"error"=>"", "exit"=>"0"}
irb(main):036:0> Yast::SCR.Read(Yast::Path.new('.network.value.em1.IPADDR'))
=> "192.168.0.1.31/192.168.0.1.31"
</code>

It is like the cookies monster that it whatever you give to him.

test/yapi_network_test.rb
```diff
@@ -0,0 +1,371 @@
...
+
+    context "with routes" do
+      # Mock the system to do not break anything
+      before do
+        allow(routing).to receive(:Read)
+        allow(routing).to receive(:Routes)
+        allow(routing).to receive(:Write)
+      end
+
+      subject { Yast::YaPI::NETWORK }
```
imobach: NP: You can move this line out of this context so you don't need to repeat this declaration through the whole file.

mchf:
LGTM. Finally is old testsuite over, yupiii ;-)

test/SCRStub.rb
```diff
@@ -58,3 +58,25 @@ def stub_scr_read(path_name)
...
+
+  def stub_clean_cache(device)
+    allow(Yast::NetworkInterfaces).to receive(:CleanCacheRead)
+    allow(Yast::NetworkInterfaces).to receive(:Add)
+    allow(Yast::NetworkInterfaces).to receive(:Edit).with(device).and_return false
+    allow(Yast::NetworkInterfaces).to receive(:Name).with(device).and_return false
+  end
+
+  def stub_write_interfaces
+    expect(Yast::NetworkInterfaces).to receive("Commit")
```
mvidner: NP: `:Commit` for consistency

mvidner:
Please run `rake osc:build` and see how the spec file still relies on the removed autotool files.

mvidner:
We may want to adapt the [yast_build and yast_install macros](https://github.com/yast/yast-devtools/blob/master/build-tools/rpm/macros.yast), or we can just do what [yast-cio](https://github.com/yast/yast-cio/blob/master/package/yast2-cio.spec) does.

teclator:
@mvidner ok, i will check, i had updated my credentials and `rake osc:build` is working again ;).

mchf:
I've tried to do ```rake osc:build``` this time and it works for me, so LGTM.

PR #380, opened 2016-01-22
Fix error when aborting the configuration of a WiFi interface (SLE12-GA)

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=950902
https://trello.com/c/niS3zpbD/420-5-sles-12-sp1-p2-950902-yast2-network-yast2-lan-error-when-configuring-wireless-interface

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan 22 13:24:15 UTC 2016 - mvidner@suse.com
+
+- Fixed an Internal error when aborting the configuration of a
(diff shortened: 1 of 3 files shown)


imobach:
LGTM as long as Travis is happy.

kobliha:
Nitpicking: The very same behavior three times in the same file? Does not look DRY at all. Does not look DRY at all. Does not look DRY at all.

mvidner:
@kobliha Yes, it is repetititive. The repeated piece is so small that abstracting it is of questionable benefit in general. This code area is relatively little used because NetworkManager stole its show so I concluded repetition was OK.

PR #381, opened 2016-01-22
Fix error when aborting the configuration of a WiFi interface (SLE12-SP1)

mvidner:
Merging #380.


mvidner:
The Travis failure looks like an uncaught bug in the new CWM API in yast2.rpm

imobach:
LGTM

PR #382, opened 2016-01-26
Fix error when aborting the configuration of a WiFi interface (master)

mvidner:
Merging #380 AKA #381 and adding a refactoring of the called method.


mvidner:
Thanks for insisting on tests. Now that I wrote them I have found that `getCurrentItem.delete` is stupid because it operates on a copy. Fixing.

imobach:
Just a minor comment. Otherwise, LGTM

PR #383, opened 2016-02-12
[WIP] Added subclass for recognize InfiniBand cards (bsc#903190)

teclator:


package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Feb 12 17:31:48 UTC 2016 - knut.anderssen@suse.com
+
+- bsc#903190
+  - Recognized netcards with subclass 02 as InfiniBand (ib) which
```
mchf: 1) AFAIK it should be 06
2) I'd go for more "user's" friendly description here. Something like "Improved detection of InfiniBand cards".

src/include/network/routines.rb
```diff
@@ -324,7 +324,9 @@ def ControllerType(hwdevice)
           return "atm"
         elsif subclass_id == 4
           return "isdn"
-        elsif subclass_id == 6
```
mchf: what about simplifying this terrible "elseif" beast ... I mean sometinhg like
```
elseif infiniband_subclass(subclass_id)
  return "ib"
```

src/include/network/routines.rb
```diff
@@ -300,88 +352,42 @@ def UsedNicName(name)
...
-      return "modem" if Ops.get_string(hwdevice, "subclass", "") == "Modem"
-      return "isdn" if Ops.get_string(hwdevice, "subclass", "") == "ISDN"
-      return "dsl" if Ops.get_string(hwdevice, "subclass", "") == "DSL"
+      class_id    = hwdevice["class_id"]
+      subclass_id = hwdevice["sub_class_id"]
 
-      subclass_id = Ops.get_integer(hwdevice, "sub_class_id", -1)
+      return "modem" if hwdevice["subclass"] == "Modem" || modem_controller?(class_id)
+      return "isdn"  if hwdevice["subclass"] == "ISDN" || isdn_controller?(class_id)
+      return "dsl"   if hwdevice["subclass"] == "DSL" || dsl_controller?(class_id)
```
mchf: I think, when refactoring this you should check if these returns are needed ... modem / isdn / dsl support was dropped from yast some time ago

src/include/network/routines.rb
```diff
@@ -450,7 +412,7 @@ def ControllerType(hwdevice)
     def ReadHardware(hwtype)
       hardware = []
 
-      Builtins.y2debug("hwtype=%1", hwtype)
+      log.debug "hwtype=#{hwtype}"
```
teclator: Uh .. i will revert this, to not change only parts of the method

src/include/network/routines.rb
```diff
@@ -300,88 +352,42 @@ def UsedNicName(name)
...
-          return ""
-        end
-      end
+      return network_subclass(hwdevice) if network_controller?(class_id)
+
       # exception for infiniband device
-      if Ops.get_integer(hwdevice, "class_id", -1) == 12
-        return "ib" if subclass_id == 6
-      end
+      return "ib" if infiniband_controller?(class_id) && subclass_id == 6
```
mchf: why is not this integrated into ```network_controller?``` method? Infiniband is kind of network controller ....

src/include/network/routines.rb
```diff
@@ -389,58 +395,14 @@ def ControllerType(hwdevice)
...
-          return ""
-        end
-      elsif Ops.get_integer(hwdevice, "class_id", -1) == 258
-        return "modem"
-      elsif Ops.get_integer(hwdevice, "class_id", -1) == 259
-        return "isdn"
-      elsif Ops.get_integer(hwdevice, "class_id", -1) == 276
-        return "dsl"
+      if interface_controller?(class_id)
+        log.info "CLASS 0x107" # this should happen rarely
```
mchf: the log message should be improved or dropped (?)

src/include/network/routines.rb
```diff
@@ -32,6 +32,58 @@ module NetworkRoutinesInclude
...
+        6   => "hsi",
+        7   => "qeth",
+        8   => "escon",
+        9   => "myri",
+        10  => "wlan",
+        11  => "xp",
+        12  => "usb",
+        129 => "sit"
+      }
+    }
```
mchf: I'm thinking if these constants belongs to a yaml file. These constants are defined by hw standards and putting it into the code makes the file unnecessarily long.

src/include/network/routines.rb
teclator: So if it was dropped i could drop all references for it and for example only support ReadHardware("netcard")

src/include/network/routines.rb
teclator: It could be splitted in other module that you can import. Is preferable a yaml file?

mchf: I was thinking of yaml file bcs you put it there and forgot of it. That is exactly what I wanted to achieve. However, if you have any other idea, I've no problem with that.

src/include/network/routines.rb
```diff
@@ -300,199 +296,69 @@ def UsedNicName(name)
...
-      # Nothing was found
-      Builtins.y2error("Unknown controller type: %1", hwdevice)
-      ""
+      ret || ""
     end
 
-    # Read HW information
-    # @param [String] hwtype type of devices to read (netcard|modem|isdn)
+    # Read HW netcards information
+    # @param [String] not needed anymore
```
jreidinger: I think correct comment is do nothing unless empty or "netcard".

src/include/network/routines.rb
```diff
@@ -513,138 +379,81 @@ def ReadHardware(hwtype)
...
+        # drivers:
+        # Although normally there is only one module
+        # (one=$[active:, module:, options:,...]), the generic
+        # situation is: one or more driver variants (exclusive),
+        # each having one or more modules (one[drivers])
+
+        # only drivers that are not marked as broken (#97540)
+        drivers = card["drivers"].select do |d|
+          # ignoring more modules per driver...
+          first_module = d.fetch("modules", []).fetch(0, []).fetch(0, []) # [module, options]
```
jreidinger: For me the last fetch looks like bug as include? below looks like expect module name as it use split on broken_modules. so it never find [] in broken_modules. so it actually works, but it is confusing.

src/include/network/routines.rb
```diff
@@ -513,138 +379,81 @@ def ReadHardware(hwtype)
...
+          brk = broken_modules.include?(first_module)
+
+          log.info "In BrokenModules, skipping: #{first_module}" if brk
+
+          !brk
+        end
+
+        if drivers == []
```
jreidinger: drivers from above is for sure array and not nil, so `if drivers.empty?` is better

src/include/network/routines.rb
```diff
@@ -513,138 +379,81 @@ def ReadHardware(hwtype)
...
+        # store the BUS type
+        bus = card["bus_hwcfg"] || card["bus"] || ""
+
+        if bus == "PCI"
+          bus = "pci"
+        elsif bus == "USB"
+          bus = "usb"
+        elsif bus == "Virtual IO"
+          bus = "vio"
+        end
```
jreidinger: btw for such translation using constant map is nicer like
```
BUS_ID_TO_NAME = {
  "PCI" => "pci",
  "USB" => "usb",
  "Virtual IO" => "vio"
}
...
one["bus"] = BUS_ID_TO_NAME[bus] || bus

```

src/include/network/routines.rb
```diff
@@ -513,138 +379,81 @@ def ReadHardware(hwtype)
...
+
+        one["bus"] = bus
+        one["busid"] = card["sysfs_bus_id"] || ""
+        one["mac"] = resource["hwaddr"][0]["addr"] || ""
```
jreidinger: this can raise exception if hwaddr is for example empty.

src/include/network/routines.rb
```diff
@@ -513,138 +379,81 @@ def ReadHardware(hwtype)
...
+        # is the cable plugged in? nil = don't know
+        one["link"] = resource["link"][0]["state"]
```
jreidinger: same here, are we sure "link" is always there with at least one element?

src/include/network/routines.rb
teclator: +1

src/include/network/routines.rb
teclator: Yeah, maybe i was too optimistic. I tried to replace the Ops dependency

Ops.get_string(resource, ["hwaddr", 0, "addr"], "")

And it take cares of nil values, so i not realized this. 

Ruby 2.3 #dig wille be nice here ;)

src/include/network/routines.rb
teclator: Same as above.

src/include/network/routines.rb
teclator: A recurrent one? xD the

https://github.com/yast/yast-network/pull/227

src/include/network/routines.rb
teclator: Yeah, it is confusing i could avoid the last one, but also i will replace it with a method like get_first_module_name to maintain this more legible.

jreidinger: not confusing but last fetch should be `.fetch(0, "")` as first_module is string and not array. So bug, which is not shown as "" nor [] is in broken modules

teclator: Yeah, this is already fixed in local, i had to commit it, with your other sugerences, i realized when you did the first review.

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Feb 23 09:56:48 UTC 2016 - knut.anderssen@suse.com
+
+- bsc#903190
+  - Recognized netcards with subclass 07 as InfiniBand (ib) which
+  now are not filtered out.
```
mvidner: This is the same description as for 3.1.143 below?!

mvidner: I see, apparently #385 was merged because this PR was taking too long?

teclator: Exactly, the fix was merged but i will continue with the clean up, just after i finish with yast/yast-snapper#40 and yast/yast-snapper#39 i'm working on it currently.

src/data/network/pci_subclasses.yml
```diff
@@ -0,0 +1,35 @@
+---
```
mvidner: Nice, this is much more readable!
And so the problem gets exposed: where does the magic mapping come from, and what do the strings mean? We should add a comment about that.
Is it a PCI standard? A kernel thing? A hwinfo thing?

teclator: We read it from hwinfo and subclasses so for network PCI class 0x02 ("network") those are the recognized ones.

https://pci-ids.ucw.cz/read/PD

teclator:
I will create a new PR that will supersede this one.

PR #386, opened 2016-02-22
Moved setting net service into 1st stage

mchf:
bnc#955260

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Feb 22 11:52:09 UTC 2016 - mfilka@suse.com
+
+- bnc#960703
+  - Moved setting network service into 1st stage
+- 3.1.140.2
+-------------------------------------------------------------------
 Fri Jan 22 13:24:15 UTC 2016 - mvidner@suse.com
 
 - Fixed an Internal error when aborting the configuration of a
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -54,12 +54,13 @@ module Yast
     # bnc#944349, so it is currently limited only on {ssh|vnc} installations.
     #
     # @param AY profile, @see e.g. Profile.current
-    def create_udevs(ay_profile)
+    def create_udevs
+      return if !Mode.autoinst
       return if !(Linuxrc.usessh || Linuxrc.vnc)
 
       log.info("Applying udev rules according AY profile")
 
-      udev_rules = ay_profile["networking"]["net-udev"]
+      udev_rules = ay_networking_section["net-udev"]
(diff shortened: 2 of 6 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Feb 22 11:52:09 UTC 2016 - mfilka@suse.com
+
+- bnc#955260
+  - Moved setting network service into 1st stage
+- 3.1.140.2
```
mvidner: The SP1 and master versioning has diverged, but you are applying a SP1 version to master.

src/lib/network/network_autoyast.rb
```diff
@@ -35,6 +47,79 @@ def merge_configs(conf)
...
+
+      return if udev_rules.nil? || udev_rules.empty?
+
+      LanItems.Read
+
+      udev_rules.each do |rule|
+        name_to = rule["name"]
+        attr = rule["rule"]
+        key = rule["value"].downcase
+        item, matching_item = LanItems.Items.find { |_, i| i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key }
```
mvidner: This line is too long

test/network_autoyast_test.rb
```diff
@@ -148,4 +148,66 @@
...
+    Yast.import "NetworkService"
+    Yast.import "Mode"
+
+    let(:network_autoyast) { Yast::NetworkAutoYast.instance }
+
+    before(:each) do
+      allow(Yast::Mode).to receive(:autoinst).and_return(true)
+    end
+
+    def product_use_nm(nm_used)
```
mvidner: I think the spec is more readable if helpers like this are named with the stub/mock prefix, like `allow_product_use_nm`, `allow_networking_section`.

package/yast2-network.changes
mchf: ok f*ck ... bad target for merging. Thanks

src/lib/network/network_autoyast.rb
mchf: ok, I can do something with that

test/network_autoyast_test.rb
mchf: ok

mvidner:
Really [bsc#955260](https://bugzilla.suse.com/show_bug.cgi?id=955260)? "autoyast install over ssh with X-forwarding halts in second stage on restarting services" is marked as Verified since 2 months.

mvidner:
I see, the card is https://trello.com/c/SUdZnkzy/416-5-sp2-finish-moving-device-renaming-into-first-stage-in-case-of-ay-installation .
Anyway, since a PR cannot change its target branch, we should continue in a new PR.

mchf:
ok close for now and start again

PR #390, opened 2016-02-29
bnc#920089 - Changed dialog window size for entering static routes

damirporobic:
Bug 920089 - [yast2-network] - jj: The ncurses window for adding a static route is ridiculously small 

- Added MinWidth for dialog window
- Added proportional sizing for elements
- Added space between elements

PR #391, opened 2016-03-06
Changed dialog window size for entering static routes + changelog

mchf:
bnc#920089
See https://github.com/yast/yast-network/pull/390 for original PR


jreidinger:
LGTM

mchf:
Patch author do not have an objection so merging.

PR #393, opened 2016-03-22
Dropped LanUdevAuto

mchf:
Several notes, TBD:
- LanUdevAuto#Import, #Write are present in the module definition but effectively dead -> removed, so DONE
- LanUdevAuto#Export to be copy-pasted elsewhere, change references -> DONE
- call createS390Devices directly (?) when importing profile on s390 -> DONE
- drop FIXME's on refactoring LanUdevAuto -> DONE


src/lib/network/install_inf_convertor.rb
```diff
@@ -44,6 +40,11 @@ def write_netconfig
       write_global_netconfig
     end
 
+    # Reports if user asked for using biosdevname pernament device names
+    def AllowUdevModify
+      /biosdevname=1/ =~ InstallInf["Cmdline"] ? false : true
```
mvidner: IMHO more readable as
```rb
InstallInf["Cmdline"] !~ /biosdevname=1/
```

src/modules/LanItems.rb
```diff
@@ -2312,8 +2192,17 @@ def enableCurrentEditButton
       end
     end
 
-    def createS390Device
-      Builtins.y2milestone("creating device s390 network device")
+    def createS390Device(rule)
```
mvidner: A YARD comment about what **rule** is?

test/install_inf_convertor_test.rb
```diff
@@ -283,4 +283,16 @@
         .to eql "LLADDR='#{HWADDR}'"
     end
   end
+
+  describe "use_biosdevname?" do
```
mvidner: `describe "AllowUdevModify" do`, or really rename the method (with inverted logic, as the name suggests)

mchf: ah I was thinking of changing the name ... but this is only left over ;-/

src/modules/LanItems.rb
```diff
@@ -2460,6 +2349,40 @@ def drop_hosts(ip)
...
+    #
+    #  It takes a whole "interfaces" section of AY profile. Keeps
+    #  only old style named interfaces (e.g. eth-bus-0.0.0000) in it
+    #  and renames them to a new name (e.g. eth0). It also produces
+    #  a list of udev rules to guarantee device naming persistency.
+    #  The rule is base on attributes described in old style name
+    #
+    #  @param [Array] list of hashes describing interfaces in AY profile
+    #  @return [Array] list of hashes for udev rules
+    def createUdevFromIfaceName(interfaces)
```
mvidner: please also use snake_case. `create_udev_from_iface_name`

src/lib/network/network_autoyast.rb
```diff
@@ -120,6 +117,26 @@ def set_network_service
...
+    # Checks if given profile contains device names in old style format
+    #
+    # OLd style device name format is eth-id-<mac> or eth-bus-<busid>.
+    # These names has to be converted into new-style (eth0) and name persistency
+    # is enabled using udev rule which is generated later.
+    #
+    # @param [Hash] configuration map
+    #
+    # @return true if configuration contains old style name
+    def oldStyle(ay)
```
mvidner: `old_style` please

src/modules/LanItems.rb
```diff
@@ -2460,6 +2357,218 @@ def drop_hosts(ip)
...
+        end
+      else
+        Builtins.foreach(
+          Convert.convert(
+            Items(),
+            from: "map <integer, any>",
+            to:   "map <integer, map <string, any>>"
+          )
+        ) do |id, row|
+          current = id
```
mvidner: Cool, RuboCop finds a real bug! This should be `@current =` or `self.current =`

src/lib/network/network_autoyast.rb
mchf: copy - pasted function. I've kept that until I grep some other modules if it is used.
Taking note to change it once done.

src/modules/LanItems.rb
mchf: same as bellow

I was also thinking about changing name and creating an alias for old one ... But didn't do so - it seems to be useless when it was moved elsewhere anyway

src/lib/network/network_autoyast.rb
```diff
@@ -15,9 +15,11 @@ class NetworkAutoYast
...
     include Logger
 
-    Yast.import "Lan"
-    Yast.import "LanItems"
-    Yast.import "Linuxrc"
+    def initialize
+      Yast.import "Lan"
+      Yast.import "LanItems"
+      Yast.import "Linuxrc"
+    end
```
jreidinger: NP: if you add it to initialize due to circular dependencies, please mention it

src/lib/network/network_autoyast.rb
```diff
@@ -48,49 +50,30 @@ def merge_configs(conf)
...
-        # for logging only
-        name_from = matching_item["ifcfg"] || matching_item["dev_name"]
-        log.info("- renaming <#{name_from}> -> <#{name_to}>")
+      # get explicit udev definition from the profile
+      ex_udev_rules = ay_networking_section["net-udev"]
+      log.info("- explicitly defined udev rules: #{ex_udev_rules}")
 
-        # selecting according device name is unreliable (selects only in between configured devices)
-        LanItems.current = item
+      no_rules &&= ex_udev_rules.nil? || ex_udev_rules.empty?
```
jreidinger: what about using
```
ex_udev_rules = ay_networking_section["net-udev"] || []
```

src/modules/LanItems.rb
```diff
@@ -2428,6 +2325,40 @@ def createS390Device
...
+    #  @param [Array] list of hashes describing interfaces in AY profile
+    #  @return [Array] list of hashes for udev rules
+    def createUdevFromIfaceName(interfaces)
+      udev_rules = []
+      attr_map = {
+        "id"  => "ATTR{address}",
+        "bus" => "KERNELS"
+      }
+
+      interfaces.keep_if do |interface|
```
jreidinger: so you modifyc interfaces parameter with keep_if. I think it really have to be documented if you want it this way. And personally I think we should avoid such side effects of methods.

src/modules/LanItems.rb
```diff
@@ -2460,6 +2391,173 @@ def drop_hosts(ip)
       Host.set_names(ip, [])
     end
 
+    # Exports udev rules for AY profile
+    def export_udevs(devices)
```
jreidinger: my hurts pain when seeing this method. I expect it is just copy and paste, but I think at least some love it deserve.

mchf: yes it is copy paste. I gave it some love, but ENOTIME.

src/modules/LanItems.rb
mchf: well, it is described in the comment. It was the minimal change how to get rid of a global variable at the time of writing. I'll re-check if after other modifications it is possible to update it to more reasonable behavior.

src/lib/network/network_autoyast.rb
```diff
@@ -48,49 +51,30 @@ def merge_configs(conf)
     end
 
     # Creates udev rules according definition from profile
-    #
-    # FIXME: Currently used only for applying udev rules during network
-    # installations (ssh, vnc, ...). It was introduced as a quick fix for
-    # bnc#944349, so it is currently limited only on {ssh|vnc} installations.
     def create_udevs
```
mvidner: It seems this method is not covered with unit tests. Is it tested in another way?

src/lib/network/network_autoyast.rb
```diff
@@ -202,5 +186,40 @@ def ay_networking_section
 
       ay_profile["networking"]
     end
+
+    # Takes a list of udev rules and assignes them to corresponding devices.
+    #
+    # If a device has an udev rule defined already, it is overwritten by new one.
+    # Note: initialization of LanItems has to be done outside of this method
+    def assign_udevs_to_devs(udev_rules)
```
mvidner: It seems method `assign_udevs_to_devs` is not covered with unit tests. Is it tested in another way?

src/modules/LanItems.rb
```diff
@@ -2343,15 +2341,13 @@ def createUdevFromIfaceName(interfaces)
         "bus" => "KERNELS"
       }
 
-      interfaces.keep_if do |interface|
+      interfaces.each do |interface|
```
jreidinger: :+1:

jreidinger:
LGTM

src/modules/LanItems.rb
```diff
@@ -2312,8 +2198,20 @@ def enableCurrentEditButton
       end
     end
 
-    def createS390Device
-      Builtins.y2milestone("creating device s390 network device")
+    # Creates eth emulation for s390 devices
+    #
+    # @param [Hash] an s390 device description as obtained from AY profile
+    def createS390Device(rule)
```
mvidner: It seems method `createS390Device` is not covered with unit tests. Is it tested in another way?
I see it was mostly copied from another file, but mistakes can happen even with copying.

src/modules/LanItems.rb
```diff
@@ -2460,6 +2388,173 @@ def drop_hosts(ip)
       Host.set_names(ip, [])
     end
 
+    # Exports udev rules for AY profile
+    def export_udevs(devices)
+      devices = deep_copy(devices)
+      ay = { "s390-devices" => {}, "net-udev" => {} }
+      if Arch.s390
```
mvidner: This branch is not covered with unit tests. And I suspect that our integration tests do not test s390 so we should write unit tests for it.

src/modules/LanUdevAuto.rb
```diff
@@ -1,480 +0,0 @@
...
-# Package:     Network configuration
-# Summary:     Udev rules for autoinstallation
-# Authors:     Michal Zugec <mzugec@suse.cz>
-#
-#
-# Representation of the configuration of network cards.
-require "yast"
-
-module Yast
-  class LanUdevAutoClass < Module
```
mvidner: RIP!!

src/lib/network/network_autoyast.rb
mchf: yes, manually and AY integration tests.

src/lib/network/network_autoyast.rb
mchf: same as above

src/modules/LanItems.rb
```diff
@@ -2341,15 +2341,19 @@ def createUdevFromIfaceName(interfaces)
         "bus" => "KERNELS"
       }
 
+      # rubocop:disable Next
+      # the check is disabled bcs the code uses capture groups. Rewriting
+      # the code would require some tricks to access these groups
       interfaces.each do |interface|
-        next if /.*-(?<attr>id|bus)-(?<value>.*)/ !~ interface["device"]
```
mvidner: This is interesting. [The documentation](http://ruby-doc.org/core-2.2.0/Regexp.html#class-Regexp-label-Capturing) says:

> When named capture groups are used with a literal regexp on the left-hand side of an expression and the `=~` operator, the captured text is also assigned to local variables with corresponding names.

So you could keep the `next` if you used `=~` instead of `!~`:

```rb
next unless /.*-(?<attr>id|bus)-(?<value>.*)/ =~ interface["device"]
```

mchf: Capture groups are nice feature but this makes their usage a bit uncomfortable.

package/yast2-network.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Mon Apr  4 17:24:13 UTC 2016 - mfilka@suse.com
+
+- bnc#955217, bnc#956605
+  - the fix was originaly enabled only for ssh / vnc installation
+    since now is udev handling completely and always handled in
+    the first stage
```
mvidner: So there is no change for the user, right? I think description is easier to understand:

> Unified udev rule handling by always doing it in the 1st stage of installation.

mvidner:
LGTM, maybe improve the changelog message

package/yast2-network.changes
mchf: ok. at least it is less noisy and goes to the point ;-)

mchf:
Thanks for reviews

PR #394, opened 2016-04-08
Do not touch VNC arguments when enabling or disabling it.

michalsrb:
YaST disables the tigervnc encryption every time it enables or disables VNC because it sets 'securitytypes' to 'none' in its xinetd configuration file.

This was originally done to help with transition from tightvnc to tigervnc. But the 'securitytypes' should now contain different values and the tigervnc package takes care of updating the xinetd file itself. So there is no need for YaST to do it at all.


mchf:
And will take care of it always when vnc is enabled / disabled or only when installing or ...
How does it handle user's updates?
Wouldn't it be better to improve tigervnc support in yast?

michalsrb:
Improving tigervnc support in YaST would be great. For example we could add UI that would allow user to configure the security, or configure some other options in user-friendly way.

But at this moment YaST is unconditionally setting the `securitytypes` parameter to `None`, which is just wrong. The defaults are `X509None,TLSNone,None`. And if user made his own modifications, YaST overwrites them too.

My main point is that the `/etc/xinetd.d/vnc` file is a configuration file that belongs to the tigervnc package. Just like any other package, tigervnc is responsible for updating it when defaults change and handling conflicts with user modifications. It is not YaST's responsibility.

The only reason why YaST was doing it was to fix the file after switch from tightvnc to tigervnc. It should have been done by the tigervnc package back then too, but unfortunately it wasn't.

kobliha:
Is there a bug nr for this? You need in in changes file to submit it to SP2 anyway. It does not depend on the fact whether it's public or not...

michalsrb:
> Is there a bug nr for this? You need in in changes file to submit it to SP2 anyway. It does not depend on the fact whether it's public or not...

https://bugzilla.suse.com/show_bug.cgi?id=939099 (orignal, private)
https://bugzilla.suse.com/show_bug.cgi?id=966694 (duplicate, public)

mchf:
thanks for bug references. I'll play with that a bit locally and merge it today.

However, I'd create a fate request on improving YaST's behavior.

mchf:
Deprecated by PR#397

PR #395, opened 2016-04-11
Fixed internal error found by OpenQA

mchf:


src/modules/LanItems.rb
```diff
@@ -47,6 +47,8 @@ class LanItemsClass < Module
 
     include Logger
 
+    require "network/install_inf_convertor"
```
jreidinger: require should be top level call unless really needed.

jreidinger: so below line 25 I will place it

jreidinger:
LGTM

mchf:
Thanks for review

kobliha:
@mchf Hmm and does it mean that we DO NOT have a unit test for the new code?

jreidinger:
@kobliha as discussed on IRC problem is that mass execution of tests in rspec preload file earlier. I think tricks with unloading class is not necessary, as it is one time fix and should not break easily. BTW running test standalone will discover it.

mvidner:
@jreidinger Hmm! We could solve this by running the tests twice:
1. without coverage measurement, and without the preloading, to catch missing imports. 
2. with coverage measurement

PR #396, opened 2016-04-12
Fixed internal error found by AY integration tests

mchf:
bnc#974910

additional info:
boot4

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -2336,6 +2336,8 @@ module Yast
     #  @param [Array] list of hashes describing interfaces in AY profile
     #  @return [Array] list of hashes for udev rules
     def createUdevFromIfaceName(interfaces)
+      return [] if !interfaces || interfaces.empty?
+
       udev_rules = []
       attr_map = {
         "id"  => "ATTR{address}",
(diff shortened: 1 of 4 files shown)


jreidinger:
LGTM

src/modules/LanItems.rb
```diff
@@ -2336,6 +2336,8 @@ def createS390Device(rule)
     #  @param [Array] list of hashes describing interfaces in AY profile
     #  @return [Array] list of hashes for udev rules
     def createUdevFromIfaceName(interfaces)
+      return [] if !interfaces || interfaces.empty?
```
jreidinger: btw it will work if you also use `interfaces ||= []` line, but it is just matter of taste

teclator:
LGTM

mchf:
Thanks for reviews

src/modules/LanItems.rb
kobliha: It's the fastest exit from the method, which easy to understand and independent on the rest of the code even in the future

jreidinger: yes, but on other hand, sanitizing input allows you to e.g. log what happen (empty interfaces) and it is easier to see that empty interfaces is same as if there is no specific device type with id or bus identifier type.

PR #397, opened 2016-04-13
Do not touch VNC arguments when enabling or disabling it.

mchf:
Updated version of: https://github.com/yast/yast-network/pull/394


mchf:
Note: I've manually tested the patch and so far I was able to do vnc installation and establish vnc connection right after the installation.

mchf:
LGTM, however I've asked for another review too since I got involved in this patch.

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Apr 19 08:01:16 UTC 2016 - mfilka@suse.com
+
+- bnc#966694
+  - Do not touch VNC arguments when enabling or disabling it.
```
mvidner: Thanks! I would mention the symptom that we are fixing:
> bnc#966694: TigerVNC Viewer Error: No matching security types
 Do not touch...

mvidner:
LGTM

package/yast2-network.changes
mchf: ok

mchf:
@michalsrb thanks for patch

PR #398, opened 2016-04-26
Setup DNS according AY profile correctly when a static net configuration is provided to linuxrc

mchf:
[bsc#971175](https://bugzilla.suse.com/show_bug.cgi?id=971175), turning L3 patch into maintenance update

TODO: wait for maintenance request for SLE-12-SP1 confirmation

### DIFF
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -247,13 +247,14 @@ module Yast
       nil
     end
 
-    # Creates DNS configuration on target
+    # Creates target's default DNS configuration
+    #
+    # It uses DNS configuration as defined in AY profile (if any) or
+    # proposes a predefined default values.
     def configure_dns
-      if Mode.autoinst
-        NetworkAutoYast.instance.configure_dns
-      else
-        NetworkAutoconfiguration.instance.configure_dns
-      end
+      ret = false
+      ret = NetworkAutoYast.instance.configure_dns if Mode.autoinst
+      ret = NetworkAutoconfiguration.instance.configure_dns if !ret
 
       DNS.create_hostname_link
     end
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -245,15 +245,27 @@ module Yast
       nil
     end
 
+    # Creates DNS configuration on target
+    def configure_dns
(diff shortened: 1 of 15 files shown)


teclator:
In general LGTM

src/clients/save_network.rb
```diff
@@ -247,15 +247,27 @@ def copy_from_instsys
       nil
     end
 
+    # Creates DNS configuration on target
+    def configure_dns
+      if Mode.autoinst
+        NetworkAutoYast.instance.configure_dns
```
mvidner: Regarding the several sets of configuration data that YaST deals with, this method assigns the "AY profile" to the "YaST working set", but does not write it to the "persistent system config" or "current system config".
OTOH NetworkAutoconfiguration::configure_dns does write to the system.
It is inconsistent, and seems to contradict the comment above, "Creates DNS configuration on target".

src/lib/network/network_autoyast.rb
```diff
@@ -120,7 +121,36 @@ def set_network_service
       NetworkService.EnableDisableNow
     end
 
-    private
+    # Initializates DNS setup according AY profile
+    #
+    # FIXME: it currently don't write DNS configuration. It is used for initialization
+    # of DNS setup according AY profile in 1st stage as part of network setup was moved
+    # here already and some parts of network configuration needs to know it. DNS write
+    # is still done in 2nd stage.
```
mvidner: If a public method returns something, we should document what the return value means.

src/clients/save_network.rb
```diff
@@ -247,15 +247,28 @@ def copy_from_instsys
       nil
     end
 
+    # Creates target's default DNS configuration
+    #
+    # It uses DNS configuration as defined in AY profile (if any) or
+    # proposes a predefined default values.
+    def configure_dns
+      ret = false
```
mvidner: "`ret`" means "the value to be returned from this method", so the name is wrong here. `configured` instead?

mvidner:
Thank you!
LGTM

mchf:
Thanks for reviews. I'll merge it once corresponding PTF is accepted.

PR #399, opened 2016-05-02
Added dependency on yast2-proxy

mchf:
bnc#977953


mvidner:
What about something better?
That is, moving [all proxy code](https://github.com/yast/yast-network/search?utf8=%E2%9C%93&q=proxy) to https://github.com/yast/yast-proxy .

mchf:
Well, I'd propose to create a new PBI for moving rest of the proxy code. Lets PO plans it according his preferences. 

This can be a quick fix to let openQA to pass.

Acceptable?

mvidner:
OK then, LGTM.

mchf:
Thanks for review. PBI created. Merging.

kobliha:
Just a hint: In next sprint of maybe even after that, I'd like to invest some time to minimize build dependencies and cycles to speed up the build process (requested by release managers). Additionally, we should think about dependencies that make the minimal installation bigger.

PR #400, opened 2016-05-04
Support vncmanager2

michalsrb:
Add support for vncmanager mode to Remote. fate#319319

Remote module can now set up 3 VNC modes: disabled, xinetd, vncmanager
* disabled
  * vncmanager systemd service is disabled
  * vnc1 and vnchttpd1 xinetd services are disabled
* xinetd
  * vncmanager systemd service is disabled
  * xinetd systemd service is enabled
  * vnc1 and vnchttpd1 xinetd services are enabled
* vncmanager
  * vncmanager systemd service is enabled
  * xinetd systemd service is enabled
  * vnchttpd1 xinetd service is enabled but vnc1 is not

Calls to Remote.Enable and Remote.Disable set xinetd and disabled modes, so
users of the old API can keep using it.

### DIFF
--- a/src/include/network/remote/dialogs.rb
+++ b/src/include/network/remote/dialogs.rb
@@ -60,9 +60,17 @@ module Yast
           # RadioButton label
           Left(
             RadioButton(
-              Id(:allow),
-              _("&Allow Remote Administration"),
-              Remote.IsEnabled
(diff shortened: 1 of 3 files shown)


michalsrb:
Oops, didn't check with the rubocop. Will fix that first.

PR #401, opened 2016-05-04
Support vncmanager 3

michalsrb:
Add support for vncmanager mode to Remote. fate#319319

Remote module can now set up 3 VNC modes: disabled, xinetd, vncmanager
* disabled
  * vncmanager systemd service is disabled
  * vnc1 and vnchttpd1 xinetd services are disabled
* xinetd
  * vncmanager systemd service is disabled
  * xinetd systemd service is enabled
  * vnc1 and vnchttpd1 xinetd services are enabled
* vncmanager
  * vncmanager systemd service is enabled
  * xinetd systemd service is enabled
  * vnchttpd1 xinetd service is enabled but vnc1 is not

Calls to Remote.Enable and Remote.Disable set xinetd and disabled modes, so
users of the old API can keep using it.

### DIFF
--- a/src/include/network/remote/dialogs.rb
+++ b/src/include/network/remote/dialogs.rb
@@ -60,9 +60,17 @@ module Yast
           # RadioButton label
           Left(
             RadioButton(
-              Id(:allow),
-              _("&Allow Remote Administration"),
-              Remote.IsEnabled
+              Id(:allow_with_vncmanager),
+              _("&Allow Remote Administration With Session Management"),
+              Remote.IsEnabled && Remote.UsesVncManager
+            )
+          ),
+          # RadioButton label
+          Left(
+            RadioButton(
+              Id(:allow_without_vncmanager),
+              _("&Allow Remote Administration Without Session Management"),
+              Remote.IsEnabled && !Remote.UsesVncManager
             )
           ),
           # RadioButton label
@@ -142,9 +150,12 @@ module Yast
       if ret == :next
         CWMFirewallInterfaces.OpenFirewallStore(firewall_widget, "", event)
 
-        allowed = Convert.to_boolean(UI.QueryWidget(Id(:allow), :Value))
+        allow_with_vncmanager = UI.QueryWidget(Id(:allow_with_vncmanager), :Value)
+        allow_without_vncmanager = UI.QueryWidget(Id(:allow_without_vncmanager), :Value)
 
-        if allowed
+        if allow_with_vncmanager
+          Remote.EnableUsingVncManager
+        elsif allow_without_vncmanager
           Remote.Enable
         else
           Remote.Disable
--- a/src/modules/Remote.rb
+++ b/src/modules/Remote.rb
@@ -153,19 +153,15 @@ module Yast
         to:   "list <map>"
       )
 
-      xinetd_vnc1_enabled = false
-      xinetd_vnchttp1_enabled = false
-      xinetd_conf.each do |m|
-        xinetd_vnc1_enabled = m["enabled"] if m["service"] == "vnc1"
-        xinetd_vnchttp1_enabled = m["enabled"] if m["service"] == "vnchttpd1"
-      end
+      xinetd_vnc1 = xinetd_conf.find { |m| m["service"] == "vnc1" }
+      xinetd_vnc1_enabled = xinetd_vnc1 ? xinetd_vnc1["enabled"] : false
 
       log.info "#{XDM_SERVICE_NAME}: #{xdm}, DM_R_A: #{dm_ra}"
-      log.info "xinetd: #{xinetd}, VNC1: #{xinetd_vnc1_enabled}, VNCHTTP1: #{xinetd_vnchttp1_enabled}"
+      log.info "xinetd: #{xinetd}, VNC1: #{xinetd_vnc1_enabled}"
 
       vncmanager = Service.Enabled(VNCMANAGER_SERVICE)
 
-      if xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled
+      if xdm && dm_ra && xinetd
         if xinetd_vnc1_enabled && !vncmanager
           Enable()
         elsif !xinetd_vnc1_enabled && vncmanager
--- a/src/modules/Remote.rb
+++ b/src/modules/Remote.rb
@@ -36,8 +36,10 @@ module Yast
 
     XDM_SERVICE_NAME = "display-manager"
     XINETD_SERVICE = "xinetd"
+    VNCMANAGER_SERVICE = "vncmanager"
 
     PKG_CONTAINING_FW_SERVICES = "xorg-x11-Xvnc"
+    PKG_CONTAINING_VNCMANAGER = "vncmanager"
 
     GRAPHICAL_TARGET = "graphical"
 
@@ -59,8 +61,8 @@ module Yast
       # Currently, all attributes (enablement of remote access)
       # are applied on vnc1 even vnchttpd1 configuration
 
-      # Allow remote administration
-      @allow_administration = false
+      # Remote administration mode, :disabled, :xinetd or :vncmanager
+      @mode = :disabled
 
       # Default display manager
       @default_dm = "xdm"
@@ -72,26 +74,37 @@ module Yast
 
     # Checks if remote administration is currently allowed
     def IsEnabled
-      @allow_administration
+      !IsDisabled()
     end
 
     # Checks if remote administration is currently disallowed
     def IsDisabled
-      !IsEnabled()
+      @mode == :disabled
     end
 
-    # Enables remote administration.
+    # Enables remote administration without vnc manager.
     def Enable
-      @allow_administration = true
+      @mode = :xinetd
+
+      nil
+    end
+
+    # Enables remote administration with vnc manager.
+    def EnableUsingVncManager
+      @mode = :vncmanager
 
       nil
     end
 
     # Disables remote administration.
     def Disable
-      @allow_administration = false
+      @mode = :disabled
 
       nil
+    end
+
+    def UsesVncManager
+      @mode == :vncmanager
     end
 
     # Reset all module data.
@@ -99,11 +112,15 @@ module Yast
       @already_proposed = true
 
       # Bugzilla #135605 - enabling Remote Administration when installing using VNC
-      @allow_administration = Linuxrc.vnc
+      if Linuxrc.vnc
+        Enable()
+      else
+        Disable()
+      end
 
       Builtins.y2milestone(
         "Remote Administration was proposed as: %1",
-        @allow_administration ? "enabled" : "disabled"
+        IsEnabled() ? "enabled" : "disabled"
       )
 
       nil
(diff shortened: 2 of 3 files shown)


src/include/network/remote/dialogs.rb
```diff
@@ -60,17 +60,25 @@ def RemoteMainDialog
...
+              Remote.Mode == :xinetd
             )
           ),
           # RadioButton label
           Left(
             RadioButton(
               Id(:disallow),
               _("&Do Not Allow Remote Administration"),
-              Remote.IsDisabled
+              Remote.Mode == :disabled
```
mchf: Basically all these ```Remote.Mode == ...``` checks breaks encapsulation. I'd stay with ```Remote.IsDisabled``` or better (according our new coding style) ```Remote.disabled?```

src/include/network/remote/dialogs.rb
```diff
@@ -91,8 +99,7 @@ def RemoteMainDialog
               "<p>If this feature is enabled, you can\n" \
               "administer this machine remotely from another machine. Use a VNC\n" \
               "client, such as krdc (connect to <tt>&lt;hostname&gt;:%1</tt>), or\n" \
-              "a Java-capable Web browser (connect to <tt>http://&lt;hostname&gt;:%2/</tt>).\n" \
-              "This form of remote administration is less secure than using SSH.</p>\n"
+              "a Java-capable Web browser (connect to <tt>https://&lt;hostname&gt;:%2/</tt>).\n"
```
mchf: I think it is not a good idea to change text which is subject of translation in this phase.

src/include/network/remote/dialogs.rb
```diff
@@ -142,12 +149,15 @@ def RemoteMainDialog
       if ret == :next
         CWMFirewallInterfaces.OpenFirewallStore(firewall_widget, "", event)
 
-        allowed = Convert.to_boolean(UI.QueryWidget(Id(:allow), :Value))
+        allowed_xinetd = Convert.to_boolean(UI.QueryWidget(Id(:allow_xinetd), :Value))
+        allowed_vncmanager = Convert.to_boolean(UI.QueryWidget(Id(:allow_vncmanager), :Value))
```
mchf: we try to avoid using helpers which were introduced during ycp -> ruby conversion in the new code. You can use just ```UI.QueryWidget(...)``` here

src/include/network/remote/dialogs.rb
```diff
@@ -142,12 +149,15 @@ def RemoteMainDialog
...
 
-        if allowed
-          Remote.Enable
+        if allowed_xinetd
+          Remote.SetMode :xinetd
+        elsif allowed_vncmanager
+          Remote.SetMode :vncmanager
         else
-          Remote.Disable
+          Remote.SetMode :disabled
```
mchf: similar as above. Pls stay with ```Remote.Disable``` or better ```Remote.disable```. Similar for other cases.

src/modules/Remote.rb
```diff
@@ -70,26 +72,38 @@ def main
       @already_proposed = false
     end
 
+    # Retrieves the current remote administration mode
+    def Mode
+      @mode
+    end
```
mchf: ```attr_reader :mode```

src/modules/Remote.rb
```diff
@@ -70,26 +72,38 @@ def main
...
-      @allow_administration = true
+      SetMode(:xinetd)
 
       nil
     end
 
     # Disables remote administration.
     def Disable
-      @allow_administration = false
+      SetMode(:disabled)
```
mchf: So, you kept ```IsDisabled``` and ```Disable``` methods. That's another reason for using them ... see above.

src/modules/Remote.rb
```diff
@@ -135,18 +149,25 @@ def Read
         from: "any",
         to:   "list <map>"
       )
-      vnc_conf = Builtins.filter(xinetd_conf) do |m|
-        s = Ops.get_string(m, "service", "")
-        s == "vnc1" || s == "vnchttpd1"
+
+      xinetd_vnc1_enabled = false
+      xinetd_vnchttp1_enabled = false
```
mchf: These assignments seems to be useless. Default value for uninitialized variables is ```nil``` and it is treated as ```false``` in ruby.

src/modules/Remote.rb
```diff
@@ -135,18 +149,25 @@ def Read
...
-        Ops.get_boolean(vnc_conf, [1, "enabled"], false)
 
       log.info "#{XDM_SERVICE_NAME}: #{xdm}, DM_R_A: #{dm_ra}"
-      log.info "xinetd: #{xinetd}, VNC: #{vnc}"
+      log.info "xinetd: #{xinetd}, VNC1: #{xinetd_vnc1_enabled}, VNCHTTP1: #{xinetd_vnchttp1_enabled}"
+
+      vncmanager = Service.Enabled(VNCMANAGER_SERVICE)
 
-      @allow_administration = xdm && dm_ra && xinetd && vnc
+      @mode = :disabled
```
mchf: ```Disable()```

src/modules/Remote.rb
```diff
@@ -179,7 +200,8 @@ def WriteXinetd
         s = Ops.get_string(m, "service", "")
         next deep_copy(m) if !(s == "vnc1" || s == "vnchttpd1")
         Ops.set(m, "changed", true)
-        Ops.set(m, "enabled", @allow_administration)
+        Ops.set(m, "enabled", @mode == :xinetd) if s == "vnc1"
+        Ops.set(m, "enabled", @mode != :disabled) if s == "vnchttpd1"
```
mchf: Similar to ```Convert``` a few comments above. Pls use just ```m["enabled"] = ...``` It helps us to clean the code ;-)

src/modules/Remote.rb
```diff
@@ -235,25 +257,36 @@ def Write
...
-        # Enable XDM
-        if !Service.Enable(XDM_SERVICE_NAME)
-          Report.Error(
-            _("Enabling service %{service} has failed") % { service: XDM_SERVICE_NAME }
-          )
-          return false
+        services = [
+          [XINETD_SERVICE, true],
+          [XDM_SERVICE_NAME, true],
+          [VNCMANAGER_SERVICE, @mode == :vncmanager]
```
mchf: You sometime use ```Mode() == :vncmanager``` and sometime ```@mode == :vncmanager```. It is inconsistent and hard to maintain in case of further changes. Pls add ```Remote#vncmanager?``` & co (or something like that, naming methods is not my favorite thing to do).

mchf:
+10 for writing tests. Thank you.

michalsrb:
Thank you for the review, I will change the new code to use the new style.

But I will keep the original API (like Remote.IsEnabled, Remote.Enable, etc) the way it is - it seem to be used in other places in yast-network and also yast-installation, so changing them would require bigger rewrites.

src/modules/Remote.rb
michalsrb: I am not so fluent in ruby, but the reason was to declare the current block so they can be accessed later. Without these lines they exist only in the inner do-end block, and attempt to access them later gives:

`NameError: undefined local variable or method `xinetd_vnc1_enabled'` at src/modules/Remote.rb:149

michalsrb:
Please check the updated version. I think I addressed all your comments.

I know the new method "Remote#UsesVncManager" is using incorrect naming convention, but all the other methods in the class have that kind of names. I can't rename them all now and it just looked too weird mixing them together.

src/modules/Remote.rb
mchf: right. I've missed that.

src/modules/Remote.rb
```diff
@@ -72,38 +74,46 @@ def main
...
     # Checks if remote administration is currently disallowed
     def IsDisabled
-      !IsEnabled()
+      @mode == :disabled
     end
 
     # Enables remote administration.
-    def Enable
-      @allow_administration = true
+    def Enable(use_vncmanager = false)
```
mchf: this boolean option is bad. It makes code hardly readable and nonintuitive as you can see few lines bellow (https://github.com/yast/yast-network/pull/401/files#diff-3e7dc301d1e2f6c63659b651f1c3ed2dR161)

I'd stay with ```Enable``` with no option for xinetd to stay backward compatible and introduce e.g. ```enable_with_persistent_sessions``` for vncmanager.

src/include/network/remote/dialogs.rb
```diff
@@ -60,9 +60,17 @@ def RemoteMainDialog
           # RadioButton label
           Left(
             RadioButton(
-              Id(:allow),
-              _("&Allow Remote Administration"),
-              Remote.IsEnabled
+              Id(:allow_with_vncmanager),
+              _("&Allow Remote Administration With Session Management"),
+              Remote.IsEnabled && Remote.UsesVncManager
```
mchf: I think that ```Remote.IsEnabled``` can be checked in ```Remote.UsesVncManager```

src/modules/Remote.rb
```diff
@@ -135,18 +145,30 @@ def Read
...
-        Ops.get_boolean(vnc_conf, [1, "enabled"], false)
 
       log.info "#{XDM_SERVICE_NAME}: #{xdm}, DM_R_A: #{dm_ra}"
-      log.info "xinetd: #{xinetd}, VNC: #{vnc}"
+      log.info "xinetd: #{xinetd}, VNC1: #{xinetd_vnc1_enabled}, VNCHTTP1: #{xinetd_vnchttp1_enabled}"
 
-      @allow_administration = xdm && dm_ra && xinetd && vnc
+      vncmanager = Service.Enabled(VNCMANAGER_SERVICE)
+
+      if xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled
```
mchf: it would be nice to hide this long condition into a private helper. So, I'd do something like: ```if remote_access_requested```

src/modules/Remote.rb
michalsrb: Alright, I'll change that.

src/include/network/remote/dialogs.rb
michalsrb: Sure it can. I made it like this for the symmetry with the next radio button.

src/modules/Remote.rb
michalsrb: I am not sure what you mean by a private helper. That expression is composed of local variables and in that method exists exactly once. How can it be meaningfully extracted into a helper?

src/modules/Remote.rb
```diff
@@ -135,18 +152,30 @@ def Read
...
-        Ops.get_boolean(vnc_conf, [1, "enabled"], false)
 
       log.info "#{XDM_SERVICE_NAME}: #{xdm}, DM_R_A: #{dm_ra}"
-      log.info "xinetd: #{xinetd}, VNC: #{vnc}"
+      log.info "xinetd: #{xinetd}, VNC1: #{xinetd_vnc1_enabled}, VNCHTTP1: #{xinetd_vnchttp1_enabled}"
+
+      vncmanager = Service.Enabled(VNCMANAGER_SERVICE)
 
-      @allow_administration = xdm && dm_ra && xinetd && vnc
+      if xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled
```
kobliha:
```ruby
if something?
   ...
end

# Returns X whether something
def something?
    xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled
end
```

michalsrb: But `xdm`, `dm_ra`, `xinetd` and `xinetd_vnchttp1_enabled` are local variables existing only inside `Remote#Read`.

mchf: I most probably still don't understand this ```if ... else``` construct correctly. What is expected combination of these flags when vncmanager should be running? According this condition it e.g. seems that xinet should be running.

michalsrb: That is correct; xinetd is used even in vncmanager mode to serve the web page with VNC applet, but not to spawn the Xvnc processes (that is done by vncmanager instead). It is described in the pull request message, but I should have made this clearer.

So:

- Enabled (vncmanager mode)
`xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled && !xinetd_vnc1_enabled && vncmanager`

- Enabled (xinetd mode)
`xdm && dm_ra && xinetd && xinetd_vnchttp1_enabled && xinetd_vnc1_enabled && !vncmanager`

- Disabled
any other combination

mchf: thanks for repeating that ...

One more thing is unclear to me. Why is ```xinetd_vnchttp1_enabled``` needed when enabling vnc1 or vncmanager service? I'd think of it as independent / alternative to these services.

Sorry, that it takes it so long. I want to use this opportunity to make some dark corners more clear to me.

michalsrb: The `vnchttp1` depends on the others. The `vnchttp1` is a simple web server that gives you a page with a java vnc viewer that connects to vnc server run by the `vnc1` service or `vncmanager`.

So running `vnchttp1` alone would be pointless. We could make `vnchttp1` optional when enabling the `vnc1` or `vncmanager`, but that would be another feature and another pull request.

mchf: That's what I've somehow expected. But written this way it seems so that ```vnc1``` and / or ```vncmanager``` depends on ```vnchttp1```. So, when e.g. user edits ```/etc/xinetd.d/vnc``` manually and disables ```vnchttpd1```  and keeps ```vnc1``` enabled then the code will disable remote access completely. 

In general I don't see a point in putting ```vnchttpd1``` check into condition for starting ```vnc1``` or ```vncmanager``` there is no relation in this direction.

michalsrb: It does behave the way you are describing. Which is the same way as it used to.

It we omit the `vnchttpd1` from this condition, it will get re-enabled after submitting the form.

I don't have any preferences either way, if you think it's better to omit it here, I will change it.

mchf: I just wanted to clarify that I get it correctly (that the code really is so messy ;-)

I'm aware of re-enabling in write.

I'd definitely drop it from here and I'd be glad if you could create a fate request which resolves the write (or asks for separate handling of ```vnchttp1``` at all). Thank you for the patience.

michalsrb:
Changed the `Remote#Read` to ignore the state of the `vnchttpd1` service.

Created [fate#320881: Make VNC java applet service optional in YaST's VNC setup](https://fate.suse.com/320881).

mchf:
Thank you for the patch, for the patience and discussion. I'll add some necessary administrative stuff and merge it.

michalsrb:
Great, thank you too.

mchf:
Updated patch submitted, so closing this one.

PR #402, opened 2016-05-11
[WIP] Do not write anything when nothing was changed

mchf:
[fate#318787](https://fate.suse.com/318787)

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -1811,7 +1811,8 @@ module Yast
     # Sets device map items related to dhclient
     def setup_dhclient_options(devmap)
       if isCurrentDHCP
-        devmap["DHCLIENT_SET_DEFAULT_ROUTE"] = TRISTATE_TO_S.fetch(@set_default_route)
+        val = TRISTATE_TO_S.fetch(@set_default_route)
+        devmap["DHCLIENT_SET_DEFAULT_ROUTE"] = val if val
       end
       devmap
     end
@@ -1825,7 +1826,7 @@ module Yast
       devmap["STARTMODE"] = @startmode
       devmap["IFPLUGD_PRIORITY"] = @ifplugd_priority.to_i if @startmode == "ifplugd"
       devmap["BOOTPROTO"] = @bootproto
-      devmap["_aliases"] = @aliases
+      devmap["_aliases"] = @aliases if @aliases && !@aliases.empty?
 
       log.info("aliases #{@aliases}")
 
@@ -2012,21 +2013,24 @@ module Yast
         newdev["INTERFACETYPE"] = @type
       end
 
-      NetworkInterfaces.Name = Ops.get_string(@Items, [@current, "ifcfg"], "")
-      NetworkInterfaces.Current = deep_copy(newdev)
+      if GetCurrentMap() != newdev
+        NetworkInterfaces.Name = Ops.get_string(@Items, [@current, "ifcfg"], "")
+        NetworkInterfaces.Current = deep_copy(newdev)
 
-      # bnc#752464 - can leak wireless passwords
-      # useful only for debugging. Writes huge struct mostly filled by defaults.
-      Builtins.y2debug("%1", NetworkInterfaces.ConcealSecrets1(newdev))
+        # bnc#752464 - can leak wireless passwords
+        # useful only for debugging. Writes huge struct mostly filled by defaults.
+        Builtins.y2debug("%1", NetworkInterfaces.ConcealSecrets1(newdev))
 
-      Ops.set(@Items, [@current, "ifcfg"], "") if !NetworkInterfaces.Commit
+        Ops.set(@Items, [@current, "ifcfg"], "") if !NetworkInterfaces.Commit
 
-      # configure bridge ports
-      if @bridge_ports
-        @bridge_ports.split.each { |bp| configure_as_bridge_port(bp) }
+        # configure bridge ports
+        if @bridge_ports
+          @bridge_ports.split.each { |bp| configure_as_bridge_port(bp) }
+        end
+
+        @modified = true
       end
 
-      @modified = true
       @operation = nil
       true
     end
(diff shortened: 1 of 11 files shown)


src/modules/LanItems.rb
```diff
@@ -2018,21 +2027,24 @@ def Commit
         newdev["INTERFACETYPE"] = @type
       end
 
-      NetworkInterfaces.Name = Ops.get_string(@Items, [@current, "ifcfg"], "")
-      NetworkInterfaces.Current = deep_copy(newdev)
+      if GetCurrentMap() != newdev
+        NetworkInterfaces.Name = Items()[@current]["ifcfg"] || ""
+        NetworkInterfaces.Current = deep_copy(newdev)
```
mvidner: The above 3 lines are still a mystery to me :-/ I'll revisit them after I review the rest.

src/modules/SuSEFirewall4Network.rb
```diff
@@ -218,28 +235,24 @@ def AnyInterfacesHandledByFirewall
...
-        SuSEFirewall.AddInterfaceIntoZone(interface, zone)
-        SuSEFirewall.SetEnableService(true)
-        SuSEFirewall.SetStartService(true)
-        # Removing protection
+      if protect_status
+        log.info("Enabling firewall because of '#{interface}' interface")
+
+        SuSEFirewall.AddInterfaceIntoZone(interface, zone) if !iface_in_zone?(interface, zone)
+
+        start_fw_service
```
mvidner: won't this set the dirty flag unnecessarily? maybe it should go inside the `if` block of `iface_in_zone?`?

mchf: no bcs functions ```SetStartService``` and ```SetEnableService``` checks if the service was started / enabled already.

Problematic functions are ```AddInterfaceIntoZone```, ```RemoveInterfaceFromZone``` and most probably others too which updates fw setup unconditionally. Most probably the check would be placed in these functions. However it changes behavior of these functions (can but need not to be a problem). It would also make implementation of this feature a bit more complex. And last but not least an additional layer of check never hurts in network ;-)

src/modules/LanItems.rb
mchf: just first line is new ;-)

1. checks if new and current sysconfig differs
2. touches ```NetworkInterfaces``` internal state machine to set proper config to update proper configuration. In my POV this should be replaced by ```NetworkInterfaces.Select```. I'll most probably do it in a later commit.
3. updates whole devmap in ```NetworkInterfaces```. In my POV it can be replaced by ```NetworkInterfaces.Change2```. As in above case I'll nost probably do so in a later commit.

src/modules/LanItems.rb
```diff
@@ -2018,21 +2029,24 @@ def Commit
         newdev["INTERFACETYPE"] = @type
       end
 
-      NetworkInterfaces.Name = Ops.get_string(@Items, [@current, "ifcfg"], "")
-      NetworkInterfaces.Current = deep_copy(newdev)
+      if GetCurrentMap() != newdev
```
mvidner: - GetCurrentMap is not documented
- it fetches the ifcfg from `NetworkInterfaces`, not from `Items`. what is the relationship between the two at this point?

mchf: It can happen that ```GetCurrentMap``` is undocumented. It is old method from YCP times.

Relationship is almost none. This part of ```Commit``` stores new devmap which is built from various "global" variables spread over the ```LanItems```. ```Items``` just provides proper ifcfg name here.

src/modules/LanItems.rb
```diff
@@ -2031,15 +2031,12 @@ def Commit
       # CanonicalizeIP is called to get new device map into the same shape as
       # NetworkInterfaces provides the current one.
       if current_map != NetworkInterfaces.CanonicalizeIP(new_map)
-        NetworkInterfaces.Name = Items()[@current]["ifcfg"] || ""
-        NetworkInterfaces.Current = deep_copy(newdev)
+        Items()[@current]["ifcfg"] = "" if !NetworkInterfaces.Change2(Items()[@current]["ifcfg"], newdev, false)
```
mvidner: please assign that `false` to a `explode_if_existing` or whatever, before use

mvidner: we can factor out the Items access if we use a mutating method on `::String`:
```rb
ifcfg_name = Items()[@current]["ifcfg"]
explode = false
ifcfg_name.replace("") if !NetworkInterfaces.Change2(ifcfg_name, newdev, explode)
```

mvidner:
LGTM.

The coverage [has increased slightly](https://coveralls.io/builds/6514433) but I've noticed that `Lan.Modified`, which is being changed, remains uncovered. It would be an easy win if you feel like keeping an eye on the numbers.

mchf:
Thanks for hint. I'm going for another 0.001% ;-) Any small step towards 50% test coverage is appreciated ;-)

mchf:
@mvidner 
I've added the testsuite and changelog. Could you pls check those two updates?

mvidner:
Once `Lan.Modified` is tested, the coverage should increase by 0.1% (7 lines of 6510). But we don't see such a change in the coveralls report.

test/lan_test.rb
```diff
@@ -159,3 +159,37 @@
...
   end
 end
+
+describe "LanClass#Modified" do
+  def expect_modification_succeedes(modname, method)
+    allow(modname)
+      .to receive(method)
+      .and_return true
+
+    expect(modname.send(method)).to be true
```
mvidner: Here you mock a method and then you call it. It does succeed but it is not testing any of production code.

mchf:
That's bcs I forgot to actually run the test on ```Lan.Modified```. Fix is on the way. Thanks

mvidner:
Yay! LGTM.

mchf:
Thank you.

PR #403, opened 2016-05-13
Smart bonding with npar and sriov interfaces (fate#315995)

teclator:
An initial study was documented [here](https://gist.github.com/teclator/c1012f264b96bc39a191a4227320bce4) although it had to be updated and some parts were redacted quickly to discuss with UX Expert.

**[Trello Card](https://trello.com/c/UKTnEvME/534-5-sle-12-sp2-p6-i-feature-315995-include-support-for-smarter-bonding-of-virtual-functions-on-npar-sr-iov-devices)**.

### Final implementation:

`Physical Port Id` has been added to the `Device Overview` and also to the `Bond Slave's description` in the `Bond Slaves` selection table. 

screenshot from 2016-05-16 09-47-53

screenshot from 2016-05-16 09-46-09

In case the user selected more than one device sharing the same ```Physical Port ID``` he will be warned about it when pressed ```Next```.

screenshot from 2016-05-16 09-47-09

screenshot from 2016-05-16 09-49-12

We had some discussions about trying to improve ```Bond Slaves``` interface but at the end only  sorting of available devices for bonding had been improved. In some cases ```available devices``` weren't sorted properly, now they are sorted by device name taking care of numerical parts.

screenshot from 2016-05-16 09-48-28

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -226,8 +226,8 @@ module Yast
           "help"          => Ops.get_string(@help, "etherdevice", "")
         },
         "BONDSLAVE"    => {
-          "widget"        => :custom,
-          "custom_widget" => Frame(
+          "widget"            => :custom,
+          "custom_widget"     => Frame(
             _("Bond Slaves and Order"),
             VBox(
               MultiSelectionBox(Id(:msbox_items), Opt(:notify), "", []),
@@ -237,18 +237,23 @@ module Yast
               )
             )
           ),
-          "label"         => _("Bond &Slaves"),
+          "label"             => _("Bond &Slaves"),
           #        "opt": [`shrinkable],
-          "init"          => fun_ref(
+          "init"              => fun_ref(
             method(:InitSlave),
             "void (string)"
           ),
-          "handle"        => fun_ref(
+          "handle"            => fun_ref(
             method(:HandleSlave),
             "symbol (string, map)"
           ),
-          "store"         => fun_ref(method(:StoreSlave), "void (string, map)"),
-          "help"          => Ops.get_string(@help, "bondslave", "")
+          "validate_type"     => :function,
+          "validate_function" => fun_ref(
+            method(:ValidateBond),
+            "boolean (string, map)"
+          ),
+          "store"             => fun_ref(method(:StoreSlave), "void (string, map)"),
+          "help"              => Ops.get_string(@help, "bondslave", "")
         },
         "BONDOPTION"   => {
           "widget" => :combobox,
@@ -576,22 +581,14 @@ module Yast
       nil
     end
 
+    # Given a device name returns the position in the bond slaves table
+    # or -1 if not present
+    #
+    # @param [String] slave name
+    # @return [Integer] index of the given slave in msbox_items table
     def getISlaveIndex(slave)
-      items = Convert.convert(
-        UI.QueryWidget(:msbox_items, :Items),
-        from: "any",
-        to:   "list <term>"
-      )
-      index = -1
-      pos = 0
-      Builtins.foreach(items) do |it|
-        if Ops.get_string(it, [0, 0], "") == slave
-          index = pos
-          raise Break
-        end
-        pos = Ops.add(pos, 1)
-      end
-      index
+      items = UI.QueryWidget(:msbox_items, :Items)
+      items.index { |it| Ops.get_string(it, [0, 0], "") == slave } || -1
     end
 
     def enableSlaveButtons
@@ -615,14 +612,14 @@ module Yast
     # Default function to init the value of slave devices box for bonding.
     # @param [String] key	id of the widget
     def InitSlave(_key)
-      Ops.set(@settings, "SLAVES", LanItems.bond_slaves)
+      @settings["SLAVES"] =  LanItems.bond_slaves || []
       UI.ChangeWidget(
         :msbox_items,
         :SelectedItems,
-        Ops.get_list(@settings, "SLAVES", [])
+        @settings["SLAVES"]
       )
 
-      Ops.set(@settings, "BONDOPTION", LanItems.bond_option)
+      @settings["BONDOPTION"] = LanItems.bond_option
 
       items = CreateSlaveItems(
         LanItems.GetBondableInterfaces(LanItems.GetCurrentName),
@@ -632,93 +629,72 @@ module Yast
       # reorder the items
       l2 = []
       l1 = []
-      Builtins.foreach(
-        Convert.convert(items, from: "list", to: "list <term>")
-      ) do |t|
-        if Builtins.contains(
-          Ops.get_list(@settings, "SLAVES", []),
-          Ops.get_string(t, [0, 0], "")
-          )
-          l1 = Builtins.add(l1, t)
+      items.each do |t|
+        if @settings["SLAVES"].include? Ops.get_string(t, [0, 0], "")
+          l1 << t
         else
-          l2 = Builtins.add(l2, t)
+          l2 << t
         end
       end
 
       items = []
-      Builtins.foreach(Ops.get_list(@settings, "SLAVES", [])) do |s|
-        Builtins.foreach(
-          Convert.convert(l1, from: "list", to: "list <term>")
-        ) do |t|
-          items = Builtins.add(items, t) if Ops.get_string(t, [0, 0], "") == s
+      @settings["SLAVES"].each do |s|
+        l1.each do |t|
+          items << t if Ops.get_string(t, [0, 0], "") == s
         end
       end
 
-      items = Builtins.union(items, l2)
+      items += l2.sort_by { |t| justify_dev_name(t[0][0]) }
+
       UI.ChangeWidget(:msbox_items, :Items, items)
       enableSlaveButtons
 
       nil
     end
 
-    def HandleSlave(_key, event)
-      event = deep_copy(event)
-      if Ops.get_string(event, "EventReason", "") == "SelectionChanged"
-        enableSlaveButtons
-      elsif Ops.get_string(event, "EventReason", "") == "Activated" &&
-          Ops.get(event, "WidgetClass") == :PushButton
-        items = Convert.convert(
-          UI.QueryWidget(:msbox_items, :Items),
-          from: "any",
-          to:   "list <term>"
-        )
-        current = Builtins.tostring(UI.QueryWidget(:msbox_items, :CurrentItem))
-        index = getISlaveIndex(current)
-        new_items = []
-        pos = 0
-        case Ops.get_symbol(event, "ID", :nil)
-        when :up
-          while Ops.greater_than(index, Ops.add(pos, 1))
-            new_items = Builtins.add(new_items, Ops.get(items, pos))
-            pos = Ops.add(pos, 1)
-          end
-          new_items = Builtins.add(new_items, Ops.get(items, index))
-          new_items = Builtins.add(
-            new_items,
-            Ops.get(items, Ops.subtract(index, 1))
-          )
-          new_items = Convert.convert(
-            Builtins.union(new_items, Builtins.sublist(items, index)),
-            from: "list",
-            to:   "list <term>"
-          )
-        when :down
-          while Ops.greater_than(index, pos)
-            new_items = Builtins.add(new_items, Ops.get(items, pos))
-            pos = Ops.add(pos, 1)
-          end
-          new_items = Builtins.add(
-            new_items,
-            Ops.get(items, Ops.add(index, 1))
-          )
-          new_items = Builtins.add(new_items, Ops.get(items, index))
-          new_items = Convert.convert(
-            Builtins.union(
-              new_items,
-              Builtins.sublist(items, Ops.add(index, 1))
-            ),
-            from: "list",
-            to:   "list <term>"
-          )
+    # A helper for sort devices by name. It justify at right with 0's numeric parts of given
+    # device name until 5 digits.
+    #
+    # ==== Examples
+    #
+    #   justify_dev_name("eth0") # => "eth00000"
+    #   justify_dev_name("eth111") # => "eth00111"
+    #   justify_dev_name("enp0s25") # => "enp00000s00025"
+    #
+    # @param name [String] device name
+    # @return [String] given name with numbers justified at right
+    def justify_dev_name(name)
+      splited_dev_name = name.scan(/\p{Alpha}+|\p{Digit}+/)
+      splited_dev_name.map! do |d|
+        if d.match(/\p{Digit}+/)
+          d.rjust(5, "0")
         else
-          Builtins.y2warning("unknown action")
+          d
+        end
+      end.join
+    end
+
+    def HandleSlave(_key, event)
+      if event["EventReason"] == "SelectionChanged"
+        enableSlaveButtons
+      elsif event["EventReason"] == "Activated" && event["WidgetClass"] == :PushButton
+        items = UI.QueryWidget(:msbox_items, :Items) || []
+        current = UI.QueryWidget(:msbox_items, :CurrentItem).to_s
+        index = getISlaveIndex(current)
+        case event["ID"]
+        when :up
+          items[index], items[index - 1] = items[index - 1], items[index]
+        when :down
+          items[index], items[index + 1] = items[index + 1], items[index]
+        else
+          log.info("unknown action")
           return nil
         end
-        items = deep_copy(new_items)
         UI.ChangeWidget(:msbox_items, :Items, items)
+        UI.ChangeWidget(:msbox_items, :CurrentItem, current)
         enableSlaveButtons
       else
-        Builtins.y2debug("event:%1", event)
+        log.debug "event:#{event}"
       end
 
       nil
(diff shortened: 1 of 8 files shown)


src/include/network/lan/address.rb
```diff
@@ -237,18 +237,23 @@ def initialize_network_lan_address(include_target)
...
-          "handle"        => fun_ref(
+          "handle"            => fun_ref(
             method(:HandleSlave),
             "symbol (string, map)"
           ),
-          "store"         => fun_ref(method(:StoreSlave), "void (string, map)"),
-          "help"          => Ops.get_string(@help, "bondslave", "")
+          "validate_type"     => :function,
+          "validate_function" => fun_ref(
+            method(:ValidateBond),
```
imobach: Please, use `downcase` names for methods (I know we have a lot of old code which is "uppercased", but we should avoid this for new code).

src/include/network/lan/address.rb
```diff
@@ -576,22 +581,14 @@ def StoreVLANSlave(_key, _event)
...
-      Builtins.foreach(items) do |it|
-        if Ops.get_string(it, [0, 0], "") == slave
-          index = pos
-          raise Break
-        end
-        pos = Ops.add(pos, 1)
-      end
-      index
+      items = UI.QueryWidget(:msbox_items, :Items)
+      items.index { |it| Ops.get_string(it, [0, 0], "") == slave } || -1
```
imobach: It would be nice if you can get rid of `Ops` methods (as long as you're refactoring the code :) ).

imobach: np: Using "single letter" variable names is preferred when using the `{}` form of a block. See https://github.com/SUSE/style-guides/blob/master/Ruby.md#parameters

src/include/network/lan/address.rb
```diff
@@ -632,93 +629,72 @@ def InitSlave(_key)
...
       l1 = []
-      Builtins.foreach(
-        Convert.convert(items, from: "list", to: "list <term>")
-      ) do |t|
-        if Builtins.contains(
-          Ops.get_list(@settings, "SLAVES", []),
-          Ops.get_string(t, [0, 0], "")
-          )
-          l1 = Builtins.add(l1, t)
+      items.each do |t|
```
imobach: np: `item` would be a better name than `t` in this case.

imobach: Maybe you could use [Array#partition](http://ruby-doc.org/core-2.1.9/Enumerable.html#method-i-partition) in this case.

src/include/network/lan/address.rb
teclator: You are right, and i changed it initially but i don't like at all 

``` it.fetch(0, []).fetch(0, "") ```

 is not so legible or 

```(it[0] || [])[0] || "" ```

It will be nice with new dig method.

File: src/include/network/lan/address.rb [hunk 4]
```diff
@@ -632,93 +629,72 @@ def InitSlave(_key)
...
+    # ==== Examples
+    #
+    #   justify_dev_name("eth0") # => "eth00000"
+    #   justify_dev_name("eth111") # => "eth00111"
+    #   justify_dev_name("enp0s25") # => "enp00000s00025"
+    #
+    # @param name [String] device name
+    # @return [String] given name with numbers justified at right
+    def justify_dev_name(name)
+      splited_dev_name = name.scan(/\p{Alpha}+|\p{Digit}+/)
```
imobach: np: Maybe in this case it does not matter, but I like to use `\A` and `\Z` to match the beginning at the end.

src/include/network/lan/address.rb
teclator: Sure, and now it's extrange also to me, i don't know why i choose **```it```** instead of  **```i```**

src/include/network/lan/address.rb
```diff
@@ -632,93 +629,72 @@ def InitSlave(_key)
...
-              new_items,
-              Builtins.sublist(items, Ops.add(index, 1))
-            ),
-            from: "list",
-            to:   "list <term>"
-          )
+          items[index], items[index + 1] = items[index + 1], items[index]
         else
-          Builtins.y2warning("unknown action")
+          log.info("unknown action")
```
imobach: Is the change from `warning` to `info` intentioned?

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
       nil
     end
 
+    # Validates created bonding. Currently just prevent the user to create a
+    # bond with more than one interface sharing the same physical port id
+    #
+    # @param [String] key the widget being validated
+    # @param [Hash] event the event being handled
+    # @return true if valid or user decision if not
+    def ValidateBond(_key, _event)
```
imobach: Maybe it's just a matter of taste, but I would extract the validation to a separated method (and let this method to handle the UI stuff).

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+        end
+      end
+
+      physical_port_ids.select! { |_k, v| v.size > 1 }
+
+      return true if physical_port_ids.empty?
+
+      message = ""
+
+      physical_port_ids.each do |port, devs|
```
imobach: From my POV a `map` would be a better option (joining the string at the end).

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+        label =  "PhysicalPortID (#{port}): "
+
+        message << wrap_text(devs.join(", "), 76, prepend_text: label)
+        message << "\n"
+      end
+
+      Popup.YesNoHeadline(
+        Label.WarningMsg,
```
imobach: Adding a note for `TRANSLATORS` to give them a clue about what `message` means would be nice.

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+        _("The interfaces selected share the same physical port and bonding " \
+          "them \nmay not have the desired effect of redundancy.\n\n%s\n" \
+          "Really continue?\n") % message
+      )
+    end
+
+    # Wrap given text breaking lines longer than given wrap size. It supports
+    # cusom separator, max number of lines to split in and cut text to add
```
imobach: `cusom` -> `custom`

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+    # as last line if cut was needed.
+    #
+    # @param [String>] text to be wrapped
```
imobach: Please, remove the `>` sign.

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+    # @param [String] wrap size
+    # @param [Hash <String>] optional parameters as separator and prepend_text.
+    # @return [String] wrap text
+    def wrap_text(text, wrap = 78, options = {})
```
imobach: Instead of `options`, you could use named arguments. Something like this:

```ruby
def wrap_text(text, wrap = 78, separator: " ", prepend_text:  " ", n_lines: 1, cut_text: false)
```

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+
+require "yast"
```
imobach: It's required in `test_helper` so it's not needed here.

src/include/network/lan/address.rb
imobach: Forget about this, I was wrong :)

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    it "returns given device name justified by 0's at right" do
+      expect(subject.justify_dev_name("em5p1")).to eq("em00005p00001")
+    end
+
+  end
+
+  describe "wrap_text" do
+    subject(:routines) { DummyClass.new }
+    let(:devices) do
```
imobach: np: you can use until 100 characters by line (maybe it fits into the same line, I haven't tried it).

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+    end
+    let(:more_devices) do
+      [
+        "enp5s0", "enp5s1", "enp5s2", "enp5s3",
+        "enp5s4", "enp5s5", "enp5s6", "enp5s7"
+      ]
+    end
+
+    context "given a text" do
+      it "returns same text if not exceed wrap size" do
```
imobach: `it does not exceed`

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+        text = "eth0, eth1, eth2, eth3, a_very_long_device_name"
+
+        expect(routines.wrap_text(devices.join(", "))).to eql(text)
+      end
+
+      context "given a line size" do
+        it "returns given text splited in lines by given line size" do
```
imobach: split*t*ed

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+          text = "eth0, eth1, eth2,\n"         \
```
imobach: Please, align `\` symbols.

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+                 "eth3,\n"                   \
+                 "a_very_long_device_name"
+
+          expect(routines.wrap_text(devices.join(", "), 16)).to eql(text)
+        end
+      end
+
+      context "given a number of lines and '...' as cut text" do
+        it "returns wrapped text until given line adding '...' as a new line" do
```
imobach: line**s**

src/include/network/routines.rb
```diff
@@ -746,6 +746,17 @@ def has_carrier?(dev_name)
       ).to_i != 0
     end
 
+    def has_physical_port_id?(dev_name)
+      !physical_port_id(dev_name).empty?
+    end
+
+    def physical_port_id(dev_name)
```
mchf: It should have some documentation.

src/modules/LanItems.rb
```diff
@@ -813,7 +813,7 @@ def GetSlaveCandidates(master, validator)
 
       result = []
 
-      Builtins.foreach(@Items) do |itemId, _attribs|
+      @Items.each do |itemId, _attribs|
```
mchf: NP: It may be better to use ```Items()```. One day it could be easier to mock thinks.

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  describe "#ValidateBond" do
+    let(:msbox_items) do
+      ["eth0", "eth1", "eth2", "eth3", "eth4", "eth5", "eth6", "eth7",
+       "enp0sp25", "enp0sp26", "enp0sp27", "enp0sp28", "enp0sp29"]
+    end
+
+    before do
+      allow(Yast::UI).to receive(:QueryWidget)
+        .with(:msbox_items, :SelectedItems) { msbox_items }
```
imobach: I think you would write:

```ruby
allow(Yast::UI).to receive(:QueryWidget)
  .with(:msbox_items, :SelectedItems)
  .and_return(msbox_items)
```
You really don't need a block to be executed there.

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+
+    context "when there is not more than one physical port id per interface" do
+      let(:msbox_items) do
+        ["eth0", "eth1", "eth2", "eth3"]
+      end
+
+      before do
+        allow(Yast::UI).to receive(:QueryWidget)
+          .with(:msbox_items, :SelectedItems) { msbox_items }
+        allow(subject).to receive(:has_physical_id?).with("eth0") { false }
```
imobach: Same as above.

src/modules/LanItems.rb
```diff
@@ -1407,9 +1407,11 @@ def BuildLanOverview
 
         mac_dev = HTML.Bold("MAC : ") + item_hwinfo["mac"].to_s + "<br>"
         bus_id  = HTML.Bold("BusID : ") + item_hwinfo["busid"].to_s + "<br>"
+        physical_port_id = HTML.Bold("PhysicalPortID : ") + physical_port_id(ifcfg_name).to_s + "<br>"
```
mchf: is ```to_s``` really needed?

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+      it "returns true" do
+        expect(subject.ValidateBond("key", "Event")).to eql(true)
+      end
+    end
+
+    context "when there is more than one physical port id per interface" do
+      before do
+        ["eth0", "eth1", "eth2", "eth3", "eth4", "eth5", "eth6", "eth7",
+         "enp0sp25", "enp0sp26", "enp0sp27", "enp0sp28", "enp0sp29"].map do |i|
+          allow(subject).to receive(:has_physical_port_id?).with(i) { true }
```
imobach: Same as above.

test/address_test.rb
```diff
@@ -0,0 +1,145 @@
...
+          allow(subject).to receive(:physical_port_id).with(i) { "00010486fd348" }
+        end
+      end
+
+      it "warns the user and request confirmation for continue" do
```
imobach: to continue

src/include/network/lan/address.rb
```diff
@@ -728,36 +704,96 @@ def HandleSlave(_key, event)
...
+      physical_port_ids = {}
+
+      items.each do |i|
+        if has_physical_port_id?(i)
+          p = physical_port_ids[physical_port_id(i)] ||= []
+          p << i
+        end
+      end
+
+      physical_port_ids.select! { |_k, v| v.size > 1 }
```
mchf: i think ```empty?``` is more descriptive than ```size```

src/include/network/lan/address.rb
mchf: this is more generic think. It may have a sense to put a ```FIXME``` or something like that with note that it can be generalized.

src/include/network/lan/address.rb
teclator: Ups. :+1:  I was influenced by ValidateBridge :)

src/include/network/lan/address.rb
imobach: But `empty?` is `size > 0` (he's checking for more than 1 element).

src/include/network/lan/address.rb
teclator: Should be warning, so i will fix it.

src/include/network/lan/address.rb
teclator: In this case i din't split responsabilites and should be fine.

src/include/network/lan/address.rb
teclator: But this check is for having more than 1 element. If there is only one interface per physical port id then it is valid.

src/include/network/lan/address.rb
teclator: I could do so, adding a extra new line as i want to finish with it.

src/include/network/lan/address.rb
teclator: About named arguments.. for me it's about taste of code also. In this case are very strange parameters, and of course you know which parameters are available without reading method implementation but for this options i don't like named arguments at all.

src/include/network/routines.rb
teclator: Yeah you are right, although it is just a line of code and the method name is very illustrative it could be not so clear for someone who doesn't know the behavior.

src/include/network/routines.rb
```diff
@@ -746,6 +746,17 @@ def has_carrier?(dev_name)
       ).to_i != 0
     end
 
+    def has_physical_port_id?(dev_name)
```
jreidinger: not ruby name. in ruby it should be `physical_port_id?` stuff like is or have/s is not needed.

src/include/network/lan/address.rb
imobach: Yes, that's the point, you need to read the code to find what parameters are available and what are thery default values. I'm fine with using a hash but, in that case, please document them.

src/modules/LanItems.rb
teclator: Not now, as the method converts to string the returned value

src/include/network/routines.rb
teclator: True

src/modules/LanItems.rb
mchf: so why is it there?

src/include/network/lan/address.rb
mchf: ah yes. Sorry for confusion.

src/modules/LanItems.rb
teclator: Because i modified the physical_port_id but didn't remove this defensive call. So thnx for comment it will be fixed in next commit ;)

src/include/network/lan/address.rb
```diff
@@ -237,18 +239,23 @@ def initialize_network_lan_address(include_target)
...
           ),
-          "store"         => fun_ref(method(:StoreSlave), "void (string, map)"),
-          "help"          => Ops.get_string(@help, "bondslave", "")
+          "validate_type"     => :function,
+          "validate_function" => fun_ref(
+            method(:validate_bond),
+            "boolean (string, map)"
+          ),
+          "store"             => fun_ref(method(:StoreSlave), "void (string, map)"),
+          "help"              => Ops.get_string(@help, "bondslave", "")
```
mchf: another ```Ops``` which can be dropped ;-)

imobach:
LGTM

mchf:
lgtm

PR #405, opened 2016-05-20
Implemented support for vncmanager

mchf:
Original patch and discussion is at https://github.com/yast/yast-network/pull/401

> Add support for vncmanager mode to Remote. [FATE#319319](https://fate.suse.com/319319)
> 
> Remote module can now set up 3 VNC modes: disabled, xinetd, vncmanager
> * disabled
>   * vncmanager systemd service is disabled
>   * vnc1 and vnchttpd1 xinetd services are disabled
> * xinetd
>   * vncmanager systemd service is disabled
>   * xinetd systemd service is enabled
>   * vnc1 and vnchttpd1 xinetd services are enabled
> * vncmanager
>   * vncmanager systemd service is enabled
>   * xinetd systemd service is enabled
>   * vnchttpd1 xinetd service is enabled but vnc1 is not
> 
> Calls to Remote.Enable and Remote.Disable set xinetd and disabled modes, so
> users of the old API can keep using it.


src/include/network/remote/dialogs.rb
```diff
@@ -60,9 +60,17 @@ def RemoteMainDialog
           # RadioButton label
           Left(
             RadioButton(
-              Id(:allow),
-              _("&Allow Remote Administration"),
-              Remote.IsEnabled
+              Id(:allow_with_vncmanager),
+              _("&Allow Remote Administration With Session Management"),
```
mvidner: We should write some explanation of the tradeoffs between the two in the help text. As an ignorant user, why would I ever want the worse option (without secession malignment or whatsit called)?

src/modules/Remote.rb
```diff
@@ -72,38 +71,53 @@ def main
...
+      if Linuxrc.vnc
+        Enable()
+      else
+        Disable()
+      end
 
       Builtins.y2milestone(
         "Remote Administration was proposed as: %1",
-        @allow_administration ? "enabled" : "disabled"
+        IsEnabled() ? "enabled" : "disabled"
```
mvidner: We should log @mode directly, to disambiguate which mode is the default for "enabled".

src/include/network/remote/dialogs.rb
```diff
@@ -60,9 +60,17 @@ def RemoteMainDialog
...
+              Remote.IsEnabled && Remote.UsesVncManager
```
mvidner: `Remote.UsesVncManager` has been renamed away. The bug is hidden even if I run `rake run[remote]` because `IsEnabled` returns false.

src/include/network/remote/dialogs.rb
```diff
@@ -142,9 +150,12 @@ def RemoteMainDialog
       if ret == :next
         CWMFirewallInterfaces.OpenFirewallStore(firewall_widget, "", event)
 
-        allowed = Convert.to_boolean(UI.QueryWidget(Id(:allow), :Value))
+        allow_with_vncmanager = UI.QueryWidget(Id(:allow_with_vncmanager), :Value)
+        allow_without_vncmanager = UI.QueryWidget(Id(:allow_without_vncmanager), :Value)
 
-        if allowed
+        if allow_with_vncmanager
+          Remote.EnableUsingVncManager
```
mvidner: undefined method `EnableUsingVncManager`

mvidner:
This looks like a really cool feature, so we should take care to explain and advertise it properly.
- should session management be enabled by default? IMHO if it works well enough it should. If we are not sure, it should be opt-in.
- write help texts for this module. So far, the unchanged text is

  > ### Remote Administration Settings

  > If this feature is enabled, you can administer this machine remotely from another machine. Use a VNC client, such as krdc (connect to `<hostname>:5901`), or a Java-capable Web browser (connect to `http://<hostname>:5801/`). This form of remote administration is less secure than using SSH.

  I looked for upstream documentation and advertisements, found that @michalsrb is the upstream author :) and that there is no documentation at https://github.com/michalsrb/vncmanager :(

How about this? 

> Without Session Management, only one user can be connected at a time to a session, and that session is terminated when the VNC clients disconnects.

> With Session Management, multiple users can interact (or view? FIXME) with a single session, and the session may persist even if noone is connected.

michalsrb:
> should session management be enabled by default? IMHO if it works well enough it should. If we are not sure, it should be opt-in.

This is first time it will be used in the wild, so lets make it opt-in.

Your suggested description is correct and sounds good to me:

> Without Session Management, only one user can be connected at a time to a session, and that session is terminated when the VNC clients disconnects.
> 
> With Session Management, multiple users can interact with a single session, and the session may persist even if noone is connected.

If this text stays as well, it should be modified to say "https" and remove the sentence about it being less secure than SSH: (I was suggesting this change in the original pull request.)

> If this feature is enabled, you can administer this machine remotely from another machine. Use a VNC client, such as krdc (connect to <hostname>:5901), or a Java-capable Web browser (connect to https://<hostname>:5801/).

mchf:
thanks for comments. I'm going to update the PR

src/include/network/remote/dialogs.rb
```diff
@@ -99,8 +99,12 @@ def RemoteMainDialog
               "<p>If this feature is enabled, you can\n" \
               "administer this machine remotely from another machine. Use a VNC\n" \
               "client, such as krdc (connect to <tt>&lt;hostname&gt;:%1</tt>), or\n" \
-              "a Java-capable Web browser (connect to <tt>http://&lt;hostname&gt;:%2/</tt>).\n" \
-              "This form of remote administration is less secure than using SSH.</p>\n"
+              "a Java-capable Web browser (connect to <tt>https://&lt;hostname&gt;:%2/</tt>).</p>\n" \
+              "<p>Without Session Management, only one user can be connected\n"\
+              "at a time to a session, and that session is terminated when the VNC clients\n" \
```
mvidner: singular: `client\n`

src/include/network/remote/dialogs.rb
mvidner: Done

src/modules/Remote.rb
mvidner: Please do this, log archeology will get easier
```rb
      Builtins.y2milestone(
        "Remote Administration was proposed as: %1",
        @mode
      )
```

src/include/network/remote/dialogs.rb
mchf: ok, going to fix it. My fault, I didn't do a proof reading just copy paste.

mvidner:
Thank you, LGTM!

mchf:
Thanks for all comments and improvements.
Thanks to @michalsrb for the patch.

PR #406, opened 2016-05-31
Enable / disable vncmanager only if it is available. 

mchf:
bnc#981842


jreidinger:
in general LGTM, just please fix tests, so travis is happy

mchf:
Confirmed even by the original patch author in the bug report. Thanks for review.

PR #409, opened 2016-06-14
Do not crash when configuring s390 devices

mchf:
bnc#983201

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -2216,8 +2216,10 @@ module Yast
     # Creates eth emulation for s390 devices
     #
     # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
-    # if not present then initialization of S390 specific LanItems' variables is
-    # expected to be done out of this function.
+    # If it contains s390 device attributes definition, then these definitions takes
+    # precendence over values assigned to corresponding LanItems' global variables
+    # before the method invocation. Hash keys are strings named after LanItems'
+    # s390 globals.
     def createS390Device(dev_attrs = {})
       Builtins.y2milestone("creating device s390 network device, #{dev_attrs}")
 
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -2215,17 +2215,24 @@ module Yast
 
     # Creates eth emulation for s390 devices
     #
-    # @param [Hash] an s390 device description as obtained from AY profile
-    def createS390Device(rule)
-      Builtins.y2milestone("creating device s390 network device, #{rule}")
+    # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
+    # if not present then initialization of S390 specific LanItems' variables is
+    # expected to be done out of this function.
+    def createS390Device(dev_attrs = {})
+      Builtins.y2milestone("creating device s390 network device, #{dev_attrs}")
 
-      Select("")
-      @type = rule["type"] || ""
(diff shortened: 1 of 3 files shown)


src/modules/LanItems.rb
```diff
@@ -2203,17 +2203,24 @@ def enableCurrentEditButton
 
     # Creates eth emulation for s390 devices
     #
-    # @param [Hash] an s390 device description as obtained from AY profile
-    def createS390Device(rule)
+    # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
+    # if not present then initialization of S390 specific LanItems' variables is
+    # expected to be done out of this function.
+    def createS390Device(dev_attrs = [])
```
jreidinger: param is hash and default value is array? it smell for me

mchf: you're right. That happens when working in hurry. Thanks for catching that.

jreidinger:
LGTM

src/modules/LanItems.rb
```diff
@@ -2203,17 +2203,24 @@ def enableCurrentEditButton
 
     # Creates eth emulation for s390 devices
     #
-    # @param [Hash] an s390 device description as obtained from AY profile
-    def createS390Device(rule)
+    # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
+    # if not present then initialization of S390 specific LanItems' variables is
+    # expected to be done out of this function.
+    def createS390Device(dev_attrs = {})
       Builtins.y2milestone("creating device s390 network device, #{rule}")
```
mvidner: NameError

src/modules/LanItems.rb
```diff
@@ -2203,17 +2203,24 @@ def enableCurrentEditButton
 
     # Creates eth emulation for s390 devices
     #
-    # @param [Hash] an s390 device description as obtained from AY profile
-    def createS390Device(rule)
+    # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
+    # if not present then initialization of S390 specific LanItems' variables is
```
mvidner: "Not present" is slightly ambiguous, I wouldn't know whether that means `empty?` or `nil?` -> "If empty then..."

src/modules/LanItems.rb
mvidner: BTW I can make ruby-lint find this bug, as well as the one this PR is fixing!
But they are hidden among 688 false positives.

src/modules/LanItems.rb
```diff
@@ -2203,17 +2203,26 @@ def enableCurrentEditButton
...
-      @type = rule["type"] || ""
-      @qeth_chanids = rule["chanids"] || ""
-      @qeth_layer2 = rule.fetch("layer2", false)
-      @qeth_portname = rule["portname"] || ""
-      @chan_mode = rule["protocol"] || ""
-      @iucv_user = rule["router"] || ""
+    # @param [Hash] an s390 device description (e.g. as obtained from AY profile).
+    # If it contains s390 device attributes definition, then these definitions takes
+    # precendence over values assigned to corresponding LanItems' global variables
+    # before the method invocation. Hash keys are symbols named after LanItems'
```
mvidner: strings, not symbols

mvidner:
Apart from the comment mistake LGTM.

mvidner:
LGTM.

mchf:
Thank you for the review.

PR #410, opened 2016-06-15
Udev methods consolidation

mchf:
Purpose is to consolidate all udev rule handling methods at one place and design a shining new class for handling that.

### DIFF
--- a/src/include/network/lan/udev.rb
+++ b/src/include/network/lan/udev.rb
@@ -84,7 +84,7 @@ module Yast
         path(".target.bash"),
         "udevadm trigger --subsystem-match=net --action=add"
       )
-      ret == 0
+      ret.zero?
     end
 
     # Removes (key,operator,value) tripplet from given udev rule.
--- a/src/include/network/lan/udev.rb
+++ b/src/include/network/lan/udev.rb
@@ -49,7 +49,7 @@ module Yast
     #
     # @return [string] value corresponding to the key or empty string
     def udev_key_value(rule, key)
-      raise ArgumentError if rule.nil?
+      raise ArgumentError, "Rule must not be nil when querying a key value" if rule.nil?
 
       rule.each do |tuple|
         # note that when using =~ then named capture groups (?<name>...) currently
--- a/src/include/network/lan/udev.rb
+++ b/src/include/network/lan/udev.rb
@@ -45,6 +45,23 @@ module Yast
       rule
     end
 
+    # Returns a value of the particular key in the rule
+    #
+    # @return [string] value corresponding to the key or empty string
+    def udev_key_value(rule, key)
+      raise ArgumentError if rule.nil?
+
+      rule.each do |tuple|
+        # note that when using =~ then named capture groups (?<name>...) currently
+        # cannot be used together with interpolation (#{})
+        # see http://stackoverflow.com/questions/15890729/why-does-capturing-named-groups-in-ruby-result-in-undefined-local-variable-or-m
+        matches = tuple.match(/#{key}={1,2}"?(?<value>[^[:space:]"]*)/)
+        return matches[:value] if matches
+      end
+
+      ""
+    end
+
     # Writes new persistent udev net rules and tells udevd to update its configuration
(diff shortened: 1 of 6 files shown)


src/include/network/lan/udev.rb
```diff
@@ -45,6 +45,23 @@ def update_udev_rule_key(rule, key, value)
       rule
     end
 
+    # Returns a value of the particular key in the rule
+    #
+    # @return [string] value corresponding to the key or empty string
+    def udev_key_value(rule, key)
+      raise ArgumentError if rule.nil?
```
jreidinger: it is good practice to print also which argument have which problem so something like `raise ArgumentError, "rule must not be nil" if rule.nil?`

src/include/network/lan/udev.rb
```diff
@@ -45,6 +45,23 @@ def update_udev_rule_key(rule, key, value)
...
+
+      rule.each do |tuple|
```
jreidinger: rule indicate one rule, but you iterate over it. A bit confusing, please document what is exactly rule param

test/udev_test.rb
```diff
@@ -101,12 +104,16 @@ def initialize
...
+    # internally used in ReplaceItemUdev, needed to be able to mock its usage
+    Yast::LanItems.current = 0
+
     updated_rule = Yast::LanItems.ReplaceItemUdev(
       "KERNELS",
       "ATTR{address}",
       "xx:01:02:03:04:05"
     )
     expect(updated_rule).to include "ATTR{address}==\"xx:01:02:03:04:05\""
+    expect(updated_rule).not_to include "KERNELS"
```
jreidinger: new method is not tested ;)

src/include/network/lan/udev.rb
mchf: well it is one rule. But in present code it is represented as an array ... e.g. ```KERNELS=pciid, NAME="eth0"``` is currently represented as ```["KERNELS=pciid", NAME="eth0"]````

src/modules/LanItems.rb
```diff
@@ -405,65 +405,43 @@ def getUdevFallback
...
-        # deep_copy on return value. However, getUdevFallback will be subect of refactoring
-        # so caution is must. Moreover new_rule is also return value so it should be
-        # copied anyway
-        new_rule = deep_copy(current_rule)
-        new_rule[i] = "#{new_key}#{operator}\"#{new_val}\""
+      if current_rule.sort != new_rule.sort
+        SetModified()
 
-        SetModified() if current_rule != new_rule
+        Items()[@current]["udev"] = { "net" => {} } if !Items()[@current]["udev"]
```
teclator: I would add a log here maybe explaining that we are adding a new udev rule instead of replacing a new one, but this is also one the things i don't like about calling ReplateItemUdev for Items without udev rules.

src/include/network/lan/udev.rb
```diff
@@ -45,6 +45,23 @@ def update_udev_rule_key(rule, key, value)
       rule
     end
 
+    # Returns a value of the particular key in the rule
+    #
+    # @return [string] value corresponding to the key or empty string
```
teclator: Document params

mchf:
```ReplaceItemUdev``` is updating device's udev. So, updating internal structures to be capable to carry the rule is fine from my POV. I assume that when someone tries to update udev rule, then he is fine with creating new one when needed.

teclator:
What i mean is that we could add the default one in case GetItemUdevRule(@current).empty? directly

mchf:
Well I tried to split udev methods to basic ones and to some with additional complexity. Generating fallback / default rules belongs to those more complex in my POV. And I'm not sure if it should be generated immediately when reading udev rules or when user asks for device renaming. E.g. It requires some more investigation in udev but I wouldn't be surprised if udev "expects" (in modern linux) to each device having own udev rule - in such case we can use first approach.

teclator:
So now i think is time to remove the [WIP] from it an update bump version

teclator:
LGTM

PR #411, opened 2016-06-21
Set keep_install_network to 'yes'

imobach:
Fixes [bsc#984146](https://bugzilla.suse.com/show_bug.cgi?id=984146).
(Internal tracking: https://trello.com/c/Q9ulwr8Q)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,8 @@
 Tue Jun 21 05:37:31 UTC 2016 - igonzalezsosa@suse.com
 
 - bsc#984146
-  - set AutoYaST keep_install_network parameter to 'yes' by default
+  - set AutoYaST keep_install_network parameter to 'true' by
+    default
 - 3.1.157
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jun 21 05:37:31 UTC 2016 - igonzalezsosa@suse.com
+
+- bsc#984146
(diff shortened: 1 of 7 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jun 21 05:37:31 UTC 2016 - igonzalezsosa@suse.com
+
+- bsc#984146
+  - set AutoYaST keep_install_network parameter to 'yes' by default
```
mvidner: The value is actually *true*. Let's use that since other config files allow 'yes' and we should not add to the confusion.

mvidner:
Since this is an incompatible change, and we will be making a Release Notes entry anyway, we should summarize the *reason* for the change also here in the .changes entry. And in the commit message.

Otherwise when we get the first bug reporting "my profile (without the k_i_n setting) no longer works", the rest of the team may not know what to do and even revert the change.

imobach:
@mvidner Actually this change makes SP2 to behave in the same way that SP1. Due to a bug, installation network configuration was available in 2nd stage (in SP1). After the bugfix, that's not the case in SP2 so those profiles that used to work in SP1 are now failing.

Anyway I agree that we have to publish a RN (and also update the AutoYaST handbook) to make clear what's the default value.

mvidner:
So, the changes file could say "... to restore the behavior of SLE12-SP1" ?

mvidner:
Thanks, LGTM.

imobach:
Thanks for the review.

mvidner:
Companion PR: https://github.com/yast/yast-autoinstallation/pull/221

PR #414, opened 2016-07-12
[Review] Request from 'schubi2' @ 'yast/yast-network/review_160712_tag_rnc_file'

schubi2:
Please review the following changes:
  * fcd2109 packaging
  * 1966f54 added dhclient_set_hostname

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 12 10:41:58 CEST 2016 - schubi@suse.de
+
+- Added entry "dhclient_set_hostname" to the RNC file.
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 12 10:41:58 CEST 2016 - schubi@suse.de
+
+- Added entry "dhclient_set_hostname" to the RNC file.
```
imobachgs: As it's a user-oriented changelog, I would replace `RNC file` by `AutoYaST schema` or something like that. What do you think?

imobachgs:
LGTM

PR #417, opened 2016-07-14
Fix broken test

ancorgs:
Trying to fix https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:H/yast2-network/standard/i586

### DIFF
--- a/test/complex_test.rb
+++ b/test/complex_test.rb
@@ -31,7 +31,7 @@ describe "NetworkLanComplexInclude::input_done?" do
 
     it "returns true for input equal to :abort in case of no user modifications" do
       allow(Yast::LanItems)
-        .to receive(:modified)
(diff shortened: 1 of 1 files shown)


mchf:
Well this doesn't explain why the test passes on x86_64 and fails only on i586

ancorgs:
That made it a fragile test in general. With potential to fail in different ways in different systems. For example, in my workstation ruby crashed after running the tests, but it does not crash any more with this fix.

mchf:
can be. In general I've nothing against this fix. It it helps it would be nice. There was no change in this code for some time and I didn't find anything arch specific what could influence the test. So LGTM once travis is happy.

mchf:
rebased and merged as PR#419

PR #418, opened 2016-07-14
update rubocop to 0.41.2 except one possible buggy part, that need ve…

jreidinger:
…rification

remaining problem was introduced by https://github.com/yast/yast-network/commit/095e1dbcad84a0075da6371bc57657506a19b327

which looks like bug

### DIFF
--- a/src/clients/routing.rb
+++ b/src/clients/routing.rb
@@ -215,6 +215,7 @@ module Yast
 
       Ops.add(Ops.add(headline, "\n"), table)
     end
+
     # Handler for action "list"
     # @param [Hash{String => String}] options action options
     def ListHandler(_options)
@@ -402,11 +403,9 @@ module Yast
       options = deep_copy(options)
       found = false
       Routing.Routes = Builtins.maplist(Routing.Routes) do |m|
-        if Ops.get(m, "destination") == Ops.get(options, "dest")
-          found = true
-        else
-          next deep_copy(m)
-        end
+        next deep_copy(m) if Ops.get(m, "destination") == Ops.get(options, "dest")
+
+        found = true
       end
 
       if found
--- a/src/clients/routing.rb
+++ b/src/clients/routing.rb
@@ -403,7 +403,7 @@ module Yast
       options = deep_copy(options)
       found = false
       Routing.Routes = Builtins.maplist(Routing.Routes) do |m|
-        next deep_copy(m) if Ops.get(m, "destination") == Ops.get(options, "dest")
+        next deep_copy(m) if Ops.get(m, "destination") != Ops.get(options, "dest")
 
         found = true
       end
--- a/src/include/network/complex.rb
+++ b/src/include/network/complex.rb
@@ -165,15 +165,15 @@ module Yast
       return [summary, links] if !split
 
       # add headers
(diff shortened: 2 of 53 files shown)


src/clients/routing.rb
```diff
@@ -402,11 +403,9 @@ def DeleteHandler(options)
       options = deep_copy(options)
       found = false
       Routing.Routes = Builtins.maplist(Routing.Routes) do |m|
-        if Ops.get(m, "destination") == Ops.get(options, "dest")
-          found = true
-        else
-          next deep_copy(m)
-        end
+        next deep_copy(m) if Ops.get(m, "destination") == Ops.get(options, "dest")
```
mchf: should be ```next deep_copy(m) if Ops.get(m, "destination") != Ops.get(options, "dest")```, shouldn't it?

src/include/network/complex.rb
```diff
@@ -378,33 +370,29 @@ def DeviceStatus(devtype, devname, devmap)
...
-            )
+            return Builtins.sformat(_("Configured as %1 with address %2"), devname, addr)
           end
-        else
-          # Network card status (%1 is device, %2 is protocol)
-          return Builtins.sformat(
-            _("Configured as %1 with %2"),
+
+          # Network card status (%1 is device, %2 is address, %3 is address)
+          Builtins.sformat(
```
mchf: missing ```return```

test/network_autoconfiguration_test.rb
```diff
@@ -66,10 +66,10 @@ def probe_netcard_factory(num)
 
     allow(Yast::SCR).to receive(:Read) do |path|
       if path.to_s =~ /^\.network\.value\."(eth\d+)".(.*)/
-        ifcfg_files.get(Regexp.last_match(1), Regexp.last_match(2))
-      else
-        raise "Unexpected Read #{path}"
+        next ifcfg_files.get(Regexp.last_match(1), Regexp.last_match(2))
```
mchf: I didn't get this

test/default_route_test.rb
```diff
@@ -29,10 +29,10 @@
 
     allow(Yast::SCR).to receive(:Read) do |path|
       if path.to_s =~ /^\.network\.value\."(eth\d+)".(.*)/
-        @ifcfg_files.get(Regexp.last_match(1), Regexp.last_match(2))
-      else
-        raise "Unexpected Read #{path}"
+        next @ifcfg_files.get(Regexp.last_match(1), Regexp.last_match(2))
```
mchf: i didn't get this

src/clients/routing.rb
jreidinger: you are right, will fix it

src/include/network/complex.rb
jreidinger: good catch

test/default_route_test.rb
jreidinger: in block next is in its context like local return. So this next do quick return of this value.

It is similar like if you have method
```
def m
  if a
    return val
  elsif b
     return valb
  end

  raise "unknown value"
end
```

mchf:
LGTM

jreidinger:
thanks

PR #422, opened 2016-08-02
Handle ifplugd priority as string

mchf:
bnc#991382

Note that default value of ```ifplugd_priority``` is a string. When fetching a value provided by user from dialog the value is converted into string. So, the conversion which is being dropped by this patch was complete nonsense.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Aug  2 12:40:53 UTC 2016 - mfilka@suse.com
+
+- bnc#991382
+  - Do not crash with internal error when using "On Cable
+    Connection" device activation mode.
+- 3.1.160
+
+-------------------------------------------------------------------
 Tue Jul 12 10:41:58 CEST 2016 - schubi@suse.de
 
 - Added entry "dhclient_set_hostname" to the AutoYaST schema file.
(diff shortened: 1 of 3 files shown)


mvidner:
I think I found the originally reported bug:
```
src/modules/LanItems.rb:1658:      @wl_default_key     = d["WIRELESS_DEFAULT_KEY"].to_i
```

Guess what string I grepped for :)

mchf:
> I think I found the originally reported bug:

> ```src/modules/LanItems.rb:1658:      @wl_default_key     = d["WIRELESS_DEFAULT_KEY"].to_i```
> Guess what string I grepped for :)

I don't think so. 
1) The bug explicitly says that WPA-PSK was used. ```WIRELESS_DEFAULT_KEY``` is for WEP.
2) "your" conversion is done when fetching from device map into internal variable which makes no trouble because the issue was caused by non-string in the device map. "my" conversion is done in opposite direction which is wrong. If you check ```LanItems::Commit``` you'll see that ```wl_default_key``` is correctly converted to string when storing back to device map there. See https://github.com/mchf/yast-network/blob/7004327e1f4b41dbf56d1d5c9c5660d9405beb11/src/modules/LanItems.rb#L1930

mvidner:
Ah I see. Sorry for the confusion.
LGTM

mchf:
Thanks for review.

PR #423, opened 2016-08-03
Do not allow adding device when device name is too long

mchf:
bnc#991486

### DIFF
--- a/test/hardware_test.rb
+++ b/test/hardware_test.rb
@@ -4,7 +4,9 @@ require_relative "test_helper"
 
 require "yast"
 
-class HardwareTestClass
+Yast.import "NetworkInterfaces"
+
+class HardwareTestClass < Yast::Module
   def initialize
     Yast.include self, "network/lan/hardware.rb"
   end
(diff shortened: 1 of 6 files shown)


teclator:
It would be nice to have a test, in that case LGTM

mchf:
Note ```ValidNicName``` is already covered by test

teclator:
I mean, `validate_hw` but is ok, just being picky

mchf:
yes, I got it ;-) That was note mainly for me and my friend Alzheimer

test/hardware_test.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it "passes for valid names only" do
+    allow(subject).to receive(:devname_from_hw_dialog).and_return valid_name
+
+    expect(subject.validate_hw(nil, nil)).to be true
+  end
+
+  # bnc#991486
+  it "fails for long names" do
+    allow(subject).to receive(:devname_from_hw_dialog).and_return long_name
```
teclator: you also could check that in this case it should set the focus

test/hardware_test.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  let(:valid_name) { "eth0" }
+  let(:long_name) { "verylongcustomnicname123" }
+
+  before(:each) do
+    allow(Yast::UI).to receive(:SetFocus)
+    expect(subject).to receive(:UsedNicName).and_return false
+  end
+
+  it "passes for valid names only" do
```
teclator: it fails for valid names and also for used nic names

teclator:
LGTM

PR #424, opened 2016-08-08
Do a wicked ifdown instead of a ip link set down (bsc#991694)

teclator:


src/include/network/routines.rb
```diff
@@ -712,9 +712,14 @@ def SetLinkUp(dev_name)
       Run("ip link set #{dev_name} up")
     end
 
-    def SetLinkDown(dev_name)
-      log.info("Setting link down for interface #{dev_name}")
-      Run("ip link set #{dev_name} down")
+    def SetLinkDown(dev_name, delete = false)
```
mchf: why is ```delete``` param needed?

src/include/network/routines.rb
```diff
@@ -712,9 +712,14 @@ def SetLinkUp(dev_name)
...
+      if delete && NetworkService.is_wicked
+        log.info("Setting link down and deleting interface #{dev_name}")
+        Run("wicked ifdown --delete #{dev_name}")
```
mchf: ```NetworkService.run_wicked```

src/include/network/routines.rb
teclator: Only if you want to call ifdown with delete option, in this case, we want to call it when renaming an interface.

mchf: I know but what is use case for forcing that? When do we need to delete the interface explicitly. In other words I have been able to always rename a device without ```delete``` being forced so far.

src/include/network/routines.rb
```diff
@@ -713,8 +713,12 @@ def SetLinkUp(dev_name)
     end
 
     def SetLinkDown(dev_name)
-      log.info("Setting link down for interface #{dev_name}")
-      Run("ip link set #{dev_name} down")
+      if NetworkService.is_wicked
+        NetworkService.run_wicked("ifdown", dev_name)
+      else
+        log.info("Setting link down for interface #{dev_name}")
```
mchf: hm this should be before ```if```

mchf:
Conclusion from irc discussion. 

It seems like wicked issue. We need to check that with wicked team.

src/include/network/routines.rb
teclator: I just maintain so because `run_wicked` already log it, but i can move up.

teclator:
Well, i don't see nothing wrong about call ```wicked``` instead of ```ip``` here, but it also seem to be a wicked bug how it is handle the workflow commented in the bug.

mtomaschewski:
+1 to use ifdown to remove previously applied config.

src/include/network/routines.rb
```diff
@@ -743,6 +745,32 @@ def SetAllLinksDown
...
     end
 
+    # Calls wicked ifdown if it is the currenct backend, set the link down
+    # otherwise
+    #
+    # @param [String] name of interface to put down
+    # @see SetLinkDown
+    def SetInterfaceDown(dev_name)
+      if NetworkService.is_wicked
+        NetworkService.run_wicked("ifdown", dev_name)
```
mchf: may be you can add a friend into this module https://github.com/yast/yast-network/blob/master/src/lib/network/wicked.rb

src/include/network/routines.rb
```diff
@@ -739,6 +745,32 @@ def SetAllLinksDown
       interfaces.all? { |i| SetLinkDown(i) }
     end
 
+    # Calls wicked ifdown if it is the currenct backend, set the link down
```
mchf: typo: currenct

mchf: there should be stated a difference between "ip set link down" and "wicked down" (the other resets wicked's cache)

src/lib/network/wicked.rb
```diff
@@ -16,5 +16,27 @@ def reload_config(devs)
 
       SCR.Execute(BASH_PATH, "wicked ifreload #{devs.join(" ")}").zero?
     end
+
+    # Reloads configuration for each device named in devs
+    #
+    # @devs [Array] list of device names
```
mvidner:
```rb
# @param devs [Array<String>] list of device names
```
(3 times in this file)

teclator: @mvidner I have updated the description with the correct behavior, but the methods just differ in the wicked call, I was thinking in use meta-programming here as it is almost the same code for the 3 methods, but by the moment I decided to use some duplication instead of other code harder to read and test. So suggestions are welcomed :wink:

src/lib/network/wicked.rb
```diff
@@ -16,5 +16,27 @@ def reload_config(devs)
 
       SCR.Execute(BASH_PATH, "wicked ifreload #{devs.join(" ")}").zero?
     end
+
+    # Does an ifdown for each device named in devs
+    #
+    # @devs [Array] list of device names
```
mvidner:
```rb
# @param devs [Array<String>] list of device names
```

I was pointing out that "@devs" is not a valid YARD tag, and this needs to be fixed 3 times in this file. The repetition is fine, no need for metamagic :)

teclator: Ups, true :) :+1:

teclator:
This PR has been replaced by  #444

PR #425, opened 2016-08-09
Completely remove old libstorage dependency

ancorgs:
Also includes a merge to sync with master.

I also commented the remaining Storage code for reference, although at this point in time all we need is a building package without the dependency, even if it contains non-working code (as long as that code is not called during our installation path).

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Aug  2 12:40:53 UTC 2016 - mfilka@suse.com
+
+- bnc#991382
+  - Do not crash with internal error when using "On Cable
+    Connection" device activation mode.
+
+-------------------------------------------------------------------
 Wed Jul 13 11:56:25 UTC 2016 - ancor@suse.com
 
 - storage-ng: commented installation-related code relying on the
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Aug  9 11:18:28 UTC 2016 - ancor@suse.com
+
+- storage-ng: commented remaining code relying on the old
+  yast2-storage
+- Removed yast2-storage dependency
+- 3.2.2
+
+-------------------------------------------------------------------
 Tue Aug  2 12:40:53 UTC 2016 - mfilka@suse.com
 
 - bnc#991382
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -45,7 +45,11 @@ module Yast
       Yast.import "String"
       Yast.import "Mode"
       Yast.import "Arch"
+# storage-ng
+# rubocop:disable Style/BlockComments
+=begin
       Yast.import "Storage"
+=end
 
       Yast.include self, "network/routines.rb"
       Yast.include self, "network/complex.rb"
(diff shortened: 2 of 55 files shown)


package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Aug  9 11:18:28 UTC 2016 - ancor@suse.com
+
+- storage-ng: commented remaining code relying on the old
+  yast2-storage
+- Removed yast2-storage dependency
+- 3.2.2
+
```
mchf: will it pass through build service with no bnc# or so?

ancorgs: That's only checked if it's submitted to Factory or SLE. So far we are not submitting anywhere, just committing to our own project.

src/clients/save_network.rb
```diff
@@ -45,7 +45,11 @@ def main
       Yast.import "String"
       Yast.import "Mode"
       Yast.import "Arch"
+# storage-ng
+# rubocop:disable Style/BlockComments
```
lslezak: NP: I'd prefer standard comments instead the block ones...

lslezak: NP: I'd prefer standard comments instead the block ones...

lslezak:
LGTM

src/clients/save_network.rb
```diff
@@ -70,6 +77,7 @@ def adjust_for_network_disks(file)
       )
 
       log.info("Network based device: #{remote_access}")
+=end
```
mchf: some time age we decided not to keep commented (disabled) code in sources. Is there any reason to keep disabled code?

src/include/network/lan/hardware.rb
```diff
@@ -829,7 +828,7 @@ def storeHW(_key, _event)
         end
         if LanItems.type == "vlan"
           # for vlan devices named vlanN pre-set vlan_id to N, otherwise default to 0
-          LanItems.vlan_id = "#{nm["vlan".size].to_i}"
+          LanItems.vlan_id = nm[VLAN_SIZE].to_i.to_s
```
mchf: when already touching this ```LanItems.vlan_id = nm[VLAN_SIZE..-1]``` looks better in my POV. At least it should remove issue when vlan id is longer than one char, which is currently present.

src/clients/save_network.rb
ancorgs: Yes. It's temporarily commented (during the migration from storage to storage-ng) and we will need to replace that code with storage-ng enabled one in the future. That's why we are commenting all the code with this system. See https://github.com/yast/yast-packager/pull/177/files/0c62bfeb3c0ae9cc036d5e49f83aab21d3505cca#r74195595

https://github.com/yast/yast-storage-ng/blob/master/doc/installer-hacks.md

src/include/network/lan/hardware.rb
mchf: ah it wasn't you who updated that ... but anyway. Could you pls fix that before it disappears in code depth again? Thanks

ancorgs: Take into account that this pull request targets storage-ng. You are viewing that change here simply because the pull request includes a merge from master to sync both branches.

The code you are pointing should be fixed in master, not here. I will do it as payment for your LGTM here. :wink:

mchf: good deal. Thanks. LGTM

src/clients/save_network.rb
mchf: ok

PR #428, opened 2016-08-11
Do not limit number of bonding slaves

mchf:
bnc#985905


src/modules/LanItems.rb
```diff
@@ -1817,6 +1803,31 @@ def setup_basic_device_options(devmap)
...
+      slaves ||= []
+
+      slave_opts = devmap.select { |k, _| k =~ /BONDING_SLAVE/ }.keys
+      slave_opts.each { |s| devmap[s] = nil }
+      slaves.each_with_index { |s, i| devmap["BONDING_SLAVE#{i}"] = s }
+
+      devmap["BONDING_MODULE_OPTS"] = options || ""
+      devmap["BONDING_MASTER"] = "yes"
+
+      deep_copy(devmap)
```
teclator: If you will return a copy, then it should be at the beginning instead of at the final, to not modify the devmap. 

So you can decide if return a modified copy of the given object or modify the object in place renaming the method to `setup_bonding!`

src/modules/LanItems.rb
```diff
@@ -1817,6 +1803,32 @@ def setup_basic_device_options(devmap)
...
+    # Sets bonding specific sysconfig options in given device map
+    #
+    # If any bonding specific option is present already it gets overwritten
+    # by new ones in case of collision. If any BONDING_SLAVEx from devmap
+    # is not set, then its value is set to 'nil'
+    #
+    # @param devmap [map] hash of a device's sysconfig variables
+    # @param slaves [array] list of strings, each string is a bond slave name
+    #
+    # @return updated device map
```
teclator: NP: `@return [Hash]`

test/bond_test.rb
```diff
@@ -102,5 +102,52 @@ module Yast
...
+
+      it "sets BONDING_MASTER and BONDING_MODULE_OPTS" do
+        expected_map = mandatory_opts
+
+        ret = LanItems.setup_bonding({}, [], option)
+
+        expect(ret.select { |k, _| k !~ /BONDING_SLAVE/ }).to match(expected_map)
+      end
+
+      it "sets BONDING_SLAVEx options according given list" do
```
teclator: according to

test/bond_test.rb
```diff
@@ -102,5 +102,52 @@ module Yast
...
       end
     end
+
+    describe "#setup_bonding" do
+      let(:bonding_map) { { "BONDING_SLAVE0" => "eth0", "BONDING_SLAVE1" => "enp0s3" } }
+      let(:mandatory_opts) { { "BONDING_MASTER" => "yes", "BONDING_MODULE_OPTS" => option } }
+      let(:option) { "bonding_option" }
+
+      it "sets BONDING_MASTER and BONDING_MODULE_OPTS" do
+        expected_map = mandatory_opts
```
teclator: is this really giving some value?

src/modules/LanItems.rb
```diff
@@ -1817,6 +1803,32 @@ def setup_basic_device_options(devmap)
...
+    # Sets bonding specific sysconfig options in given device map
+    #
+    # If any bonding specific option is present already it gets overwritten
+    # by new ones in case of collision. If any BONDING_SLAVEx from devmap
+    # is not set, then its value is set to 'nil'
+    #
+    # @param devmap [Hash] hash of a device's sysconfig variables
+    # @param slaves [array] list of strings, each string is a bond slave name
+    #
+    # @return [Hash] updated device map
```
jreidinger: it should be mentioned it is copy of that device map, not just updated

src/modules/LanItems.rb
```diff
@@ -1817,6 +1803,32 @@ def setup_basic_device_options(devmap)
...
+    def setup_bonding(devmap, slaves, options)
+      raise ArgumentError, "Device map has to be provided." if devmap.nil?
+
+      devmap = deep_copy(devmap)
+      slaves ||= []
+
+      slave_opts = devmap.select { |k, _| k =~ /BONDING_SLAVE/ }.keys
```
jreidinger: what about using `start_with?` method?

src/modules/LanItems.rb
```diff
@@ -1812,14 +1812,14 @@ def setup_basic_device_options(devmap)
     # @param devmap [Hash] hash of a device's sysconfig variables
     # @param slaves [array] list of strings, each string is a bond slave name
     #
-    # @return [Hash] updated device map
+    # @return [Hash] a copy of updated device map
```
jreidinger: well, it is still not clear that it do not modify original hash...so maybe `updated copy of device map`?

jreidinger:
LGTM

mchf:
Thanks for reviews

PR #430, opened 2016-08-18
Fixing issue found by openQA

mchf:
Bnc#991486
issue: https://openqa.suse.de/tests/515222#step/yast2_lan_restart/69

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Aug 18 08:52:41 UTC 2016 - mfilka@suse.com
+
+- bnc#991486
+  - extended fix for this bug from 3.1.163. Do not report invalid
+    names incorrectly
+- 3.1.166  
+
+-------------------------------------------------------------------
 Thu Aug 11 19:02:36 UTC 2016 - mfilka@suse.com
 
 - bnc#985905 
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -393,7 +393,7 @@ module Yast
       Ops.set(
         @widget_descr_local,
         "HWDIALOG",
-        Ops.get(@widget_descr_hardware, "HWDIALOG", {})
+        Ops.get(widget_descr_hardware, "HWDIALOG", {})
       )
     end
 
--- a/src/include/network/lan/hardware.rb
+++ b/src/include/network/lan/hardware.rb
@@ -63,6 +63,8 @@ module Yast
         "help"          => initHelp
       }
 
+      # validation function currently checks user's input in :ifcfg_name widget
+      # However this widget is present only when adding new device
       if isNewDevice
         widget_descr["validate_type"] = :function
         widget_descr["validate_function"] = fun_ref(method(:validate_hw), "boolean (string, map)")
--- a/src/include/network/lan/hardware.rb
+++ b/src/include/network/lan/hardware.rb
@@ -55,12 +55,12 @@ module Yast
 
     def widget_descr_hardware
       widget_descr = {
-        "widget"            => :custom,
-        "custom_widget"     => ReplacePoint(Id(:hw_content), Empty()),
-        "init"              => fun_ref(method(:initHwDialog), "void (string)"),
-        "handle"            => fun_ref(method(:handleHW), "symbol (string, map)"),
-        "store"             => fun_ref(method(:storeHW), "void (string, map)"),
-        "help"              => initHelp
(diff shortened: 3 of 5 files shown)


src/include/network/lan/address.rb
```diff
@@ -393,7 +393,7 @@ def initialize_network_lan_address(include_target)
       Ops.set(
         @widget_descr_local,
         "HWDIALOG",
-        Ops.get(@widget_descr_hardware, "HWDIALOG", {})
+        Ops.get(widget_descr_hardware, "HWDIALOG", {})
```
jreidinger: above is `@widget_descr_local` I found quite confusing to have such mixture. Is it at least covered by some testsuite?

jreidinger:
LGTM, if you use object CWM widgets, then you can easily have two variants, one with validation and one without ;)

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Aug 18 08:52:41 UTC 2016 - mfilka@suse.com
+
+- bnc#991486
+  - extended fix for this bug from 3.1.163. Do not report invalid
+    names incorrectly
```
mvidner: This description means that there is an invalid name and the report that we produce about it is incorrect. It has completely misled me about the actual nature of the bugfix.

IIUC the real bug is that we report "Configuration name is invalid" even if there is no name to edit. So what about:
> Do not report "Configuration name is invalid"  inNetworkCard Setup / Hardware when there is no name to edit.

src/include/network/lan/hardware.rb
```diff
@@ -51,19 +51,26 @@ def initialize_network_lan_hardware(include_target)
...
+        "custom_widget" => ReplacePoint(Id(:hw_content), Empty()),
+        "init"          => fun_ref(method(:initHwDialog), "void (string)"),
+        "handle"        => fun_ref(method(:handleHW), "symbol (string, map)"),
+        "store"         => fun_ref(method(:storeHW), "void (string, map)"),
+        "help"          => initHelp
       }
+
+      # validation function currently checks user's input in :ifcfg_name widget
+      # However this widget is present only when adding new device
+      if isNewDevice
```
mvidner: This is introducing an inconsistency in layers. The "init" method calls `isNewDevice`, so the flag should be checked *inside* `validate_hw`, not around it.

package/yast2-network.changes
mchf: well there is name to edit :-/

I wanted to say something like that we're getting "false positives" here

src/include/network/lan/hardware.rb
mchf: it depends .... from other perspective init method takes dialog definition as provided by ```widget_descr_hardware``` and rebuilds dialog completely according ```isNewDevice```. From this point of view I want to get checks on ```isNewDevice``` out of dialog handlers to be able to split dialog creation completely. In other words I'd like to refactor the code so that dialog is build according ```isNewDevice``` at one place and no additional hacks in handlers are needed. It is doable bcs you know before creating dialog which version you need.

package/yast2-network.changes
teclator: Well, the current behaviour is more like:

> Do not validate existing interfaces hardware.
or 
> Validate only new interfaces hardware.

src/include/network/lan/hardware.rb
teclator: We are modifying the widget in case of a NewDevice not adding a validation, so I don't see an inconsistency in that part.

We could also just return true in `validation_hw` in case of `!isNewDevice` just in case for some reason that method could be called in other scenarios but that method is so binding to the widget that don't know if should be needed.

mchf: short summary:
- ```validate_hw``` is not needed in case of editing device
- always return ```true``` is thing I wanted to avoid. Validation function which always returns true is nonsense from my POV
- current place was chosen to provide widget definition according its purpose in time to avoid various widgets removal / adding during init phase.

mvidner: OK, you have convinced me.

package/yast2-network.changes
mvidner: "Do not validate existing interfaces hardware" (or the converse) is too generic, I would not understand what is being fixed.
Another try:

>   Do not report "Configuration name is invalid" in Network Card Setup / Hardware when editing an existing configuration

mvidner:
LGTM (the changes improvement is NP)

mchf:
Thanks for reviews

PR #431, opened 2016-08-19
Init and store IFPLUGD_PRIORITY

mchf:
Partially related to [bsc#991382](https://bugzilla.suse.com/show_bug.cgi?id=991382)


src/include/network/lan/address.rb
```diff
@@ -1174,7 +1174,7 @@ def initIfplugdPriority(_key)
       UI.ChangeWidget(
         Id("IFPLUGD_PRIORITY"),
         :Value,
-        Builtins.tointeger(Ops.get_string(@settings, "IFPLUGD_PRIORITY", "0"))
+        @settings["IFPLUGD_PRIORITY"].to_i
```
mvidner: I don't see the difference. Maybe we should update the docs of tointeger:
http://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast/Builtins#tointeger-class_method
http://www.rubydoc.info/stdlib/core/String#to_i-instance_method

mvidner: Is it the difference between `tointeger(nil)` -> `nil` but `nil.to_i` -> `0`?

teclator: It seems so and also `tointeger("") -> nil`

teclator:
And apart of that, the problem is that the value of the `ifplugd_priority` is only save if we are in the **General Tab**
https://github.com/yast/yast-network/blob/master/src/include/network/lan/address.rb#L1507

with this change it solve all the issues if I'm not wrong

```
        if LanItems.startmode == "ifplugd"
          ifp_prio =
            if UI.WidgetExists(Id("IFPLUGD_PRIORITY"))
              UI.QueryWidget(Id("IFPLUGD_PRIORITY"), :Value)
            else
              @settings["IFPLUGD_PRIORITY"]
            end
          LanItems.ifplugd_priority = ifp_prio.to_s if !ifp_prio.nil?
        end
```

or just

```
        if LanItems.startmode == "ifplugd"
          ifp_prio = @settings["IFPLUGD_PRIORITY"]

          LanItems.ifplugd_priority = ifp_prio.to_s if !ifp_prio.nil?
        end
```

src/include/network/lan/address.rb
mchf: yes

src/include/network/lan/address.rb
```diff
@@ -1408,7 +1414,8 @@ def AddressDialog
                          " then we need a way to decide which interface to take up. Therefore we have to\n" \
                          " set the priority of each interface.  </p>\n"
                      ),
-        "init"    => fun_ref(method(:initIfplugdPriority), "void (string)")
+        "init"    => fun_ref(method(:initIfplugdPriority), "void (string)"),
+        "store"    => fun_ref(method(:store_ifplugd_priority), "void (string, map)")
```
teclator: Better approach :+1:

teclator:
LGTM if you make rubocop happy ;)

mchf:
it's impossible to have him happy ... for longer time ;-)

mchf:
Thanks for review

PR #432, opened 2016-08-24
Perform explicit device activation in case of ssh install. 

mchf:
bnc#993587

tested manually


src/modules/Lan.rb
```diff
@@ -31,6 +31,7 @@
 # Input and output routines.
 require "yast"
 require "network/confirm_virt_proposal"
+require "network/wicked_handlers"
```
jreidinger: why?

src/lib/network/wicked_handlers.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+
+require "yast"
+
+module Yast
+  module Wicked
```
jreidinger: inconsistent module name with file name

src/modules/Lan.rb
mchf: bcs I forgot to remove it when refactoring the patch ;-)

src/lib/network/wicked_handlers.rb
mchf: well, I tried to find a ruby style guide about this and found nothing so far.

And in fact e.g. module name ```WickedHandlers``` vs file name ```wicked_handlers``` is not much better than this one.

jreidinger:
LGTM

mchf:
Thanks for the review ... merging before you change your mind ;-)

PR #433, opened 2016-09-01
Handle errors related to non utf-8 characters in /etc/sysctl.conf more carefully

mchf:
[bsc#992821](https://bugzilla.suse.com/show_bug.cgi?id=992821)

This should be merged into master/SP3 branch only once SP2 is forked off

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -225,28 +225,29 @@ module Yast
         }
       }
 
-      Builtins.foreach(methods) do |which, method|
-        filelist = Ops.get_list(method, "filelist", [])
-        filepath = Ops.get_string(method, "filepath", "")
-        regexp = Ops.get_string(method, "regexp", "")
-        Builtins.foreach(filelist) do |file|
-          filename = Builtins.sformat("%1/%2", filepath, file)
-          if FileUtils.Exists(filename)
-            Builtins.foreach(
-              Builtins.splitstring(
-                Convert.to_string(SCR.Read(path(".target.string"), filename)),
-                "\n"
-              )
-            ) do |row|
-              if Ops.greater_than(
-                Builtins.size(
-                  Builtins.regexptokenize(String.CutBlanks(row), regexp)
-                ),
-                0
-              )
-                Builtins.y2milestone("IPv6 is disabled by '%1' method.", which)
-                @ipv6 = false
-              end
+      methods.each do |which, method|
+        filelist = method["filelist"] || []
+        filepath = method["filepath"] || ""
+        regexp = method["regexp"] || ""
+
+        filelist.each do |file|
+          filename = "#{filepath}/#{file}"
+          next if !FileUtils.Exists(filename)
+
+          Builtins.foreach(
+            Builtins.splitstring(
+              Convert.to_string(SCR.Read(path(".target.string"), filename)),
+              "\n"
+            )
+          ) do |row|
+            if Ops.greater_than(
+              Builtins.size(
+                Builtins.regexptokenize(String.CutBlanks(row), regexp)
+              ),
+              0
+            )
+              log.info("IPv6 is disabled by '#{which}' method.")
+              @ipv6 = false
             end
           end
         end
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -217,19 +217,13 @@ module Yast
     def readIPv6
       @ipv6 = true
 
-      methods =
-        #         "module" : $[
-        #                 "filelist" : ["ipv6", "50-ipv6.conf"],
-        #                 "filepath" : "/etc/modprobe.d/",
-        #                 "regexp"   : "^[[:space:]]*(install ipv6 /bin/true)"
-        #         ]
-        {
-          "builtin" => {
-            "filelist" => ["sysctl.conf"],
-            "filepath" => "/etc/",
-            "regexp"   => "^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1"
-          }
+      methods = {
+        "builtin" => {
+          "filelist" => ["sysctl.conf"],
+          "filepath" => "/etc",
+          "regexp"   => "^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1"
         }
+      }
 
       Builtins.foreach(methods) do |which, method|
         filelist = Ops.get_list(method, "filelist", [])
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -216,7 +216,7 @@ module Yast
 
     # Checks local configuration if IPv6 is allowed
     #
-    # return [Boolean] true when IPv6 configuration is loaded properly
+    # return [Boolean] true when IPv6 is enabled in the system
     def readIPv6
       ipv6 = true
 
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -237,14 +237,22 @@ module Yast
           filename = "#{filepath}/#{file}"
           next if !FileUtils.Exists(filename)
 
-          lines = (SCR.Read(path(".target.string"), filename) || []).split("\n")
-          @ipv6 = false if lines.any? { |row| row != regexp }
+          begin
+            lines = (SCR.Read(path(".target.string"), filename) || []).split("\n")
+          rescue ArgumentError => e
+            log.error("readIPv6: an error when parsing #{filename}")
+            log.error(e.message)
+
+            return false
+          end
+
+          @ipv6 = false if lines.any? { |row| row =~ regexp }
         end
       end
 
       log.info("readIPv6: IPv6 is #{@ipv6 ? "enabled" : "disabled"}")
 
-      nil
+      true
     end
 
     # Read all network settings from the SCR
@@ -360,7 +368,10 @@ module Yast
       ProgressNextStage(_("Reading network configuration...")) if @gui
       NetworkConfig.Read
 
-      readIPv6
+      if !readIPv6
+        msg = _("A configuration file is corrupted. Do you want to continue?")
+        return false if !@gui || !Popup.ContinueCancel(msg)
+      end
 
       Builtins.sleep(sl)
 
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -220,35 +220,25 @@ module Yast
     def readIPv6
       @ipv6 = true
 
-      methods = {
-        "builtin" => {
-          "filelist" => ["sysctl.conf"],
-          "filepath" => "/etc",
-          "regexp"   => /^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1/
-        }
-      }
+      sysctl_path = "/etc/sysctl.conf"
+      ipv6_regexp = /^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1/
 
-      methods.each do |_, method|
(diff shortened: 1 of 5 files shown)


src/modules/Lan.rb
```diff
@@ -214,51 +214,45 @@ def isAnyInterfaceDown
...
-                Builtins.size(
-                  Builtins.regexptokenize(String.CutBlanks(row), regexp)
-                ),
-                0
-              )
-                Builtins.y2milestone("IPv6 is disabled by '%1' method.", which)
-                @ipv6 = false
-              end
-            end
+      methods.each do |_, method|
```
jreidinger: what is a goal of having methods as hash when you do not use key? why not having it array?

src/modules/Lan.rb
```diff
@@ -214,51 +214,45 @@ def isAnyInterfaceDown
...
+          filename = "#{filepath}/#{file}"
+          next if !FileUtils.Exists(filename)
+
+          begin
+            lines = (SCR.Read(path(".target.string"), filename) || []).split("\n")
+          rescue ArgumentError => e
+            log.error("readIPv6: an error when parsing #{filename}")
+            log.error(e.message)
+
+            return false
```
jreidinger: so if we cannot parse then silently expect it is false? what about exception or popup mentioning, that given file cannot be parse as it contain something wrong?

src/modules/Lan.rb
```diff
@@ -214,51 +214,45 @@ def isAnyInterfaceDown
...
+        filelist = method["filelist"] || []
+        filepath = method["filepath"] || ""
+        regexp = method["regexp"] || ""
+
+        filelist.each do |file|
+          filename = "#{filepath}/#{file}"
+          next if !FileUtils.Exists(filename)
+
+          begin
+            lines = (SCR.Read(path(".target.string"), filename) || []).split("\n")
```
jreidinger: btw instead of `SCR.Read(path(".target.string"), filename)` you can use `TargetFile.read(filename)` from https://github.com/yast/yast-yast2/blob/master/library/system/src/lib/yast2/target_file.rb

src/modules/Lan.rb
```diff
@@ -214,51 +214,45 @@ def isAnyInterfaceDown
...
           end
+
+          @ipv6 = false if lines.any? { |row| row =~ regexp }
         end
       end
 
-      nil
+      log.info("readIPv6: IPv6 is #{@ipv6 ? "enabled" : "disabled"}")
+
+      true
```
jreidinger: ugh, why always true? why not return @ipv6? at least it is against what is written in method description

src/modules/Lan.rb
mchf: well that is leftover from previous implementation.

src/modules/Lan.rb
mchf: I wanted to have this method without UI. So, popup is raised elsewhere.

src/modules/Lan.rb
mchf: and what is the benefit of having just one use case of ```SCR::Read``` covered by specialized method?

src/modules/Lan.rb
mchf: hm, I wanted to remove setting ```@ipv6``` from this method, but obviously not finished that.

src/modules/Lan.rb
mchf: in general i'm still not decided if I should drop the hash completely and parse just that one file which is currently the only used or not.

src/modules/Lan.rb
jreidinger: Well, it use ruby native method, so it can be faster and shorter ;) and this class is used for CFA as file reader.

src/modules/Lan.rb
jreidinger: in such case I recommend to use exception

src/modules/Lan.rb
jreidinger: as it goes to master and not SP2 I propose to clean it up :wink:

src/modules/Lan.rb
mchf: I remembered my final decision in this (I've tried several approaches, so I've a little confused myself ;-)

```true / false``` from this method means if the configuration of IPv6 was / was not read successfully.  That's also reason why I decided to handle exception here. And that's also reason why I didn't remove setting ```@ipv6``` from this method.

Based on ```true / false``` result upper layers can decide if it is critical error for them or not. When running in GUI, user is asked (in an upper layer which uses that). When running in non-interactive mode (AY), then it should fail.

Comment (jreidinger) [inline src/modules/Lan.rb]: in such case at least change method documentation, even if I prefer exception solution as it force to handle it somehow and do something if return value is not handled

File: src/modules/Lan.rb (hunk 1 above)
Comment (mchf) [inline src/modules/Lan.rb]: I need to check if I should worry of other places or if these are historical only ;-)

File: src/modules/Lan.rb (hunk 2 above)
Comment (mchf) [inline src/modules/Lan.rb]: ok, i'll re-think it and update if needed

Comment (teclator) [inline src/modules/Lan.rb]: But in case of an exception, we will consider `@ipv6 = true` anyway?

File: src/modules/Lan.rb [hunk 5]
```diff
@@ -214,51 +214,45 @@ def isAnyInterfaceDown
...
-        #                 "filepath" : "/etc/modprobe.d/",
-        #                 "regexp"   : "^[[:space:]]*(install ipv6 /bin/true)"
-        #         ]
-        {
-          "builtin" => {
-            "filelist" => ["sysctl.conf"],
-            "filepath" => "/etc/",
-            "regexp"   => "^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1"
-          }
+      methods = {
```
teclator: Is this really needed? why we need an extra level?

src/modules/Lan.rb
mchf: yes if user wants to continue, that currently is the default value.

src/modules/Lan.rb
mchf: as stated elsewhere, return value is not value of ```@ipv6``` internal variable.

src/modules/Lan.rb
```diff
@@ -374,7 +368,10 @@ def Read(cache)
       ProgressNextStage(_("Reading network configuration...")) if @gui
       NetworkConfig.Read
 
-      readIPv6
+      if !readIPv6
```
mchf: described /explained elsewhere

src/modules/Lan.rb
mchf: as described elsewhere ... that is leftover ... already dropped

mvidner:
Sorry, this approach is wrong.

In the reported bug, the problem is that the customer's whole system is using a non-unicode ecoding:

> \# /etc/sysconfig/language
```sh
RC_LANG="en_US.ISO-8859-1"  # Latin-1
ROOT_USES_LANG="ctype"
AUTO_DETECT_UTF8="no"
```

and then we try to read a file which has a comment including the Latin-1 characters.

YaST ignores the RC_LANG setting and assumes the system always uses unicode.

This PR deals with the problem only for a single configuration file, in the wrong way (by declaring the file contents invalid).

We should instead

a) Declare non-unicode systems outright unsupported, not fixing anything.
b) Make YaST work with non-unicode, piece by piece. This is supported in Ruby by having a different Internal Encoding (stays UTF-8) from the External Encoding (read from LANG).

mchf:
> This PR deals with the problem only for a single configuration file, in the wrong way (by declaring the file > contents invalid).
> 
> We should instead
> 
> a) Declare non-unicode systems outright unsupported, not fixing anything.

I'm for this one at this time. At least bcs I don't know any configuration option which is written in non-ASCII chars. However, I disagree that we should not fix anything. From UX perspective the software should be robust as much as possible, so reading "corrupted" configuration file should not crash whole application with an "Internal Error" message. I also prefer best effort approach (possibility to continue even with corrupted config), but that's another case.

mchf:
> This PR deals with the problem only for a single configuration file, in the wrong way (by declaring the file contents invalid).

Well, the message is probably written incorrectly, it should state something that "configuration cannot be loaded"

mvidner:
I have [asked on research@](https://mailman.suse.de/mlarch//SuSE/research/2016/research.2016.09/msg00056.html) whether we can declare it as unsupported.
If we want to be nice, we could make a "global rescue" like we do with Internal Error but detect a Unicode problem and present a "Sorry, but..." message.

src/modules/Lan.rb
```diff
@@ -214,51 +214,27 @@ def isAnyInterfaceDown
       down
     end
 
+    # Checks local configuration if IPv6 is allowed
+    #
+    # return [Boolean] true when IPv6 configuration is loaded properly
```
mvidner: "true when IPv6 configuration is loaded properly" is ambiguous. I understood it as "true if the configuration settings have been read without error, and IPv6 may be on or off" but it in fact means "true if IPv6 is enabled".

test/SCRStub.rb
```diff
@@ -51,11 +51,32 @@ def written_value_for(key)
   # and named after the yast path (without the leading '.').
   #
   # @return Object
-  def stub_scr_read(path_name)
```
mvidner:
```rb
# @param path_name [String] a SCR path, eg. ".etc.xinetd_conf.services"
```

test/SCRStub.rb
```diff
@@ -51,11 +51,32 @@ def written_value_for(key)
...
     info = YAML.load_file(file)
-    path = Yast::Path.new(path_name)
-    allow(Yast::SCR).to receive(:Read).with(path).and_return info
+    mock_path(path_name, info)
+  end
+
+  # Stubs calls to SCR.Read. Returns file content as a list of lines.
+  #
+  # Should be equivalent to ycp's SCR::Read(".target.string", path)
+  #
```
mvidner:
```rb
# @param path_name [String] an absolute filesystem path, eg. "/etc/sysctl.conf"
```

test/data/scr_read/etc/sysctl.conf
```diff
@@ -0,0 +1,6 @@
+kernel.sysrq = 0
+net.ipv4.ip_forward = 0
+net.ipv4.tcp_syncookies = 1
+net.ipv6.conf.all.forwarding = 0
+net.ipv6.conf.all.disable_ipv6 = 1
```
mvidner: this line is duplicated, is that OK?

src/modules/Lan.rb
```diff
@@ -214,51 +214,27 @@ def isAnyInterfaceDown
...
-        #         ]
-        {
-          "builtin" => {
-            "filelist" => ["sysctl.conf"],
-            "filepath" => "/etc/",
-            "regexp"   => "^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1"
-          }
-        }
+      sysctl_path = "/etc/sysctl.conf"
+      ipv6_regexp = /^[[:space:]]*(net.ipv6.conf.all.disable_ipv6)[[:space:]]*=[[:space:]]*1/
```
mvidner: NP: the dots should be escaped in the regexp as `\.` but in practice this will not hurt

mvidner:
Thanks for the cleanup! Just a couple of minor points.

test/data/scr_read/etc/sysctl.conf
mchf: most probably copy paste issue. Thanks

src/modules/Lan.rb
mchf: I think it is more readable this way ...

mvidner:
Thanks! LGTM

mchf:
Thanks all for review.

PR #434, opened 2016-09-03
Disable DHCP autoconfig in installer when a configuration is already provided

mchf:
bnc#988157

### DIFF
--- a/src/clients/inst_setup_dhcp.rb
+++ b/src/clients/inst_setup_dhcp.rb
@@ -1,12 +1,24 @@
 require "yast"
 require "network/network_autoconfiguration"
 
-def main
-  Yast::NetworkAutoconfiguration.instance.configure_dhcp
+module Yast
+  class SetupDhcp
+    include Singleton
+    include Logger
 
-  # if this is not wrapped in a def, ruby -cw says
-  # warning: possibly useless use of a literal in void context
-  :next
+    def main
+      nac = Yast::NetworkAutoconfiguration.instance
+      if !nac.any_iface_active?
+        nac.configure_dhcp
+      else
+        log.info("Automatic DHCP configuration not started - an interface is already configured")
+      end
+
+      # if this is not wrapped in a def, ruby -cw says
+      # warning: possibly useless use of a literal in void context
+      :next
+    end
+  end
 end
 
-main
+Yast::SetupDhcp.instance.main
--- a/src/include/network/devices.rb
+++ b/src/include/network/devices.rb
@@ -88,51 +88,5 @@ module Yast
       Builtins.y2debug("Free device=%1", ret)
       ret
     end
-
-    # Update Devices map
-    # @param [String] type device type
-    # @param [String] device device number
-    # @param [Hash] newdev new device map
-    # @param [Boolean] check if check if device already exists
-    # @return true if success
-    def ChangeDevice(type, device, newdev, check)
-      newdev = deep_copy(newdev)
-      Builtins.y2debug("ChangeDevice(%1,%2,%3,%4)", type, device, newdev, check)
-      Builtins.y2debug("Devices=%1", @Devices)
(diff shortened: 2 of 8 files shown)


src/include/network/devices.rb
```diff
@@ -88,51 +88,5 @@ def GetFreeDevice(type)
...
-      end
-      # remove(devmap, dev);
-      devmap = Builtins.remove(devmap, dev)
-      Ops.set(@Devices, type, devmap)
-      @DeletedDevices = Builtins.add(@DeletedDevices, dev)
-
-      Builtins.y2debug("Devices=%1", @Devices)
-      Builtins.y2debug("DeletedDevices=%1", @DeletedDevices)
-      true
-    end
```
imobachgs: Nice clean-up!

test/network_autoconfiguration_test.rb
```diff
@@ -151,3 +151,27 @@ def probe_netcard_factory(num)
     expect(result["eth"]["eth1"]["DHCLIENT_SET_DEFAULT_ROUTE"]).to eq nil
   end
 end
+
+describe "NetworkAutoconfiguration#any_iface_active?" do
```
imobachgs: This `describe` block should be inside the general `describe` block. I mean something like:

```ruby
describe Yast::NetworkAutoconfiguration do
   # already existing code

  describe "#any_iface_active?" do
    it "returns true..."
  end
end
```

src/clients/inst_setup_dhcp.rb
```diff
@@ -1,12 +1,24 @@
 require "yast"
 require "network/network_autoconfiguration"
 
-def main
-  Yast::NetworkAutoconfiguration.instance.configure_dhcp
+module Yast
+  class SetupDhcp
```
imobachgs: This class should live in `network/clients` (see [inst_update_installer](https://github.com/yast/yast-installation/blob/master/src/lib/installation/clients/inst_update_installer.rb) as example.

imobachgs:
Apart from some minor code organization comments, I miss a unit-test for `Yast::SetupDhcp#main` client (although it's pretty simple). IMHO we should move `SetupDhcp` client to its own file (maybe we should call it `networks/clients/inst_setup_dhcp`, calling the class `InstSetupDhcp`).

src/clients/inst_setup_dhcp.rb
mchf: Could be. These (networking) clients were created in early times of rewriting installer when the structure was not fixed. 

As I should move more clients, I'd leave it for SP3.

test/network_autoconfiguration_test.rb
mchf: ok, I can try it ;-)

test/network_autoconfiguration_test.rb
```diff
@@ -44,110 +44,136 @@ def probe_netcard_factory(num)
...
-      case path.to_s
-      when ".network.section"
-        ifcfg_files.sections
-      when /^\.network\.value\."(eth\d+)"$/
-        ifcfg_files.keys(Regexp.last_match(1))
-      when ".modules.options", ".etc.install_inf"
-        []
-      else
-        raise "Unexpected Dir #{path}"
+  describe "it sets DHCLIENT_SET_DEFAULT_ROUTE properly" do
```
imobachgs: np: You're describing `#configure_dhcp` and the description about what it does should be in an `it` block. I think the test need some refactoring but I guess it's out of the scope of this PR.

imobachgs:
LGTM

test/network_autoconfiguration_test.rb
mchf: well, I'm not the author but ...
... this test does a lot but its main focus is on just one very specific part of automatic dhcp configuration. Since I also think it deserve more love it is named as is to highlight its incompleteness ;-)

mchf:
Thanks for review

PR #436, opened 2016-09-07
Reload even virtual interfaces when updating config in installer

mchf:
[bsc#993587](https://bugzilla.suse.com/show_bug.cgi?id=993587)


mvidner:
What is a "virtual interface" here? I thought "bridge" or "vlan" but the bug is about s390.

Does the docs need to be fixed? Compare:

http://www.rubydoc.info/github/yast/yast-network/Yast/NetworkRoutinesInclude:GetAllInterfaces
Return list of all interfaces present in the system (not only configured ones as NetworkInterfaces::List does).

http://www.rubydoc.info/github/yast/yast-network/Yast%2FLanItemsClass%3AgetNetworkInterfaces
get list of all configurations for "netcard" macro in NetworkInterfaces module

http://www.rubydoc.info/github/yast/yast-network/Yast%2FLanItemsClass%3AGetNetcardInterfaces
Creates list of all known netcard items
It means list of item ids of all netcards which are detected and/or configured in the system

mchf:
This is extension of previous fix (https://github.com/yast/yast-network/pull/432) as mentioned in changelog. Details were explained during calls and in respective trello card. Basically this issue is not s390 specific. The key presumption for this issue is (e.g.) ssh installation.

Main difference is that ```GetNetcardInterfaces``` returns list of all interfaces physically present in the system (in /sys). The other function lists all interfaces with configuration. Difference between these functions is the best visible on virtual devices which are not present in /sys until they're created.

The note about virtual interfaces refers to user's report who reported that for vlan.

mvidner:
LGTM, thank you!

mchf:
Thanks for review

PR #438, opened 2016-09-13
Fix sorting and cleanup of code. Backport from master (bsc#991968)

teclator:
This changes come from this commit https://github.com/yast/yast-network/pull/403/commits/132ba1e991bfc186849fa8d7548c429a63a6d8b9 and fix the sorting issue.

### Before the fix
screencast2

### After the fix
screencast


mchf:
NP: It would be nice to have bug number in a commit message - it helps in case of archaeological code investigation and so on.

otherwise LGTM

mchf:
Thanks, LGTM

PR #439, opened 2016-09-21
Update networking using AY in running system

mchf:
[bsc#978463](https://bugzilla.suse.com/show_bug.cgi?id=978463) for SP4


src/routines/routines.ycp
```diff
@@ -912,12 +912,12 @@ define list<string> GetAllInterfaces()
...
 boolean SetLinkUp( string dev_name)
 {
-    return Run( sformat("ip link set %1 up", dev_name) );
+    return Run( sformat("ifup %1", dev_name) );
 }
 
 boolean SetLinkDown( string dev_name)
 {
-    return Run( sformat("ip link set %1 down", dev_name) );
+    return Run( sformat("ifdown %1", dev_name) );
```
mvidner: "Use complete interface up / down instead of link up / down " describes the WHAT of the change but not the WHY.

Besides, the functions are explicitly named Set**Link**Up/Down so it is wrong to make them do something else.

mvidner:
I understand that this is a conversion of a PTF to a MU. Am I right in assuming that the commits before September were all part of the PTF?

mvidner:
Does this correspond to some bug or PR that we did for SLE12? Comparing the fixes would help in the review.

src/modules/LanUdevAuto.ycp
```diff
@@ -210,7 +211,7 @@ global define boolean Write()
 
         string dev = rule["rule"]:"" == "KERNELS" ? FindBy(MatchBusId, rule["value"]:"") : FindBy(MatchMac, rule["value"]:"");
 
-        if( !IsEmpty(dev)))
```
mvidner: This fix should go in a commit preceding this one, saying that it is a fix that *did* make it to the PTF which the customer actually tested.

src/modules/LanUdevAuto.ycp
```diff
@@ -182,12 +182,11 @@ global string FindBy(boolean(map, string) validator, string attr)
 {
     string ret = "";
 
-    foreach(integer index, map devmap, (map<integer, map<string, any> >)LanItems::Items,
+    foreach(map hwmap, ReadHardware("netcard"),
     {
-        // we're interested in configured devices only => return ifcfg name
```
mvidner: Why are we making this change comparing to what the customer has tested?

mchf:
This doesn't correspond to particular PR in SLE-12. This is kind of very minimal and specific implementation of implementation of SLE-12's functionality which was done in several commits. In SLE-12 there is more robust mechanism for device renaming. Moreover, LanUdevAuto module was (finally) dropped from SLE-12. That is one of biggest differences - in SLE-12 is renaming handled at one place / module. In SLE-11 in two (LanItems, LanUdevAuto) depending on way of invocation.

mchf:
Key features implemented in this PR are:
1) udev rules are imported in lan_auto client. In SLE-11 is ```LanUdevAuto::Import``` buried in autoyast code. However, it doesn't work in when applying a profile in target.
2) udev rules are actually written (there was a bug in SCR handling in ```LanUdevAuto```)
3) device's old ifcfg file is dropped

src/modules/LanUdevAuto.ycp
mchf: bcs the patch didn't worked in both cases (installation / running system) properly.

src/routines/routines.ycp
mchf: bcs setting link up / down is just about setting a flag. But we need to reset the device's configuration completely. Otherwise it could happen that when switching from DHCP -> static, the DHCP address could stay on the device.

This part of the patch will be ported into SLE-12.

mvidner: OK, then we should add this to the commit message:
> Otherwise it could happen that when switching from DHCP -> static, the DHCP address could stay on the device.

But what about the other users of SetLinkUp/Down:

```
src/modules/LanUdevAuto.ycp:218:            SetLinkDown( dev);
src/modules/LanUdevAuto.ycp:238:            SetLinkUp(dev);
# ^ these two are newly added, they should be fine
src/modules/LanItems.ycp:402:            SetLinkDown( dev_name);
src/routines/routines.ycp:936:        ret = SetLinkUp( ifc) && ret;
```

will they work fine with this change?

mvidner:
> 3) device's old ifcfg file is dropped

I don't see that done explicitly. Is that part of `LanUdevAuto::Write();`?

mchf:
> I don't see that done explicitly. Is that part of LanUdevAuto::Write();?

You're right. I've removed this in the end. When I tried to implement it using NetworkInterfaces means I was only able to do it when touching (reading) devices cache (the cache contains AY's data at the moment, not the system ones)

mchf:
@mvidner 
There should be:
3) ```keep_install_network=true``` is ignored when importing profile in running system

mvidner:
OK. Maybe you want to make the points 1) 2) 3) more explicit in the changelog, but "apply udev rules on running system" is a good enough summary already.

mchf:
Thanks for review.

PR #440, opened 2016-09-21
Optimize build

jreidinger:


test/test_helper.rb
```diff
@@ -11,6 +11,12 @@
   c.include SCRStub
 end
 
+# stub module to prevent its Import
+# Useful for modules from different yast packages, to avoid build dependencies
+def stub_module(name)
```
imobachgs: np: Maybe having this as part of the `Yast::RSpec` library would be nice so it can be used in other modules.

jreidinger: agree, will do it in future

PR #441, opened 2016-09-21
Bridge handling improved (bsc#962824)

teclator:
I have unified the configuration overview with bridges involved with the bonding one as you can see in the screenshots. There have been fixed also some small issue.

## Bond overview
enslavedbond

## Bridge overview
enslavedbridge

## Old bridge configuration
old_bridge

## New bridge configuration
new_bridge

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Sep 21 11:56:38 UTC 2016 - kanderssen@suse.com
+
+- Bridge handling has been improved (bsc#962824).
+  - "NONE" is shown instead of 0.0.0.0 for old bridge configuration
+  - The bridge master is shown in the enslaved interface.
+  - The interfaces overview is updated after a bridge is modified
+- 3.1.170.1
+
+-------------------------------------------------------------------
 Wed Sep  7 11:23:16 UTC 2016 - mfilka@suse.com
 
 - bnc#993587
--- a/src/include/network/complex.rb
+++ b/src/include/network/complex.rb
@@ -408,7 +408,7 @@ module Yast
       bootproto = devmap["BOOTPROTO"] || "static"
 
       if bootproto.empty? || bootproto == "static"
-        devmap["IPADDR"]
+        devmap["IPADDR"].to_s
       else
         bootproto.upcase
       end
--- a/src/include/network/complex.rb
+++ b/src/include/network/complex.rb
@@ -408,6 +408,8 @@ module Yast
       bootproto = devmap["BOOTPROTO"] || "static"
 
       if bootproto.empty? || bootproto == "static"
+        return "NONE" if devmap["IPADDR"] == "0.0.0.0"
+
         devmap["IPADDR"].to_s
       else
         bootproto.upcase
--- a/src/include/network/complex.rb
+++ b/src/include/network/complex.rb
@@ -401,18 +401,17 @@ module Yast
     # @param [Hash] devmap device map
     # @return textual device protocol
     def DeviceProtocol(devmap)
-      devmap = deep_copy(devmap)
-      if Ops.get_string(devmap, "STARTMODE", "") == "managed"
+      if devmap["STARTMODE"] == "managed"
         # Abbreviation for "The interface is Managed by NetworkManager"
         return _("Managed")
       end
-      ip = Ops.get_string(devmap, "BOOTPROTO", "static")
-      ip = if ip.nil? || ip == "" || ip == "static"
-        Ops.get_string(devmap, "IPADDR", "")
+      bootproto = devmap["BOOTPROTO"] || "static"
+
+      if bootproto.empty? || bootproto == "static"
+        devmap["IPADDR"]
       else
-        Builtins.toupper(ip)
+        bootproto.upcase
       end
-      ip
     end
   end
 end
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -475,25 +475,9 @@ module Yast
     # @param [String] key	id of the widget
     # @param [String] key id of the widget
     def StoreBridge(key, _event)
-      Ops.set(
-        @settings,
-        "BRIDGE_PORTS",
-        String.CutBlanks(
-          Builtins.mergestring(
-            Convert.convert(
-              UI.QueryWidget(Id("BRIDGE_PORTS"), :SelectedItems),
-              from: "any",
-              to:   "list <string>"
-            ),
-            " "
-          )
-        )
-      )
-      Builtins.y2milestone(
-        "store bridge %1 : %2",
-        key,
-        Ops.get_string(@settings, "BRIDGE_PORTS", "")
-      )
+      @settings["BRIDGE_PORTS"] = get_selected_bridges.join(" ")
+
+      log.info("store bridge with ports: #{@settings['BRIDGE_PORTS']}")
 
       nil
     end
@@ -705,7 +689,7 @@ module Yast
         !configured_slaves.include? slave
       end
 
-      Lan.bond_autoconf_slaves = (Lan.bond_autoconf_slaves + new_slaves).uniq.sort
+      Lan.autoconf_slaves = (Lan.autoconf_slaves + new_slaves).uniq.sort
 
       nil
     end
@@ -1595,6 +1579,10 @@ module Yast
     end
 
   private
+
+    def get_selected_bridges
+      UI.QueryWidget(Id("BRIDGE_PORTS"), :SelectedItems) || []
+    end
 
     def get_selected_slaves
       UI.QueryWidget(:msbox_items, :SelectedItems) || []
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -475,6 +475,8 @@ module Yast
     # @param [String] key	id of the widget
     # @param [String] key id of the widget
     def StoreBridge(key, _event)
+      selected_bridge_ports = UI.QueryWidget(Id("BRIDGE_PORTS"), :SelectedItems) || []
+
       @settings["BRIDGE_PORTS"] = selected_bridge_ports.join(" ")
 
       LanItems.bridge_ports = @settings["BRIDGE_PORTS"]
@@ -601,6 +603,7 @@ module Yast
     # @param [String] key	id of the widget
     def InitSlave(_key)
       @settings["SLAVES"] = LanItems.bond_slaves || []
+
       UI.ChangeWidget(
         :msbox_items,
         :SelectedItems,
@@ -679,6 +682,8 @@ module Yast
     def StoreSlave(_key, _event)
       configured_slaves = @settings["SLAVES"] || []
 
+      selected_slaves = UI.QueryWidget(:msbox_items, :SelectedItems) || []
+
       @settings["SLAVES"] = selected_slaves
 
       @settings["BONDOPTION"] = UI.QueryWidget(Id("BONDOPTION"), :Value).to_s
@@ -703,6 +708,8 @@ module Yast
     # @param [Hash] event the event being handled
     # @return true if valid or user decision if not
     def validate_bond(_key, _event)
+      selected_slaves = UI.QueryWidget(:msbox_items, :SelectedItems) || []
+
       physical_ports = repeated_physical_port_ids(selected_slaves)
 
       physical_ports.empty? ? true : continue_with_duplicates?(physical_ports)
@@ -1559,17 +1566,6 @@ module Yast
       nil
     end
 
-    # Obtains the ports selected in the 'BRIDGE_PORTS' widget
-    # @return [Array<String>] selected ports
-    def selected_bridge_ports
-      UI.QueryWidget(Id("BRIDGE_PORTS"), :SelectedItems) || []
-    end
-
-    # Obtains the slaves selected in the 'msbox_items' widget
-    # @return [Array<String>] selected slaves
-    def selected_slaves
-      UI.QueryWidget(:msbox_items, :SelectedItems) || []
-    end
 
     # Initializes the Address Dialog @settings with the corresponding LanItems values
     def initialize_address_settings
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1548,6 +1548,12 @@ module Yast
             Routing.RemoveDefaultGw
           end
         end
+
+        if LanItems.type == "br"
+          LanItems.bridge_ports.split(" ").each do |ifcfg_name|
+            configure_as_bridge_port(ifcfg_name) if old_bridge_config?(ifcfg_name)
+          end
+        end
       end
 
       if LanItems.type == "vlan"
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1332,7 +1332,7 @@ module Yast
       end
       @hostname_initial = String.FirstChunk(Ops.get(host_list, 0, ""), " \t")
 
-      check_bridge_ports_config
+      check_bridge_ports_config if LanItems.type == "br"
 
       initialize_address_settings
 
@@ -1550,15 +1550,13 @@ module Yast
     # of a bridge. In case that some configuration is not correct then adapt it
     # if the user accepts.
     def check_bridge_ports_config
-      return false unless LanItems.type == "br"
-
       ports = LanItems.bridge_ports.split(" ").select { |p| old_bridge_port_config?(p) }
 
-      unless ports.empty?
-        ports.each { |p| configure_as_bridge_port(p) } if adapt_bridge_port_config?(ports)
-      end
+      return if ports.empty?
 
-      true
+      ports.each { |p| configure_as_bridge_port(p) } if adapt_bridge_port_config?(ports)
+
+      nil
     end
 
     # Obtains the ports selected in the 'BRIDGE_PORTS' widget
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -475,7 +475,7 @@ module Yast
     # @param [String] key	id of the widget
     # @param [String] key id of the widget
     def StoreBridge(key, _event)
-      @settings["BRIDGE_PORTS"] = get_selected_bridges.join(" ")
+      @settings["BRIDGE_PORTS"] = selected_bridge_ports.join(" ")
 
       LanItems.bridge_ports = @settings["BRIDGE_PORTS"]
 
@@ -679,7 +679,7 @@ module Yast
     def StoreSlave(_key, _event)
       configured_slaves = @settings["SLAVES"] || []
 
-      @settings["SLAVES"] = get_selected_slaves
+      @settings["SLAVES"] = selected_slaves
 
       @settings["BONDOPTION"] = UI.QueryWidget(Id("BONDOPTION"), :Value).to_s
 
@@ -703,7 +703,7 @@ module Yast
     # @param [Hash] event the event being handled
     # @return true if valid or user decision if not
     def validate_bond(_key, _event)
-      physical_ports = repeated_physical_port_ids(get_selected_slaves)
+      physical_ports = repeated_physical_port_ids(selected_slaves)
 
       physical_ports.empty? ? true : continue_with_duplicates?(physical_ports)
     end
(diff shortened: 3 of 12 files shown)


src/modules/Lan.rb
```diff
@@ -1039,7 +1042,6 @@ def Packages
         # for wlan require iw instead of wireless-tools (bnc#539669)
         "wlan" => "iw",
         "vlan" => "vlan",
-        "br"   => "bridge-utils",
```
teclator: As we are using iptools2 and wicked, this dependency should be removed as this way of configure bridges is currently deprecated.

src/include/network/lan/address.rb
```diff
@@ -1553,8 +1555,6 @@ def AddressDialog
         LanItems.vlan_id = Builtins.tostring(
           Ops.get_integer(@settings, "VLAN_ID", 0)
         )
-      elsif LanItems.type == "br"
-        LanItems.bridge_ports = Ops.get_string(@settings, "BRIDGE_PORTS", "")
```
teclator: Removing this to `StoreBridge` avoid current ports in case that the `bridge ports tab` is not visited.
## Previously

bridge_ports_removed

## After the fix

bridge_ports_removed

src/include/network/lan/address.rb
```diff
@@ -1355,6 +1355,19 @@ def AddressDialog
...
 
+      if LanItems.type == "br"
+        ports = LanItems.bridge_ports.split(" ").select { |p| old_bridge_config?(p) }
+
+        unless ports.empty?
+          message = format(_("The bridge ports listed below are configured in the old way.\n\n" \
+                            "Bridge ports: %s\n\n" \
+                            "Do you want to adapt them now?"), ports.join(", "))
+          if Popup.YesNoHeadline(Label.WarningMsg, message)
+            ports.each { |p| configure_as_bridge_port(p) }
```
teclator: We are just adapting it when the bridge is edited as we can see in this gif:

ask_for_adapt


We could also ask for adapt the configuration when the enslaved interface is edited.

package/yast2-network.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Sep 21 11:56:38 UTC 2016 - kanderssen@suse.com
+
+- Bridge handling has been improved (bsc#962824).
+  - "NONE" is shown instead of 0.0.0.0 for old bridge configuration
+  - The bridge master is shown in the enslaved interface.
+  - The interfaces overview is updated after a bridge is modified
+- 3.1.170.1
```
mchf: if I remember correctly, SP3 didn't diverged so far.

teclator: This change https://github.com/yast/yast-network/pull/440 was the reason I choose that version.

src/include/network/complex.rb
```diff
@@ -401,18 +401,17 @@ def DeviceStatus(devtype, devname, devmap)
...
         # Abbreviation for "The interface is Managed by NetworkManager"
         return _("Managed")
       end
-      ip = Ops.get_string(devmap, "BOOTPROTO", "static")
-      ip = if ip.nil? || ip == "" || ip == "static"
-        Ops.get_string(devmap, "IPADDR", "")
+      bootproto = devmap["BOOTPROTO"] || "static"
+
+      if bootproto.empty? || bootproto == "static"
+        devmap["IPADDR"]
```
teclator: In case of `nil` `bootproto` is set to `"static"` as you can see in line 408. Now we have a unit test for `DeviceProtocol` method and we can see that for `nil`, `""` and `"static"` we return the IP address.

src/include/network/lan/address.rb
```diff
@@ -1596,6 +1593,10 @@ def AddressDialog
 
   private
 
+    def get_selected_bridges
```
mchf: missing documentation for new method

src/include/network/lan/address.rb
```diff
@@ -475,25 +475,11 @@ def InitBridge(key)
...
-            " "
-          )
-        )
-      )
-      Builtins.y2milestone(
-        "store bridge %1 : %2",
-        key,
-        Ops.get_string(@settings, "BRIDGE_PORTS", "")
-      )
+      @settings["BRIDGE_PORTS"] = get_selected_bridges.join(" ")
```
mchf: misleading function name. Something like ```selected_bridge_ports``` would be better in my POV.

src/include/network/lan/address.rb
```diff
@@ -1369,6 +1355,19 @@ def AddressDialog
         Ops.set(@settings, "VLAN_ID", Builtins.tointeger(LanItems.vlan_id))
       end
 
+      if LanItems.type == "br"
```
mchf: i think that this branch deserves own method

src/include/network/lan/bridge.rb
```diff
@@ -34,16 +34,43 @@ def initialize_network_lan_bridge(_include_target)
...
+      ip == "0.0.0.0" ? "" : ip.to_s
+    end
+
+    # Checks if the given interface is using the old way of config enslaved
+    # interfaces with bootproto as static and 0.0.0.0 as IPADDR
+    #
+    # @param [String] interface name
+    # @return [Boolean] returns true if given enslaved interface is configured
+    # in the old way
+    def old_bridge_config?(ifcfg_name)
```
mchf: no bridge but bridge port config

src/include/network/lan/address.rb
mchf: nice ;-) Very common pattern in network code :-/

src/modules/LanItems.rb
```diff
@@ -1311,25 +1339,26 @@ def BuildLanOverview
...
-            ifcfg_name,
-            ifcfg_conf
-          )
+          status = DeviceStatus(ifcfg_type, ifcfg_name, ifcfg_conf)
 
           bullets << _("Device Name: %s") % ifcfg_name
           bullets += startmode_overview(key)
+
+          if bridge_index[ifcfg_name]
+            ip = bridge_ip(ip).empty? ? "NONE" : bridge_ip(ip)
```
mchf: AFAIK this is not correct - bridge ports must not have ip addr (I mean other than 0.0.0.0 in case of pre-wicked configs)

src/modules/LanItems.rb
```diff
@@ -2382,6 +2406,34 @@ def self.publish_variable(name, type)
...
+
+      format("%s: %s", desc, slaves.join(" "))
+    end
+
+    # Check if the given interface is enslaved in a bond or in a bridge
+    #
+    # @return [Boolean] true if enslaved
+    def enslaved?(ifcfg_name)
+      bond_index = BuildBondIndex()
+      return true if bond_index[ifcfg_name] || bridge_index[ifcfg_name]
```
mchf: I'd keep these separated.

src/include/network/lan/address.rb
teclator: I was thinking also on it so just need to decide how to name it

src/modules/LanItems.rb
teclator: If the interface is edited before removed from the bridge for example, in that case the IP will be shown. So in this case we are just replacing `0.0.0.0` in other case there could be some human error or not expected configuration and for that should not be hidden anyway.

src/modules/LanItems.rb
teclator: You always can check directly if it's enslaved in a bridge or in a bond checking the associated hash, so in some way it is separated and this method was just to check if we have to add something to the Note column reusing code, but I can separate it if it makes the code more clear and clean.

mchf: I know. I just thought that it is a bit overkill to ask for the reverse mapping for bond when in fact checking for bridge and vice versa. One of those is always useless the device can be either bridged or bonded, never both.

src/modules/LanItems.rb
mchf: if the interface is edited before removing from bridge, then yast should remove it from bridge. Otherwise you could put system into inconsistent / incorrect state.

src/include/network/lan/address.rb
mchf: yeah .. the hardest part ;-)

src/include/network/complex.rb
mchf: I've addressed ```devmap["IPADDR"]``` here

package/yast2-network.changes
mchf: ok, i didn't remember well ;-)

src/modules/LanItems.rb
teclator: @mchf of course this is also true, but i mean, if there is also ans IP by whatever reason, for example manual edition of the file, then there could be a wrong IP, what to do in that case, show the IP, show ERROR.

The values should be `0.0.0.0` or `NONE` in all the cases, but we never know, and in that unknown scenarios, how you prefer to handle it?

src/modules/LanItems.rb
teclator: So, as you prefer I can remove that method and rewrite the `Note` part of code checking both indexes and writing the note column according to them.

src/include/network/lan/address.rb
```diff
@@ -1593,14 +1546,72 @@ def AddressDialog
 
   private
 
-    def get_selected_bridges
+    # Checks if the current LanItem is a bridge or an enslaved interface part
+    # of a bridge. In case that some configuration is not correct then adapt it
+    # if the user accepts.
+    def check_bridge_ports_config
+      return false unless LanItems.type == "br"
```
mchf: As I try to drop dependency on these "global" status variables I prefer to have new functions stateless (e.g. checking device type on-the-fly as you never know who forgot to reset or set it somewhere to something reasonable). It should also ease "big bang" refactoring of ```LanItems``` one day.

src/include/network/lan/address.rb
```diff
@@ -1593,14 +1546,72 @@ def AddressDialog
...
+
+      ports = LanItems.bridge_ports.split(" ").select { |p| old_bridge_port_config?(p) }
+
+      unless ports.empty?
+        ports.each { |p| configure_as_bridge_port(p) } if fix_bridge_port_config?(ports)
```
mchf: you really want to raise a popup for every old fashioned bridge port? Poor SLE-11 user with 10 ports bridge ;-)

src/include/network/lan/bridge.rb
```diff
@@ -60,6 +87,21 @@ def configure_as_bridge_port(device)
 
       NetworkInterfaces.Commit
       NetworkInterfaces.Add
+
+      NetworkInterfaces.Current = selected_interface
+    end
+
+    # Asks the user about adapt the current bridge port config in case that it
+    # is configured in the old way (BOOTPROTO == static) & (IPADDR == 0.0.0.0).
+    def fix_bridge_port_config?(ports)
```
teclator: Maybe `adapt_bridge_port_config?` could be a better name for this method, and then instead of configure_as_bridge_port define this method as it is an adaption and not a new configuration at all:

```
def adapt_bridge_port_config?(ports)
  ...
end
 
def adapt_bridge_port_config!(ifcfg_name)
  item_id = LanItems.get_configured(ifcfg_name)
  devmap = LanItems.GetDeviceMap(item_id)

  return false unless devmap

  devmap["IPADDR"] = ""
  devmap["NETMASK"] = ""
  devmap["BOOTPROTO"] = "none"
  # take out PREFIXLEN from old configuration (BNC#735109)
  devmap["PREFIXLEN"] = ""

  LanItems.SetDeviceMap(item_id, devmap)

  true
end
```

@mchf what do you think?

src/include/network/complex.rb
teclator: true .. :)

src/include/network/lan/address.rb
teclator: It is only raised once with all the ports in the question

src/include/network/complex.rb
teclator: fixed

src/include/network/lan/address.rb
teclator: Added documentation

src/include/network/lan/address.rb
teclator: Renamed and also renamed get_selected_slaves

src/include/network/lan/bridge.rb
teclator: Renamed

src/include/network/lan/address.rb
teclator: This was exactly the reason I proposed the new method for adapt the config, because the other ones depends on the NetworkInterfaces instance, and it also changed the Current to the last port edited what could produce some unexpected issues.

src/include/network/complex.rb
mchf: thx

src/modules/LanItems.rb
mchf: well if there is no (```IPADDR=0.0.0.0``` and ```BOOTPROTO=static```) or (```BOOTPROTO=none``` and possibly ```IPADDR=<whatever>```) then wicked won't accept that anyway. So in first case is having an ip set invalid and we shouldn't mask it, in second case it doesn't depend and we do not need to take care of it.

src/modules/LanItems.rb
mchf: well, my idea behind this is that one day there will be a generic ```iface_type``` class with ```bridge_type``` and ```bond_type``` successors. If you rewrite it into two, we can copy paste methods as they are into these classes with almost no change.

If it makes sense to you split it. If you think that I'm overengineering things keep it as it is.

teclator: In the future could be as simple as ```"Enslaved in #{interface.master.name}" if interface.master`` or something like that

src/include/network/lan/address.rb
mchf: Just thinking ...
... using ```NetworkInterfaces.GetType``` makes no sense when creating new bridge until new configuration is commited into ```NetworkInterfaces```
... so you can just move the type check out of this method.

src/include/network/lan/address.rb
mchf: sure, my fault.

src/include/network/lan/bridge.rb
```diff
@@ -34,16 +34,43 @@ def initialize_network_lan_bridge(_include_target)
...
+      selected_interface = NetworkInterfaces.Current
       log.info("Adapt device #{device} as bridge port")
 
       # when using wicked every device which can be bridged
       # can be set to BOOTPROTO=none. No workaround with
       # BOOTPROTO=static required anymore
-      NetworkInterfaces.Edit(device)
+      unless NetworkInterfaces.Edit(device)
+        Lan.autoconf_slaves += [device] unless Lan.autoconf_slaves.include? device
+        return false
```
mchf: inconsistent return value

src/include/network/lan/bridge.rb
```diff
@@ -60,6 +87,21 @@ def configure_as_bridge_port(device)
 
       NetworkInterfaces.Commit
       NetworkInterfaces.Add
+
+      NetworkInterfaces.Current = selected_interface
```
mchf: why to return original ```Current```?

src/include/network/lan/bridge.rb
mchf: well it better describes current use case, but i don't see any benefit in having such one purpose method. So do you have any other pros? ;-)

src/include/network/lan/bridge.rb
```diff
@@ -34,16 +34,43 @@ def initialize_network_lan_bridge(_include_target)
...
     def configure_as_bridge_port(device)
+      selected_interface = NetworkInterfaces.Current
       log.info("Adapt device #{device} as bridge port")
 
       # when using wicked every device which can be bridged
       # can be set to BOOTPROTO=none. No workaround with
       # BOOTPROTO=static required anymore
-      NetworkInterfaces.Edit(device)
+      unless NetworkInterfaces.Edit(device)
+        Lan.autoconf_slaves += [device] unless Lan.autoconf_slaves.include? device
```
mchf: well this autoconf stuff is part of a fate request for bonding devices where it was requested to do an udev update for slaves. As it was not requested for bridges, it shouldn't be implemented for bridges in my POV.

teclator: Ok, as we was talking some time ago about unifying bonding and bridge management as much as possible was what I tried but this is going too further and some changes should not be necessary and also invalid, so taking this in account I'm going to modify the PR in concordance.

test/bridge_test.rb
```diff
@@ -6,87 +6,167 @@
...
 
-    let(:expected_bridgeable) do
-      [
-        "bond0",
-        "eth4",
-        "eth11",
-        "eth12",
-        "tap0"
-      ]
+  describe "#old_bridge_port_config?" do
```
mvidner: please add a test for the "eth1" case , which has `  { "BOOTPROTO" => "none" } ` and so should return false

test/bridge_test.rb
```diff
@@ -6,87 +6,167 @@
...
+        "eth5" => { "BOOTPROTO" => "static", "STARTMODE" => "nfsroot" },
+        "eth6" => { "BOOTPROTO" => "static", "STARTMODE" => "ifplugd" }
+      },
+      "tun"  => {
+        "tun0" => {
+          "BOOTPROTO" => "static",
+          "STARTMODE" => "onboot",
+          "TUNNEL"    => "tun"
+        }
+      },
```
mvidner: This case is not documented in the new yard comment

src/include/network/lan/bridge.rb
```diff
@@ -34,32 +34,97 @@ def initialize_network_lan_bridge(_include_target)
...
+        end
+        NetworkInterfaces.Current["_aliases"] = {}
+
+        NetworkInterfaces.Commit
+        NetworkInterfaces.Add
 
-      # remove all aliases (bnc#590167)
-      aliases = NetworkInterfaces.Current["_aliases"] || {}
-      aliases.each do |alias_name, alias_ip|
-        NetworkInterfaces.DeleteAlias(device, alias_name) if alias_ip
```
mvidner: Sorry, I would be totally baffled as a user, and would not know what to answer. Why are we even asking?

src/include/network/lan/address.rb
mvidner: What is there for the user to decide? IMHO this pop-up is harmful, burdening the user with a decision they cannot understand ("old way, new way"?)
I propose that we simply adapt the config and just log the action.

src/include/network/lan/bridge.rb
mvidner: @teclator what is this method supposed to return in this case, and in the normal case? Add a @return tag?

src/modules/LanItems.rb
mvidner: The string `"NONE"` is wrong (I guess you have blindly copied it from the bug reporter) because it is not translated and it is not an abbreviation. We should either use `""` ( it will be used in a table cell so that is OK) or `_("none")`.

src/include/network/lan/address.rb
```diff
@@ -475,25 +475,11 @@ def InitBridge(key)
...
-        )
-      )
-      Builtins.y2milestone(
-        "store bridge %1 : %2",
-        key,
-        Ops.get_string(@settings, "BRIDGE_PORTS", "")
-      )
+      @settings["BRIDGE_PORTS"] = selected_bridge_ports.join(" ")
+
+      LanItems.bridge_ports = @settings["BRIDGE_PORTS"]
```
mvidner: NP: So `@settings["BRIDGE_PORTS"]` is a String but `@settings["SLAVES"]` is an Array. WTF. I know this is a bit out of focus, but could we at least document it and add a FIXME to unify it?

src/modules/LanItems.rb
```diff
@@ -732,7 +732,7 @@ def IsBondable(bondMaster, itemId)
     # @param [Fixnum] itemId        index into LanItems::Items
     # TODO: bridgeMaster is not used yet bcs detection of bridge master
```
mvidner: This TODO is no longer valid

src/include/network/lan/address.rb
teclator: @mvidner 

It was a doubt I have  at the beginning of the PBI as it was also commented in the card. After discussing it with @mchf the conclusion was that automatic conversion does not means not ask the user about it.

We could do it automatically, in that case when? when the interface is edited and saved or also do it although the user press cancel in the edition. If should be converted so why not just have a method that checks current interface configuration when the client is called and adapt old config style?

We just decided to prompt the user just to verify he really wanted to adapt the current configuration and also being a litle conservative taking in account that if something change in current workflow the network is restarted and should not restarted if not (so maybe the user does not expect changes when `OK` pressed but then he realized that the network was restarted and some files changed magically.

src/modules/LanItems.rb
teclator: I just used the same was used in bonding that is the BOOTPROTO upcased

https://github.com/yast/yast-network/blob/master/src/modules/LanItems.rb#L1272

So I could adapt both for use `_("none")` in that case, or just with the enslaved note should be enough and just let that column as blank

src/include/network/lan/address.rb
teclator: Sure, and is how differs also in configuration per interface. As bonding master is configured as:
```
BONDING_SLAVE0='eth0'
BONDING_SLAVE1='eth1'
```
while bridge_ports are configured as 
```
BRIDGE_PORTS='eth0 eth1'
```

src/include/network/lan/bridge.rb
teclator: At the end i decided to just let it as it was, and it always returns true and just add the changes if `NetworkInterfaces.Edit(device)` success and also remember the previous Interfaces selected by NetworkInterface to avoid let it in a different one what could produce not desired results.

https://github.com/teclator/yast-network/blob/bridge_overview_improvement/src/include/network/lan/bridge.rb#L92

As an alternative I added the adapt_old_port_config method which avoid to work over NetworkInterfaces and where we could also add the logic to write the config to the files without restarting the network

https://github.com/teclator/yast-network/blob/bridge_overview_improvement/src/include/network/lan/bridge.rb#L113

src/include/network/lan/bridge.rb
teclator: is not returned anymore, i let it as it was returning true again

teclator:
I have doubts about what to do with the Popup and also about refreshing all the interfaces that were added to the bridge, as currently just the interfaces configuration was written but the interface was not updated, and you could see them as not configured and things like that.

So i was in doubt if just duplicate current behavior or try to unify it with bonding.

src/include/network/lan/bridge.rb
```diff
@@ -34,32 +34,97 @@ def initialize_network_lan_bridge(_include_target)
       textdomain "network"
     end
 
+    # It returns an empty string if given IP is 0.0.0.0, otherwise returns
+    # the given IP
+    #
+    # @param [String] ip
+    # @return [String] empty string if given ip is 0.0.0.0; given ip
+    # otherwise
+    def bridge_ip(ip)
```
mchf: it should be ```bridge_port_ip```

teclator:
Have been replaced by #448

PR #444, opened 2016-09-29
Do a wicked ifdown instead of a ip link set down (bsc#991694) 2nd edition

mchf:
Extending https://github.com/yast/yast-network/pull/424


src/include/network/routines.rb
```diff
@@ -712,68 +712,44 @@ def GetAllInterfaces
...
+    # Calls wicked ifup if it is the currenct backend, set the link up
+    # otherwise
     #
-    # @param [String] name of interface to 'set link up'
-    def SetLinkUp(dev_name)
-      log.info("Setting link up for interface #{dev_name}")
-      Run("ip link set #{dev_name} up")
+    # @param [String] name of interface to put down
+    def SetIfaceUp(dev_name)
+      log.info("Setting interface #{devname} up")
```
teclator: just decide if you want to use dev_name or devname

src/include/network/routines.rb
```diff
@@ -712,68 +712,44 @@ def GetAllInterfaces
...
+    # Calls wicked ifdown if it is the current backend, set the link down
+    # otherwise
     #
-    # @param [String] name of interface to 'set link down'
-    def SetLinkDown(dev_name)
-      log.info("Setting link down for interface #{dev_name}")
-      Run("ip link set #{dev_name} down")
+    # @param [String] name of interface to put down
+    def SetIfaceDown(dev_name)
+      log.info("Setting interface #{devname} down")
```
teclator: Same as before

src/include/network/routines.rb
```diff
@@ -712,68 +712,44 @@ def GetAllInterfaces
...
 
-      interfaces.all? { |i| SetLinkDown(i) }
-    end
-
-    # Calls wicked ifdown if it is the current backend, set the link down
-    # otherwise
-    #
-    # @param [String] name of interface to put down
-    # @see SetLinkDown
-    def SetInterfaceDown(dev_name)
```
teclator: The idea behind this was to maintain compatible with iptools2 in case that wicked was not installed and `ifdown` not used replaced by `ip link set ` as we was using until now.

But I'm fine just doing an ifdown ang giving that responsibility to wicked.

src/lib/network/wicked.rb
```diff
@@ -16,27 +16,5 @@ def reload_config(devs)
 
       SCR.Execute(BASH_PATH, "wicked ifreload #{devs.join(" ")}").zero?
     end
-
-    # Does an ifdown for each device named in devs
-    #
-    # @param [Array] list of device names
-    # @return [Boolean] true if devices bring down; false otherwise
-    def bring_down(devs)
```
teclator: It makes sense to remove both methods if we are not using them, and in the future will see. Also don't know exactly if there are some benefits between called it with multiple interfaces or calling it once per interface.

src/include/network/routines.rb
```diff
@@ -711,36 +712,44 @@ def GetAllInterfaces
...
     #
     # @return [boolean] false if some of interfaces cannot be set up
-    def SetAllLinksUp
+    def SetAllIfacesUp
       interfaces = GetAllInterfaces()
 
       return false if interfaces.empty?
 
-      interfaces.all? { |i| SetLinkUp(i) }
+      interfaces.all? { |i| SetIfaceUp(i) }
```
teclator: When wicked is the backend maybe could be interested to call `wicked ifup interfaces.join(" ")` from a performance point of view.

test/routines_test.rb
```diff
@@ -156,3 +156,59 @@ def initialize
...
+  context "when the NetworkService is not wicked" do
+    it "sets down the link of interface with iproute2 utils" do
+      expect(routines).to receive(:SetLinkDown).with(interface)
+
+      routines.SetInterfaceDown(interface)
+    end
+  end
+end
+
+describe "SetInterfaceUp" do
```
teclator: Not needed anymore, or should be renamed.

test/routines_test.rb
```diff
@@ -156,3 +156,59 @@ def initialize
     expect(routines.physical_port_id?("eth0")).to eql(true)
   end
 end
+
+describe "SetInterfaceDown" do
```
teclator: As commented below

test/wicked_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  it "returns true if given devices reload successfully" do
+    expect(Yast::SCR).to receive(:Execute)
+      .with(DummyNetwork::BASH_PATH, "wicked ifreload eth0 eth1").and_return(0)
+
+    expect(subject.reload_config(["eth0", "eth1"])).to eql(true)
+  end
+end
+
+describe "#bring_up" do
```
teclator: Method was removed

test/wicked_test.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  it "returns true if given devices bring up successfully" do
+    expect(Yast::SCR).to receive(:Execute)
+      .with(DummyNetwork::BASH_PATH, "wicked ifup eth0 eth1").and_return(0)
+
+    expect(subject.bring_up(["eth0", "eth1"])).to eql(true)
+  end
+end
+
+describe "bring_down" do
```
teclator: Method was removed

src/modules/LanItems.rb
```diff
@@ -543,7 +543,7 @@ def WriteUdevItemsRules
           # removing is less error prone when tracking name changes, so it was chosen.
           item_udev_net = RemoveKeyFromUdevRule(item_udev_net, "KERNEL")
           # setting links down during AY is forbidden bcs it can freeze ssh installation
-          SetLinkDown(dev_name) if !Mode.autoinst
+          SetInterfaceDown(dev_name) if !Mode.autoinst
```
teclator: This call has to be renamed to `SetIfaceDown`

src/include/network/routines.rb
```diff
@@ -28,6 +28,7 @@
 #
 module Yast
   module NetworkRoutinesInclude
+    include Wicked
```
teclator: Is not needed.

teclator:
LGTM

src/modules/LanItems.rb
```diff
@@ -430,7 +430,13 @@ def ReplaceItemUdev(replace_key, new_key, new_val)
       operator = new_key == "NAME" ? "=" : "=="
       current_rule = getUdevFallback
```
teclator: What I don't link about this method is that calls getUdevFallback which creates a rule if not exist inside a method called ReplaceItemUdev, and then calling again RemovekeFromUdevRule over (getUdevFallback) and then comparing again new rule with current rule is alitle strange for me.

mchf: yes it is. However, the code somehow relies on that and I don't plan to add another change into this PR. I don't want to hijack the PR to far from original issue ;-)

teclator: I'm fine with that POV so just check that we have a test for this case and we don't let some corner case uncovered, I can check if if you want.

teclator: Tested manually, and also has unit tests so LGTM.

mchf:
Thanks for review and finally merging.

PR #445, opened 2016-10-05
Install WIFI packages during system installation bug#1002700

ancorgs:


src/include/network/lan/wizards.rb
```diff
@@ -55,11 +55,12 @@ def LanSequence
       aliases = {
         "read"     => [-> { ReadDialog() }, true],
         "main"     => -> { MainSequence("") },
-        "packages" => [-> { PackagesInstall(Lan.Packages) }, true],
         "write"    => [-> { WriteDialog() }, true]
```
mvidner: This change makes RuboCop fail with "Align the elements of a hash literal if they span more than one line."
W!T!F!

package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        3.1.171
+Version:        3.2.1
```
mchf: different version number than in spec

ancorgs: I had to "correct history" (like dictators do) because the previous patch didn't jump to 3.2.0 as expected.

mchf: ah sorry ... for second view I can see ...

PR #446, opened 2016-10-06
Clean up - Updated spec according build service

mchf:


package/yast2-network.spec
```diff
@@ -23,13 +23,13 @@ Release:        0
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
 
-Group:          System/YaST
-License:        GPL-2.0
 # should be required by devtools
-BuildRequires:  perl-XML-Writer pkgconfig rpm
+BuildRequires:  perl-XML-Writer
+BuildRequires:  pkgconfig
```
jreidinger: why it is needed? I do not see any pkgconfig usage

package/yast2-network.spec
```diff
@@ -23,13 +23,13 @@ Release:        0
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
 
-Group:          System/YaST
-License:        GPL-2.0
 # should be required by devtools
-BuildRequires:  perl-XML-Writer pkgconfig rpm
+BuildRequires:  perl-XML-Writer
```
jreidinger: I think it can be dropped

package/yast2-network.spec
```diff
@@ -23,13 +23,13 @@ Release:        0
...
+BuildRequires:  pkgconfig
+BuildRequires:  rpm
```
jreidinger: useless, can be removed

mchf:
In general I have no idea about those requires you highlighted. It is just what I got back from BS ;-) I'll update it.

jreidinger:
@mchf LGTM, in general if you remove BuildRequire and still can build package, then it is not needed :wink:

package/yast2-network.spec
```diff
@@ -23,13 +23,10 @@ Release:        0
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
 
-Group:          System/YaST
-License:        GPL-2.0
 # should be required by devtools
-BuildRequires:  perl-XML-Writer pkgconfig rpm
```
mvidner: Sorry, now you're mixing up formatting changes done by `/usr/lib/obs/service/format_spec_file.files/prepare_spec` with semantic changes to the dependencies. I do not care if we do either of them, but I do care to not mix them in a single commit.

package/yast2-network.spec
```diff
@@ -24,9 +24,6 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 Source0:        %{name}-%{version}.tar.bz2
 
 # should be required by devtools
```
mvidner: The comment belongs to the deleted lines, delete it too.

mchf:
thanks for reviews

PR #448, opened 2016-10-06
Alternative improve bridge handling (bsc#962824)

teclator:
This in alternative to #441 and which depends on yast/yast-yast2#500. 

The idea is to adapt the configuration internally when the `ifcfg-*` files are read instead of when the bridge or interface enslaved where edited. The files will only be written in case that the user change some configuration but if he edit the enslaved interface or just in the **LanOverview** table he will not see the static configuration.

I have unified the configuration overview with bridges involved with the bonding one as you can see in the screenshots. There have been fixed also some small issue.

## Bond overview
enslavedbond

## Bridge overview
enslavedbridge

## Old bridge configuration
old_bridge

## New bridge configuration
new_bridge

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,15 @@
 -------------------------------------------------------------------
+Mon Oct 10 11:56:38 UTC 2016 - kanderssen@suse.com
+
+- Bridge handling has been improved (bsc#962824).
+  - "NONE" is shown instead of 0.0.0.0 for old bridge configuration
+  - The bridge master is shown in the enslaved interface.
+  - The interfaces overview is updated after a bridge is modified
+  - The interfaces enlsaved are not lost when save in a different
+    tab.
+- 3.1.170.3
+
+-------------------------------------------------------------------
 Fri Oct  7 15:16:11 UTC 2016 - kanderssen@suse.com
 
 - If an interface is not configured yet then just set the interface
--- a/src/data/network/sysconfig_defaults.yml
+++ b/src/data/network/sysconfig_defaults.yml
@@ -42,3 +42,4 @@ BONDING_MODULE_OPTS: mode=active-backup miimon=100
 TUNNEL_SET_OWNER: ''
 TUNNEL_SET_GROUP: ''
 IPOIB_MODE: connected
+BRIDGE_PORTS: ''
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -125,6 +125,11 @@ module Yast
       @wl_default_key = 0
       @wl_nick = ""
 
+      # FIXME: We should unify bridge_ports and bond_slaves variables
+
+      # interfaces attached to bridge (list delimited by ' ')
+      @bridge_ports = ""
+
       # bond options
       @bond_slaves = []
       @bond_option = ""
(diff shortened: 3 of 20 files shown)


package/yast2-network.changes
```diff
@@ -1,10 +1,19 @@
...
+
+-------------------------------------------------------------------
 Mon Sep 12 09:40:57 UTC 2016 - kanderssen@suse.com
 
-- Use full interface up / down instead of setting just link 
+- Use full interface up / down instead of setting just link
   up / down to allow proper update of interface's configuration.
   (bsc#991694)
-- 3.1.171
+- 3.1.170.1
```
mchf: in my POV you don't need the fourth number as we decided that SP3 will use 3.2.x format

src/modules/LanItems.rb
```diff
@@ -1005,53 +1011,26 @@ def Read
...
-        end
-        if pos.nil?
-          pos = Builtins.size(@Items)
-          Ops.set(@Items, pos, {})
+      interfaces.each do |confname|
+        @Items.each do |key, value|
+          match = value.fetch("hwinfo", {}).fetch("dev_name", "") == confname
+          @Items[key]["ifcfg"] = confname if match
         end
-        Ops.set(@Items, [pos, "ifcfg"], confname)
```
teclator: Previously we always set the "ifcfg" attribute for Items with hwinfo and also for Items without it, so the loop below does nothing at the end.

So we could keep the method above just with some cleanup, or this alternative that only sets the ifcfg attribute as the interface name if match, and makes the loop below useful.

mchf: nice

test/complex_test.rb
```diff
@@ -13,56 +14,95 @@ def initialize
...
+  describe "#DeviceProtocol" do
+    let(:ipaddr) { "192.168.0.120" }
+    let(:managed) { { "STARTMODE" => "managed" } }
+    let(:static) { { "BOOTPROTO" => "static", "IPADDR" => ipaddr } }
+    let(:empty) { { "BOOTPROTO" => "", "IPADDR" => ipaddr } }
+    let(:no_bootproto) { { "IPADDR" => ipaddr } }
+    # IPADDR are set just to show that returns the protocol instead of the IP
+    let(:dhcp) { { "BOOTPROTO" => "dhcp", "IPADDR" => ipaddr } }
+    let(:none) { { "BOOTPROTO" => "none", "IPADDR" => ipaddr } }
+    let(:zero) { { "BOOTPROTO" => "static", "IPADDR" => ipaddr } }
```
mchf: never used

teclator: true, i copied and pasted to added to use it with 0.0.0.0 address but just used merge at the end, so I will remove or fix it. Thnx

mchf:
@teclator I've added one simple update

src/data/network/sysconfig_defaults.yml
```diff
@@ -42,3 +42,4 @@ BONDING_MODULE_OPTS: mode=active-backup miimon=100
 TUNNEL_SET_OWNER: ''
 TUNNEL_SET_GROUP: ''
 IPOIB_MODE: connected
+BRIDGE_PORTS: ''
```
teclator: :+1:

test/complex_test.rb
teclator: Removed

mchf:
lgtm

mchf:
wait a second please ... one more test came to my mind ;-)

mchf:
ok, worked

package/yast2-network.changes
teclator: It diverged from master in version 170, as the 3.1.171 was not backported, and it was before the change to 3.2 version. This has also been discussed in the mailing list.

PR #449, opened 2016-10-07
We cannot just remove the handle of link state (bsc#991694)

teclator:
If an interface is not configured yet then if we try to bring it up with wicked and maybe with NetworkManager will complain. We have to set the link up and them will handle the event.

### DIFF
--- a/src/include/network/routines.rb
+++ b/src/include/network/routines.rb
@@ -727,20 +727,20 @@ module Yast
       Run("ip link set #{dev_name} down")
     end
 
-    # Calls wicked ifdown with the given interface
-    #
-    # @param [String] name of interface to put down
-    def SetIfaceDown(dev_name)
-      log.info("Setting interface #{dev_name} down")
-      Run("ifdown #{dev_name}")
-    end
-
     # Calls wicked ifup with the given interface
     #
     # @param [String] name of interface to put down
     def SetIfaceUp(dev_name)
       log.info("Setting interface #{dev_name} up")
       Run("ifup #{dev_name}")
+    end
+
+    # Calls wicked ifdown with the given interface
+    #
+    # @param [String] name of interface to put down
+    def SetIfaceDown(dev_name)
+      log.info("Setting interface #{dev_name} down")
+      Run("ifdown #{dev_name}")
     end
 
     # Tries to set all interfaces up
--- a/src/include/network/routines.rb
+++ b/src/include/network/routines.rb
@@ -711,22 +711,36 @@ module Yast
       end
     end
 
-    # Calls wicked ifup if it is the currenct backend, set the link up
-    # otherwise
+    # Wrapper to call 'ip link set up' with the given interface
     #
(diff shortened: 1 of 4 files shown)


src/include/network/routines.rb
```diff
@@ -751,6 +765,28 @@ def SetAllIfacesDown
...
 
+    # Tries to set up the link of all available interfaces
+    #
+    # @return [boolean] false if some of interfaces cannot be set up
+    def SetAllLinksUp
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
+
+      interfaces.all? { |i| SetLinkUp(i) }
```
ancorgs: This will stop trying to get interfaces up as soon as the one fails

src/include/network/routines.rb
```diff
@@ -751,6 +765,28 @@ def SetAllIfacesDown
...
+
+    # Tries to set the link down of all available interfaces
+    #
+    # @return [boolean] false if some of interfaces cannot be set down
+    def SetAllLinksDown
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
+
+      interfaces.all? { |i| SetLinkDown(i) }
```
ancorgs: Same comment than above

src/include/network/routines.rb
```diff
@@ -733,22 +747,56 @@ def SetIfaceDown(dev_name)
...
+    end
+
+    # Tries to set the link down of all available interfaces
+    #
+    # @return [boolean] false if some of interfaces cannot be set down
+    def SetAllLinksDown
+      all_up = true
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
```
ancorgs: Same than above

src/include/network/routines.rb
```diff
@@ -733,22 +747,56 @@ def SetIfaceDown(dev_name)
...
+    #
+    # @return [boolean] false if some of interfaces cannot be set up
+    def SetAllLinksUp
+      all_up = true
       interfaces = GetAllInterfaces()
 
       return false if interfaces.empty?
 
-      interfaces.all? { |i| SetIfaceDown(i) }
+      interfaces.each { |i| all_up = false if !SetLinkUp(i) }
```
ancorgs: use unless

src/include/network/routines.rb
```diff
@@ -733,22 +747,56 @@ def SetIfaceDown(dev_name)
...
+    # Tries to set the link down of all available interfaces
+    #
+    # @return [boolean] false if some of interfaces cannot be set down
+    def SetAllLinksDown
+      all_up = true
+      interfaces = GetAllInterfaces()
+
+      return false if interfaces.empty?
+
+      interfaces.each { |i| all_up = false if !SetLinkDown(i) }
```
ancorgs: Same than before, `unless` just if you want

ancorgs:
LGTM

PR #451, opened 2016-10-07
Merge changes from SLE12-SP2 (#449)

teclator:


mchf:
lgtm

src/include/network/routines.rb
```diff
@@ -742,24 +756,62 @@ def SetIfaceDown(dev_name)
 
     # Tries to set all interfaces up
     #
-    # @return [boolean] false if some of interfaces cannot be set up
+    # @return [boolean] false if interfaces are empty or some
+    # interface cannot be bring up
```
mvidner: Sorry, but this is plain wrong. This is not how one handles the empty set. And you would be able to see it if there was an actual use of this API, which there isn't.

src/include/network/routines.rb
```diff
@@ -742,24 +756,62 @@ def SetIfaceDown(dev_name)
...
     def SetAllIfacesUp
+      all_up = true
       interfaces = GetAllInterfaces()
 
       return false if interfaces.empty?
 
-      interfaces.all? { |i| SetIfaceUp(i) }
+      interfaces.each { |i| all_up = false if !SetIfaceUp(i) }
+
+      all_up
```
teclator: As you commented only `SetAllLinksUp()` is currently in use. At least in yast2-network code and I also don't expect that someone is using `SetAllLinksDown()`. `SetAllIfacesUp` and `SetAllIfacesDown()` was introduced in the reverted commit and I fixed it basically based in CR and as they use the same code structure and also the main purpose of the PR was to revert the removement of `set link down`.

src/include/network/routines.rb
teclator: I just tried to comment what the API was doing in case of no interfaces, that could be unexpected as for example @ancorgs expected `true` instead of false if there was no interfaces to bring up. 

And we don't want to change the API at least for the SP2 release but we could consider it in master.

src/include/network/routines.rb
mchf: bcs @teclator did a "revert". ```SetLinksDown``` used to be used but as time passed it was replaced by per interface approach and the ```SetLinksDown``` has become to be useless.

src/include/network/routines.rb
mchf: The API is in fact (almost) created from scratch bcs it's original usage was replaced by ```SetIface*``` methods

src/include/network/routines.rb
mvidner: OK, this is a difference in fixing a maintenance branch and cleaning up for master. I now see that of the 8 methods (All or Single, Interface or Link, Up or Down), only 3 are used, but the others could be kept for symmetry. If we make up our minds about the return values (see my other comment) they should become more readable along the way.

mvidner: My proposal is: if a call fails, we cannot react to it anyway. Discard the return value and let the `SetAll*` methods become oneliners:

```rb
# Tries to set all interfaces up. Errors are logged.
# @return [void]
def SetAllIfacesUp
  GetAllInterfaces().each { |i| SetIfaceUp(i) }
end
```

teclator:
In general, it needs a discussion about the API, because as you commented it is used as a procedural but declared as a function ignoring the return method, but from other point of view it could be useful.

And in the per interface method, the result of calling Run is returned but should be documented.

mvidner:
I thought about documenting the return value of the single interface methods, but then I proposed to discard them anyway ;-)

src/include/network/routines.rb
```diff
@@ -722,44 +722,41 @@ def GetAllInterfaces
...
-    #
-    # @return [boolean] false if some of interfaces cannot be set down
-    def SetAllIfacesDown
-      interfaces = GetAllInterfaces()
-
-      return false if interfaces.empty?
-
-      interfaces.all? { |i| SetIfaceDown(i) }
+    # Tries to set the link up of all available interfaces
+    def SetAllLinksUp
```
teclator: This method will be also removed in #453 when the only client that is using it be removed.

PR #454, opened 2016-10-11
Optimize hosts read

jreidinger:
WIP, will be blog entry

rationale for this pr is bug report https://bugzilla.suse.com/show_bug.cgi?id=877047
even with this change I get to less then 20% of original time and found more possible optimization in cfa that will even speed it up.

content for blog:
We get bug report, that yast2 network is slow when there is a lot of entries in /etc/hosts file. We take it as opportunity to tost new profiler support in yast. Profiler shows that problem is in SCR calls, which took a lot of time, so another oppurtinity to test new framework cfa with augeas parser. Augeas have already lense for /etc/hosts file, so it is quite easy to add it. In short here are results for network CLI with /etc/hosts containing ~ 10 000 entries:

before:
```
time yast2 lan.rb list
id	name, 			bootproto
0	Intel Ethernet controller, DHCP


real	1m15.735s
user	1m15.076s
sys	0m0.164s

```

after:
```
time yast2 lan list
id	name, 			bootproto
0	Intel Ethernet controller, DHCP


real	0m19.079s
user	0m18.348s
sys	0m0.244s
```

We also improved cfa itself
<https://github.com/config-files-api/config_files_api/pull/10> which we speed up by 50%.


src/modules/Host.rb
```diff
@@ -121,27 +110,20 @@ def EnsureHostnameResolvable
...
-          names = Convert.convert(
-            SCR.Read(
-              Builtins.topath(Builtins.sformat(".etc.hosts.\"%1\"", host))
-            ),
-            from: "any",
-            to:   "list <string>"
-          )
-          next { host => names } if names != []
-        end
+      if Ops.greater_than(SCR.Read(path(".target.size"), CFA::Hosts::PATH), 0)
```
mchf: when doing cleanup this should be replaced too ;-)

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,147 @@
...
+      aliases.each do |a|
+        aliases_col.add(a)
+      end
+      data.add(unique_id, entry_line)
+    end
+
+    def remove_hostname(hostname)
+      entries = data.select(hostname_matcher(hostname))
+      entries.each do |entry|
+        entry = entry[:value]
```
mchf: overwriting variable used for iterating in loop is usually not considered to be a good idea.

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,147 @@
...
+          reduced_aliases.delete(hostname)
+          set_host(entry["ipaddr"], entry["canonical"], reduced_aliases)
+        end
+      end
+    end
+
+  private
+
+    def ip_matcher(ip)
+      Matcher.new { |k, v| v["ipaddr"] == ip }
```
mchf: ```k``` is not used so it should be ```_```

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    def hostname_matcher(hostname)
+      Matcher.new do |k, v|
```
mchf: as above

src/modules/Host.rb
jreidinger: well, cfa currently do not have method to recognize if file is not empty or exists. It raise exception. Of course I can use some cli tool to get it

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,147 @@
...
+  class Hosts < BaseModel
+    PARSER = AugeasParser.new("hosts.lns")
+    PATH = "/etc/hosts".freeze
+    include Yast::Logger
+
+    def initialize(file_handler: nil)
+      super(PARSER, PATH, file_handler: file_handler)
+    end
+
+    def hosts
```
jreidinger: ok

src/modules/Host.rb
```diff
@@ -215,34 +165,40 @@ def Write
...
     # @return autoinstallation settings
     def Export
-      return {} if @hosts.empty?
+      exported_hosts = @hosts.hosts
+      return {} if exported_hosts.empty?
 
       # Filter out IPs with empty hostname (so that valid autoyast
       # profile is created)(#335120)
       # FIXME: this also removes records with empty names from @hosts. Such
       # side effect is unexpected and should be removed.
```
mchf: if I get it correctly this FIXME is no longer true

jreidinger: you are right

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,147 @@
+require "yast"
+require "yast2/target_file"
+
+require "cfa/base_model"
+require "cfa/matcher"
+require "cfa/augeas_parser"
+
+module CFA
+  class Hosts < BaseModel
```
mvidner: For the people that did not design CFA::BaseModel, a comment with the full URL of its API docs will be useful: http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/BaseModel

jreidinger: updated

test/data/hosts
```diff
@@ -0,0 +1,24 @@
...
+::1             localhost ipv6-localhost ipv6-loopback
+
+fe00::0         ipv6-localnet
+
+ff00::0         ipv6-mcastprefix
+ff02::1         ipv6-allnodes
+ff02::2         ipv6-allrouters
+ff02::3         ipv6-allhosts
+10.100.128.72   pepa.labs.suse.cz pepa pepa2
+10.254.128.01   pepa1.labs.suse.cz pepa1
```
mvidner: The "01" is incidental complexity, make that "1".

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,168 @@
...
+    PARSER = AugeasParser.new("hosts.lns")
+    PATH = "/etc/hosts".freeze
+    include Yast::Logger
+
+    def initialize(file_handler: nil)
+      super(PARSER, PATH, file_handler: file_handler)
+    end
+
+    # returns old format of hosts used in Host module.
+    # @return [Hash<Array<String>>] format is hash where key is ip and value
```
mvidner: I will rewrite this piece

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,168 @@
...
+        data.delete(e[:key])
+      end
+    end
+
+    # replaces or adds new host entry. If more then one entry with given ip exists
+    # then replaces the last instance
+    def set_host(ip, canonical, aliases = [])
+      entries = data.select(ip_matcher(ip))
+      if entries.empty?
+        log.info "adding new entry for ip #{ip}"
```
mvidner: Ahem, `add_entry(ip, canonical, aliases)` instead?

jreidinger: hmm, that cause another search thrue tree searching for ip

jreidinger:
jenkins now working, travis have strange issue with augeas bindings, I suggest to ignore it

src/lib/cfa/hosts.rb
mvidner: What? The body of this `if` block is an exact copy of the body of that method.

jreidinger: ah, right, add_entry do not search for existing ips, so I will adapt

src/modules/Host.rb
```diff
@@ -27,6 +27,8 @@
 # Authors:	Michal Svec <msvec@suse.cz>
```
jreidinger: done

src/lib/cfa/hosts.rb
jreidinger: done

src/lib/cfa/hosts.rb
jreidinger: done

src/lib/cfa/hosts.rb
jreidinger: done

mvidner:
Regarding the API of CFA::Hosts: the method naming is quite bad! The class is named Hosts (that's OK) and every single public instance method includes the word "host". But those methods are not dealing with hosts! Instead, they deal with **ips**, **names**, and whole ip+names **entries**

- add_host -> add_entry
- delete -> delete_by_ip
- remove_hostname -> delete_by_name
- set_host -> set_entry

Then we have the legacy API which supports the doubly nested list (`["log.example.org log", "sql.example.org sql"]`)
- host
- hosts

but we are missing what I thought would be obvious candidates,
- find_by_ip
- find_by_name

src/lib/cfa/hosts.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      entries.each do |e|
+        log.info "deleting record #{e.inspect}"
+        data.delete(e[:key])
+      end
+    end
+
+    # Replaces or adds a new host entry.
+    # If more than one entry with the given ip exists
+    # then it replaces the last instance.
```
jreidinger: you are right, Initially I wrongly understand code. But still there is a question. Does it make sense at all? so after this change from three entries:
```
1.0.0.0 hostA aliasA
1.0.0.0 hostB aliasB
1.0.0.0 hostC aliasC
```

you get
```
1.0.0.0 hostD aliasD
1.0.0.0 hostD aliasD
1.0.0.0 hostD aliasD
```
which looks bogus

jreidinger: in fact I expect that such method always have oldhn and always have newhn and as result it just replace old hn with new hn...and maybe their aliases?

jreidinger:
regarding API, I agree with naming and will change it.

But why are you missing those two methods? Does it needed on any place? When I design API I try to avoid adding method that noone need as it is hard to properly design it. E.g. what will be return value? array? or first matching entry? And how it look like? string like now? or hash with canonical and aliases? or only canonical? and for find_by_hostname ( name is confusing for me ) does it return ip or aliases? or hash with all three ( canonical, ip and aliases) ? So if such method someone find useful, then I am not against adding it, when use case is clear.

jreidinger:
@mvidner also thinking about remove_hostname and delete_by_hostname. I also found it confusing. as `delete_by_ip` removes entry where is given ip. That make sense and is intuitive. But `delete_by_hostname` do not delete entry, only in specific case. It just remove given hostname from entries where it is available, but keep ip if it have other hostnames.

so for 
```
1.0.0.0 siteA siteB siteC
```
if I call `delete_by_hostname("siteC")` it can indicate that it remove whole this line, but it instead do just
```
1.0.0.0 siteA siteB
```

src/lib/cfa/hosts.rb
mvidner: It is bogus, you should get
```
1.0.0.0 hostD aliasD
```

Ah, now that i think of it, which of hostA hostB hostC is the canonical name anyway? We need to test that.

jreidinger: @mvidner do you see it intuitive? I really do not expect such behavior. Currently it is used only in one place `Host.Update` which is in fact also quite confusing. So what Update did? it removed old hostname and adds new hostname to given ip. So from my non-expert point of view I expect that in fact this set_entry is not needed and instead add_entry should be used in this case after `remove_hostname` (which is currently used ).

Also according to googling it looks like all of them are canonical and all of them will be resolved. So if you use hostA.suse.cz or hostB.suse.cz or hostC.suse.cz it should point you to same ip. And reverse it is first occurence, according to debian discussion - http://lists.debian.org/debian-devel/2004/06/msg00468.html so host for ip will get hostA.suse.cz

mvidner:
BTW regarding the [`/etc/hosts` editor](https://github.com/yast/yast-network/blob/master/src/include/network/services/host.rb) (Network Services →Hostnames, `yast2 host`) - it uses a "bulk" API: read everything via `Host.name_map`, and then `Host.clear` and `Host.add_name` in a loop.

jreidinger:
@mvidner I know, it is a bit misoptimal as it kills comment placement in file. Better way would be direct operation on hosts model and adding, removing and replacing entries there.

mvidner:
I have now run the code and it says Internal Error, undefined method set_host.

src/lib/cfa/hosts.rb
mvidner: OK, I have tested changing
- just the IP
- just the name
- both IP and name at the same time
and the hosts file came out right in all cases (after I've fixed the leftover set_host→set_entry occurrences)

mvidner:
`delete_hostname` has no tests, that's why the `set_host` has escaped. `set_entry` also has no tests BTW.

mvidner:
Regarding the augeas failure:

ruby-augeas-0.5.0 needs augeas-1.0.0: https://github.com/hercules-team/ruby-augeas/commit/9025d887eb80af9dcf8ecdb3209f52526437d0a7
and it uses this setup in its Travis file: (https://github.com/hercules-team/ruby-augeas/commit/7a69aa19141e7450ed4518b78968e1850c2117b1)
```
  - sudo add-apt-repository -y ppa:raphink/augeas
  - sudo apt-get update
  - sudo apt-get install libaugeas-dev libxml2-dev
```

jreidinger:
thanks

mchf:
I did some tests. And I get an Internal Error sometimes:
cfa_host

mchf:
The issue was raised in installer with already merged version of the patch.

jreidinger:
checking

jreidinger:
@mchf fix is in https://github.com/yast/yast-network/pull/456

PR #455, opened 2016-10-16
 Merge branch 'SLE-12-SP2' into master

teclator:
* Improved bridge handling (bsc#962824).

### DIFF
--- a/src/clients/inst_setup_dhcp.rb
+++ b/src/clients/inst_setup_dhcp.rb
@@ -1,24 +1,3 @@
-require "yast"
-require "network/network_autoconfiguration"
-
-module Yast
-  class SetupDhcp
-    include Singleton
-    include Logger
-
-    def main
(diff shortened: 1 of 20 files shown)


src/clients/inst_setup_dhcp.rb
```diff
@@ -1,24 +1,3 @@
...
-      else
-        log.info("Automatic DHCP configuration not started - an interface is already configured")
-      end
-
-      # if this is not wrapped in a def, ruby -cw says
-      # warning: possibly useless use of a literal in void context
-      :next
-    end
-  end
-end
```
teclator: What you mean? I don't get you. I just moved the client to lib, but no take any decision about it.

mchf: I've overlooked that. Sorry for confusion.

mchf:
lgtm

PR #457, opened 2016-10-21
Use 2 as forward delay for bridges created by virt proposal (bsc#988278)

teclator:


src/modules/Lan.rb
```diff
@@ -951,7 +951,7 @@ def ProposeVirtualized
           Ops.set(NetworkInterfaces.Current, "BRIDGE", "yes")
           Ops.set(NetworkInterfaces.Current, "BRIDGE_PORTS", ifcfg)
           Ops.set(NetworkInterfaces.Current, "BRIDGE_STP", "off")
-          Ops.set(NetworkInterfaces.Current, "BRIDGE_FORWARDDELAY", "0")
+          Ops.set(NetworkInterfaces.Current, "BRIDGE_FORWARDDELAY", "2")
```
kobliha: Doesn't explain why

package/yast2-network.changes
```diff
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Oct 21 11:01:04 UTC 2016 - kanderssen@suse.com
+- Use '2' as the default value for forward delay in bridges created
+  by the virt proposal (bsc#988278)
```
kobliha: The changes file describes the change being done, but it should say something like: preventing from doing this or that by ... doing. Or just an information about the change, not that you just changed 0 to 2 for unknown reason.

teclator: I will close it by now until more information is given, as the default not enable STP and in that case seems that the forwarddelay is not taking in cosideration, and I thought it could be independent of STP. The kernel min value is '2' although the IEEE range starts in 4.

More info here

https://github.com/openSUSE/wicked/blob/master/src/bridge.c#L293-L356

And the bug has been also updated by @mtomaschewski . Thnx @mtomaschewski !!

PR #458, opened 2016-10-24
Improved adding routes

mchf:
[bsc#927629](https://bugzilla.suse.com/show_bug.cgi?id=927629)  "yast lan does not allow IPv6 and CIDR subnet masks in static routes"

### DIFF
--- a/src/include/network/lan/help.rb
+++ b/src/include/network/lan/help.rb
@@ -98,8 +98,12 @@ module Yast
                                ) +
                                  _(
                                    "<p>For each route, enter destination network IP address, gateway address,\n" \
-                                     "and netmask. To omit any of these values, use a dash sign \"-\". Select\n" \
-                                     "the device through which the traffic to the defined network will be routed.\"-\" is an alias for any interface.</p>\n"
+                                     "and netmask. You can use either IPv4 netmask or prefix length when defining\n" \
+                                     "network part of route. Prefix length has to be prefixed using '/'.\n" \
+                                     "To omit any of these values, use a dash sign \"-\". Select\n" \
+                                     "the device through which the traffic to the defined network will be routed.\"-\" is an alias for any interface.\n" \
+                                     "Please note that in case of IPv6 networks only prefix length is accepted\n" \
+                                     "for netmask definition.</p>\n"
                                  ) +
                                  # Routing dialog help 2/2
                                  _(
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -144,7 +144,7 @@ module Yast
       return true if Netmask.Check(netmask)
       return true if netmask.start_with?("/") && netmask[1..-1] =~ /\A\d+\z/
 
-      return false
+      false
     end
 
     # Route edit dialog
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -424,15 +424,43 @@ module Yast
       false
     end
 
+    # Mangles route definition for storing in routes file
+    #
+    # Basically netmask field is obsolete, so it converts
+    # the record to use CIDR notation if netmask is defined.
+    #
+    # For details see "man routes".
+    def convert_route_conf(route)
+      return route if route["netmask"].empty? || route["netmask"] == "-"
+
+      dest = route["destination"]
+      netmask = route["netmask"]
+
+      if Netmask.Check4(netmask)
+        cidr = Netmask.ToBits(netmask)
+      elsif netmask.start_with?("/")
+        cidr = netmask[1..-1]
+      else
+        # if it is netmask then long netmask is not supported for IPv6
+        # if it is prefix length (CIDR), then prefix it has to be prefixed by '/'
+        raise ArgumentError, "Invalid netmask or prefix length"
+      end
+
+      route["destination"] = "#{dest}/#{cidr}"
+      route["netmask"] = "-"
+
+      route
+    end
+
     def storeRouting(_key, _event)
       route_conf = @r_items.map do |route|
-        {
+        convert_route_conf(
           "destination" => route[1].to_s,
           "gateway"     => route[2].to_s,
           "netmask"     => route[3].to_s,
           "device"      => route[4].to_s,
           "extrapara"   => route[5].to_s
-        }
+        )
       end
 
       @defgw = UI.QueryWidget(Id(:gw), :Value)
@@ -440,27 +468,21 @@ module Yast
       @defgw6 = UI.QueryWidget(Id(:gw6), :Value)
       @defgwdev6 = UI.QueryWidget(Id(:gw6dev), :Value)
 
-      if @defgw != ""
-        route_conf = Builtins.add(
-          route_conf,
-          "destination" => "default",
-          "gateway"     => @defgw,
-          "netmask"     => "-",
-          "device"      => @defgwdev
-        )
-      end
+      route_conf << {
+        "destination" => "default",
+        "gateway"     => @defgw,
+        "netmask"     => "-",
+        "device"      => @defgwdev
+      } if !@defgw.empty?
 
-      if @defgw6 != ""
-        route_conf = Builtins.add(
-          route_conf,
-          "destination" => "default",
-          "gateway"     => @defgw6,
-          "netmask"     => "-",
-          "device"      => @defgwdev6
-        )
-      end
+      route_conf << {
+        "destination" => "default",
+        "gateway"     => @defgw6,
+        "netmask"     => "-",
+        "device"      => @defgwdev6
+      } if !@defgw6.empty?
 
-      Routing.Routes = deep_copy(route_conf)
+      Routing.Routes = route_conf
       Routing.Forward_v4 = UI.QueryWidget(Id(:forward_v4), :Value)
       Routing.Forward_v6 = UI.QueryWidget(Id(:forward_v6), :Value)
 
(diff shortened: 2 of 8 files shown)


src/include/network/services/routing.rb
```diff
@@ -140,6 +140,23 @@ def initialize_network_services_routing(include_target)
...
+    # prefix length. If prefix length is used it has to start with '/'.
+    # For IPv6 network, only prefix length is allowed.
+    #
+    # param [string] netmask or /<prefix length>
+    def valid_netmask?(netmask)
+      return false if netmask.nil? || netmask.empty?
+      return true if Netmask.Check4(netmask)
+
+      prefix = netmask[1..-1].to_i
+      return true if netmask.start_with?("/") && prefix.between?(1, 128)
```
jreidinger: well, I think this code is more readable, as prefix is needed only if it start with "/"
```
if netmask.start_with?("/")
  return netmask[1..-1].to_i.between?(1, 128)
end

false
```

src/include/network/lan/help.rb
```diff
@@ -98,8 +98,12 @@ def initialize_network_lan_help(_include_target)
                                ) +
                                  _(
                                    "<p>For each route, enter destination network IP address, gateway address,\n" \
-                                     "and netmask. To omit any of these values, use a dash sign \"-\". Select\n" \
-                                     "the device through which the traffic to the defined network will be routed.\"-\" is an alias for any interface.</p>\n"
+                                     "and netmask. You can use either IPv4 netmask or prefix lenght when defining\n" \
```
mchf: If that is the only grammar issue, I have to congratulate myself ;-) 

Thanks, fixing.

src/include/network/services/routing.rb
```diff
@@ -258,7 +271,7 @@ def RoutingEditDialog(id, entry, devs)
         end
         route = Builtins.add(route, val)
         val = Convert.to_string(UI.QueryWidget(Id(:genmask), :Value))
-        if val != "-" && val != "0.0.0.0" && !Netmask.Check(val)
+        if val != "-" && val != "0.0.0.0" && !valid_netmask?(val)
```
mvidner: How am I supposed to know the difference between "check the netmask" and "is the netmask valid"?
Why are we not using/fixing the Netmask library instead? Observe these:
- http://www.rubydoc.info/github/yast/yast-yast2/Yast/NetmaskClass
- http://www.rubydoc.info/github/yast/yast-network/Yast/NetworkLanAddressInclude:ValidateNetmask
- http://www.rubydoc.info/github/yast/yast-network/Yast/NetworkRoutinesInclude:validPrefixOrNetmask

src/include/network/services/routing.rb
```diff
@@ -424,43 +437,65 @@ def validateRouting(_key, _event)
       false
     end
 
+    # Mangles route definition for storing in routes file
```
mvidner: "Mangling" is not something you want to introduce. See http://www.thefreedictionary.com/mangle (C++ mangling is, well, C++, not a good role model)
"Adapt" or "convert".

src/include/network/services/routing.rb
```diff
@@ -424,43 +438,65 @@ def validateRouting(_key, _event)
...
+      netmask = route["netmask"]
+
+      if Netmask.Check4(netmask)
+        cidr = Netmask.ToBits(netmask)
+      elsif netmask.start_with?("/")
+        cidr = netmask[1..-1]
+      else
+        # if it is netmask then long netmask is not supported for IPv6
+        # if it is prefix length (CIDR), then prefix it has to be prefixed by '/'
+        raise ArgumentError, "Invalid netmask or prefix length"
```
jreidinger: it is nice to add value to exception, so immediatelly visible like `"Invalid netmask or prefix length: '#{netmask}'"

src/include/network/services/routing.rb
```diff
@@ -424,43 +438,65 @@ def validateRouting(_key, _event)
...
+        "netmask"     => "-",
+        "device"      => @defgwdev
+      } if !@defgw.empty?
+
+      route_conf << {
+        "destination" => "default",
+        "gateway"     => @defgw6,
+        "netmask"     => "-",
+        "device"      => @defgwdev6
+      } if !@defgw6.empty?
```
jreidinger: just question, why @defgw6 and others are instance variables and not local ones?

src/modules/Routing.rb
```diff
@@ -205,6 +205,35 @@ def WriteIPForwarding
...
+
+        if !prefix.nil?
+          route["destination"] = subnet
+          route["netmask"] = "/#{prefix}"
+        end
+
+        route
+      end
+
+      routes
```
jreidinger: Very nasty sideeffect as it looks like it do nothing as you do not assign map result to routes variable or use `map!`. But in fact change hash in map, so even each do that stuff. Also it is not clear that it modify parameter itself.

src/include/network/services/routing.rb
mchf: most probably bcs of conversion from ycp.

src/include/network/services/routing.rb
mchf: ok

src/include/network/services/routing.rb
mchf: ok

jreidinger:
LGTM

src/modules/Routing.rb
```diff
@@ -205,6 +205,35 @@ def WriteIPForwarding
       write_ipv6_forwarding(@Forward_v6)
     end
 
+    # Converts routes config as read from system into well-defined format
```
mvidner: UGH! Why do we introduce both `normalize_routes(routes)` and `convert_route_conf(route)`?

mchf:
overall comment, for reference;
Purpose of this PR is to 
1) allow creating route using CIDR notation
2) normalize routes representation. Basically "netmask" field is obsolete in routes conf and CIDR should be preferred.
3) follow convention when using netmask in IPv6 networks is inconvenient and prefix should be used there.

(1) is the reported bug, (2) and (3) is covered e.g. by "man routes"

src/include/network/services/routing.rb
mchf: bcs the input value need not be only netmask but also prefix. I agree that ```valid_netmask?``` is most probably not the best name here. I've used "netmask" as a reference to the column name which is being validated.

src/include/network/services/routing.rb
mchf: ok

src/modules/Routing.rb
mchf: apart of that ```normalize_routes``` works on top of an array of routes, the basic difference is that ```normalize_routes``` should turn route definition into form which is expected in ```convert_route_conf``` - that "destination" field carries only network ip and netmask or prefix is placed in "destination"

mvidner:
We are adding two methods to change the routing settings, to convert between two formats of these, but we have no documentation of even one of these formats.

Forget the legacy code for a while. If we design the model class like follows,
we can say: `convert_route_conf` changes the hash so that raw_netmask is "-"
and `normalize_routes` changes the hash so that raw_netmask contains
a quad or a slash-number.

Or do it some other way, but do have one place that has a Yardoc URL where I can see what a routing entry is. 

```rb
class Yast::New::Routing      # I cannot remember the new namespacing style :(
  # @return [Array<RoutingEntry>]
  attr :entries
  # @return [Boolean]
  attr :forward_ipv4
  # @return [Boolean]
  attr :forward_ipv6

  class RoutingEntry
    # @return [String]  eg. "10.20.0.0"
    def destination_pfx; end
    # @return [Integer] eg. 16
    def destination_len; end

    # @return [String] always in the form "10.20.0.0/16", or "default"
    def destination; end

    # either as length or as a mask, after a slash or as a separate argument
    # destination = "default"
    # destination = "10.20.0.0/16"
    # destination = "10.20.0.0/255.255.0.0"
    # destination = ["10.20.0.0", 16]
    # destination = ["10.20.0.0", "255.255.0.0"]
    def destination=(pfx, len_or_mask=nil); end

    # @return [String]
    attr :gateway
    # @return [String] "eth0"
    attr :device
    # @return [String]
    attr :extrapara

    # @param [Hash{String=>String}] h the keys are
    #   "destination", "gateway", "netmask", "device", "extrapara"
    def self.from_anyagent(h); end
    # @return [Hash{String=>String}] h the keys are
    #   "destination", "gateway", "netmask", "device", "extrapara"
    def to_anyagent; end

    private

    # @return [String] The raw value of the 1st column of routes(5)
    def raw_destination; end
    # @return [String] The raw value of the 3rd column of routes(5)
    def raw_netmask; end
  end
end
```

mchf:
nice example, I'll go through it. However, I'm afraid that when going this way I cannot handle it as a part of "Kill as many bugs as possible in a given time" task ;-)

mvidner:
If we want to get this over with quickly, I suggest this option:

>  do have one place that has a Yardoc URL where I can see what a routing entry is.

mvidner:
@teclator do you have an opinion on this?

teclator:
In case that we accept it for the quick bug fixing I would consider the @mvidner suggestion as soon as possible. We have some bugs related to static routes created/applied by YaST as for example https://bugzilla.suse.com/show_bug.cgi?id=1005478 so could be a good oportunity to improve current code.

mvidner:
@mchf can you pull in https://github.com/yast/yast-network/commit/7667199617851d330714b12d1109cc79ba9e48ae from https://github.com/yast/yast-network/tree/bnc927629-routes-ipv6 please?

mchf:
@mvidner pulled, rebased

PR #459, opened 2016-10-25
Do not hardcode eth0 in the CLI test.

mvidner:
https://openqa.opensuse.org/tests/286685#step/yast2_cmdline/17


ancorgs:
LGTM, but travis disagrees.

ancorgs:
Well, looks almost good. The changelog entry is missing. And if we want it to go to openQA, then also the version increase.

PR #460, opened 2016-10-27
Fix summary

jreidinger:


src/modules/Lan.rb
```diff
@@ -726,44 +726,35 @@ def Summary(mode)
...
       header_nm = _("Network Mode")
 
       if NetworkService.is_network_manager
         href_nm = "lan--nm-disable"
         # network mode: the interfaces are controlled by the user
         status_nm = _("Interfaces controlled by NetworkManager")
-        # disable NetworkManager applet
-        link_nm = Hyperlink(href_nm, _("Disable NetworkManager"))
+        # switch from network manager to wicked
+        link_nm = Hyperlink(href_nm, _("Switch to Wicked"))
```
mchf: NP: I'd use something like "use wicked". But for grammar questions is better to ask @mvidner

mchf:
One NP, LGTM otherwise.

src/modules/Lan.rb
jreidinger: well, if I took whole sentence, then `Interfaces controlled by NetworkManager(Switch to Wicked)` looks slightly better then `Interfaces controlled by NetworkManager(use wicked)`
@mvidner  what do you think?

mvidner: The whole point of the PR is to be consistent with other proposals, which have the form "This is a description (do this)". So I think it is better to start with a lowercase letter: "switch to Wicked", "switch to NetworkManager".

jreidinger:
updated, according to @mvidner suggest, so merging.

PR #461, opened 2016-11-09
DHCLIENT_SET_HOSTNAME per interface

mchf:
- https://trello.com/c/pN6f3kzr (screenshots!)
- [bsc#984890](https://bugzilla.suse.com/show_bug.cgi?id=984890)


src/modules/LanItems.rb
```diff
@@ -846,14 +846,32 @@ def GetBridgeableInterfaces(bridgeMaster)
...
+    # @return [Array] list of NIC names which are configured to use (any) dhcp
+    def find_dhcp_ifaces
+      find_by_sysconfig do |ifcfg|
+        ["dhcp4", "dhcp6", "dhcp", "dhcp+autoip"].include?(ifcfg["BOOTPROTO"])
+      end
+    end
+
+    # Finds all devices which has DHCLIENT_SET_HOSTNAME set to "yes"
+    #
+    # @return [Array] list of NIC names which has the option set to "yes"
```
mvidner: even better, `Array<String>`

src/modules/LanItems.rb
```diff
@@ -2575,6 +2595,22 @@ def export_udevs(devices)
       deep_copy(ay)
     end
 
+    # Finds the devices for whose sysconfig options the given block evaluates to true
+    #
+    # It searches all known devices and passes sysconfig configuration
+    # into given block.
+    #
+    # @return [Array] list of device names
```
mvidner: See http://www.rubydoc.info/gems/yard/file/docs/Tags.md#yieldparam
```rb
# @yieldparam [Hash] ifcfg
# @yieldreturn [Boolean]
```

src/include/network/lan/help.rb
```diff
@@ -123,7 +123,9 @@ def initialize_network_lan_help(_include_target)
                                  "also a hostname via DHCP.  The hostname will be set automatically by the DHCP client.\n" \
                                  "However, changing the hostname at runtime may confuse the graphical desktop. \n" \
                                  "Therefore, disable this option if you connect to different networks that assign \n" \
-                                 "different hostnames.</p> "
+                                 "different hostnames. To avoid collisions in multi homed systems, you should enable the \n" \
+                                 "option for just one interface. Default value is used for interfaces which don't have the \n" \
```
mvidner: "Default value" should read "A default value" or "The default value". In any case, it does not tell the user what the default value is, does it?

src/include/network/routines.rb
```diff
@@ -970,5 +970,34 @@ def confirmed_detection(hwtype)
...
+    # A popup informing user that incorrent DHCLIENT_SET_HOSTNAME was detected
+    def fix_dhclient_warning(devs)
+      message = _(
+        fix_dhclient_msg(devs) + _("Use Hostname/DNS tab to fix it.")
+      )
+
+      Popup.Warning(message)
+    end
+
+    def fix_dhclient_yesno(devs)
```
mvidner:
```
# A yes-no pop-up
# @return [Boolean]
```

src/include/network/routines.rb
```diff
@@ -970,5 +970,34 @@ def confirmed_detection(hwtype)
...
+          "DHCLIENT_SET_HOSTNAME option is set for more than one interface.\n" \
+          "This causes indeterministic behavior.\n" \
+          "Affected devices:\n" \
+          "%s\n",
+          devs.join(" ")
+        )
+      )
+    end
+
+    # A popup informing user that incorrent DHCLIENT_SET_HOSTNAME was detected
```
mvidner: `# @return [void]`

src/include/network/summary.rb
```diff
@@ -91,35 +91,5 @@ def OverviewTableContents(caption, header, contents, first)
...
     end
-
-    # Create table widget for the overview screens
-    # @param [Yast::Term] header table header
-    # @param [Array] contents table contents
-    # @return table widget
-    def OverviewTable(header, contents)
-      header = deep_copy(header)
-      contents = deep_copy(contents)
-      VBox(VSpacing(0.5), OverviewTableContents("", header, contents, true))
```
mvidner: Yay for removed code!
Even better, you can remove also the above-defined `OverviewTableContents`.

mchf: how could I overlook that ... pressing delete.

src/modules/LanItems.rb
```diff
@@ -909,37 +927,6 @@ def FindDeviceIndex(device)
       ret
     end
 
-    # It finds a new style device name for device name in old fashioned format
-    #
-    # It goes through currently present devices and tries to mach it to given
-    # old fashioned name
-    #
-    # @returns [String] new style name in case of success. Given name otherwise.
-    def getDeviceName(oldname)
```
mchf: ah, overlooked. I'll revert this change. Thanks

src/include/network/services/dns.rb
```diff
@@ -128,22 +112,25 @@ def initialize_network_services_dns(include_target)
...
+              "",
+              []
+            ),
             ReplacePoint(Id("dh_host_text"), Empty())
           ),
-          # help
-          "help"          => Ops.get_string(@help, "dhcp_hostname", ""),
-          "init"          => fun_ref(method(:InitDhcpHostname), "void (string)")
+          "init"          => fun_ref(method(:InitDhcpIfaces), "void (string)"),
+          "store"         => fun_ref(method(:StoreDhcpIfaces), "voide (string, map)")
```
mvidner: voide → void

src/include/network/routines.rb
```diff
@@ -989,5 +989,31 @@ def confirmed_detection(hwtype)
...
       Confirm.Detection(hwstring, nil)
     end
+
+    # Returns a generic message informing user that incorrect DHCLIENT_SET_HOSTNAME
+    # setup was detected.
+    #
+    # @param [Array<String>] list of incorrectly configured devices
+    # @return [String] a message stating that incorrect DHCLIENT_SET_HOSTNAME setup was detected
+    def fix_dhclient_msg(devs)
+      _(
```
mvidner: Translating a variable string will not work.

src/include/network/services/dns.rb
```diff
@@ -128,22 +112,25 @@ def initialize_network_services_dns(include_target)
             Hostname.ValidDomain
           )
         },
-        "DHCP_HOSTNAME"   => {
+        "DHCP_IFACES"     => {
```
mvidner: "DHCP_IFACES" entirely hides the fact that the widget is related to the hostname.

src/include/network/services/dns.rb
```diff
@@ -431,23 +409,74 @@ def StoreHnWidget(key, _event)
...
+      if !DNS.valid_dhcp_cfg?
+        fix_dhclient_warning(hostname_ifaces)
+
+        selected = NO_CHANGE_LABEL
+        items = [Item(Id(NO_CHANGE_LABEL), "", true)]
+      elsif hostname_ifaces.size == 1
+        selected = hostname_ifaces.first
+      end
+      # translators: no device selected placeholder
+      items << Item(Id(NONE_LABEL), _(NONE_LABEL), selected == NONE_LABEL)
```
mvidner: This won't work. The translation method would run, but the string extraction would not find the string.

src/include/network/services/dns.rb
```diff
@@ -431,23 +409,74 @@ def StoreHnWidget(key, _event)
...
+      # translators: placeholder for "set hostname via any DHCP aware device"
+      items << Item(Id(ANY_LABEL), _("yes: " + ANY_LABEL), selected == ANY_LABEL)
```
mvidner: Even if the resulting string is constant, the POT extraction would not work. We must use literal strings as arguments of `_()`.

src/include/network/services/dns.rb
```diff
@@ -431,23 +409,74 @@ def StoreHnWidget(key, _event)
...
+
+      items += LanItems.find_dhcp_ifaces.sort.map do |iface|
+        # translators: label is in form yes: <device name>
+        Item(Id(iface), _(format("yes: %s", iface)), iface == selected)
```
mvidner: The correct way is `format(_("blah blah %s"), arg)`

src/include/network/services/dns.rb
```diff
@@ -431,23 +409,74 @@ def StoreHnWidget(key, _event)
...
-    # @param [String] key	the widget receiving the event
-    # @param event	the event being handled
-    # @return nil so that the dialog loops on
-    def InitDhcpHostname(_key)
-      UI.ChangeWidget(Id("DHCP_HOSTNAME"), :Enabled, @has_dhcp)
-      if !@has_dhcp
-        UI.ReplaceWidget(Id("dh_host_text"), Label(_("No interface with dhcp")))
+    # Checks whether setting hostname via DHCP is allowed
+    def use_dhcp_hostname?
+      UI.QueryWidget(Id("DHCP_IFACES"), :Value) != "none"
```
mvidner: No. `"no"`. Or `NONE_LABEL`.

src/include/network/routines.rb
```diff
@@ -989,5 +989,31 @@ def confirmed_detection(hwtype)
...
+        format(
+          "DHCLIENT_SET_HOSTNAME option is set for more than one interface,\n" \
+          "or global default is set incorrectly.\n" \
+          "This causes indeterministic behavior.\n\n" \
```
mvidner: Oh nice. So what should I do as the user?

src/include/network/services/dns.rb
```diff
@@ -431,23 +409,74 @@ def StoreHnWidget(key, _event)
...
+
+      hostname_ifaces = LanItems.find_set_hostname_ifaces
+      selected = DNS.dhcp_hostname ? ANY_LABEL : NONE_LABEL
+      items = []
+
+      if !DNS.valid_dhcp_cfg?
+        fix_dhclient_warning(hostname_ifaces)
+
+        selected = NO_CHANGE_LABEL
+        items = [Item(Id(NO_CHANGE_LABEL), "", true)]
```
mvidner: Do we really want an empty description here? Combined with the vague "indeterministic!" popup this would leave me really confused as to what will happen and what I should do.

src/modules/DNS.rb
```diff
@@ -668,6 +668,16 @@ def create_hostname_link
       nil
     end
 
+    # Checks if system DHCLIENT_SET_HOSTNAME is valid
+    #
+    # @return [Boolean]
+    def valid_dhcp_cfg?
+      return false if LanItems.find_set_hostname_ifaces.size > 1
```
mvidner: This new method is the only one that refers to LanItems (without even importing it). Maybe it should be moved there.

src/modules/LanItems.rb
```diff
@@ -2426,6 +2435,50 @@ def createUdevFromIfaceName(interfaces)
...
+    # @param [String] device name where should be hostname configuration active
+    # @return [Boolean] false when the configuration cannot be set for a device
+    def conf_set_hostname(device)
+      dhcp_devs = find_dhcp_ifaces
+
+      return false if !dhcp_devs.include?(device)
+
+      ret = (dhcp_devs - [device]).all? do |dev|
+        item_id = find_configured(dev)
+        SetItemSysconfigOpt(item_id, "DHCLIENT_SET_HOSTNAME", "no")
```
mvidner: We said that the implementation would be
> ETHn: global=no, ifcfg[n]=yes, ifcfg[others]=nil

Is there an advantage to using "no" over nil? I think it just makes it harder to switch to "yes: any" manually.

mvidner:
Thanks! I think I have figured out a better popup text + widget label.

.rubocop.yml
```diff
@@ -13,7 +13,7 @@ Metrics/BlockNesting:
 # Offense count: 5
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 1916
+  Max: 2000
```
teclator: I agree, and this is a good idea to add to the current documentation PBI proposing how we could redesign it, and maybe not just `LanItems` but also `NetworkInterfaces` as the two main classes/modules which involves network configuration and seems to be the responsables of too many things.

BTW, thnx for the graph, I think that is very useful.

src/modules/LanItems.rb
mchf: well it is mainly leftover from previous implementation which I overlooked and it also solves switching from ifcfg[n] -> ifcfg[m]

.rubocop.yml
mchf: Apart of others I've a plan to create a class for "lan item" and then use a kind of container with meaningful iterators instead of "LanItems" that should help with all those "GetItem.." and "SetItem..." and "GetCurrent..." and "SetCurrent...". However, I'd like to do these changes for SLE13.

mchf: another plan, which is already partly started, is to separate udev stuff.

mvidner:
LGTM! Thank you!

mchf:
Thanks for review.

PR #462, opened 2016-11-14
Udev Rule: Use also the dev_port when bus_id match is in use (bsc#1007172)

teclator:
When a interface is enslaved in a bond, then specially for easier card replacement the udev rule is modified using the bus_id instead of the mac address, but the bus_id could be not unique (multiport cards with one network function). In that cases also the dev_port should be used if it is exported via sysfs.

Here is the current systemd code which takes care of predictable network device names https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L21 if you want to know more.

And also this blog post https://major.io/2015/08/21/understanding-systemds-predictable-network-device-names/ could be interesting to clarify the current situation as well as how systemd choose the current name: https://github.com/systemd/systemd/blob/master/src/udev/net/link-config.c#L413

**Note:** There was also a bug in hwinfo obtaining the bus_id only for the first interface using the same pci path, it has been fixed in this PR: https://github.com/openSUSE/hwinfo/pull/36. We now require at least `hwinfo-21.35`

### DIFF
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -285,6 +285,25 @@ module Yast
       nil
     end
 
+    # Updates the udev rule of the current Lan Item, adding the dev_port and
+    # using the bus_id instead ot the mac address.
+    def use_udev_rule_for_bonding!
+      # Update or insert the dev_port if the sysfs dev_port attribute is present
+      LanItems.ReplaceItemUdev(
+        "ATTR{dev_port}",
+        "ATTR{dev_port}",
+        LanItems.dev_port(LanItems.GetCurrentName)
+      ) if LanItems.dev_port?(LanItems.GetCurrentName)
+
+      # Iff particular bond slave uses mac based persistency,
+      # overwrite to bus id based one. Don't touch otherwise.
+      LanItems.ReplaceItemUdev(
+        "ATTR{address}",
+        "KERNELS",
+        LanItems.getCurrentItem.fetch("hwinfo", {}).fetch("busid", "")
+      )
+    end
+
     # Automatically configures slaves when user enslaves them into a bond or bridge device
     def UpdateSlaves
       current = LanItems.current
@@ -309,12 +328,7 @@ module Yast
         case LanItems.GetDeviceType(current)
         when "bond"
           LanItems.startmode = "hotplug"
-          # if particular bond slave uses mac based persistency, overwrite to bus id based one. Don't touch otherwise.
-          LanItems.ReplaceItemUdev(
-            "ATTR{address}",
-            "KERNELS",
-            Ops.get_string(LanItems.getCurrentItem, ["hwinfo", "busid"], "")
-          )
+          use_udev_rule_for_bonding!
         when "br"
           LanItems.ipaddr = ""
         end
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -285,25 +285,6 @@ module Yast
       nil
     end
 
-    # Updates the udev rule of the current Lan Item, adding the dev_port and
-    # using the bus_id instead ot the mac address.
-    def use_udev_rule_for_bonding!
-      # Update or insert the dev_port if the sysfs dev_port attribute is present
-      LanItems.ReplaceItemUdev(
-        "ATTR{dev_port}",
-        "ATTR{dev_port}",
-        LanItems.dev_port(LanItems.GetCurrentName)
-      ) if LanItems.dev_port?(LanItems.GetCurrentName)
-
-      # Iff particular bond slave uses mac based persistency,
-      # overwrite to bus id based one. Don't touch otherwise.
-      LanItems.ReplaceItemUdev(
-        "ATTR{address}",
-        "KERNELS",
-        LanItems.getCurrentItem.fetch("hwinfo", {}).fetch("busid", "")
-      )
-    end
-
     # Automatically configures slaves when user enslaves them into a bond or bridge device
     def UpdateSlaves
       current = LanItems.current
@@ -328,7 +309,8 @@ module Yast
         case LanItems.GetDeviceType(current)
         when "bond"
           LanItems.startmode = "hotplug"
-          use_udev_rule_for_bonding!
+
+          LanItems.update_item_udev_rule!(:bus_id)
         when "br"
           LanItems.ipaddr = ""
         end
@@ -352,7 +334,7 @@ module Yast
     # into bond device and persistence based on bus id is required, then some configuration changes
     # are required in ifcfg and udev. It used to be needed to do it by hand before.
     def AutoUpdateOverview
-      # TODO: allow disabling. E.g. iff bus id based persistency is not requested.
+      # TODO: allow disabling. E.g. if bus id based persistency is not requested.
       UpdateSlaves()
 
       nil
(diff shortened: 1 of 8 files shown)


src/include/network/lan/complex.rb
```diff
@@ -285,6 +285,25 @@ def enableDisableButtons
       nil
     end
 
+    # Updates the udev rule of the current Lan Item, adding the dev_port and
+    # using the bus_id instead ot the mac address.
+    def use_udev_rule_for_bonding!
```
mchf: well, ```use_udev_rule_for_bonding``` sounds like when udev rule is used for bonding only. I'd use something like ```create_...``` or ```update_...```

also you should either document return value or return nil explicitly (I'd prefer the later one)

src/include/network/lan/complex.rb
```diff
@@ -285,6 +285,25 @@ def enableDisableButtons
...
+      # Update or insert the dev_port if the sysfs dev_port attribute is present
+      LanItems.ReplaceItemUdev(
+        "ATTR{dev_port}",
+        "ATTR{dev_port}",
+        LanItems.dev_port(LanItems.GetCurrentName)
+      ) if LanItems.dev_port?(LanItems.GetCurrentName)
+
+      # Iff particular bond slave uses mac based persistency,
```
mchf: typo

src/include/network/routines.rb
```diff
@@ -832,6 +832,25 @@ def physical_port_id?(dev_name)
       !physical_port_id(dev_name).empty?
     end
 
+    # Dev port of of the given interface from /sys/class/net/$dev_name/dev_port
```
mchf: typo - two "of"

src/lib/network/edit_nic_name.rb
```diff
@@ -63,12 +63,16 @@ def run
...
+        current_rule = LanItems.GetItemUdevRule(LanItems.current)
+
         if UI.QueryWidget(:udev_type, :CurrentButton) == :mac
           rule_key = MAC_UDEV_ATTR
           rule_value = @mac
+          LanItems.RemoveKeyFromUdevRule(current_rule, "ATTR{dev_port}")
         else
           rule_key = BUSID_UDEV_ATTR
           rule_value = @bus_id
+          LanItems.ReplaceItemUdev("ATTR{dev_port}", "ATTR{dev_port}", dev_port(@old_name))
```
mchf: so you should think of generalization of ```use_udev_rule_for_bonding!``` e.g. turning it into ```..._for_busid``` or even some more intrusive refactoring of this piece of sh.. errr code. At least you do not check if ```dev_port``` returns something none empty here.

test/complex_test.rb
```diff
@@ -104,4 +104,40 @@ def initialize
...
+          "hwinfo" => {
+            "dev_name" => "test0",
+            "busid"    => "00:08:00"
+          },
+          "udev"   => {
+            "net" => ["ATTR{address}==\"01:02:03:04:05\"", "KERNEL==\"eth*\"", "NAME=\"test0\""]
+          }
+        }
+      }
+      allow(Yast::LanItems).to receive(:dev_port).and_return("0")
```
mchf: pls also add test(s) when ```dev_port``` is not available.

src/include/network/lan/complex.rb
teclator: The two options you are given in the first sentence was what made me also doubt about which name method to use, because it currently use ReplaceItem for both (dev_port and mac_address) so basically it will update_or_create the dev_port attribute in the udev_rulle (if exists) with the read value and will replace the mac_address with the bus_id.

src/lib/network/edit_nic_name.rb
teclator: I added the dev_port?(@old_name) locally and planned to update the PR with the first CR, I was also thinking in use directly the method and maybe move it to some Udev class but thinking to do it as part of the documentation PBI.

src/lib/network/edit_nic_name.rb
```diff
@@ -63,18 +63,11 @@ def run
...
-        if UI.QueryWidget(:udev_type, :CurrentButton) == :mac
-          rule_key = MAC_UDEV_ATTR
-          rule_value = @mac
-        else
-          rule_key = BUSID_UDEV_ATTR
-          rule_value = @bus_id
-        end
+        udev_type = UI.QueryWidget(:udev_type, :CurrentButton)
 
-        # update udev rules and other config
```
teclator: Removed by error :-1:  will add it back after the code review with the changes that could be needed.

src/modules/LanItems.rb
```diff
@@ -413,6 +413,66 @@ def GetItemUdev(key)
...
+        ) if LanItems.dev_port?(LanItems.GetCurrentName)
+
+        # If the current rule is mac based, overwrite to bus id. Don't touch otherwise.
+        LanItems.ReplaceItemUdev(
+          "ATTR{address}",
+          "KERNELS",
+          LanItems.getCurrentItem.fetch("hwinfo", {}).fetch("busid", "")
+        )
+      else
+        log.error("The key given for udev rule #{based_on} is not mac or bus_id.")
```
mchf: NP: shouldn't it raise an exception?

mchf:
i still miss a test when dev_port is not present

src/modules/LanItems.rb
teclator: Added `ArgumentError` exception

test/lan_items_helpers_test.rb
```diff
@@ -166,3 +166,104 @@ def check_GetItemUdev(key, expected_value)
...
+describe "#update_item_udev_rule!" do
+  let(:hwinfo) { { "dev_name" => "test0", "busid" => "00:08:00", "mac" => "01:02:03:04:05" } }
+  let(:udev_net) { ["ATTR{address}==\"01:02:03:04:05\"", "KERNEL==\"eth*\"", "NAME=\"test0\""] }
+  let(:rule) { { 0 => { "hwinfo" => hwinfo, "udev" => { "net" => udev_net } } } }
+
+  before do
+    Yast::LanItems.Items = rule
+    Yast::LanItems.current = 0
+    # The current item hasn't got a dev_port
+    allow(Yast::LanItems).to receive(:dev_port).and_return("")
```
teclator: @mchf so by default I'm not exporting the dev_port and below we have a test that export it explicitly, the same is for `:mac` so, is one of this tests where you are missing some case? I have replaced the mock of `dev_port?` what was in use before the last changes.

test/lan_items_helpers_test.rb
```diff
@@ -166,3 +166,104 @@ def check_GetItemUdev(key, expected_value)
...
+    let(:rule) { { 0 => { "ifcfg" => "eth0" } } }
+
+    it "returns nil" do
+      expect(Yast::LanItems.RemoveItemUdev("KEY_TO_DELETE")).to eql(nil)
+    end
+  end
+end
+
+describe "#update_item_udev_rule!" do
+  let(:hwinfo) { { "dev_name" => "test0", "busid" => "00:08:00", "mac" => "01:02:03:04:05" } }
```
mchf: NP: is bus id and mac intentionally written incorrectly or not? ;-)

test/lan_items_helpers_test.rb
mchf: no just overlooked the change, thanks

mchf:
lgtm

test/lan_items_helpers_test.rb
teclator: Yes it is, but of course valid values are preferable, so already replaced.

mchf:
lgtm

PR #463, opened 2016-11-14
Refresh description

mchf:
Bnc#1009931


src/modules/LanItems.rb
```diff
@@ -1280,17 +1280,17 @@ def ip_overview(ip)
       when /DHCP/
         bullets << format("%s %s", _("IP address assigned using"), ip)
       else
-        prefixlen = NetworkInterfaces.Current["PREFIXLEN"]
-        if prefixlen
+        prefixlen = LanItems.prefix
```
jreidinger: it looks a bit inconsistent. Is it prefix or length of prefix? variable is len, but method is only prefix

src/modules/LanItems.rb
```diff
@@ -1366,26 +1366,32 @@ def startmode_overview(item_id)
       [startmode_descr]
     end
 
-    def ip_overview(ip)
+    # Creates details for device overvies based on ip configuration type
```
teclator: overvie**w**

src/modules/LanItems.rb
```diff
@@ -1366,26 +1366,32 @@ def startmode_overview(item_id)
...
+    #
+    # Produces list of strings. Strings are intended for "bullet" list, e.g.:
+    # * <string1>
+    # * <string2>
+    #
+    # return [Array] list of strings, one string is intended for one "bullet"
```
teclator: please, document also the param

src/modules/LanItems.rb
```diff
@@ -1418,15 +1424,16 @@ def BuildLanOverview
 
         if !ifcfg_name.empty?
           ifcfg_conf = GetDeviceMap(key)
+          log.error("BuildLanOverview: devmap for #{key}/#{ifcfg_name} is nil") if ifcfg.nil?
```
mchf: yeah, details ;-) fixed

mchf: when you know ```key```, you can get hwinfo from logs already.

teclator:
LGTM

mchf:
Thanks for reviews

PR #464, opened 2016-11-15
[RFC] document generic options

jreidinger:
https://trello.com/c/QNlLs9GW
not finished yet

### DIFF
--- a/doc/features.md
+++ b/doc/features.md
@@ -50,3 +50,12 @@ Devices
 - for wireless it allows to configure mode in which operating, encryption, essid of network and so on
 - for infiniband it allows set IPoIB connected x datagram
 
+
+s390 Devices
+============
+
+- allows layer 2 and 3 devices
+- support channel ids for devices
+- allow layer 2 devices to be in bond
+- recognized types - qeth, lcs, ctc
+
--- a/doc/features.md
+++ b/doc/features.md
@@ -1,0 +1,23 @@
+Network Module Features
+-----------------------
+
+Network Services
+======================
+
+- wicked
+- Network Manager
+- no network service
+
+IPv6
+==============
+
+- enable/disable it
+
+DHCP Client options
+====================
+
+- set hostname according to dhcp response
+- set machine identifier for asking dhcp
+- allow to set specific option for hostname to send for dhcpd client ( how often it is used? )
+
+
--- a/doc/features.md
+++ b/doc/features.md
@@ -73,3 +73,13 @@ Network Devices
 
 - ability to check what device is on mount point and if device is network one
 
+
+Shared Features ( in yast2 )
+============================
+
+- Internet test
+- Firewall CWM widgets
+- Bind9 YaPI interface
+- mapping of ports number and its aliases by IANA
+- Part ranges support ( well, it looks like object in ycp way )
+- firewall part support susefirewall and also firewalld
--- a/doc/features.md
+++ b/doc/features.md
@@ -7,6 +7,7 @@ Network Services
 - wicked
 - Network Manager
 - no network service
+- still support old netconfig ( but no guarantie )
 
 IPv6
 ==============
@@ -16,16 +17,15 @@ IPv6
 DHCP Client options
 ====================
 
-- set hostname according to dhcp response
+- send back to dhcp server hostname
 - set machine identifier for asking dhcp
 - allow to set specific option for hostname to send for dhcpd client ( how often it is used? )
 
 DNS
 ===================
 
-- set hostname
-- set domain
-- change hostname with dhcp ( looks same as in dhcp client options )
+- set hostname and domain, combining them into FQDN writting into /etc/hostname
+- change hostname with dhcp
 - assign hostname to local ip ( so own hostname is always resolvable )
 - modify /etc/resolv.conf, keep it manage automatic or modify it manually with selected values
 
@@ -40,7 +40,7 @@ Routing
 Devices
 =======
 
-- allows to setup it without setup (for bondings), dhcp or static ( with ip, hostname, mask and iBFT )
+- allows to setup it without setup (for bondings and iBFT), dhcp or static ( with ip, hostname, mask )
 - manage udev names, allowing renaming of devices when needed
 - blinking of device via ethtool for easier device identification
 - allows to add additional ips to device
@@ -59,7 +59,8 @@ s390 Devices
 - allows layer 2 and 3 devices
 - support channel ids for devices
 - allow layer 2 devices to be in bond
-- recognized types - qeth, lcs, ctc
+- recognized types - qeth, lcs, ctc, iucv
+- allows to set logical link address for layer 2 qeth devices
 
 Firewall parts
 ==============
--- a/doc/features.md
+++ b/doc/features.md
@@ -20,4 +20,28 @@ DHCP Client options
 - set machine identifier for asking dhcp
 - allow to set specific option for hostname to send for dhcpd client ( how often it is used? )
 
+DNS
+===================
+
+- set hostname
+- set domain
+- change hostname with dhcp ( looks same as in dhcp client options )
+- assign hostname to local ip ( so own hostname is always resolvable )
+- modify /etc/resolv.conf, keep it manage automatic or modify it manually with selected values
+
+Routing
+=========
+
+- allows to assign gateway for ipv4 and ipv6 and also assign via which device it can be reached
+- allows to enable ipv4 and ipv6 forwarding
+- allows manual edit of routing table
+
+
+Devices
+=======
+
+- allows to setup it without setup (for bondings), dhcp or static ( with ip, hostname, mask )
+- allows to add additional ips to device
+- supported devices: eth, wlan, token ring
+- for eth and token ring - allow to specify kernel module needed for device, if it is pcmci or usb and ethtool options
 
--- a/doc/features.md
+++ b/doc/features.md
@@ -42,6 +42,11 @@ Devices
 
 - allows to setup it without setup (for bondings), dhcp or static ( with ip, hostname, mask )
 - allows to add additional ips to device
-- supported devices: eth, wlan, token ring
+- supported devices: eth, wlan, token ring, infiniband, tun, tap, bridge,vlan, bond, arcnet, bluetooth, dummy, fddi, usb, myrinet (usb and bluetooth looks strange )
 - for eth and token ring - allow to specify kernel module needed for device, if it is pcmci or usb and ethtool options
+- for each devices it can be specified when activate device, ifplugd priority, firewall zone and MTU
+- for bridge it allow to specify any physical devices ( so e.g. tap and tun are filtered out )
+- for bonding it allows to specify over which device it bonding
+- for wireless it allows to configure mode in which operating, encryption, essid of network and so on
+- for infiniband it allows set IPoIB connected x datagram
 
--- a/doc/features.md
+++ b/doc/features.md
@@ -59,3 +59,9 @@ s390 Devices
 - allow layer 2 devices to be in bond
 - recognized types - qeth, lcs, ctc
 
+Firewall parts
+==============
+
+- stage1 proposal + write it down ( firewall enable/disable, sshd enable/disable, ssh port open/close, vnc enable/disable)
+- assign network device to firewall zone
+
--- a/doc/features.md
+++ b/doc/features.md
@@ -1,85 +1,85 @@
 Network Module Features
------------------------
+=======================
 
 Network Services
-======================
+----------------
 
 - wicked
 - Network Manager
 - no network service
-- still support old netconfig ( but no guarantie )
+- still support old netconfig ( but no guarantee )
 
 IPv6
-==============
+----
 
 - enable/disable it
 
-DHCP Client options
-====================
+DHCP Client Options
+-------------------
 
-- Set hostname which is send back to the DHCP server
-- set machine identifier for asking dhcp
+- Set hostname which is sent back to the DHCP server
+- set machine identifier for asking DHCP
 - allow to set specific option for hostname to send for dhcpd client ( how often it is used? )
 
 DNS
-===================
+---
 
-- set hostname and domain, combining them into FQDN writting into /etc/hostname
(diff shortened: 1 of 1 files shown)


doc/installation.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+and network/config:
+```
+NETCONFIG_DNS_STATIC_SEARCH_LIST="suse.de"
+NETCONFIG_DNS_STATIC_SERVERS="192.168.122.1"
+```
+
+
+## inst_install_inf
```
mchf: in fact this is going to be removed (almost) completely soon.

doc/features.md
```diff
@@ -0,0 +1,67 @@
...
+
+- allows to assign gateway for ipv4 and ipv6 and also assign via which device it can be reached
+- allows to enable ipv4 and ipv6 forwarding
+- allows manual edit of routing table
+
+
+Devices
+=======
+
+- allows to setup it without setup (for bondings), dhcp or static ( with ip, hostname, mask )
```
mchf: and iBFT

doc/installation.md
teclator: It should :), checking the interface I realized that most of the code is useless currently.

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+
+IPv6
+==============
+
+- enable/disable it
+
+DHCP Client options
+====================
+
+- set hostname according to dhcp response
```
mchf: i don't fully understand this note. If we talk about "Hostname to Send" then this is about hostname which is being send back to dhcp server.

doc/features.md
```diff
@@ -0,0 +1,74 @@
+Network Module Features
+-----------------------
+
+Network Services
+======================
+
+- wicked
+- Network Manager
+- no network service
```
mchf: we're able to handle even old netconfig .. when it is installed. But i don't know if it is still doable in SLE-12.

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+
+- set hostname according to dhcp response
+- set machine identifier for asking dhcp
+- allow to set specific option for hostname to send for dhcpd client ( how often it is used? )
+
+DNS
+===================
+
+- set hostname
+- set domain
```
mchf: this is bad. hostname and domain is actually combined into FQDN and written into /etc/hostname. This feature is a bit controversial. See https://trello.com/c/so0amSNp/1176-sle-12-sp3-fate-yast2-network-clarify-hostname-setup

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+- change hostname with dhcp ( looks same as in dhcp client options )
```
mchf: no, that is something else

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+
+- allows to assign gateway for ipv4 and ipv6 and also assign via which device it can be reached
+- allows to enable ipv4 and ipv6 forwarding
+- allows manual edit of routing table
+
+
+Devices
+=======
+
+- allows to setup it without setup (for bondings), dhcp or static ( with ip, hostname, mask and iBFT )
```
mchf: ehm no. That note about iBFT belongs to "None" configuration - it means that configuration should be read from firmware of iBFT aware device

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+- for infiniband it allows set IPoIB connected x datagram
+
+
+s390 Devices
+============
+
+- allows layer 2 and 3 devices
+- support channel ids for devices
+- allow layer 2 devices to be in bond
+- recognized types - qeth, lcs, ctc
```
mchf: and iucv

doc/features.md
```diff
@@ -0,0 +1,74 @@
...
+- for eth and token ring - allow to specify kernel module needed for device, if it is pcmci or usb and ethtool options
+- for each devices it can be specified when activate device, ifplugd priority, firewall zone and MTU
+- for bridge it allow to specify any physical devices ( so e.g. tap and tun are filtered out )
+- for bonding it allows to specify over which device it bonding
+- for wireless it allows to configure mode in which operating, encryption, essid of network and so on
+- for infiniband it allows set IPoIB connected x datagram
+
+
+s390 Devices
+============
```
mchf: allows to set logical link address for L2 qeth device

doc/features.md
```diff
@@ -0,0 +1,75 @@
...
+
+IPv6
+==============
+
+- enable/disable it
+
+DHCP Client options
+====================
+
+- send back to dhcp server hostname
```
mchf: I'd rephrase it e.g. like this: "Set hostname which is send back to the DHCP server"

doc/features.md
```diff
@@ -0,0 +1,75 @@
...
+
+- set hostname and domain, combining them into FQDN writting into /etc/hostname
+- change hostname with dhcp
+- assign hostname to local ip ( so own hostname is always resolvable )
+- modify /etc/resolv.conf, keep it manage automatic or modify it manually with selected values
+
+Routing
+=========
+
+- allows to assign gateway for ipv4 and ipv6 and also assign via which device it can be reached
```
mchf: default gateway

doc/autoinstallation.md
```diff
@@ -0,0 +1,99 @@
...
+`setup_before_proposal` has to be set to `true` as commented in the [autosetup section](#ausosetup), 
+more details about network 
+[here](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Network)
+
+#### ssh_settings_finish
+
+## Second Stage
+
+The _SecondStage_, also known as `continue` in the control file, is basically 
+the stage where the configuration should be definitly written.
```
jreidinger: I am missing here important option that keeps network config from first stage

doc/autoinstallation.md
```diff
@@ -0,0 +1,99 @@
...
+
+### autoconfigure
+
+This client will read the 
+[desktop configuration files](https://yastgithubio.readthedocs.io/en/latest/autoyast-development/#desktop-configuration-file)
+and is very probably that will write the configuration of our network calling
+the lan_auto client with the information read in the networking section, but
+as commented, it depents on the information of the desktop files to do its job.
+
+
```
jreidinger: why so many empty lines?

doc/installation.md
```diff
@@ -0,0 +1,98 @@
+## Introduction
+
+The workflow of the installation is customized via the [control
+file](https://github.com/yast/yast-installation/blob/master/doc/control-file.md), 
+and below we have the principal steps concerning network configuration:
+
+  - Linuxrc
+  - inst_install_inf (What is still needed)
+  - inst_dhcp
+  - manual_configuration (openSUSE only in Addons)
```
jreidinger: I think in SLE it is in registration, so in both you have chance to set it, not?

doc/autoinstallation.md
teclator: It is described in network_finish method.

doc/autoinstallation.md
teclator: Sorry, added the new line explicitly, fixing it.

doc/installation.md
teclator: In SLE you can launch the `lan` client manually in `disks_activation`, `registration` and `addons' like you can see [here](https://gist.github.com/teclator/35a6aee2cd387605b6ef#final-implementation), so basically is the same without `registration`. I will rewrite it.

doc/autoinstallation.md
```diff
@@ -0,0 +1,99 @@
...
+  - network_finish
+    - save_network
+
+#### network_finish
+
+This client calls `save_network` which copies udev rules and ifcfg files from
+from the running system which should be the linuxrc config and not the profile
+network configuration. 
+
+Take in account that the copy of the ifcfg files will be done only if
```
mchf: not only ifcfg files

doc/autoinstallation.md
mchf: this client is not part of yast2-network but autoyast itself. Autoyast just generates a list of clients to call. The list is generated using "I-always-forget-option-name" from desktop file(s). So to get to the point, I think that this paragraph is useless and doesn't belong into network's documentation.

lan_auto client is THE client which uses AutoYaST api (Read, Write, Import, Export, ... and other calls) - entry point for AutoYaST either in autoinstallation even in running system.

doc/installation.md
```diff
@@ -0,0 +1,98 @@
...
+  - finish 
+    - network_finish
+      - save_network
+
+Note: If you are interested in know more about installation process check 
+[this](https://github.com/yast/yast-installation/blob/master/doc/installation_overview.md) 
+documentation.
+
+### Linuxrc
```
mchf: I think that this is not intended to document linuxrc here.

doc/installation.md
```diff
@@ -0,0 +1,98 @@
...
+[here](https://www.suse.com/documentation/sled-12/singlehtml/book_sle_deployment/book_sle_deployment.html#sec.i.yast2.network) 
+and also will be available in some steps during the installation (in Leap 42.2 
+only for Addons).
+
+## inst_finish
+
+In this step is when the configuration is really copied from the running
+system to the installed one. This client calls more specialized clients to
+procceed with the configuration, in case of networking it will call
+`network_finish` with will call `save_network`.
```
mchf: I'd mention differences in case of "upgrade"

doc/autoinstallation.md
```diff
@@ -0,0 +1,94 @@
...
+like it is in the installation mode but the principal steps are not exactly
+the same, and also while in a normal installation the [Second
+Stage](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#overviewandconcept)
+is not needed, in autoinstallation the installation proccess is divided in
+two and in the _Second Stage_ is where and when the system configuration is
+really done.
+
+There are two ways to give a profile to _AutoYaST_, with (`autoyast` or with
+`autoyast2` parameters), the principal difference from networking point of 
+view is that `autoyast` will not fetch the profile which means that doesn't
```
mvidner: "`autoyast` will not fetch the profile" confusingly suggests that the profile is ignored. How about this:

> [the difference is that] `autoyast` leaves the fetching of the profile to YaST which means Linuxrc does not need to configure the network, while for `autoyast2` Linuxrc fetches the profile and may need need to configure the network.

doc/autoinstallation.md
```diff
@@ -0,0 +1,94 @@
...
+#### network_finish
+
+This client calls `save_network` which copies udev rules and ifcfg files
+from the running system which should be the linuxrc config and not the profile
+network configuration. 
+
+Take in account that the copying of the ifcfg files will be done only if 
+`keep_install_network` has not been set to `false` and in case that we need to
+configure the network before _SecondStage_ then `setup_before_proposal` has to
+be set to `true` as commented in the [autosetup section](#ausosetup) more
```
mvidner: #au**t**osetup

doc/autoinstallation.md
```diff
@@ -0,0 +1,94 @@
...
+details about network
+[here](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Network)
```
mvidner: This paragraph is hard to understand.

doc/autoinstallation.md
```diff
@@ -0,0 +1,94 @@
...
+## Second Stage
+
+The _SecondStage_, also known as `continue` in the control file, is basically 
+the stage where the configuration should be definitively written.
+
+### autoconfigure
+
+This client will read the [desktop configuration
+files](https://yastgithubio.readthedocs.io/en/latest/autoyast-development/#desktop-configuration-file)
+and is very probably that will write the configuration of our network calling
```
mvidner: "Probably"? I would like my documentation to be more deterministic. Can we remove the uncertainty?

doc/autoinstallation.md
```diff
@@ -0,0 +1,94 @@
...
+the `lan_auto` client with the information read in the `networking` section, but
+as mentioned, it depends on the information of the desktop files to do its
```
mvidner: "Depends on the desktop files." Well OK, but they change as rarely as the code. It's a different game than the AY profile or the control file. So let's assume them constant.

doc/installation.md
```diff
@@ -0,0 +1,100 @@
+## Introduction
+
+The workflow of the installation is customized via the [control
+file](https://github.com/yast/yast-installation/blob/master/doc/control-file.md), 
+and below we have the principal steps concerning network configuration:
+
+  - Linuxrc
+  - inst_install_inf (What is still needed)
```
mvidner: "What is still needed" sounds like an editorial TODO. IMHO it can be left out of this overview

doc/installation.md
```diff
@@ -0,0 +1,100 @@
...
+This client will try to configure dhcp in all the connected cards that haven't 
+been configured yet but only in the case that linuxrc has not activated one
+previously (i.e. with some parameter that implies a remote connection).
+
+## Manual Configuration (lan client)
+
+In case that the network was not configured then the **lan** client will be
+launched after the welcome dialog as you can see
+[here](https://www.suse.com/documentation/sled-12/singlehtml/book_sle_deployment/book_sle_deployment.html#sec.i.yast2.network) 
+and also will be available in some steps during the installation (in Leap 42.2
```
mvidner: "will be available in **some** steps" is not as clear and definite as I'd like. Does this depend on the control file? If yes, can we describe the typical choices?  Like: "for SLE, the registration needs the network, therefore it is set up at $this point, but for Leap..."

doc/autoinstallation.md
teclator: Quite better now

doc/autoinstallation.md
teclator: OK, make sense

teclator:
I think that maybe make sense to split this PR in two, just keep this one with the generic features and a new one for installation and autoinstallation workflows. I could add more examples there and make it more clear if needed. 
Just an example of a typical bug: https://bugzilla.suse.com/show_bug.cgi?id=1010611

PR #465, opened 2016-11-16
Backport udev improvements for SP1 (bsc#1005721)

teclator:
It is a backport of this https://github.com/yast/yast-network/pull/462, I have cherry-picked commits which where related to udev.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -81,3 +81,40 @@ Style/AccessorMethodName:
 Style/TrivialAccessors:
   Enabled: false
 
+# All the configuration below this comment tries to avoid offenses because of backported
+# code that uses different rubocop style.
+#
+# See https://github.com/yast/yast-network/pull/465
+
+Style/AccessModifierIndentation:
+  EnforcedStyle: outdent
+
+Style/ElseAlignment:
+  Exclude:
+    - src/modules/LanItems.rb
+    - src/include/network/lan/address.rb
+    - src/include/network/widgets.rb
+    - src/clients/firewall_stage1_proposal.rb
+    - src/modules/NetworkStorage.rb
+    - src/clients/inst_do_net_test.rb
+    - src/modules/SuSEFirewall4Network.rb
+
+Lint/EndAlignment:
+  Exclude:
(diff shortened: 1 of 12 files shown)


teclator:
Travis is complaining because of rubocop, I'm my laptop I'm using yast2-devtools 3.1.46-6.1 and for SP1 it is 3.1.38, as it is a backport I think that does not make to adapt the style to just pass that check.

mchf:
Well I tent to agree with you, but on the other hand, we will have to fight with that even in future maintenance requests. So it would be better to fix it by hand here, to have all green.

mchf:
apart of rubocop LGTM. I'd wait with merge till monday for a result of discussing this with the rest of the team.

teclator:
@mchf thnx!, ok I will wait, it is always better to take decisions on Monday than on Friday

imobachgs:
@mchf I think that the rubocop issue is something we should discuss within the team. We've narrowed the problem: in the past (until October) for SP1 we were using the `YaST:Head:Travis` repository, which contained a new version of `yast2-devtools`. So we validated the changes using that version.

@lslezak [adapted the script](https://github.com/yast/yast-devtools/commit/059d987aeeac690cd5fabb4ae708564db4069fe6) to use the right repo (`YaST:SLE-12:SP1:Travis`) and, since then, we started to use the old rules again for SP1 (`yast2-devtools` 3.1.38).

IMHO, as this is a L3, I would merge this change and start discussing within the team how to resolve this issue for the future.

mchf:
@imobachgs the patch for L3 was provided by @teclator already so merging is not in hurry. Thanks for details.

.rubocop.yml
```diff
@@ -81,3 +81,40 @@ Style/AccessorMethodName:
 Style/TrivialAccessors:
   Enabled: false
 
+# All the configuration below this comment tries to avoid offenses because of backported
```
teclator: because of backported code :)

Not really, these are rules and files that were failing, so just exclude them.

mchf: ok

mchf:
lgtm

PR #469, opened 2016-11-22
Bsc 1010944

imobachgs:
Fix for [bsc#1010994](https://bugzilla.suse.com/show_bug.cgi?id=1010994). BTW, I've added some tests for the `CFA::Hosts` class.


jreidinger:
LGTM

PR #470, opened 2016-11-24
Merged SLE-12-SP2 into master

mchf:


package/yast2-network.spec
```diff
@@ -35,10 +35,13 @@ BuildRequires:  rubygem(yast-rake)
 BuildRequires:  yast2 >= 3.1.86
 # Network: Adapt old enslaved interface config
 Requires:       yast2 >= 3.1.206.2
+<<<<<<< HEAD
```
teclator: This is not needed :)

mchf: dammed ... I always forget something

mchf:
Thanks for reviews and tip for the tool.

PR #473, opened 2016-11-30
Let yast-network produces noarch package

mchf:
since there is no arch dependency


package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Nov 30 09:04:58 UTC 2016 - mfilka@suse.com
+
+- bnc#1012764
+  - made rpm architecture independent (noarch)
```
jreidinger: can we do such change between SP? otherwise LGTM

mchf: I've talked with @kobliha and he accepted that for SP3, I'll ask him for confirmation even here.

kobliha: yes, approved

PR #475, opened 2016-12-07
Fixed internal error when building device overview

mchf:
[bsc#1013684](https://bugzilla.suse.com/show_bug.cgi?id=1013684)

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -1382,7 +1382,7 @@ module Yast
       if ip =~ /DHCP/
         bullets << format("%s %s", _("IP address assigned using"), ip)
       elsif IP.Check(ip)
-        prefixlen = dev_map["PREFIXLEN"]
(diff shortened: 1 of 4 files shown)


src/modules/LanItems.rb
```diff
@@ -1382,7 +1382,7 @@ def ip_overview(dev_map)
       if ip =~ /DHCP/
         bullets << format("%s %s", _("IP address assigned using"), ip)
       elsif IP.Check(ip)
-        prefixlen = dev_map["PREFIXLEN"]
+        prefixlen = dev_map["PREFIXLEN"] || @prefix
```
mvidner: This seems wrong: `ip_overview` operates on its argument, but here we reach out for some global state.

This method has no tests. How did the `dev_map` look like for the reported bug? Make a new test with that?

Test coverage shows that the method does get some testing, but the `if` bodies are not covered.

mchf:
Thanks for review

PR #476, opened 2016-12-09
dhclient set hostname tweaks

mchf:
[bsc#1013605](https://bugzilla.suse.com/show_bug.cgi?id=1013605) openQA test fails: Hostname Error Pop-Up During Installation

\+ listbox is enabled only when wicked service is selected

### DIFF
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -45,6 +45,7 @@ module Yast
       Yast.import "LanItems"
       Yast.import "Popup"
       Yast.import "Map"
+      Yast.import "NetworkService"
 
       Yast.include include_target, "network/routines.rb"
       Yast.include include_target, "network/widgets.rb"
@@ -420,7 +421,7 @@ module Yast
 
     # Init handler for DHCP_HOSTNAME
     def InitDhcpHostname(_key)
-      UI.ChangeWidget(Id("DHCP_HOSTNAME"), :Enabled, has_dhcp?)
+      UI.ChangeWidget(Id("DHCP_HOSTNAME"), :Enabled, has_dhcp? && NetworkService.is_wicked)
 
       hostname_ifaces = LanItems.find_set_hostname_ifaces
       selected = DNS.dhcp_hostname ? ANY_LABEL : NONE_LABEL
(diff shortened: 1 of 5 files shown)


src/include/network/services/dns.rb
```diff
@@ -430,7 +431,7 @@ def InitDhcpHostname(_key)
         fix_dhclient_warning(LanItems.invalid_dhcp_cfgs)
 
         selected = NO_CHANGE_LABEL
-        items << [Item(Id(NO_CHANGE_LABEL), _("keep current settings"), true)]
+        items << Item(Id(NO_CHANGE_LABEL), _("keep current settings"), true)
```
mvidner: `- fixed displaying the "keep current settings" option`

src/lib/network/network_autoyast.rb
```diff
@@ -22,6 +22,19 @@ def initialize
       Yast.import "Linuxrc"
     end
 
+    # Clears ifcfg files to be ready for proposals done later by the installer
+    #
+    # @return [void]
+    def clear_ifcfgs
```
mvidner: The name and the comment make it sound like the entire contents of the ifcfgs are emptied. Compare http://www.rubydoc.info/search/stdlib/core?q=clear

src/include/network/services/dns.rb
mchf: Ah this was the one pointed during meeting. I didn't check that yet :-/ Good to know that it is going to be fixed too ;-)

src/lib/network/network_autoyast.rb

mchf:
1) I plan to touch linuxrc too - learning opportunity 
2) I don't know If I found all flows which can lead to such inconsistent confirg - one never know what can linuxrc do next time, so
3) I think that implementing that in yast makes installer more robust. Especially when all new code fits into approx. five lines

mvidner:
Proposed linuxrc fix: https://github.com/openSUSE/linuxrc/pull/131

okurz:
hi, the corresponding linuxrc PR has been merged but we still see problems like e.g. https://openqa.opensuse.org/tests/324003#step/yast2_lan/12
I assume this PR here intends to fix it?

nilxam:
ah, linuxrc update is pending in staging A now https://build.opensuse.org/request/show/446728 , will get it into Factory asap

mchf:
@okurz no, the fix was moved into linuxrc completely. This PR fixes some minor issues only.

mvidner:
Closing this one in favor of #477.

@okurz no, that problem is because the linuxrc fix is not yet being used.

PR #482, opened 2016-12-22
Sle 12 sp2 bnc 1016164

schubi2:
An array produces with SCR multiline entries. (Same behavior as fstab).

I have tried Write(.etc.hosts."10.20.1.29", "beholder.suse.cz beholder") which produces an error.
So the docu is false too.


schubi2:
After a discussion with Knut we will let this behavior and are declaring it as a "feature"

PR #483, opened 2017-01-04
Forbid the addition of wlan interfaces to bridge proposal (bsc#956755)

teclator:


src/modules/Lan.rb
```diff
@@ -818,154 +818,76 @@ def UseNetworkManager
 
     def IfcfgsToSkipVirtualizedProposal
       skipped = []
-      Builtins.foreach(LanItems.Items) do |current, _config|
-        ifcfg = Ops.get_string(LanItems.Items, [current, "ifcfg"], "")
-        if NetworkInterfaces.GetType(ifcfg) == "br"
+
+      LanItems.Items.each do |_current, config|
+        ifcfg = config["ifcfg"]
+        ifcfg_type = NetworkInterfaces.GetType(ifcfg)
```
mchf: np: ifcfg_type is (almost) unused

src/modules/Lan.rb
```diff
@@ -818,154 +818,76 @@ def UseNetworkManager
...
+
+        case ifcfg_type
+        when "br"
           NetworkInterfaces.Edit(ifcfg)
```
mchf: no reason for that in my POV - ifcfg is being skipped

mchf: Exactly speaking, you need it just because of accessing ```NetworkInterfaces.Current``` a few lines bellow. As this call is incorrect itself (it directly access internal structure of NetworkInterfaces) and use of Edit is misleading here (no editation is done) I propose to use ```LanItems.GetDeviceMap``` instead of ```NetworkInterfaces.Current```. ```NetworkInterfaces.Edit``` can be dropped in such case.

src/modules/Lan.rb
```diff
@@ -818,154 +818,76 @@ def UseNetworkManager
...
+            log.info("For interface #{ifcfg} found slave #{slave}")
+            skipped << slave
           end
-        end
-        # Skip also usb device as it is not good for bridge proposal (bnc#710098)
-        if NetworkInterfaces.GetType(ifcfg) == "usb"
+
+        # Skip also usb and wlan devices as they are not good for bridge proposal (bnc#710098)
+        when "usb", "wlan"
           NetworkInterfaces.Edit(ifcfg)
```
mchf: useless

src/modules/Lan.rb
teclator: True. I just adapted the code doing a small cleanup. In general was thinking in use LanItems.IsBridgeable instead of this method as them check almost the same.

src/modules/Lan.rb
```diff
@@ -818,155 +818,64 @@ def UseNetworkManager
...
+    end
+
+    def ProposeVirtualized
       # then each configuration (except bridges) move to the bridge
       # and add old device name into bridge_ports
-      Builtins.foreach(LanItems.Items) do |current, _config|
-        ifcfg = Ops.get_string(LanItems.Items, [current, "ifcfg"], "")
-        if Builtins.contains(skipped, ifcfg)
-          Builtins.y2milestone("Skipping interface %1", ifcfg)
+      LanItems.Items.each do |current, config|
```
mchf: I'd avoid use of ```current``` name as it has special meaning in LanItems

src/modules/Lan.rb
```diff
@@ -1088,6 +997,68 @@ def activate_network_service
...
+      return false if !NetworkInterfaces.Edit(ifcfg)
+
+      old_config = deep_copy(NetworkInterfaces.Current)
+      log.debug("Old Config #{ifcfg}\n#{old_config}")
+
+      log.info("old configuration #{ifcfg}, bridge #{bridge_name}")
+
+      NetworkInterfaces.Name = bridge_name
+
+      # from bridge interface remove all bonding-related stuff
```
mchf: why only bonding stuff is being removed? Why not even vlan, ptp and so on? I think that using ```LanItems::SetDeviceMap``` would shorten this function a bit and it should remove direct NetworkInterfaces cache hacking.

src/modules/Lan.rb
```diff
@@ -1088,6 +997,68 @@ def activate_network_service
...
+      if hwinfo.fetch("type", "") == "wlan"
+        log.warn("not proposing WLAN interface")
+
+        return false
+      end
+
+      LanItems.ProposeItem
+    end
+
+    def refresh_lan_items
```
mchf: this function is a bit dangerous as it forces complete service restart

teclator: I haven't changed the current behavior just make it more readable, with a little cleanup. As commented, this is a first step just for what are you suggesting, which things are still valid, which could be removed/reimplemented and which needs a better implementation. 

I just wanted to fix the original issue with a small cleanup, but opening the discussion for improvements.

teclator:
@mchf thnx, as commented and for record, this is just a small cleanup with a fix for the bug, but not touch or refactorize current implementation.

PR #486, opened 2017-01-11
Create udev rule when not exist a previous one. (#996879)

teclator:
During an installation some interfaces are ignored by the udev rules generator like KVM, VMWare and Hyper-V virtual interfaces as you can see here:

 http://git.kernel.org/cgit/linux/hotplug/udev.git/tree/src/rule_generator/75-persistent-net-generator.rules#n35

In that cases the current code does not assign correctly new udev rules for that interfaces and this PR solves that.

I tested an autoinstallation with VMWare mac addresses with and without renaming interfaces successfully.

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -485,6 +485,9 @@ module Yast
     # @param new_val     [string] value for new key
     # @return updated rule when replace_key is found, current rule otherwise
     def ReplaceItemUdev(replace_key, new_key, new_val)
+      Items()[@current]["udev"] = { "net" => {} } if !Items()[@current]["udev"]
+      new = Items()[@current]["udev"]["net"].to_a.empty?
+
       # =    for assignment
       # ==   for equality checks
       operator = new_key == "NAME" ? "=" : "=="
@@ -498,12 +501,15 @@ module Yast
       new_rule = AddToUdevRule(rule, "#{new_key}#{operator}\"#{new_val}\"")
       new_rule.push(name_tuple)
 
-      log.info("ReplaceItemUdev: new udev rule = #{new_rule}")
-
-      if current_rule.sort != new_rule.sort
+      if new || current_rule.sort != new_rule.sort
         SetModified()
 
-        Items()[@current]["udev"] = { "net" => {} } if !Items()[@current]["udev"]
+        if new
+          log.info("Creating a new udev rule for #{LanItems.GetCurrentName}: #{rule}")
+        else
+          log.info("ReplaceItemUdev: new udev rule = #{new_rule}")
+        end
+
         Items()[@current]["udev"]["net"] = new_rule
       end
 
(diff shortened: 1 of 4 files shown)


src/modules/LanItems.rb
```diff
@@ -485,6 +485,9 @@ def update_item_udev_rule!(based_on = :mac)
     # @param new_val     [string] value for new key
     # @return updated rule when replace_key is found, current rule otherwise
     def ReplaceItemUdev(replace_key, new_key, new_val)
+      Items()[@current]["udev"] = { "net" => [] } if !Items()[@current]["udev"]
```
mchf: well I believe that this line is not needed bcs of this https://github.com/yast/yast-network/blob/master/src/modules/LanItems.rb#L1085

src/modules/LanItems.rb
```diff
@@ -485,6 +485,9 @@ def update_item_udev_rule!(based_on = :mac)
...
+      new = Items()[@current]["udev"]["net"].to_a.empty?
```
mchf: I'd avoid using ```new```

teclator: completely agree

src/modules/LanItems.rb
teclator: It is just a save check, and was part of the method just moved to the beginning and used and empty array instead of a hash as the default value. 

If I remember correctly there was some issue in the past because of not ["udev"] when renaming , removed or something like that so... 

as for example check added in https://github.com/yast/yast-network/pull/284 or bug https://bugzilla.suse.com/show_bug.cgi?id=914833 I could assume that, but I'm being conservative.

teclator: Other question is consider nil as a value to replace or not.. I mean, move that responsibility to an specific method.

mchf: I know, it was there.Let's keep it. I've no big problem with that.

src/modules/LanItems.rb
```diff
@@ -488,7 +488,7 @@ def ReplaceItemUdev(replace_key, new_key, new_val)
       # =    for assignment
       # ==   for equality checks
       operator = new_key == "NAME" ? "=" : "=="
-      current_rule = getUdevFallback
+      current_rule = GetItemUdevRule(@current)
```
teclator: Just obtain the current rule without generating a new one if not exist should fix it.

mchf: +1, beat getUdevFallback everywhere you can ;-)

mchf:
i think that testsuite deserves some love bcs there is lot of mocking of ```getUdevFallback``` which should not be needed anymore, so please double check it.

teclator:
Well, `getUdevFallback` is still used for the `new_rule`, when replacing the key, only was removed from the `current_rule`.

PR #488, opened 2017-01-16
Added LanItems summary

teclator:
This is just the firs request for comments, I could make it simpler if needed or if we consider it as over-engineering.

I decided to use the LanItems.summary but permitting to used different types of summaries being the Default based on Summary class as it is in others networks modules like:

https://github.com/yast/yast-network/blob/master/src/modules/Host.rb#L263
https://github.com/yast/yast-network/blob/master/src/modules/DNS.rb#L532
https://github.com/yast/yast-network/blob/master/src/modules/DNS.rb#L532

So for CASP one click dialog we could use:

 `Yast::LanItems.summary("one_line")`
or
`Yast::LanItemsSummary.new.one_line`

> === OLD ===
> 
>  `Yast::LanItems.summary(:type => "one_line")`
> or
> `Yast::LanItemsSummary.new(:type => "one_line").summary`
> or
>   `Yast::LanItemsSummary::OneLine.new.summary`
> 

And in case of a proposal in the future just `Yast::LanItems.summary`

Here screenshots with the current CASP one dialog overview: yast/yast-installation#487

### With one interface configured 
networkoneinterface

### With multiple interfaces and different bootproto
networkmultipleinterfaces

### DIFF
--- a/src/lib/network/lan_items_summary.rb
+++ b/src/lib/network/lan_items_summary.rb
@@ -1,0 +1,132 @@
+# encoding: utf-8
+
+# Copyright (c) [2017] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+
+require "yast"
+
+module Yast
+  # This class creates a summary of the configured lan items supporting
+  # different types of summaries.
+  #
+  # @example
+  #   LanItemsSummary.new.summary
+  #   => "<ul><li><p>eth0<br>DHCP</p></li><li><p>eth1<br>NONE</p></li></ul>"
+  # @example
+  #   LanItemsSummary.new(:type => 'one_line').summary
+  #   => "Multiple Interfaces"
+  class LanItemsSummary
+    attr_accessor :summary_type, :options
+
+    def initialize(options = {})
+      Yast.import "LanItems"
+      Yast.import "Summary"
+
+      @summary_type =
+        case options[:type]
+        when "", nil
+          Default
+        when "one_line"
+          OneLine
+        else
+          raise NotImplementedError,
+            "The LanItems summary #{options[:type]} has not been implemented yet"
+        end
+      @options = options.reject { |k, _| k == :type }
+    end
+
+    # Delegates the summary to the specific class depending on the specific.
+    # Uses Default if no type is given.
+    def summary
+      @summary_type.send(:new, @options).summary
+    end
+
+    class Base
+      include I18n
+
+      attr_accessor :options
+
+      def initialize(options)
+        @options = options
+      end
+    end
+
+    # This class generates a one line text summary.
+    class OneLine < Base
+      def initialize(options = {})
+        @options = options
+      end
+
+      def summary
+        protocols  = []
+        configured = []
+        output     = []
+
+        Yast::LanItems.Items.each do |item, conf|
+          next if !LanItems.IsItemConfigured(item)
+
+          ifcfg = LanItems.GetDeviceMap(item) || {}
+
+          protocol = LanItems.DeviceProtocol(ifcfg)
+
+          protocols <<
+            if protocol =~ /DHCP/
+              "DHCP"
+            elsif IP.Check(protocol)
+              "STATIC"
+            else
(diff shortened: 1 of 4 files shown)


imobachgs:
I'm fine with the general approach (having a new class that takes care of generating the summary), but I would simplify it a little bit:

* Using classes for each type of summary looks overkill to me (after all, logic is not that complex and you are not maintaining state). Maybe I would implement different methods and that's all.
* I would avoid using a generic `@options` hash. I rather prefer to use real parameters (using keywords, for example). If we're going to have a lot of options, then `@options` would be a reasonable approach.

Just my opinion :)

mchf:
for the first view it seems to be a bit over engineered, I'll give it even second look ;-)

teclator:
I tried to make it general and as much open as possible. For example with the **OneLine** type currently we only show on interface or multiple interfaces, but maybe we could change the number of the items to show per line being an option to pass, but don't like to use as a specific option in LanItem, but as commented.

The current types are mostly like presenters, I could by now instead of create separate class just call LanItemsSummary.send(@summary_type) and move the logic to methods as @imobachgs has suggested.

src/lib/network/lan_items_summary.rb
```diff
@@ -0,0 +1,132 @@
...
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
```
mvidner: The license is trunca

teclator: Fixed

teclator:
@mchf @mvidner ok, as no updates I will modify current PR based on @imobachgs review.

test/lan_items_summary_test.rb
```diff
@@ -0,0 +1,91 @@
...
+      { "IPADDR" => "1.2.3.4", "NETMASK" => "255.255.255.0" }
+    ].freeze
+  end
+
+  let(:items) do
+    {
+      0 => { "ifcfg" => "eth0" },
+      1 => { "ifcfg" => "eth1" },
+      2 => { "ifcfg" => "br0" }
+    }
```
imobachgs: .freeze?

test/lan_items_summary_test.rb
```diff
@@ -0,0 +1,91 @@
...
+  before do
+    allow(Yast::LanItems).to receive(:Items).and_return(items)
+    allow(Yast::LanItems).to receive(:IsItemConfigured).and_return(true)
+    dhcp_maps.each_with_index do |item, index|
+      allow(Yast::LanItems).to receive(:GetDeviceMap).with(index).and_return(item)
+    end
+  end
+
+  describe "#default" do
+    it "retuns a Richtext summary of the configured interfaces" do
```
imobachgs: returns

test/lan_items_summary_test.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  describe "#one_line" do
+    it "returns a plain text summary of the configured interfaces in one line" do
+      expect(subject.one_line).to eql "Multiple Interfaces"
+    end
+
+    context "when there are no configured interfaces" do
+      let(:items) { {} }
+      it "returns 'Not configured'" do
```
imobachgs: Above you're referencing to `Yast::Summary.NotConfigured`. Please, keep consistency.

teclator: Yes, but Summary.NotConfigured returns richtext 

```Yast::Summary.NotConfigured => <ul><li>Not configured yet.</li></ul>```

Comment (imobachgs) [inline test/lan_items_summary_test.rb]:
```ruby
Yast.import "Summary"
Yast::Summary.NotConfigured #=> "Not configured yet."
```

src/lib/network/lan_items_summary.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      output << protocols.first if protocols.uniq.size == 1
+
+      case configured.size
+      when 0
+        return _("Not configured")
+      when 1
+        output << configured.join(", ")
+      else
+        output << "Multiple Interfaces"
```
kobliha: PLS mark text for translators

teclator: Updated!

test/lan_items_summary_test.rb
teclator: Sorry, I was thinking in the previous output that was based on Yast::Summary.DevicesList([]), updated!, thnx for check it.

test/lan_items_summary_test.rb
```diff
@@ -19,9 +19,10 @@
       0 => { "ifcfg" => "eth0" },
       1 => { "ifcfg" => "eth1" },
       2 => { "ifcfg" => "br0" }
-    }
+    }.freeze
   end
   let(:no_interfaces) { _("Not configured") }
+  let(:multiple_interfaces) { _("Multiple Interfaces") }
```
imobachgs: Maybe these ones are just constants.

PR #489, opened 2017-01-17
Added LanItemsSummary as part of #Fate322328

teclator:
This PR superseeds #488.

I have backported some code that was used by the original PR and which fixes some bugs for TW.

The commits picked are from #463 & #475 .

Maybe could be interesting also the changes related to dhcp hostnames in #477 at least for CASP branch.

PR #491, opened 2017-01-19
Added network proposal for CASP (Fate#322328)

teclator:
- https://trello.com/c/XDgl2dj7/816-13-casp-new-all-in-one-dialog
- Main PR: https://github.com/yast/yast-installation/pull/487

Uses the new summary API added in #489 and uses the new proposal_client flag ["label_proposal"] introduced in yast-yast2#530

### DIFF
--- a/test/network_proposal_test.rb
+++ b/test/network_proposal_test.rb
@@ -1,0 +1,64 @@
+require_relative "test_helper"
+
+require "network/clients/network_proposal"
+
+class DummyLanItems
+  def summary(params)
+    params == "one_line" ? "one line summary" : "rich_text_summary"
+  end
+end
+
+class DummyLan
+  def Summary(_)
+    ["rich_text_summary"]
+  end
+end
+
+describe Yast::NetworkProposal do
+  subject { Yast::NetworkProposal.new }
+
+  let(:lan_items_mock) { DummyLanItems.new }
+  let(:lan_mock) { DummyLan.new }
+
+  before do
+    stub_const("Yast::LanItems", lan_items_mock)
+    stub_const("Yast::Lan", lan_mock)
+  end
+
+  describe "#description" do
+    it "returns a map with id, menu_title and rich_text_title " do
+      expect(subject.description).to include("id", "menu_title", "rich_text_title")
+    end
+  end
+
+  describe "#make_proposal" do
+    it "returns a map with 'label_proposal' as an array with one line summary'" do
+      expect(subject.make_proposal({})["label_proposal"]).to eql(["one line summary"])
+    end
+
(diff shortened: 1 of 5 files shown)


jreidinger:
code looks fine for me, but what about some tests? We should improve test coverage in network 😉

teclator:
@jreidinger added test for network proposal.

test/network_proposal_test.rb
```diff
@@ -0,0 +1,64 @@
...
+  def Summary(_)
+    ["rich_text_summary"]
+  end
+end
+
+describe Yast::NetworkProposal do
+  subject { Yast::NetworkProposal.new }
+
+  let(:lan_items_mock) { DummyLanItems.new }
+  let(:lan_mock) { DummyLan.new }
```
jreidinger: what about mocking real methods instead of using fake methods? And if you really want to use this approach. I really suggest to use double instead of defining own global classes. e.g. for DummyLan you can use `stub_const("Yast::Lan", double(Summary: ["rich_text_summary"])`

test/network_proposal_test.rb
```diff
@@ -0,0 +1,64 @@
...
+    it "returns a map with 'preformatted_proposal' as an array with the network summary'" do
+      expect(subject.make_proposal({})["preformatted_proposal"]).to eql(["rich_text_summary"])
+    end
+  end
+
+  describe "#ask_user" do
+    Yast.import "Wizard"
+
+    before do
+      allow(Yast::Wizard)
```
jreidinger: allow what? what this syntax do? I never see it

test/network_proposal_test.rb
teclator: Mocked real methods.

test/network_proposal_test.rb
teclator: Just thinking in allow any Wizard method to not mock the open and close dialog, so used null object at the end.

test/network_proposal_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  Yast.import "LanItems"
+  Yast.import "Lan"
+
+  before do
+    stub_const("Yast::Wizard", double.as_null_object)
+    allow(Yast::LanItems).to receive(:summary).with("one_line").and_return("one_line_summary")
+    allow(Yast::Lan).to receive(:Summary).with("summary").and_return(["rich_text_summary"])
+  end
+
+  subject { Yast::NetworkProposal.new }
```
jreidinger: this should work out of box without this line. Why it is needed?

teclator: Not needed

PR #492, opened 2017-01-20
Fixing version (#491) and submit to 'casp10' target

teclator:


package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        3.1.174
+Version:        3.1.173.1
```
imobachgs: Just for the records, the changes file looks fine. This line fixes a problem in #491.

PR #477, opened 2016-12-16
Hostname dhcp fixes

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=1013605
https://trello.com/c/pN6f3kzr

PR #495, opened 2017-01-24
Backport dhcp hostnames fixes from #477 (bnc#1013605)

teclator:
Some fixes regarding to setting the hostname by dhcp were adding in yast/yast-network#461, and later also some fixes concerning linurxc but only in master yast/yast-network#477. This is the related trello card https://trello.com/c/pN6f3kzr.

So this PR will backport that fixes into CASP because linuxrc has been already updated with the changes of this PR: openSUSE/linuxrc#131.

That changes are not yet in SLE-12-SP2 so should wait until then.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,10 +1,8 @@
 require "yast/rake"
 
+Yast::Tasks.submit_to :casp10
+
 Yast::Tasks.configuration do |conf|
   # lets ignore license check for now
   conf.skip_license_check << /.*/
-  conf.obs_api = "https://api.suse.de/"
-  conf.obs_target = "CASP_1.0"
-  conf.obs_sr_project = "SUSE:SLE-12-SP2:Update:Products:CASP10"
-  conf.obs_project = "Devel:YaST:CASP:1.0"
 end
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -1,10 +1,8 @@
 require "yast/rake"
 
+Yast::Tasks.submit_to :casp10
```
teclator: Use the 'casp10' target added in https://github.com/yast/yast-rake/pull/38 as it is already in CASP.

mvidner:
> That changes are not yet in SLE-12-SP2 so should wait until then.

Sorry, which changes are not there, and what should wait, and until when? Until merging the former to SP?

teclator:
linuxrc changes are not in SP2 yet, I could also add the backport to SP2 but as linuxrc has not been updated there yet and this branch already diverged I think that is preferable to backport in CASP and then in SP2 but as you also mentioned in trello, the package was not part of the installation media which is where the bug should arise

mchf:
LGTM

mvidner:
> as you also mentioned in trello, the package was not part of the installation media which is where the bug should arise

Yes, but I had not realized that this would affect the CaaSP installation too

teclator:
@mvidner it really could be also merged into SP2 without problem, but also with the network proposal added in #491 or only this fixes?

teclator:
@mvidner @mchf so, I will merge just in CASP by now and if decided later not diverge adding also the network proposal in SP2.

PR #496, opened 2017-01-25
Backport dhcp hostname fixes #477 and added network_proposal #491 (bnc#1013605)

teclator:
If you want to know more, you can check #495, so basically the dhcp hostname fixes added in #477 were backported to CASP and the network proposal added was added to CASP in #491.

Both PRs can also be backported to SP2 which means that we could not diverge yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ before_install:
     # disable rvm, use system Ruby
     - rvm reset
     - wget https://raw.githubusercontent.com/yast/yast-devtools/master/travis-tools/travis_setup.sh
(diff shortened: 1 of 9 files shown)


mchf:
ok, why do we need network proposal in SP2?

teclator:
We don't need it, is just for being it as synced with CASP as possible for being easier in the future to merge changes.

The network proposal is not used in SP2 but not hurts and is not expected to be used.

mchf:
I'm just thinking if it can somehow make troubles with duds in the future and how big deal it is. However as proposal needs also an update in control file and there is no much other updates. So, lets go.

PR #500, opened 2017-02-22
Fixed internal error raised for incomplete udev rule in AY profile

mchf:
bnc#1020074


mchf:
DUD required so closing until explicitly requested

PR #501, opened 2017-02-22
Fixed internal error raised for incomplete udev rule in AY profile

mchf:
bnc#1020074

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -218,6 +218,15 @@ module Yast
       ay_profile["networking"]
     end
 
+    # Checks if the udev rule is valid for renaming a NIC
+    def valid_rename_udev_rule?(rule)
+      return false if rule["name"].nil? || rule["name"].empty?
+      return false if rule["rule"].nil? || rule["rule"].empty?
+      return false if rule["value"].nil? || rule["value"].empty?
+
+      true
+    end
+
     # Takes a list of udev rules and assignes them to corresponding devices.
     #
     # If a device has an udev rule defined already, it is overwritten by new one.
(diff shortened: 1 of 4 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -228,7 +237,11 @@ def assign_udevs_to_devs(udev_rules)
       udev_rules.each do |rule|
         name_to = rule["name"]
         attr = rule["rule"]
-        key = rule["value"].downcase
+        key = rule["value"]
+
+        next if !valid_rename_udev_rule?(rule)
```
mvidner: You can keep the original line if you move this one before it.

test/network_autoyast_test.rb
```diff
@@ -319,4 +319,33 @@ def keep_install_network_value(value)
       end
     end
   end
+
+  describe "#valid_rename_udev_rule?" do
+    it "fails when the rule do not contain new name" do
+      rule = { "rule" => "exists", "value" => "as_well" }
```
mvidner: Please use meaningful values when it does not hinder the test. Copying the values from the bug should do it.
>        <rule>
>          <name>eth0</name>
>          <rule>ATTR{address}</rule>
>          <value>00:02:29:69:d3:63</value>
>        </rule>

mvidner:
From a high level perspective, I dislike that this logic is added to the AutoYaST part of the code, but it checks the internal consistency of a udev rule. Instead, we should have a udev rule model class that has a `from_ay` constructor and a `valid?` method. But that may not be appropriate for an SP2 update.

mchf:
> From a high level perspective, I dislike that this logic is added to the AutoYaST part of the code, but it checks the internal consistency of a udev rule.

depends. I've though of it as a kind of "user input check". Mostly bcs it checks requirements we have for udev read from a profile. Generic udev rule check should be more detailed.

test/network_autoyast_test.rb
mchf: well the bug was about missing attributes ;-) but ok, I'll use these meaningful values.

src/lib/network/network_autoyast.rb
mchf: i know. This was a kind of cosmetic decision. Is it blocker?

mchf:
Thanks for review

PR #502, opened 2017-03-02
Merged SLE-12-SP2

mchf:


src/lib/network/network_autoyast.rb
```diff
@@ -218,6 +218,15 @@ def ay_networking_section
       ay_profile["networking"]
     end
 
+    # Checks if the udev rule is valid for renaming a NIC
+    def valid_rename_udev_rule?(rule)
+      return false if rule["name"].nil? || rule["name"].empty?
+      return false if rule["rule"].nil? || rule["rule"].empty?
+      return false if rule["value"].nil? || rule["value"].empty?
+
+      true
```
jreidinger: what about more dry check like:
```
["name", "rule", "value"].each do |key|
  return false if rule[key].nil? || rule[key].empty?
end

return true
```

src/lib/network/network_autoyast.rb
```diff
@@ -228,7 +237,11 @@ def assign_udevs_to_devs(udev_rules)
       udev_rules.each do |rule|
         name_to = rule["name"]
         attr = rule["rule"]
-        key = rule["value"].downcase
+        key = rule["value"]
+
+        next if !valid_rename_udev_rule?(rule)
+        key.downcase!
```
jreidinger: well, you modify now content if udev_rules, it is ok?

src/lib/network/network_autoyast.rb
mchf: matter of preference. My "simpler"  version is more readable for me.

src/lib/network/network_autoyast.rb
mchf: there should be no problem with that.

mchf:
Thanks for review

PR #510, opened 2017-04-11
use real test coverage

jreidinger:


mvidner:
Real numbers for test coverage are good!

But Travis fails ("git does not know about *.pot"), and it seems to be a result of stricter checks that you have enabled: https://github.com/yast/docker-yast-ruby/pull/4 . Can you please resolve that?

PR #512, opened 2017-05-17
Do not raise an internal error in installation when updating /etc/hosts

mchf:
[bsc#1038521](https://bugzilla.suse.com/show_bug.cgi?id=1038521)

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -292,10 +292,11 @@ module Yast
     # make sure /etc/hosts resolves it to our, bnc#664929
     def ResolveHostnameToStaticIPs
       static_ips = StaticIPs()
-      if Ops.greater_than(Builtins.size(static_ips), 0)
-        fqhostname = Hostname.MergeFQ(DNS.hostname, DNS.domain)
-        Update(fqhostname, fqhostname, static_ips)
-      end
+      return if static_ips.empty?
+
+      fqhostname = Hostname.MergeFQ(DNS.hostname, DNS.domain)
+
+      static_ips.reject(&:empty?).each { |sip| Update(fqhostname, fqhostname, sip) }
 
       nil
     end
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -279,12 +279,14 @@ module Yast
     def StaticIPs
       NetworkInterfaces.Read
       devs = NetworkInterfaces.Locate("BOOTPROTO", "static")
-      devs = Builtins.filter(devs) { |dev| dev != "lo" }
-      ips = Builtins.maplist(devs) do |dev|
-        NetworkInterfaces.GetValue(dev, "IPADDR")
-      end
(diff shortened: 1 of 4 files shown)


src/modules/Host.rb
```diff
@@ -292,10 +291,11 @@ def StaticIPs
...
       static_ips = StaticIPs()
-      if Ops.greater_than(Builtins.size(static_ips), 0)
-        fqhostname = Hostname.MergeFQ(DNS.hostname, DNS.domain)
-        Update(fqhostname, fqhostname, static_ips)
-      end
+      return if static_ips.empty?
+
+      fqhostname = Hostname.MergeFQ(DNS.hostname, DNS.domain)
+
+      static_ips.reject(&:empty?).each { |sip| Update(fqhostname, fqhostname, sip) }
```
mvidner: `StaticIPs` is defined in this file and this method is the only user. My conclusion is to instead fix that method not to return invalid data in the first place.

mvidner:
> This issue was a regression. The Host#Updated used to accept an array of
static IPs in the past. This feature / API was dropped during
refactoring. However, ResolveHostnameToStaticIPs still used the old API.

Regarding this I found a comment in the original refactoring PR (https://github.com/yast/yast-network/pull/454#discussion_r83167590)
> I have a suspicion that with the refactoring of Host.Update we have broken something. I will do some manual tests and report back.

It turns out the manual tests were not enough. Apparently we did have coveralls set up back then but it did not report in the PR. Even in this PR it does not report. I need to check this.

And, static type checking would have caught this too.

Cc @jreidinger

jreidinger:
I will check coveralls

jreidinger:
@mvidner in fact it is configured and reported also here. What exactly you missing?

mvidner:
Duh, the coverage report was there and it is still there! I was looking in the wrong place (the check+cross marks) but it is a regular comment. Sorry for the noise.

mchf:
>> I have a suspicion that with the refactoring of Host.Update we have broken something. I will do some  manual tests and report back.

> It turns out the manual tests were not enough. Apparently we did have coveralls set up back then but it did not report in the PR. Even in this PR it does not report. I need to check this.

> And, static type checking would have caught this too.

yes, I obviously did manual tests wrong or incomplete :-/

src/modules/Host.rb
mvidner: OK, and now we no longer need `:empty` here!

mchf:
empty ips was not the main / only problem. The problem (also) was that module tried to write multiple IPs for one entry - that caused the issue in the lens.

you can reproduce the issue even this way - boot sp3 installation with "ifcfg=eth0=1.1.1.1 ifcfg=eth1=2.2.2.2 hostname=sles.suse.de" - and you'll get the issue too.

mchf:
Thanks for the reviews

PR #513, opened 2017-05-18
bsc#1036440 and bsc#994471: Command "yast remote list" should not alter the system.

gilsonsouza:


lslezak:
The Travis fix is here: https://github.com/gilsonsouza/yast-network/pull/1

src/clients/remote.rb
```diff
@@ -56,8 +56,6 @@ def main
         ),
         "id"         => "remote",
         "guihandler" => fun_ref(method(:RemoteGUI), "any ()"),
-        "initialize" => fun_ref(Remote.method(:Read), "boolean ()"),
-        "finish"     => fun_ref(Remote.method(:Write), "boolean ()"),
         "actions"    => {
           "list"  => {
```
lslezak: Does the `list` command still work properly? The initialization has been moved from the global section only to the `allow` command. The question is do the other commands still work when the initialization is skipped?

gilsonsouza: You're right... `list` cannot know the status of the system without initialization.
I'll fix this.

src/clients/remote.rb
```diff
@@ -159,8 +158,9 @@ def SetRAHandler(options)
       else
         Remote.Disable
       end
+      success = Remote.Write
 
-      true
+      success
```
lslezak: NP: You can avoid storing the result to `success` variable, you can directly return the `Remote.Write` value without a temporary variable.

PR #517, opened 2017-05-25
Fixed updating of /etc/hosts during installation

mchf:
[bsc#1039532](https://bugzilla.suse.com/show_bug.cgi?id=1039532)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Thu May 25 09:27:09 UTC 2017 - mfilka@suse.com
+
+- bnc#1039532
+  - do not write hostname into /etc/hosts duplicitly for one IP
+  - configure static IPs without own hostname to resolve to system
+    wide hostname during installation
+- 3.2.28
+
+-------------------------------------------------------------------
 Mon May 22 13:19:22 UTC 2017 - gsouza@suse.com
 
 - bsc#1036440 and bsc#994471
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -289,14 +289,22 @@ module Yast
       static_ips
     end
 
-    # if we have a static address,
-    # make sure /etc/hosts resolves it to our, bnc#664929
+    # Configure system to resolve static ips without hostname to system wide hostname
+    #
+    # It is expected to be used during installation only. If user configures static
+    # ips during installation and do not assign them particular hostname, then such
+    # ips are configuret to resolve to the system wide hostname (see Hostname module,
+    # /etc/HOSTNAME)
+    #
+    # Originally implemented as a fix for bnc#664929, later extended for bnc#1039532
     def ResolveHostnameToStaticIPs
-      static_ips = StaticIPs()
+      # reject those static ips which have particular hostnames already configured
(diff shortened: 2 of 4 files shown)


src/modules/Host.rb
```diff
@@ -289,14 +292,22 @@ def StaticIPs
...
+    # It is expected to be used during installation only. If user configures static
+    # ips during installation and do not assign them particular hostname, then such
+    # ips are configuret to resolve to the system wide hostname (see Hostname module,
+    # /etc/HOSTNAME)
+    #
+    # Originally implemented as a fix for bnc#664929, later extended for bnc#1039532
     def ResolveHostnameToStaticIPs
-      static_ips = StaticIPs()
+      # reject those static ips which have particular hostnames already configured
+      static_ips = StaticIPs().reject { |sip| @hosts.include_ip?(sip) }
```
jreidinger: so looks like variable name is no longer descriptive as it is not static_ips, but static ips without hostname?

package/yast2-network.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Thu May 25 09:27:09 UTC 2017 - mfilka@suse.com
+
+- bnc1039532
```
jreidinger: missing `#`

src/modules/Host.rb
mchf: hostname has never been taken into account here

package/yast2-network.changes
mchf: fixed ;-)

package/yast2-network.changes
```diff
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Thu May 25 09:27:09 UTC 2017 - mfilka@suse.com
+
+- bnc#1039532
+  - do not write hostname into /etc/hosts duplicitly for one IP
+  - configure static IPs to resolve to system wide hostname during
+    installation
```
mvidner: But this was the case already before this fix wasn't it? We still call `ResolveHostnameToStaticIPs`.

mvidner:
From this PR I don't understand what the problem is. Maybe test cases would help?

package/yast2-network.changes
mchf: no, before the hostname was assigned to every static ip - even that which was configured separately.

mchf: e.g. you could assign ```hostname1``` to ip 1.1.1.1 when editing particular device and ```hostname2``` on Hostname/DNS tab and the result was ```1.1.1.1 hostname1 hostname2```. After this PR it is ```1.1.1.1 hostname1```

mchf:
Referenced issue was partly fixed by previous https://github.com/yast/yast-network/pull/516. Code changes done here are just improvements.

For example before, when hostname was defined via edit device dialog, the /etc/hosts' line looked like
```<ip> <hostname> <hostname>```. Now it is ```<ip> <hostname>```. Other details are described above.

package/yast2-network.changes
mchf: ah I finally see ... I forgot to write half of sentence as I already wrote it 2 or 3 times ;-)

mvidner:
@mchf I have a spaghetti feeling about this. Please add a test for the fix.

mchf:
Thanks for the reviews.

PR #518, opened 2017-05-30
Backported fix for /etc/hosts

mchf:
See https://github.com/yast/yast-network/pull/517


mchf:
Thanks for the review

PR #520, opened 2017-06-09
[Sle15] AY setup written in first stage

mchf:
* DNS supported
* /etc/hosts supported
* Routing supported

See https://github.com/yast/aytests-tests/pull/77 (DNS test)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,9 +1,9 @@
 -------------------------------------------------------------------
 Fri Jul 11 09:40:11 UTC 2017 - mfilka@suse.com
 
-- Moving network setup in AY into first stage
-  - DNS configuration is written when second stage is disabled in
-    the profile
+- bnc#1049814
+  - Moving network setup in AY into first stage completely when
+    the second stage is disabled.
 - 3.3.3
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Jul 11 09:40:11 UTC 2017 - mfilka@suse.com
+
+- Moving network setup in AY into 1s stage
+  - DNS configuration is written when second stage is disabled in
+    the profile
+- 3.3.3
+
+-------------------------------------------------------------------
 Mon Jul 10 15:03:09 UTC 2017 - jreidinger@suse.com
 
 - Fix crash during write if Host.Read and Host.Import is called
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Fri Jul 11 09:40:11 UTC 2017 - mfilka@suse.com
 
-- Moving network setup in AY into 1s stage
+- Moving network setup in AY into first stage
   - DNS configuration is written when second stage is disabled in
     the profile
 - 3.3.3
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -269,6 +269,10 @@ module Yast
       NetworkAutoconfiguration.instance.configure_hosts if !configured
     end
 
+    # Invokes configuration of parts which are in charge of Lan module
+    #
+    # Currently it handles just AutoYaST installation. It just exits in case
+    # of common installation as there currently is nothing to do.
     def configure_lan
       return if !Mode.autoinst
 
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -259,6 +259,16 @@ module Yast
       DNS.create_hostname_link
     end
 
+    # Creates target's /etc/hosts configuration
+    #
+    # It uses hosts' configuration as defined in AY profile (if any) or
+    # proceedes according the proposal
+    def configure_hosts
+      configured = false
+      configured = NetworkAutoYast.instance.configure_hosts if Mode.autoinst
+      NetworkAutoconfiguration.instance.configure_hosts if !configured
+    end
+
     # It does an automatic configuration of installed system
     #
     # Basically, it runs several proposals.
@@ -268,7 +278,7 @@ module Yast
       configure_dns
 
       # this depends on DNS configuration
-      NetworkAutoconfiguration.instance.configure_hosts
+      configure_hosts
 
       set_network_service
 
--- a/src/clients/save_network.rb
+++ b/src/clients/save_network.rb
@@ -281,19 +281,19 @@ module Yast
       NetworkAutoYast.instance.configure_nics
     end
 
+    def configure_lan
+      NetworkAutoYast.instance.configure_lan
+    end 
+
     # It does an automatic configuration of installed system
     #
     # Basically, it runs several proposals.
     def configure_target
       NetworkAutoconfiguration.instance.configure_virtuals
 
-      configure_nics
-      configure_dns
+      configure_lan
 
-      # this depends on DNS configuration
-      configure_hosts
-      # this depends on NICs configuration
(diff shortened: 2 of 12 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -106,24 +108,63 @@ def set_network_service
...
+    # @return [Boolean] true when configuration was present and loaded from the profile
+    def configure_hosts(write: false)
+      log.info("NetworkAutoYast: Hosts configuration")
+
+      # expected format for Hosts.Import is { "ip" => [list, of, names] }
+      hosts_config = (ay_host_section["hosts"] || {}).map do |host|
+        # we neet to guarantee order of the items here
+        [host["host_address"] || "", host["names"] || []]
+      end
+      hosts_config = hosts_config.to_h.delete_if { |k, v| k.empty? || v.empty? }
```
mvidner: This non-trivial FromAY-like section should go into its own method (class?) and be covered with tests.

src/modules/DNS.rb
```diff
@@ -691,6 +615,90 @@ def read_hostname_from_install_inf
...
+    # present => nil
+    def dhclient_set_hostname
+      SYSCFG_TO_BOOL[SCR.Read(path(".sysconfig.network.dhcp.DHCLIENT_SET_HOSTNAME"))]
+    end
+
+    # Reads value of WRITE_HOSTNAME_TO_HOSTS and translates it to boolean
+    #
+    # return {true, false, nil} "yes" => true, "no" => false, otherwise or not
+    # present => nil
+    def write_hostname_to_hosts
```
mvidner: A method named `write_*` that in fact calls `SCR.Read`?! I know it copies the sysconfig name but here I would prefix it with `should_` or something.

src/modules/DNS.rb
```diff
@@ -381,65 +330,40 @@ def Write
...
-        HOSTNAME_PATH,
-        Ops.add(fqhostname, "\n")
-      )
-
-      create_hostname_link
-
-      Builtins.sleep(sl)
+      ProgressNextStage(_("Writing hostname...")) if gui
+      update_hostname
+      Builtins.sleep(0) if gui
```
mvidner: Ugh, what?!

src/modules/DNS.rb
```diff
@@ -691,6 +615,90 @@ def read_hostname_from_install_inf
...
+        # @dhcp_hostname and @wrote_hostname can currently be nil only when
+        # not present in original file. So, do not add it in such case.
+        SCR.Write(
+          path(".sysconfig.network.dhcp.DHCLIENT_SET_HOSTNAME"),
+          @dhcp_hostname ? "yes" : "no"
+        ) if !@dhcp_hostname.nil?
+        SCR.Write(
+          path(".sysconfig.network.dhcp.WRITE_HOSTNAME_TO_HOSTS"),
+          @write_hostname ? "yes" : "no"
+        ) if !@write_hostname
```
mvidner: Bug.

src/lib/network/network_autoyast.rb
mchf: well this is almost everything what ```configure_hosts``` does

src/modules/DNS.rb
mchf: hm, consequence of some massive updating and code shifting and so on. I'll drop it.

src/clients/save_network.rb
```diff
@@ -259,16 +259,41 @@ def configure_dns
...
+      NetworkAutoYast.instance.configure_routing
+    end
+
+    def configure_nics
+      return if !Mode.autoinst
+
+      NetworkAutoYast.instance.configure_nics
+    end
+
     # It does an automatic configuration of installed system
```
mchf: upgrade scenario ignores network setup ... but i'm not sure about autoupgrade - will check

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Jun  9 09:40:11 UTC 2017 - mfilka@suse.com
+
+- Moving network setup in AY into 1s stage
```
mchf: currently not aware of any - the PBI contains no fate number 

@kobliha: is there any fate request?

src/lib/network/network_autoyast.rb
mvidner: Ok, now that the other configure_* methods are gone it makes less sense to split it off. Still, with all the "or empty" logic in here, tests are needed.

Also, how does this code relate to [Import in host_auto](https://github.com/mchf/yast-network/blob/f7eb35908554123042b8c570fa8f5f6517bea923/src/clients/host_auto.rb#L85)? Do we need to "guarantee item order" also there?

mvidner:
does host_auto#Import need similar code like configure_hosts?

src/lib/network/network_autoyast.rb
mchf: Item order is most probably bad description here. I refer to "host_address" and "names". Each of these hash keys refer to a xml node. And this nodes can be ordered differently in the profile. We must guarantee that "host_address" is at first place.

mchf:
yes, see https://github.com/yast/yast-network/blob/master/src/clients/host_auto.rb#L86

src/lib/network/network_autoyast.rb
mvidner: I think that this duplicates `Host.Import` . Is that one obsolete then?

mvidner: Ah, sorry, `configure_submodule` does call `Import`

src/lib/network/network_autoyast.rb
```diff
@@ -104,26 +106,40 @@ def set_network_service
...
+    # If the installer is running in 1st stage mode only, then the configuration
+    # is also written
+    #
+    # @param [Boolean] write forces instant writing of the configuration
+    # @return [Boolean] true when configuration was present and loaded from the profile
+    def configure_hosts(write: false)
+      log.info("NetworkAutoYast: Hosts configuration")
 
-      true
+      # expected format for Hosts.Import is { "ip" => [list, of, names] }
```
mvidner: OK, and this must be documented *there*, in the API docs for Host.Import (not Hosts).

mchf:
Writing in first stage is experimental feature from my POV and is used only when second stage is disabled. Othewise _auto clients are used in second stage. See AY integration tests for some more details.

src/lib/network/network_autoyast.rb
```diff
@@ -206,16 +222,53 @@ def merge_routing(instsys_routing, ay_routing)
...
+
+    # Returns host section of the current AY profile
+    #
+    # Note that autoyast transforms the host's subsection
+    # into:
+    # {
+    #   hosts => [
+    #     # first <host_entry>
+    #     {
+    #       "hosts_address" => <ip>,
```

src/lib/network/network_autoyast.rb
```diff
@@ -206,16 +222,53 @@ def merge_routing(instsys_routing, ay_routing)
...
 
       ay_profile = Profile.current
 
       return {} if ay_profile.nil? || ay_profile.empty?
-      return {} if ay_profile["networking"].nil?
+      ay_profile
+    end
+
+    # Returns networking section of current AY profile
+    def ay_networking_section
```
mchf: regarding ```Host.Import``` see bellow

src/lib/network/network_autoyast.rb

src/lib/network/network_autoyast.rb
mchf: in my POV "networking" section is used quite often so it deserved own method as a shortcut.

mchf:
Thanks for the review

PR #521, opened 2017-06-20
Fix detection of Mellanox ConnectX-3 boards with bad subclass (bsc#10…

nmorey:
…44982)

ConnectX-3 boards badly use the subclass 128 which causes them to be ignored by wicked.
This patch adds a check for the 3 specific boards known to have this issue.

Signed-off-by: Nicolas Morey-Chaisemartin <NMoreyChaisemartin@suse.com>

### DIFF
--- a/src/include/network/routines.rb
+++ b/src/include/network/routines.rb
@@ -363,6 +363,12 @@ module Yast
         when 145
           return "usb" # #22739
         when 128
+          # Mellanox ConnectX-3 series badly uses the 128 sub class
+          # Check for the specific known boards with bad subclass.
+          if Ops.get_integer(hwdevice, "vendor_id", 0) == 71_091
+            device_id = Ops.get_integer(hwdevice, "device_id", -1)
+            return "ib" if [69_635, 69_636, 69_639].include?(device_id)
+          end
           # Nothing was found
           Builtins.y2error("Unknown network controller type: %1", hwdevice)
           Builtins.y2error(


mchf:
master branch is for future development. If you want the fix in SLE-12-SP3, then merge against respective branch.

src/include/network/routines.rb
```diff
@@ -363,6 +363,12 @@ def ControllerType(hwdevice)
         when 145
           return "usb" # #22739
         when 128
+          # Mellanox ConnectX-3 series badly uses the 128 sub class
+          # Check for the specific known boards with bad subclass.
+          if Ops.get_integer(hwdevice, "vendor_id", 0) == 71_091
```
mchf: ```Ops``` is leftover from ycp -> ruby conversion times. So, feel free to use ```hwdevice["vendor_id"] || 0```

src/include/network/routines.rb
```diff
@@ -363,6 +363,12 @@ def ControllerType(hwdevice)
...
+            device_id = Ops.get_integer(hwdevice, "device_id", -1)
```
mchf: See above

src/include/network/routines.rb
```diff
@@ -363,6 +363,12 @@ def ControllerType(hwdevice)
...
+            return "ib" if [69_635, 69_636, 69_639].include?(device_id)
```
mchf: Could you pls help us and put here a comment with some examples of concrete cards?

src/include/network/routines.rb
nmorey: I'll fix this

src/include/network/routines.rb
nmorey: You want extra comments in the code with the card list ?
BTW why do the device_id have on extra bit compared to the lspci value ?

mchf: just a few most common examples or a link to a datasheet or something like that if possible. Reporters usually use business names, so every piece of invormation is a speed up for debugging the problem.

mchf: i don't know why there is an extra bit. This piece of code is dark magic approximately 15 years old. It is being investigated per bug report bases only ;-)

src/include/network/routines.rb
mchf: thanks

src/include/network/routines.rb
nmorey: Ok. I'll list the board and the expect lspci value when I resubmit

nmorey:
See PR #522 with updated patch and against SLE12-SP3.

PR #524, opened 2017-06-29
Report error when 2nd stage is disabled (bsc#1046198)

lslezak:
- Report a warning when the imported AY settings require the seconds stage but it is disabled in the profile
- Fix for https://bugzilla.suse.com/show_bug.cgi?id=1046198
- Tested with the profile attached at in the bug report

ay_dns_warning

### DIFF
--- a/src/modules/DNS.rb
+++ b/src/modules/DNS.rb
@@ -52,6 +52,9 @@ module Yast
       Yast.import "Service"
       Yast.import "String"
       Yast.import "FileUtils"
+      Yast.import "Stage"
+      Yast.import "Mode"
+      Yast.import "Report"
 
       Yast.include self, "network/routines.rb"
       Yast.include self, "network/runtime.rb"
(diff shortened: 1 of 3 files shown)


src/modules/DNS.rb
```diff
@@ -479,6 +482,21 @@ def Import(settings)
 
       @nameservers = Builtins.eval(Ops.get_list(settings, "nameservers", []))
       @searchlist = Builtins.eval(Ops.get_list(settings, "searchlist", []))
+
+      # check for AY unsupported scenarios, the name servers and the search domains
+      # are written in the 2nd stage, if is disabled then it cannot work (bsc#1046198)
```
mchf: So, we're fixing a bug which is not a bug. Somebody tried to use the module by an unsupported way and we'll take care of it. 

In my POV the issue should be closed as invalid. Instead of adding a check for one of several unsupported cases we should improve documentation for SP3 to make more obvious what is and is not supported.

lslezak: Yes, I agree that it's not a real bug. But on the other hand nobody reads the documentation, if we print an error we might report false bug reports...

The tricky part is that the profile validates, and unless you know this limitation (or read the documentation) there is no way to find out that the profile is not correct.

schubi2:
"The Import is called twice during installation, is it OK to display the error twice? Or should I add some logic so it is displayed only once?"
Hm, not sure. SP3 will be the base of CAASP and there is no second stage at all. So the "error" could be
shown very often there. For SLES12-SP3 it would be OK because this situation is really rare.

src/modules/DNS.rb
schubi2: I have the same opinion as Lada here. Just inform the user. But I would take a warning or info level.

PR #527, opened 2017-07-04
Do not update sysconfig when no IP is given in case of static bootproto

mchf:
bnc#1046735

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1482,33 +1482,32 @@ module Yast
           LanItems.device = NetworkInterfaces.device_num(ifcfgname)
         end
 
-        LanItems.bootproto = Ops.get_string(@settings, "BOOTPROTO", "")
-        if LanItems.bootproto == "static" # #104494
-          ip_changed = LanItems.ipaddr !=
-            Ops.get_string(@settings, "IPADDR", "")
-          if ip_changed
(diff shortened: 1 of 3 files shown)


src/include/network/lan/address.rb
```diff
@@ -1482,33 +1482,32 @@ def AddressDialog
...
-                @hostname_initial,
-                Ops.get_string(@settings, "HOSTNAME", ""),
-                Ops.get_string(@settings, "IPADDR", "")
-              )
-            end
+          if (@hostname_initial != hostname && !hostname.empty?) || ip_changed
+            Host.Update(@hostname_initial, hostname, LanItems.ipaddr)
           end
-        else
+        elsif bootproto != "static"
```
jreidinger: so what happen if not ip is set?

mchf: hopefully nothing. Intention is not to update config bcs ```BOOTPROTO=static``` makes no sense without IP defined

jreidinger: so why not use return instead like
`return nil if bootproto == "static" && ipaddr.empty?`
and maybe log it out?

jreidinger: or to be correct `return ret`

jreidinger: and all in all. Is it valid at all? Maybe it is bug in validation of ip address widget, not?

mchf:
Thanks for the review

PR #529, opened 2017-07-10
Added storage-ng to check if disk is in network

gilsonsouza:


package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Jul 10 07:06:17 UTC 2017 - gsouza@suse.com
+
+- Added support to storage-ng so it can verify if a disk is in
+network.
```
jreidinger: bad indentation here. it should be aligned with "Added"

src/lib/network/clients/save_network.rb
```diff
@@ -0,0 +1,287 @@
...
+      Yast.include self, "network/routines.rb"
+      Yast.include self, "network/complex.rb"
+
+      # for update system don't copy network from inst_sys (#325738)
+      if !Mode.update
+        save_network
+      else
+        Builtins.y2milestone("update - skip save_network")
+      end
+      # EOF
```
jreidinger: this is not EOF. I am sure :)

test/SCRStub.rb
```diff
@@ -17,8 +17,6 @@
 #     end
 #
 module SCRStub
-  DATA_PATH = File.join(File.expand_path(File.dirname(__FILE__)), "data")
```
jreidinger: I see a lot of methods in this file use this constant, where it is now defined?

package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        3.4.0
```
ancorgs: Strictly speaking, we don't need to increase version with every change to the storage-ng branch. We are not creating submit requests of this branch to any product, so there is no need to increase version . It can bring problems and confusion with no gain. Of course, the changelog entries are still needed, just leaving the number alone unless we need it in order to introduce a dependency in another package.

src/lib/network/clients/save_network.rb
```diff
@@ -0,0 +1,287 @@
...
+      # storage-ng
+      # Check if installation is targeted to a remote destination.
+      devicegraph = Y2Storage::StorageManager.instance.y2storage_staging
+      is_disk_in_network = devicegraph.filesystem_in_network?(
+        Installation.destdir
+      )
+
+      return unless is_disk_in_network
+
+      log.info("Directory \"#{Installation.destdir}\" is on a network based device")
```
ancorgs: The old code looked like this (with the log message before the return):
```ruby
log.info("Network based device: #{remote_access}")
return if remote_access == :no
```

Of course, the new log message is nicer. But the old code always wrote something to the logs (no matter if the device is in network of not). That's better because it allows us to know that the check was performed.

test/data/devicegraphs/mixed_disks.yml
```diff
@@ -0,0 +1,82 @@
+---
+- disk:
```
ancorgs: I believe this whole file is not needed. See below.

test/save_network_test.rb
```diff
@@ -0,0 +1,40 @@
...
+
+require "yast"
+require "network/clients/save_network"
+describe Yast::SaveNetworkClient do
+  describe "#adjust_for_network_disks" do
+    before do
+      allow(Yast::SCR).to receive(:Execute).and_return("exit" => 0, "stdout" => "", "stderr" => "")
+      fake_scenario(scenario)
+    end
+    let(:scenario) { "mixed_disks" }
```
ancorgs: You are mocking the staging devicegraph and the only relevant call to it below. Thus, you don't need a YAML file describing the scenario, you don't need `fake_scenario` and, of course, you don't need all the infrastructure you introduced around it. This should be enough:

```ruby
describe Yast::SaveNetworkClient do
  describe "#adjust_for_network_disks" do
    before do
      Y2Storage::StorageManager.create_test_instance

      allow(Yast::SCR).to receive(:Execute).and_return("exit" => 0, "stdout" => "", "stderr" => "") 
      staging = Y2Storage::StorageManager.instance.y2storage_staging
      allow(staging).to receive(:filesystem_in_network?).and_return(in_network)
    end 

    context "when installation directory is in a network device" do
      let(:in_network) { true }

      it "tunes ifcfg file for remote filesystem" do
        expect(Yast::SCR).to receive(:Execute).with(anything, /nfsroot/).once
        subject.main
      end 
    end 

    context "when installation directory is in a local device" do
      let(:in_network) { false }

      it "does not touch any configuration file" do
        expect(Yast::SCR).to_not receive(:Execute).with(anything, /nfsroot/)
        subject.main
      end 
    end 
  end 
end
```

The rest of your test will keep working as-is. And you can revert all the changes done to the other tests and kill the YAML file and `storage_helpers.rb`.

src/lib/network/clients/save_network.rb
gilsonsouza: I just moved the code without checking it... I'm going to remove this line!

package/yast2-network.spec
gilsonsouza: okay!

src/lib/network/clients/save_network.rb
gilsonsouza: The problem to keep the same message is that I don't know which device is used anymore (usb? iscsi?...) and it is not necessary for this code. Do you think we still need the same message or should we create a new message displaying a  true/false status? What do you suggest?

test/save_network_test.rb
gilsonsouza: Great! I'll do it!

src/lib/network/clients/save_network.rb
ancorgs: Whatever message. I just want to leave a trace in the log telling us that we did the check. Logs are the only tool we have to figure out the execution workflow in most bug reports. Even if it's as simple and repetitive as:
```ruby
log.info("Checking whether \"#{Installation.destdir}\" is on a network based device")
return unless is_disk_in_network
log.info("\"#{Installation.destdir}\" is in network")
```
(although sure you can come up with something better) :-)

src/lib/network/clients/save_network.rb
```diff
@@ -40,7 +39,10 @@ def adjust_for_network_disks(file)
         Installation.destdir
       )
 
-      return unless is_disk_in_network
+      unless is_disk_in_network
```
ancorgs: [According to SUSE's Ruby style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md#statements) we should not use `unless` in multiline format. I.e. this is considered acceptable
```ruby
do_whatever unless is_disk_in_network
```
But this is not :-(
```ruby
unless is_disk_in_network
  do_whatever
end
```
So I guess we should do this to be compliant:
```ruby
if !is_disk_in_network
  log.info("...")
  return
end
```

package/yast2-network.changes
```diff
@@ -2,9 +2,76 @@
 Mon Jul 10 07:06:17 UTC 2017 - gsouza@suse.com
 
 - Added support to storage-ng so it can verify if a disk is in
-network.
+  network.
 - Updated dependencies to include storage-ng.
-- 3.3.1
+
+-------------------------------------------------------------------
```
ancorgs: Wow, wow! Now I see a lot of duplication in the changelog. According to this diff you are re-adding a lot of entries that are already below.

PR #533, opened 2017-07-12
Avoid creating duplicit udev rules in AutoYaST installation

mchf:
[bsc#1038717](https://bugzilla.suse.com/show_bug.cgi?id=1038717)

not confirmed to fix reporter's issue, but tested manually by me and requested by Frederic for SP3

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jul 12 05:37:30 UTC 2017 - mfilka@suse.com
+
+- bnc#1038717
+  - avoid creating duplicit udev rules in AutoYaST installation
+- 3.1.178
+
+-------------------------------------------------------------------
 Thu May 18 10:11:01 UTC 2017 - gsouza@suse.com
 
 - bsc#1036440 and bsc#994471
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Wed Jul 12 05:37:30 UTC 2017 - mfilka@suse.com
 
 - bnc#1038717
-  - avoid creating duplicit udev rules in AutoYaST installation
+  - avoid creating duplicate udev rules in AutoYaST installation
 - 3.1.178
 
 -------------------------------------------------------------------
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -227,6 +227,24 @@ module Yast
       true
     end
 
+    def rename_lan_item(item, name_to, attr=nil, key=nil)
+      # selecting according device name is unreliable (selects only in between configured devices)
+      LanItems.current = item
+
+      if !attr.nil? && !key.nil?
+        # find out what attribude is currently used for setting device name and
+        # change it if needed. Currently mac is used by default. So, we check is it is
+        # the other one (busid). If no we defaults to mac.
+        bus_attr = LanItems.GetItemUdev("KERNELS")
+        current_attr = bus_attr.empty? ? "ATTR{address}" : "KERNELS"
+
+        # make sure that we base renaming on defined attribute with value given in AY profile
+        LanItems.ReplaceItemUdev(current_attr, attr, key)
+      end
+
+      LanItems.rename(name_to)
+    end
+
     # Takes a list of udev rules and assignes them to corresponding devices.
     #
     # If a device has an udev rule defined already, it is overwritten by new one.
@@ -242,27 +260,25 @@ module Yast
         next if !valid_rename_udev_rule?(rule)
         key.downcase!
 
+        # find item which matches to the given rule definition
         item, matching_item = LanItems.Items.find do |_, i|
           i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key
         end
         next if !matching_item
 
-        # for logging only
         name_from = matching_item["ifcfg"] || matching_item["dev_name"]
         log.info("Matching device found - renaming <#{name_from}> -> <#{name_to}>")
 
-        # selecting according device name is unreliable (selects only in between configured devices)
-        LanItems.current = item
+        # find rule in collision
+        colliding_item, item_map = LanItems.Items.find do |i, _|
+          LanItems.GetDeviceName(i) == name_to
+        end
 
-        # find out what attribude is currently used for setting device name and
-        # change it if needed. Currently mac is used by default. So, we check is it is
-        # the other one (busid). If no we defaults to mac.
-        bus_attr = LanItems.GetItemUdev("KERNELS")
-        current_attr = bus_attr.empty? ? "ATTR{address}" : "KERNELS"
+        # rename item in collision
+        rename_lan_item(colliding_item, name_from)
 
-        # make sure that we base renaming on defined attribute with value given in AY profile
-        LanItems.ReplaceItemUdev(current_attr, attr, key)
-        LanItems.rename(name_to)
+        # rename matching item
+        rename_lan_item(item, name_to, attr, key)
       end
     end
   end
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -250,9 +250,13 @@ module Yast
 
         # make sure that we base renaming on defined attribute with value given in AY profile
         LanItems.ReplaceItemUdev(current_attr, attr, key)
+      elsif attr.nil? ^ key.nil? # xor
+        raise ArgumentError, "Not enough data for udev rule definition"
       end
 
       LanItems.rename(name_to)
+
+      nil
     end
 
     # Takes a list of udev rules and assignes them to corresponding devices.
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -227,7 +227,7 @@ module Yast
       true
     end
 
-    def rename_lan_item(item, name_to, attr=nil, key=nil)
+    def rename_lan_item(item, name_to, attr = nil, key = nil)
       return if item.nil? || item.empty? || item < 0 || item >= LanItems.Items.size
       return if name_to.nil? || name_to.empty?
 
@@ -273,7 +273,7 @@ module Yast
         log.info("Matching device found - renaming <#{name_from}> -> <#{name_to}>")
 
         # find rule in collision
-        colliding_item, item_map = LanItems.Items.find do |i, _|
+        colliding_item, _item_map = LanItems.Items.find do |i, _|
           LanItems.GetDeviceName(i) == name_to
         end
 
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -228,6 +228,9 @@ module Yast
     end
 
     def rename_lan_item(item, name_to, attr=nil, key=nil)
+      return if item.nil? || item.empty? || item < 0 || item >= LanItems.Items.size
+      return if name_to.nil? || name_to.empty?
+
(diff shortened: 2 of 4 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -227,6 +227,27 @@ def valid_rename_udev_rule?(rule)
       true
     end
 
+    def rename_lan_item(item, name_to, attr = nil, key = nil)
+      return if item.nil? || item.empty? || item < 0 || item >= LanItems.Items.size
```
jreidinger: please document type of item in documentation. as empty and comparison with 0 looks strange as I have no clue how it can work unless it is own special object:

```
jreidinger@linux-vvcf:~> irb
irb(main):001:0> 0.empty?
NoMethodError: undefined method `empty?' for 0:Fixnum
	from (irb):1
	from /usr/bin/irb:11:in `<main>'
irb(main):002:0> "1" < 0
ArgumentError: comparison of String with 0 failed
	from (irb):2:in `<'
	from (irb):2
	from /usr/bin/irb:11:in `<main>'
irb(main):003:0> [1] < 1
NoMethodError: undefined method `<' for [1]:Array
	from (irb):3
	from /usr/bin/irb:11:in `<main>'
```

src/lib/network/network_autoyast.rb
```diff
@@ -227,6 +227,27 @@ def valid_rename_udev_rule?(rule)
...
+      return if name_to.nil? || name_to.empty?
```
jreidinger: name_to looks like string, so please document it. Ideally all arguments what type it expects.

src/lib/network/network_autoyast.rb
jreidinger: and that brings me to another big topic. I am missing test case for this method. I think it is doable.

src/lib/network/network_autoyast.rb
```diff
@@ -242,27 +263,25 @@ def assign_udevs_to_devs(udev_rules)
...
-        # for logging only
         name_from = matching_item["ifcfg"] || matching_item["dev_name"]
         log.info("Matching device found - renaming <#{name_from}> -> <#{name_to}>")
 
-        # selecting according device name is unreliable (selects only in between configured devices)
-        LanItems.current = item
+        # find rule in collision
+        colliding_item, _item_map = LanItems.Items.find do |i, _|
+          LanItems.GetDeviceName(i) == name_to
+        end
```
jreidinger: so if I get code it always find collision? does here missing check for nil? I would like to see test case also for this method.

src/lib/network/network_autoyast.rb
```diff
@@ -227,6 +227,27 @@ def valid_rename_udev_rule?(rule)
...
     end
 
+    def rename_lan_item(item, name_to, attr = nil, key = nil)
+      return if item.nil? || item.empty? || item < 0 || item >= LanItems.Items.size
+      return if name_to.nil? || name_to.empty?
+
+      # selecting according device name is unreliable (selects only in between configured devices)
+      LanItems.current = item
+
+      if !attr.nil? && !key.nil?
```
jreidinger: if only one of this argument is set, I expect argument error here, as it is obviously wrong usage.

src/lib/network/network_autoyast.rb
mchf: ok, my mistake (copy+paste )

src/lib/network/network_autoyast.rb

src/lib/network/network_autoyast.rb
mchf: ```nil``` is eaten (ignored) later in ```remove_lan_item```

jreidinger: so please document it as I found it quite confusing for reader of code.

jreidinger:
updates now make code at least visually workable, but as it goes to GMC I really expect unit tests for new method and all scenarios covered.

package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jul 12 05:37:30 UTC 2017 - mfilka@suse.com
+
+- bnc#1038717
+  - avoid creating duplicit udev rules in AutoYaST installation
```
mvidner: "duplicit" is not a word. Please consider https://twitter.com/mvidner/status/884388828701700096
- http://www.thefreedictionary.com/duplicity
- http://www.thefreedictionary.com/duplicate

src/lib/network/network_autoyast.rb
```diff
@@ -227,6 +227,36 @@ def valid_rename_udev_rule?(rule)
...
+    # @param [String] value for the given udev attribute. Optional parameter.
+    def rename_lan_item(item, name_to, attr = nil, key = nil)
+      return if item.nil? || item < 0 || item >= LanItems.Items.size
+      return if name_to.nil? || name_to.empty?
+
+      # selecting according device name is unreliable (selects only in between configured devices)
+      LanItems.current = item
+
+      if !attr.nil? && !key.nil?
+        raise ArgumentError, "Missing udev attr / value" if attr.nil? || key.nil?
```
mvidner: This condition is impossible given the condition one line above it. I demand unit tests.

src/lib/network/network_autoyast.rb
mchf: done

package/yast2-network.changes
mchf: fixed

src/lib/network/network_autoyast.rb
mchf: true, working on tests

mchf:
@mvidner, @jreidinger I've added a test

test/network_autoyast_test.rb
```diff
@@ -348,4 +348,53 @@ def keep_install_network_value(value)
...
+        .and_return({ 0 => { "ifcfg" => "eth0" }})
+    end
+
+    context "valid arguments given" do
+      it "renames the item with no udev attribute change" do
+        expect(Yast::LanItems)
+          .to receive(:rename)
+          .with("new_name")
+        expect(Yast::LanItems)
+          .not_to receive(:ReplaceItemUdev)
```
jreidinger: what about expectation that proper old device is selected?

test/network_autoyast_test.rb
```diff
@@ -348,4 +348,53 @@ def keep_install_network_value(value)
...
+        network_autoyast.send(:rename_lan_item, 0, "")
+      end
+
+      it "do not try to rename an item when given item id is invalid" do
+        expect(Yast::LanItems)
+          .not_to receive(:rename)
+
+        network_autoyast.send(:rename_lan_item, nil, "new_name")
+        network_autoyast.send(:rename_lan_item, -1, "new_name")
+        network_autoyast.send(:rename_lan_item, 100, "new_name")
```
jreidinger: to be honest I expect here that exception is raised. Silently ignore failures ( and especially programming failures ) is not my cup of tea.

src/lib/network/network_autoyast.rb
```diff
@@ -250,9 +250,13 @@ def rename_lan_item(item, name_to, attr = nil, key = nil)
 
         # make sure that we base renaming on defined attribute with value given in AY profile
         LanItems.ReplaceItemUdev(current_attr, attr, key)
+      elsif attr.nil? ^ key.nil? # xor
```
jreidinger: JFYI `^` is bitwise operator, so if you do not have boolean only ( not this case ) it can produce funny results like:
```
irb(main):002:0> true ^ 1
=> false
```

test/network_autoyast_test.rb
mchf: That is a test for another method in my POV

test/network_autoyast_test.rb
```diff
@@ -395,6 +395,15 @@ def keep_install_network_value(value)
...
       end
+
+      it "raise an exception when udev definition is incomplete" do
+        expect {
+          network_autoyast.send(:rename_lan_item, 0, "new_name", "KERNELS", nil)
+        }.to raise_error(ArgumentError)
+        expect {
+          network_autoyast.send(:rename_lan_item, 0, "new_name", nil, "0000:00:03.0")
+        }.to raise_error(ArgumentError)
+      end
```
jreidinger: :+1:

test/network_autoyast_test.rb
mchf: I'm not that sure here. At least first two cases can mean (and means in case of the current code) that the value mean "not found" case. Ignoring such cases can lead to more readable code:

```
rename_lan_item(get_item_for("eth0"), "new_name")
```

vs

```
item = get_item_for("eth0")
if !item.nil? && item > 0 && item <= get_max_item do
  rename_lan_item(item, "new_name")
end
```

So, I prefer exceptions for cases when the code is not able to reliable (trivially) recover.

mvidner:
I could criticize several aspects of `rename_lan_item` but that is not the most important point. It is: What is the actual bug, and how does this PR fix it?
This piece, calling the code being modified, could interact with it:
```rb
      # implicitly defined udev rules are overwritten by explicit ones in
      # case of collision.
      assign_udevs_to_devs(im_udev_rules)
      assign_udevs_to_devs(ex_udev_rules)
```

mchf:
bug is that when you wish to rename a device to an existing name, then the code created duplicate rule. Say that someone wishes to have a particular device named eth0. Before the fix, when the system defined the device e.g. eth1 then after renaming there were 2 rules for eth0. The fix is that names in collisions are found and names are swapped.

So say that user wanted "2nd device to be named eth0"
old code:
```
before rename   after rename
eth0          ->   eth0
eth1          ->   eth0
```

new code:
```
eth0          ->    eth1
eth1          ->    eth0
```

src/lib/network/network_autoyast.rb
```diff
@@ -242,27 +274,25 @@ def assign_udevs_to_devs(udev_rules)
...
+          LanItems.GetDeviceName(i) == name_to
+        end
 
-        # find out what attribude is currently used for setting device name and
-        # change it if needed. Currently mac is used by default. So, we check is it is
-        # the other one (busid). If no we defaults to mac.
-        bus_attr = LanItems.GetItemUdev("KERNELS")
-        current_attr = bus_attr.empty? ? "ATTR{address}" : "KERNELS"
+        # rename item in collision
+        rename_lan_item(colliding_item, name_from)
```
mvidner: But omitting the attr and key arguments means that ReplaceItemUdev is not called, so we still end up with duplicate names.
And THAT is the thing we should have a test for.
The current tests just test the complicated API which only exists because `rename_lan_item` seems to serve two different purposes.

mvidner:
@mchf Please show me the logs of your manual test. Then we may be able to write a meaningful automated test for this.

mchf:
@mvidner automated test for what?  @schubi2 helped me with AY integration test for that - see the PBI https://trello.com/c/ge080F5O/1037-3-sle-12-sp2-1038717-l3-question-autoyast-net-udev-rules-priority-and-requirements

mvidner:
An automated test that this fix works at all: have N nics, M<N rules for them, check that the rules have no duplicates. Just like the test case in the bug.

Thanks for the pointer to the trello card, but even there I see no link to a test.

Anyway, it is unlikely this will go into SP3 at this point, so you should enjoy the vacation :)

mchf:
@mvidner that is task for an integration test ... so I'll either check with @schubi2 if he has something, or I'll create some based on the profile I've used for testing.

I'm doing some stuff around house, so coding is rest ;-)

mchf:
An integration test https://github.com/yast/aytests-tests/pull/92

mchf:
Thanks for the reviews

PR #540, opened 2017-07-27
Load /etc/hosts entries before add new ones (bsc#1039851)

teclator:
Just make it backward compatible.

- trello card: https://trello.com/c/OWAkiSkH

### DIFF
--- a/test/host_test.rb
+++ b/test/host_test.rb
@@ -114,10 +114,52 @@ describe Yast::Host do
     end
   end
 
+  describe ".Import" do
+    let(:file) do
+      file_path = File.expand_path("../data/default_hosts", __FILE__)
+      CFA::MemoryFile.new(File.read(file_path))
+    end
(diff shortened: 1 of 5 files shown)


test/host_test.rb
```diff
@@ -114,10 +114,52 @@
...
+
+    it "merges current entries with the given ones" do
+      Yast::Host.Import("hosts" => { "10.20.1.29" => ["beholder"] })
+
+      expect(Yast::Host.name_map).to eql(etc_hosts.merge("10.20.1.29" => ["beholder"]))
+    end
+  end
+
   describe ".Export" do
+    let(:file) do
```
imobachgs: np: I guess you could move `file` to the outside block.

teclator: well, it is out but with other path, so I could just define the file_path instead of the whole file

PR #543, opened 2017-08-01
Use proper API for querying device name during AY installation

mchf:
bnc#1037727

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -280,7 +280,7 @@ module Yast
         end
         next if !matching_item
 
-        name_from = matching_item["ifcfg"] || matching_item["dev_name"]
+        name_from = LanItems.GetDeviceName(item)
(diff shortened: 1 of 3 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -280,7 +280,7 @@ def assign_udevs_to_devs(udev_rules)
         end
         next if !matching_item
 
-        name_from = matching_item["ifcfg"] || matching_item["dev_name"]
+        name_from = LanItems.GetDeviceName(item)
```
teclator: So it basically does not fix anything at all but just to clean up it?

teclator: OK, so it basically fix the matching_item["hwinfo"]["dev_name"] case

mchf: fixes an issue when ifcfg is not present but udev rule is defined by user in the profile. The rule cannot be paired properly in such case.

PR #545, opened 2017-08-02
Fix default value for sysconfig network configuration when read. (bsc#1051624)

teclator:


src/modules/DNS.rb
```diff
@@ -282,8 +282,8 @@ def ProposeHostname
     def Read
       return true if @initialized == true
 
-      @dhcp_hostname = dhclient_set_hostname
-      @write_hostname = get_write_hostname_to_hosts
+      @dhcp_hostname = dhclient_set_hostname || false
```
mchf: np: i'm thinking of adding a note that "no" is default value in sysconfig in both cases.

src/modules/DNS.rb
```diff
@@ -282,8 +282,8 @@ def ProposeHostname
     def Read
       return true if @initialized == true
```
mchf: you can even drop ```== true`` when tuning return value already ;-)

teclator: even :)

src/modules/DNS.rb
teclator: Added a note, with bug a reference to the bug ;)

teclator:
Thnx 👍

package/yast2-network.changes
```diff
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Aug  2 13:01:48 UTC 2017 - knut.anderssen@suse.com
+
+- bsc#1051624
+  - Fix default values for not existent sysconfig network variables 
+    when read. (bsc#1051624)
```
mchf: nice to have bug reference twice ;-)

teclator: xDDD

teclator: Removed, added by error again after modified the changelog 👎

PR #552, opened 2017-08-14
Crash on nonexistent /etc/hosts file

mchf:
bnc#1047615

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -109,20 +109,18 @@ module Yast
     # @return true if success
     def Read
       return true if @initialized
+      return true if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-      # read /etc/hosts
-      if Ops.greater_than(SCR.Read(path(".target.size"), CFA::Hosts::PATH), 0)
-        @hosts = CFA::Hosts.new
-        @hosts.load
-      end
+      @hosts = CFA::Hosts.new
+      @hosts.load
 
       # save hosts to check for changes later
       @hosts_init = CFA::Hosts.new
       @hosts_init.load
 
       Builtins.y2debug("hosts=#{@hosts.inspect}")
+
       @initialized = true
-      true
     end
 
     # Write hosts settings and apply changes
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -41,6 +41,7 @@ module Yast
       Yast.include self, "network/routines.rb"
 
       # Data was modified?
+      # TODO: Drop the flag. It is useless since we have @hosts and @hosts_init
       @modified = false
 
       @initialized = false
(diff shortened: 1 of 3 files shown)


src/modules/Host.rb
```diff
@@ -111,14 +111,14 @@ def Read
       return true if @initialized
 
       # read /etc/hosts
-      if Ops.greater_than(SCR.Read(path(".target.size"), CFA::Hosts::PATH), 0)
+      if File.exist?(CFA::Hosts::PATH)
```
jreidinger: This won't work in installation with switched SCR.

mchf: true ... but AFAIK not needed. Configured /etc/hosts is copied to the target at the end of installation.

src/modules/Host.rb
```diff
@@ -109,20 +109,18 @@ def EnsureHostnameResolvable
...
+      @hosts = CFA::Hosts.new
+      @hosts.load
 
       # save hosts to check for changes later
       @hosts_init = CFA::Hosts.new
       @hosts_init.load
 
       Builtins.y2debug("hosts=#{@hosts.inspect}")
+
       @initialized = true
```
jreidinger: pull request looks good, just it looks like comment for method is not longer valid, as it cannot return false. So I suggest to document that it always return true and that it can raise exception if file is malformed.

mvidner:
@jreidinger First we should formulate a consistent policy of CFA with respect to non-existent files. Should all users take care of it, like here? Or should CFA abstract that away?

jreidinger:
@mvidner plan is to add method to CFA to ask for existence of file in future

mchf:
@mvidner I thought that in some cases non existence of a file can be real error, that's why I've handled it here. In the future I'd prefer to have two methods in CFA - exception safe and unsafe ...

mvidner:
I still think this is a layering violation. We have defined a library and a concrete class for accessing and parsing that config file, yet we must manually handle the case where the file is missing?

Host.Import may still fail. Does that illustrate my point?

[BaseModel#initialize](http://www.rubydoc.info/github/config-files-api/config_files_api/CFA/BaseModel#initialize-instance_method) already presets `@data` to parser.empty. In which situation do we need to handle a missing config file differently?

jreidinger:
@mvidner problem is `BaseModel#load` which call `FileHandler#read` which result in exception. So load is problematic call.

jreidinger:
@mvidner do you really expect that load silently ignore that file cannot be loaded?

mvidner:
I claim that the model should by default treat a missing file as an empty model. That's what eg. [yast-bootloader does](https://github.com/yast/yast-bootloader/search?utf8=%E2%9C%93&q=ENOENT&type=)
Or can you point me to a place in our code where we handle such a case differently?

jreidinger:
@mvidner please notice that in bootloader we hace treat 2 of 5 files as optional, so if it is missing we continue with empty. For remaining three exception is raised and catched by different exception handler that report it to user and in such case offer to propose configuration from scratch as mandatory file is missing. This is something that CFA cannot know.

jreidinger:
@mvidner e.g. this read is not catched by this specific handler https://github.com/yast/yast-bootloader/blob/99236316e7ad0e97f077dcf2ef5352a47bdf7265/src/lib/bootloader/grub2base.rb#L80

mchf:
Thanks for the reviews

PR #553, opened 2017-08-14
Merge branch 'storage-ng' into 'master' and release version 3.3.7

wfeldt:


src/modules/LanItems.rb
```diff
@@ -2249,11 +2253,14 @@ def ProposeItem
...
       @netmask = ""
       @bootproto = "dhcp"
+# storage-ng
+=begin
       # #176804
       if Storage.IsDeviceOnNetwork(NetworkStorage.getDevice("/")) != :no
         @startmode = "nfsroot"
         Builtins.y2milestone("startmode nfsroot")
       end
+=end
```
ancorgs: Hmmm, this is suspicious. In theory this functionality was already taken back by @gilsonsouza, so I wouldn't expect commented code anymore. Let me check

ancorgs: Ok, so it looks like we forgot about this file when bringing the functionality back. So this is ok from the merge POV... but we need to fix it ASAP because it looks like https://trello.com/c/qlK9iJpH/1031-3-new-storage-network-based-root-check-readded is not as done as we though.

ancorgs:
Looks like other fixes were introduced in master meanwhile... which leave us with a situation similar to yast2-packager. If we rebase and fix today, those fixes will not reach TW until the whole storage-ng merge does it (months). @mchf How bad would that be?

ancorgs:
According to @kobliha , attracting https://build.opensuse.org/request/show/516888 into the storage-ng vortex is not that bad. A missing `/etc/hosts` is really unusual, so delaying the fix in Tumbleweed a couple of months is acceptable. So please rebase so we can move on.

PR #555, opened 2017-08-17
use storage-ng code to determine if "/" is on a network filesystem

wfeldt:
See also https://github.com/yast/yast-network/pull/529.

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -23,6 +23,7 @@
 # **************************************************************************
 require "yast"
 require "yaml"
+require "y2storage"
 require "network/install_inf_convertor"
 require "network/wicked"
 require "network/lan_items_summary"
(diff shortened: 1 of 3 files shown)


mchf:
regarding new storage I've to believe you, however that one line looks good to me anyway. However since it is re-enabling a feature it deserves a version update in my POV.

src/modules/LanItems.rb
```diff
@@ -2253,14 +2249,11 @@ def ProposeItem
       @ipaddr = ""
       @netmask = ""
       @bootproto = "dhcp"
-# storage-ng
-=begin
-      # #176804
-      if Storage.IsDeviceOnNetwork(NetworkStorage.getDevice("/")) != :no
+      devicegraph = Y2Storage::StorageManager.instance.staging
```
ancorgs: Please, don't delete the original bug number `#176804`

wfeldt:
This is for factory. They can't expect a bug number for everything we do there.

ancorgs:
As @mchf said, I worry about the bnc/fate number. But other than that (we'll see if that's a problem), LGTM.

PR #556, opened 2017-08-17
restored changes lost when merged storage-ng patch

mchf:
Some updates got lost when https://github.com/yast/yast-network/pull/553 was merged.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 17 19:42:51 UTC 2017 - mfilka@suse.com
+
+- restored part of version 3.3.3 (bnc#1049814) which was 
+  accidentaly dropped in subsequent versions
+- 3.3.10
+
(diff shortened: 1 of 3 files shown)


mchf:
@ancorgs Could you pls check if everything needed from storage-ng is here?

package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Aug 17 19:42:51 UTC 2017 - mfilka@suse.com
+
+- reverted part of version 3.3.3 which was accidentaly dropped in
```
mvidner: s/reverted/restored/

mvidner:
I'm just wondering how we found this. Did a test break?
The changed code looks covered by tests that pass before *and* after this fix??

mchf:
@mvidner I found it when I was working on another issue and missed some new calls.

mchf:
thanks for the reviews.

PR #557, opened 2017-08-21
Do not clear /etc/hosts when writing virtual proposal

mchf:
bnc#1039656


teclator:
LGTM. 

NP: Would be nice to have a test

mchf:
Thanks for the review

PR #558, opened 2017-08-21
Do not display an AY warning (bsc#1054400)

lslezak:
...when the hostname setting is read out of profile.

- Fix for https://bugzilla.suse.com/show_bug.cgi?id=1054400
- The hostname can be also read from `install.inf` (passed via boot command line), in that case the warning is wrong as the hostname is not specified in the profile.
- 3.2.38

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Aug 21 07:02:32 UTC 2017 - lslezak@suse.cz
+
+- AutoYaST: Do not display a warning about disabled second stage
+  when the hostname setting is read out of profile (bsc#1054400)
+- 3.2.39
+
+-------------------------------------------------------------------
 Mon Aug 21 06:57:26 UTC 2017 - mfilka@suse.com
 
 - bnc#1039656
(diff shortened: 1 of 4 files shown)


src/modules/DNS.rb
```diff
@@ -68,6 +68,9 @@ def main
       # Domain Name (not including the host part)
       @domain = ""
 
+      # the hostname and the domain has been imported from an AY profile
+      @hostname_from_profile = nil
```
mchf: you don't need a class variable for that - used in ```Import``` only

mchf: not a class variable but instance one of course

package/yast2-network.changes
```diff
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Aug 21 07:02:32 UTC 2017 - lslezak@suse.cz
+
+- AutoYaST: Do not display a warning about disabled second stage
+  when the hostname setting is read out of profile (bsc#1054400)
```
mchf: ... and second stage is enabled

lslezak: No, the second stage is actually disabled, but the setting is read out of profile so it is not an user fault.

mchf: ah ok.

mchf:
lgtm

PR #563, opened 2017-09-21
Ensure Host entries are read before write (bsc#1058396)

teclator:
- [Trello Card](https://trello.com/c/I0zqz1UB/1179-3-sles12-sp3-1058396-l3-etc-hosts-blanked-out-after-autoyast-installation)
- [bugzilla](https://bugzilla.suse.com/show_bug.cgi?id=1058396)

## Description of the problem (already in Trello)

- When there is no `<host>` section defined then the host_auto client is not called and the Host configuration is not read or imported, then if there is a `<networking>` section the lan_auto client write down the **Host** configuration without be initialized or read previously, which means that it blank out the /etc/hosts file.

- There was a similar bug in a common installation when virtual proposal was used https://github.com/yast/yast-network/pull/557, in this case, as only **LanItems** configuration was needed the call to write the **Lan** was simply replaced.

## Solution

- As it is AutoYaST specific, we ensure that /etc/hosts is read when importing.

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -558,6 +558,9 @@ module Yast
       # reload/restart network before this to put correct resolv.conf from dhcp-backup
       orig = Progress.set(false)
       DNS.Write
+      # Ensure that the /etc/hosts has been read to no blank out it (bsc#1058396)
+      # The method just returns true in case of initialized.
+      Host.Read
       Host.EnsureHostnameResolvable
       Host.Write
       Progress.set(orig)
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -558,9 +558,6 @@ module Yast
       # reload/restart network before this to put correct resolv.conf from dhcp-backup
       orig = Progress.set(false)
       DNS.Write
-      # Ensure that the /etc/hosts has been read to no blank out it (bsc#1058396)
-      # The method just returns true in case of initialized.
-      Host.Read
       Host.EnsureHostnameResolvable
       Host.Write
       Progress.set(orig)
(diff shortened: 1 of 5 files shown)


imobachgs:
In general it looks good. About when to call `Host.Read`, I would prefer doing it during `Import` but I do not have an strong opinion. @schubi2 what is your POV?

schubi2:
I would also prefer the Import function. This is a AY bug only and the Write function will be used in other cases too whereas the Import function will be used by AY only.

teclator:
@schubi2 exactly because it's an AY bug I was thinking in just adding into the Import method, but I saw the other bug related to virual proposal, was not sure and the Write was more defensive approach but I also don't like it at all.

mchf:
@teclator the "virtual" case happened in common installation.

teclator:
@mchf ok, just please summarize what you tested and your research to maybe have it documented for future improments although I suppose that most of them were done for SLE15. 

BTW, as commented it is mainly for AY with the particular problem of no <host> section. In a normal installation whatever workflow we follow will need a Host.Read before anything.

mchf:
@teclator: Could you just add 693ef04cf94c3a24c720f3b5f5147a269a87765f from my repo to this PR?

Short summary:
Basic problem is that lan and host auto clients are separated and some issues arise from this. I've tried various hacks to bypass that but nothing worked reliably in all cases. The above commit is only think which left from these experiments. It adds "ordering" into invocation of these clients.

teclator:
@mchf great, added the requirement to ensure the ordering calls.

src/modules/Lan.rb
```diff
@@ -659,6 +659,10 @@ def Import(settings)
       DNS.Import(settings["dns"] || {})
       Routing.Import(settings["routing"] || {})
 
+      # Ensure that the /etc/hosts has been read to no blank out it in case of
+      # not defined <host> section (bsc#1058396)
+      Host.Read
```
mchf: hm. Now when ordering is guaranteed, you can change it to

```
Host.Read if !Host.initialized
```

mchf: ah bad idea ```Host`` already checks that, so no need to do it here.

mchf:
lgtm

src/modules/Lan.rb
teclator: Exactly, if it is already initialized then it just returns.

teclator:
@mchf , @imobachgs, @schubi2 thnx for review it

PR #565, opened 2017-09-25
Update Remote module to handle systemd based VNC.

michalsrb:
The VNC service switched from xinetd to systemd.
VNC over HTTP service can be enabled/disabled separately.

### DIFF
--- a/src/modules/Remote.rb
+++ b/src/modules/Remote.rb
@@ -35,11 +35,13 @@ module Yast
     include Yast::Logger
 
     XDM_SERVICE_NAME = "display-manager".freeze
-    XINETD_SERVICE = "xinetd".freeze
+    XVNC_SERVICE = "xvnc.socket".freeze
+    XVNC_NOVNC_SERVICE = "xvnc-novnc.socket".freeze
     VNCMANAGER_SERVICE = "vncmanager".freeze
 
     PKG_CONTAINING_FW_SERVICES = "xorg-x11-Xvnc".freeze
     PKG_CONTAINING_VNCMANAGER = "vncmanager".freeze
+    PKG_CONTAINING_XVNC_NOVNC = "xorg-x11-Xvnc-novnc".freeze
 
     GRAPHICAL_TARGET = "graphical".freeze
 
@@ -61,8 +63,11 @@ module Yast
       # Currently, all attributes (enablement of remote access)
       # are applied on vnc1 even vnchttpd1 configuration
 
-      # Remote administration mode, :disabled, :xinetd or :vncmanager
+      # Remote administration mode, :disabled, :xvnc or :vncmanager
       @mode = :disabled
+
+      # Whether a web based vnc viewer should be served
+      @web_vnc_enabled = false
 
       # Remote administration has been already proposed
       # Only force-reset can change it
@@ -81,7 +86,7 @@ module Yast
 
     # Enables remote administration without vnc manager.
     def Enable
-      @mode = :xinetd
+      @mode = :xvnc
 
       nil
     end
@@ -102,6 +107,18 @@ module Yast
 
     def EnabledVncManager
       @mode == :vncmanager
+    end
+
+    def IsWebVncEnabled
+      @web_vnc_enabled
+    end
+
+    def EnableWebVnc
+      @web_vnc_enabled = true
+    end
+
+    def DisableWebVnc
+      @web_vnc_enabled = false
     end
 
     # Reset all module data.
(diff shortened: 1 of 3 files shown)


src/modules/Remote.rb
```diff
@@ -104,6 +109,18 @@ def EnabledVncManager
       @mode == :vncmanager
     end
 
+    def IsWebVncEnabled
```
mchf: well, I'm not such big expert in ruby but this ```web_vnc?``` is more according to the ruby style guide.

src/modules/Remote.rb
```diff
@@ -131,54 +148,29 @@ def Propose
     # Read the current status
     # @return true on success
     def Read
-      # are the proper services enabled in xinetd?
-      xinetd_conf = SCR.Read(path(".etc.xinetd_conf.services")) || {}
-      xinetd_vnc1_enabled = xinetd_conf.any? { |c| c["service"] == "vnc1" && c["enabled"] }
-
-      log.info "VNC1 enabled: #{xinetd_vnc1_enabled}"
+      display_manager_remote_access = SCR.Read(path(".sysconfig.displaymanager.DISPLAYMANAGER_REMOTE_ACCESS")).to_s == "yes"
```
mchf: you don't need ```to_s``` here

src/modules/Remote.rb
```diff
@@ -131,54 +148,29 @@ def Propose
...
 
+      xdm        = Service.Enabled(XDM_SERVICE_NAME)
       vncmanager = Service.Enabled(VNCMANAGER_SERVICE)
-      vnc_service_running = xinetd_vnc1_enabled || vncmanager
+      xvnc       = Service.Enabled(XVNC_SERVICE)
+      xvnc_novnc = Service.Enabled(XVNC_NOVNC_SERVICE)
```
mchf: is this variable ever used?

mchf: ah yes it is ...

src/modules/Remote.rb
michalsrb: Probably yes. I was just trying to replicate the style of the rest of the file. Function as `IsEnabled`, `IsDisabled`, `EnabledVncManager`.

Should I keep it as it is, or use `web_vnc?` for this new function, or maybe rename them all?

src/modules/Remote.rb
michalsrb: Ok, I'll remove that. It was like that in `xinet_ready?`.

src/modules/Remote.rb
michalsrb: Few lines below: `EnableWebVnc() if xvnc_novnc`

src/modules/Remote.rb
```diff
@@ -131,54 +148,29 @@ def Propose
...
+          EnableVncManager()
+        end
 
-      if xinet_ready? && vnc_service_running
-        Enable() if xinetd_vnc1_enabled
-        EnableVncManager() if vncmanager
+        EnableWebVnc() if xvnc_novnc
       else
         Disable()
+        DisableWebVnc()
```
mchf: how should I know from the above condition that just web vnc should be disabled?

src/modules/Remote.rb
mchf: btw ```NOVNC``` in name is really funny when setting vnc access, so what is this service responsible for?

src/modules/Remote.rb
michalsrb: I am not sure if I understand your question.
Web VNC depends on VNC. If the conditions for VNC are not fulfilled, Web VNC is not enabled either.

src/modules/Remote.rb
michalsrb: I did not name it. :-)
https://github.com/novnc/noVNC

It is VNC viewer written in javascript and HTML5. The service serves the web page with the viewer, so the remote desktop can be optionally accessed using browser.

src/modules/Remote.rb
michalsrb: The `to_s` is removed now.

michalsrb:
By the way this changes are part of fate#323880 and bnc#1058460.

michalsrb:
Is there something missing for this to be merged? The VNC configuration in YaST is going to be broken in SLE15 without this...

src/modules/Remote.rb
lslezak: Although the `web_vnc?` better fits the usual Ruby style I'd keep it as it is to have consistent names.

teclator:
@lslezak @mchf @michalsrb 

I think that we could merge it by now, and could just clean current code in a new PR, I already have one in progress but have to finish unit tests.

It will use y2remote namespace under src/lib, new widgets and dialogs and the only thing I will probably let almost as it is is the current client because the CommandLine.

Just a note:

@michalsrb when I was rewriting current code I realized or at least seems so, that if you switch from vnc to vncmanager (without and with session) it does not stop any service (shouldn't it stop the one that will not be used?). And also probably the checkbox for enable Wed should be disabled if disabled is selected.

That are the only things I have seen about the current code, about method namings I have the same opinion than lslezak, for current code just follow the same convention that is currently used.

michalsrb:
> @michalsrb when I was rewriting current code I realized or at least seems so that if you switch from vnc to vncmanager (without and with session) it does not stop any service (shouldn't it stop the one that will not be used?) and also probably the checkbox for enable Web should be disabled if disabled is selected.

Switching from vnc to vncmanager:
Before this change: no service needed to be stopped, only xinetd configuration had to be changed (disable the vnc services, keep the http-vnc services).
After this change: xvnc.socket should be stopped, vncmanager.service should be started. This should be happening.

PR #569, opened 2017-10-03
Improved #564 test coverage

teclator:
- [Trello Card](https://trello.com/c/I0zqz1UB/1179-3-sles12-sp3-1058396-l3-etc-hosts-blanked-out-after-autoyast-installation)
- Improves #563 test coverage.

PR #570, opened 2017-10-03
Merge SLE-12-SP3

teclator:
- It merges #569


teclator:
Yep, that is why I didn't merged it, I was debugging it and there are some symptoms about wrong mocking.

https://github.com/yast/yast-network/blob/master/test/save_network_test.rb#L23
https://github.com/yast/yast-network/blob/master/src/lib/network/clients/save_network.rb#L26

We are calling save_network (this call should be allowed in rspec).

Previously the imported config was flushed so does not affect at all as the second test do.

https://github.com/yast/yast-network/pull/570/files#diff-cc5a7158462444de5d7be71f2be3bc77R166

PR #575, opened 2017-10-04
Added mocks to avoid call save_network

teclator:
- I want to fix mocks for this test before merge #574.


teclator:
@mchf thnx ;)

PR #576, opened 2017-10-06
[WIP] SuSEFirewall4Network.IsInstalled returns true only if installed (bnc#1037214)

teclator:
This is an alternative patch if we only want to modify yast2-network package with a simpler workaround.

- https://github.com/yast/yast-yast2/pull/630


mvidner:
@teclator IIUC now that https://github.com/yast/yast-yast2/pull/630 is merged, this one is not necessary anymore.

teclator:
@mvidner true, closing it ;) Thnx!

PR #577, opened 2017-10-10
Update record in /etc/hosts properly

mchf:
[bsc#1052042](https://bugzilla.suse.com/show_bug.cgi?id=1052042)
https://trello.com/c/MHxlV1Lp

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1620,10 +1620,8 @@ module Yast
     #
     # When hostname is empty, then old IP's record is cleared from /etc/hosts and
     # new is not created.
-    # When hostname is changed then canonical name and all aliases in the record
+    # Otherwise the canonical name and all aliases in the record
     # are replaced by new ones.
-    # When NIC's IP was changed than the canonical name and all aliases are assigned
-    # the new IP (or replaced by new ones if the hostname was changed as well)
     #
     # @param [String] ip address
     # @param [String] new hostname
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1614,6 +1614,17 @@ module Yast
 
     # Performs hostname update
     #
+    # This handles ip and hostname change when editing NIC properties.
+    # The method relies on old NIC's IP which is set globally at initialization
+    # of NIC edit dialog (@see LanItems#ipaddr)
+    #
+    # When hostname is empty, then old IP's record is cleared from /etc/hosts and
+    # new is not created.
+    # When hostname is changed then canonical name and all aliases in the record
+    # are replaced by new ones.
+    # When NIC's IP was changed than the canonical name and all aliases are assigned
+    # the new IP (or replaced by new ones if the hostname was changed as well)
+    #
     # @param [String] ip address
     # @param [String] new hostname
     def update_hostname(ipaddr, hostname)
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1494,20 +1494,19 @@ module Yast
         if bootproto == "static"
           ip_changed = LanItems.ipaddr != ipaddr
           hostname = @settings.fetch("HOSTNAME", "")
+          hostname_changed = @hostname_initial != hostname
 
-          if ip_changed || hostname.empty?
+          if ip_changed || hostname_changed || hostname.empty?
             log.info("Dropping record for #{LanItems.ipaddr} from /etc/hosts")
+
             Host.remove_ip(LanItems.ipaddr)
+            Host.Update(@hostname_initial, hostname, ipaddr) if !hostname.empty?
           end
 
           LanItems.ipaddr = ipaddr
           LanItems.netmask = Ops.get_string(@settings, "NETMASK", "")
           LanItems.prefix = Ops.get_string(@settings, "PREFIXLEN", "")
           LanItems.remoteip = Ops.get_string(@settings, "REMOTEIP", "")
-
-          if (@hostname_initial != hostname && !hostname.empty?) || ip_changed
-            Host.Update(@hostname_initial, hostname, LanItems.ipaddr)
-          end
         else
           LanItems.ipaddr = ""
           LanItems.netmask = ""
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -69,8 +69,6 @@ module Yast
 
       @fwzone_initial = ""
 
-      @hostname_initial = ""
-
       @force_static_ip = ProductFeatures.GetBooleanFeature(
         "network",
         "force_static_ip"
@@ -1331,14 +1329,6 @@ module Yast
 
       @fwzone_initial = fwzone
 
-      host_list = Host.names(LanItems.ipaddr)
-      if Ops.greater_than(Builtins.size(host_list), 1)
-        Builtins.y2milestone(
-          "More than one hostname for single IP detected, using the first one only"
-        )
-      end
-      @hostname_initial = String.FirstChunk(Ops.get(host_list, 0, ""), " \t")
-
       initialize_address_settings
 
       wd = Convert.convert(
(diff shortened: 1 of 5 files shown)


src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,36 @@ def repeated_physical_port_ids(slaves)
...
+      log.info("Dropping record for #{LanItems.ipaddr} from /etc/hosts")
+
+      Host.remove_ip(LanItems.ipaddr)
+      Host.Update(hostname_initial, hostname, ipaddr) if !hostname.empty?
+
+      nil
+    end
+
+    # Returns canonical hostname for the given ip
+    def hostname_initial(ipaddr)
```
jreidinger: 1. naming is a bit inconsistent. The first one is `update_hostname` and second is `hostname_initial`.
2. 11 new lines added to already long file
3. file is include, so even worse.
4. both method have hostname in its name. Maybe oppurtinity to start with proper hostname object?

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,36 @@ def repeated_physical_port_ids(slaves)
...
+
+      log.info("Dropping record for #{LanItems.ipaddr} from /etc/hosts")
+
+      Host.remove_ip(LanItems.ipaddr)
+      Host.Update(hostname_initial, hostname, ipaddr) if !hostname.empty?
+
+      nil
+    end
+
+    # Returns canonical hostname for the given ip
```
jreidinger: so why not name it `hostname_for_ip`? In object language `Hostname.for_ip(ip)`

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,36 @@ def repeated_physical_port_ids(slaves)
 
       physical_port_ids
     end
+
+    # Performs hostname update
```
jreidinger: this definitively needs more elaboration what it do as I see many side effect like that 

1. depends on current LangItems.ipaddr
2. takes original hostname for currnet ipaddr
3. it removes any hostnames for current ipaddr
4. what happens with original hostname for original ipaddr?
5. it adds new hostname to new upaddr, but only if hostname is not empty
6. behavior when hostname is empty is not documented

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,36 @@ def repeated_physical_port_ids(slaves)
...
+      hostname_changed = hostname_initial != hostname
+
+      return if !(ip_changed || hostname_changed || hostname.empty?)
+
+      log.info("Dropping record for #{LanItems.ipaddr} from /etc/hosts")
+
+      Host.remove_ip(LanItems.ipaddr)
+      Host.Update(hostname_initial, hostname, ipaddr) if !hostname.empty?
+
+      nil
```
jreidinger: new method and you do not use return statement, so this is not needed ( legacy ycp conversion )

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,36 @@ def repeated_physical_port_ids(slaves)
...
+    end
+
+    # Returns canonical hostname for the given ip
+    def hostname_initial(ipaddr)
+      host_list = Host.names(ipaddr)
+      if Ops.greater_than(Builtins.size(host_list), 1)
+        Builtins.y2milestone(
+          "More than one hostname for single IP detected, using the first one only"
```
jreidinger: what about logging what hostnames is defined for given ip?

src/include/network/lan/address.rb
mchf: ad (1) hm, hostname_initial copied original name of instance variable. I'll change it.
ad (2) just moving lines out of pretty long method, resulting file is in fact approx 3 lines shorter (not counting added docu and some white spaces)
ad (4) may be, but it is maintenance update and I don't want to create another bug by overlooking some nasty side effect.

src/include/network/lan/address.rb
mchf: documentation improved

src/include/network/lan/address.rb
```diff
@@ -1614,6 +1614,17 @@ def repeated_physical_port_ids(slaves)
...
+    # This handles ip and hostname change when editing NIC properties.
+    # The method relies on old NIC's IP which is set globally at initialization
+    # of NIC edit dialog (@see LanItems#ipaddr)
+    #
+    # When hostname is empty, then old IP's record is cleared from /etc/hosts and
+    # new is not created.
+    # When hostname is changed then canonical name and all aliases in the record
+    # are replaced by new ones.
+    # When NIC's IP was changed than the canonical name and all aliases are assigned
+    # the new IP (or replaced by new ones if the hostname was changed as well)
```
jreidinger: and it still clear old IP from /etc/hosts if I get code.

mchf: yes, bcs it is change of ip in the particular configuration - no sense to keep it there.

jreidinger:
@mvidner see my first comment, I also complain about it, but on other hand it is maintenance update, so do it properly in master, but in SP3 we should be a bit conservative.

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,47 @@ def repeated_physical_port_ids(slaves)
...
+    #
+    # This handles ip and hostname change when editing NIC properties.
+    # The method relies on old NIC's IP which is set globally at initialization
+    # of NIC edit dialog (@see LanItems#ipaddr)
+    #
+    # When hostname is empty, then old IP's record is cleared from /etc/hosts and
+    # new is not created.
+    # When hostname is changed then canonical name and all aliases in the record
+    # are replaced by new ones.
+    # When NIC's IP was changed than the canonical name and all aliases are assigned
```
mvidner: No, with the current implementation, the custom aliases are lost.

src/include/network/lan/address.rb
```diff
@@ -1631,5 +1611,47 @@ def repeated_physical_port_ids(slaves)
...
+    # Performs hostname update
+    #
+    # This handles ip and hostname change when editing NIC properties.
+    # The method relies on old NIC's IP which is set globally at initialization
+    # of NIC edit dialog (@see LanItems#ipaddr)
+    #
+    # When hostname is empty, then old IP's record is cleared from /etc/hosts and
+    # new is not created.
+    # When hostname is changed then canonical name and all aliases in the record
+    # are replaced by new ones.
```
mvidner: OK, but then this test case becomes misleading, because it shows a variant where aliases are preserved, but the code is never called that way. Is that intentional?
https://github.com/yast/yast-network/blob/a3e23d6af35defe2ae61e9aae8cb8ef5302d5de3/test/host_test.rb#L237;L245

src/include/network/lan/address.rb
mchf: Hm, may be I described that wrong. All aliases mean that when you assign FQDN (e.g. sles15.suse.com) then even alias gets created (in this case sles15). Id didn't work in some cases before. And if it works this way then it is in my POV ok to erase old aliases when new FQDN is added. Even also bcs old aliases are not directly editable in yast.

src/include/network/lan/address.rb
mchf: that test testes behavior under the hood - the old API. In this PR I've tried to create a new method which should handle the behavior in a defined way. Of course that it is possible to do more with underlying API.

src/include/network/lan/address.rb
mvidner: You are describing the *one* alias that YaST knows about but I was concerned about the other aliases, for example `dmz` and `int` here:
```
10.0.0.1    foobar-dmz.example.com foobar-dmz dmz
192.168.0.1 foobar-int.mylan foobar-int int
```

mchf: yes, that is the last sentence about. Aliases are not directly editable nor even mentioned in yast. So, what is the right behavior here when user decides to change canonical name? How should we know if he / she wants to keep old ones or expects them to be removed too? I see arguments for both cases.

teclator: Expectations are always hard to face, no one expect the same.. The current help only says it will be written and that is all.

> Optionally, you can enter a fully qualified hostname for this IP address. The hostname will be written to /etc/hosts</quote>

Why you say old aliases are not directly editable?, I hope you mean directly through yast-network but we can edit them directly with `yast2 host`

I would expect only the short alias derived from the FQDN to be deleted but not the other ones as @mvidner commented.

mchf: yes I've talked about yast2 network (lan)

mvidner: I guess dropping them is fine. But then the code comment "When NIC's IP was changed..." is false.

mvidner:
@mchf please apply https://github.com/mvidner/yast-network/commit/ebc2b219b0c93a2039bc399bb751527626c57150

mchf:
thanks for the reviews

PR #579, opened 2017-11-01
Merged SLE-12-SP3

mchf:
mainly bnc#1052042 + some small cleanup which didn't deserve own PR.


mchf:
thanks for the review

PR #583, opened 2017-11-07
Improved device name collision recognition in autoinstallation

mchf:
[bsc#1056109](https://bugzilla.suse.com/show_bug.cgi?id=1056109)
https://trello.com/c/A0ujYsUt
Previously on this topic: #533 

This PR solves issues when ay udev rules leads to a collision in device renaming. For example when AY profile defines rule eth0 -> eth1 (-> means "rename to") and eth1 exists in the system.

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -280,20 +280,34 @@ module Yast
         end
         next if !matching_item
 
-        name_from = LanItems.GetDeviceName(item)
+        name_from = item_name(item)
         log.info("Matching device found - renaming <#{name_from}> -> <#{name_to}>")
 
-        # find rule in collision
-        colliding_item, _item_map = LanItems.Items.find do |i, _|
-          LanItems.GetDeviceName(i) == name_to
-        end
-
         # rename item in collision
-        rename_lan_item(colliding_item, name_from)
+        rename_lan_item(colliding_item(name_to), name_from)
 
         # rename matching item
         rename_lan_item(item, name_to, attr, key)
       end
+    end
+
+    # Returns items id, taking into account that item could be renamed
+    #
+    # @return [String] device name
+    def item_name(item_id)
+      LanItems.renamed?(item_id) ? LanItems.renamed_to(item_id) : LanItems.GetDeviceName(item_id)
+    end
+
+    # Finds a LanItem which name is in collision to the provided name
+    #
+    # @param [String] a device name (eth0, ...)
+    # @return [Integer] item id (see LanItems::Items)
+    def colliding_item(name)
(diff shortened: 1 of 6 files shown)


test/network_autoyast_test.rb
```diff
@@ -353,7 +353,7 @@ def keep_install_network_value(value)
     before(:each) do
       allow(Yast::LanItems)
         .to receive(:Items)
-        .and_return(0 => { "ifcfg" => "eth0" })
+        .and_return(0 => { "ifcfg" => "eth0", "udev" => { "net" => ["ATTR{address}==\"24:be:05:ce:1e:91\"", "KERNEL==\"eth*\"", "NAME=\"eth0\""] } })
```
mvidner: Is *this* supposed to be a test for this PR?

What about a new test case that fails without the fix and succeeds with the fix?
What about summarizing the 62 Bugzilla comments so far so that a reviewer gets an idea of what the bug even is?

mchf: this is necessary fix to keep the testsuite working after the patch. I'll add testsuite for new methods.

test/network_autoyast_test.rb
```diff
@@ -406,4 +406,58 @@ def keep_install_network_value(value)
...
   end
+
+  context "When creating udev rules based on the AY profile" do
+    def mock_lan_item(renamed_to: nil)
+      allow(Yast::LanItems)
+        .to receive(:Items)
+        .and_return(
+          0 => {
+            "ifcfg"      => "eth0",
+            "renamed_to" => !renamed_to.nil? ? renamed_to : nil,
```
mvidner: `"renamed_to" => renamed_to,` does work :)

src/lib/network/network_autoyast.rb
```diff
@@ -280,20 +281,34 @@ def assign_udevs_to_devs(udev_rules)
         end
         next if !matching_item
 
-        name_from = LanItems.GetDeviceName(item)
+        name_from = item_name(item)
```
mvidner: OK, I think I can even see that this is the bug fix: realizing that `name_from` may also have been renamed. Right?
So we should be able to write a test case that mimics the bug report, fails without this fix and succeeds with it.

test/network_autoyast_test.rb
mchf: ehm ... right.

src/lib/network/network_autoyast.rb
mchf: yes. I'd bet that I have an AY integration test for it ... but cannot find it. I'll do something with that.

mchf:
thanks for the review

PR #586, opened 2017-11-09
Do not depend on SuSEFirewall to set firewalld interface zone (fate#323460)

teclator:
- https://trello.com/c/Yw5lJeGN/1092-5-firewall-firewalld-is-the-new-susefirewall-support-for-network-installer
- It depends on yast/yast-yast2#644


src/include/network/lan/address.rb
```diff
@@ -1621,5 +1602,19 @@ def initial_hostname(ipaddr)
 
       String.FirstChunk(Ops.get(host_list, 0, ""), " \t")
     end
+
+    def firewall_zones
+      @firewalld ||= Y2Firewall::Firewalld.instance
```
imobachgs: NP: Just a suggestion, why not creating a method like this?
```ruby
def firewalld
  @firewalld ||= Y2Firewall::Firewalld.instance
end
```

And use it forgetting about initialization. Just thinking ahead...

PR #588, opened 2017-11-10
[WIP] Drop replace direct access to LanItems::Items whenever it is possible.

mchf:


teclator:
@mchf it was created during the HW, but probably you was not able to finish it, and even some class like NetDevice were added for having some kind of discussion. 

If we do not plan to work on it we probably should close it

mchf:
well i work on it, but do not submit;-)

teclator:
@mchf so please add a [WIP] to the current PR.

mchf:
obsolete by https://github.com/yast/yast-network/pull/796

PR #591, opened 2017-11-15
Removed unnecesary SuSEFirewall.Write (bnc#1066982)

teclator:
The problem is that as the progress is not disabled the values of the Remote allowing widgets are not read correctly and are written as disabled.

BTW the call is not needed as Remote.Write already writes the SuSEFirewall configuration.

https://github.com/yast/yast-network/blob/master/src/modules/Remote.rb#L219

This PR fix the issue.

Also for record, the origin of the problem was that the SuSEFirewall.Read was moved from Remote.Read method to the Write part in this PR:

 https://github.com/yast/yast-network/pull/513

### DIFF
--- a/src/include/network/remote/dialogs.rb
+++ b/src/include/network/remote/dialogs.rb
@@ -159,8 +159,6 @@ module Yast
       end
       if ret == :next
         CWMFirewallInterfaces.OpenFirewallStore(firewall_widget, "", event)
-        # OpenFirewallStore just sets variables, but we need real firewall write
-        SuSEFirewall.Write
 
         allow_with_vncmanager = UI.QueryWidget(Id(:allow_with_vncmanager), :Value)
         allow_without_vncmanager = UI.QueryWidget(Id(:allow_without_vncmanager), :Value)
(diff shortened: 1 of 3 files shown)


src/include/network/remote/dialogs.rb
```diff
@@ -159,8 +159,6 @@ def RemoteMainDialog
       end
       if ret == :next
         CWMFirewallInterfaces.OpenFirewallStore(firewall_widget, "", event)
-        # OpenFirewallStore just sets variables, but we need real firewall write
```
jreidinger: well, maybe keep some comment here, so it is obvious that it is not needed to call.

jreidinger:
@teclator also question, when Write contain Read, does not reset it value set in Firewall widget?

teclator:
@jreidinger it shouldn't as many of YaST modules, it does not read again if already read.

https://github.com/yast/yast-yast2/blob/master/library/network/src/lib/network/susefirewall2.rb#L1506

And I already tested it manually

src/include/network/remote/dialogs.rb
teclator: if makes sense to you then it's fine for me, what I see there is that no Write method is called, just stored the SuSEFirewall config and the Remote config and then in the return of the dialog is where the Write is done, if it is not clear just reading the code then of course a comment is needed.

PR #592, opened 2017-11-22
Do not display a warning about disabled second stage

teclator:
- https://trello.com/c/9aMgl3vG/1197-autoyast-leanos-check-if-dns-warn-is-still-usefull

Once yast-network configures already the DNS when no second stage is available the warn does not make sense anymore.

Here is also the aytest that confirm it is working:

https://github.com/yast/aytests-tests/blob/master/aytests/first_stage_dns.rb

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Wed Nov 22 11:59:22 UTC 2017 - knut.anderssen@suse.com
+
+- AutoYaST: Do not display a warning about second stage being
+  disabled when importing the DNS configuration (added in 3.3.1)
+  once  the network configuration is done during the first stage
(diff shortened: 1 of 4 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Nov 22 11:59:22 UTC 2017 - knut.anderssen@suse.com
+
+- AutoYaST: Do not display a warning about second stage being
+  disabled when importing the DNS configuration (added in 3.3.1)
+  once  the network configuration is done during the first stage
+  (since version 3.3.3) (bsc#1069761)
+- 4.0.9
```
lslezak: I cannot see a version change in the spec file, it is still 4.0.8...

PR #595, opened 2017-11-24
Backported patch

mchf:
bnc#1061306


mchf:
replaced by: https://github.com/yast/yast-yast2/pull/645

PR #596, opened 2017-11-29
Bump version & changelog of (#565)

teclator:
- It increases version & provides changelog for #565


teclator:
y2firewall/firewalld/zone.rb was missing in Makefile, fixed in yast/yast-yast2#649

PR #598, opened 2017-12-04
Use improved api

mchf:
Bnc#1061306 

depends on https://github.com/yast/yast-yast2/pull/652


mchf:
thanks for the review

PR #600, opened 2018-01-17
Removed useless code

mchf:
In firewalld each device is always assigned to a zone. Either explicitly
or implicitly to a default one provided by firewalld.


mchf:
Thanks for the review

PR #603, opened 2018-01-30
Do not crash with internal error if /etc/hosts is corrupted

mchf:
[bsc#1077435](https://bugzilla.suse.com/show_bug.cgi?id=1077435)

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -354,25 +354,26 @@ module Yast
   def load_hosts
     return false if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-    begin
-      @hosts = CFA::Hosts.new
-      @hosts.load
+    @hosts = CFA::Hosts.new
+    @hosts.load
 
-      # save hosts to check for changes later
-      @hosts_init = CFA::Hosts.new
-      @hosts_init.load
-    rescue IOError, SystemCallError, RuntimeError => error
-      log.error("Loading /etc/hosts failed with exception #{error.inspect}")
-
-      # get clean environment, crashing due to exception is no option here
-      @hosts = CFA::Hosts.new
-      @hosts_init = nil
-
-      # reraise the exception - let the gui takes care of it
-      raise
-    end
+    # save hosts to check for changes later
+    @hosts_init = CFA::Hosts.new
+    @hosts_init.load
 
     true
+
+  # rescuing only those exceptions which are related to /etc/hosts access
+  # (e.g. corrupted file, file access error, ...)
+  rescue IOError, SystemCallError, RuntimeError => error
+    log.error("Loading /etc/hosts failed with exception #{error.inspect}")
+
+    # get clean environment, crashing due to exception is no option here
+    @hosts = CFA::Hosts.new
+    @hosts_init = nil
+
+    # reraise the exception - let the gui takes care of it
+    raise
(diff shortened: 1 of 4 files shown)


kobliha:
Unit test would be more than nice to have...

mvidner:
So the situation is that a config file contains garbage (which manifests as a parsing error down in CFA). It's not random bytes but YaST can't make sense of it.

If we say "Reading /etc/hosts failed" the user will think that the file is not there or the permissions are wrong. We should clearly say that we do not understand the contents.

src/modules/Lan.rb
```diff
@@ -370,7 +370,10 @@ def Read(cache)
       # Progress step 6/9
       ProgressNextStage(_("Reading hostname and DNS configuration...")) if @gui
       DNS.Read
-      Host.Read
+      if !Host.Read
+        ret = @gui && Popup.ContinueCancel(_("Reading /etc/hosts failed.\nThe file content is not in expected format.\nIf you continue any existing configuration will be overwritten."))
```
mvidner:
```
This                          line                                        is                                                                                                                                                             too long
```

src/modules/Lan.rb
```diff
@@ -370,7 +370,10 @@ def Read(cache)
       # Progress step 6/9
       ProgressNextStage(_("Reading hostname and DNS configuration...")) if @gui
       DNS.Read
-      Host.Read
+      if !Host.Read
```
mvidner: `yast2 host` will fail because it does not check the return value or Host.Read, and there are other callers too.

src/modules/Host.rb
```diff
@@ -354,14 +354,25 @@ def set_names(address, names)
...
-    @hosts.load
+    ret = true
+    begin
+      @hosts = CFA::Hosts.new
+      @hosts.load
+
+      # save hosts to check for changes later
+      @hosts_init = CFA::Hosts.new
+      @hosts_init.load
+    rescue RuntimeError => error
```
mvidner: Two problems with this:
1. `RuntimeError` is too generic
2. cannot see where the parse error occurred

In detail:

1. How about having an exception class across YaST, `GarbageDataError`, meaning that YaST cannot understand the config files that it's supposed to edit. My point is that we should not let the user proceed and instead have a global rescue, similar to the (500) Internal Error, but labeled (400) "It's Your Fault!!", or "Lost In Translation", or "Cannot Process Such Data".

2. The actual error is like
    > 'Augeas parsing/serializing error: Iterated lens matched less than it should at /usr/share/augeas/lenses/dist/hosts.aug:23.12-.42:' (RuntimeError)

    and a specific problem for Augeas is that the line location refers to the *lens*, or grammar, and not the actual unparseable data :( Line 23 in hosts.aug is

          let lns = ( empty | comment | record ) *

    We probably cannot solve that in this PR. See also <https://github.com/hercules-team/augeas/wiki/Tracking-down-errors>

jreidinger: ideally we shoud log content of file and then we can easily reproduce it.

jreidinger:
@mvidner in bootloader I am solving it with this: https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/config_dialog.rb#L23

mchf:
backported updates from https://github.com/yast/yast-network/pull/621

PR #604, opened 2018-01-30
[WIP] Do not enable xvnc.socket

schubi2:


schubi2:
Should be done in yast2-installation

PR #605, opened 2018-02-14
Fix bridge proposal (bsc#1062596)

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1062596

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -1037,7 +1037,7 @@ module Yast
       # first configure all connected unconfigured devices with dhcp (with default parameters)
       hwinfo = config.fetch("hwinfo", {})
       if hwinfo.fetch("link", false) == true
-        log.warn("item number #{LanItems.current} has link:false detected")
+        log.warn("Ifcfg #{config["ifcfg"]} has link:false detected")
         return false
       end
       if hwinfo.fetch("type", "") == "wlan"
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -1035,7 +1035,7 @@ module Yast
       end
 
       hwinfo = config.fetch("hwinfo", {})
-      unless hwinfo.fetch("link", false) == true
+      unless hwinfo.fetch("link", false)
         log.warn("Lan item #{item} has link:false detected")
         return false
       end
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -866,14 +866,14 @@ module Yast
 
         bridge_name = format("br%s", NetworkInterfaces.GetFreeDevice("br"))
 
-        if !LanItems.IsBridgeable(bridge_name, current)
-          log.info "The interface #{ifcfg} is not bridgeable."
+        if !LanItems.IsBridgeable(bridge_name, current) || !proposable_as_bridge?(config)
+          log.info "The interface #{ifcfg} can not be proposed as bridge."
           next
         end
 
         LanItems.current = current
 
-        propose_current_item!(config) if !LanItems.IsCurrentConfigured
+        LanItems.ProposeItem if !LanItems.IsCurrentConfigured
 
         next unless configure_as_bridge!(ifcfg, bridge_name)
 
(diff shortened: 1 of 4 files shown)


src/modules/Lan.rb
```diff
@@ -1033,23 +1033,18 @@ def configure_as_bridge!(ifcfg, bridge_name)
       NetworkInterfaces.Commit
     end
 
-    def propose_current_item!(config)
+    def proposable_as_bridge?(config)
       # first configure all connected unconfigured devices with dhcp (with default parameters)
       hwinfo = config.fetch("hwinfo", {})
-
       if hwinfo.fetch("link", false) == true
         log.warn("item number #{LanItems.current} has link:false detected")
```
teclator: This log has to be change, current has not been changed yet.

src/modules/Lan.rb
```diff
@@ -1033,23 +1033,18 @@ def configure_as_bridge!(ifcfg, bridge_name)
...
 
-    def propose_current_item!(config)
+    def proposable_as_bridge?(config)
       # first configure all connected unconfigured devices with dhcp (with default parameters)
       hwinfo = config.fetch("hwinfo", {})
-
       if hwinfo.fetch("link", false) == true
-        log.warn("item number #{LanItems.current} has link:false detected")
-
+        log.warn("Ifcfg #{config["ifcfg"]} has link:false detected")
```
mchf: and what if the device is not configured?

src/modules/Lan.rb
```diff
@@ -866,14 +866,14 @@ def ProposeVirtualized
...
-          log.info "The interface #{ifcfg} is not bridgeable."
+        if !LanItems.IsBridgeable(bridge_name, current) || !proposable_as_bridge?(config)
+          log.info "The interface #{ifcfg} can not be proposed as bridge."
           next
         end
 
         LanItems.current = current
 
-        propose_current_item!(config) if !LanItems.IsCurrentConfigured
+        LanItems.ProposeItem if !LanItems.IsCurrentConfigured
```
mchf: I'd put ```!LanItems.IsCurrentConfigured``` into ```!proposable_as_bridge?(config)``` and name it somehow like ```propose_bridge?``` and result use in this condition

src/modules/Lan.rb
```diff
@@ -866,14 +866,14 @@ def ProposeVirtualized
 
         bridge_name = format("br%s", NetworkInterfaces.GetFreeDevice("br"))
 
-        if !LanItems.IsBridgeable(bridge_name, current)
-          log.info "The interface #{ifcfg} is not bridgeable."
+        if !LanItems.IsBridgeable(bridge_name, current) || !proposable_as_bridge?(config)
+          log.info "The interface #{ifcfg} can not be proposed as bridge."
```
mchf: np: but you don't propose ```ifcfg``` to be a bridge. You're checking if given device can be inserted into a bridge (as a slave)

teclator: Yep, that is true, the naming was not correct as it was just proposing the item config when not configured and then it uses some properties for the bridge itself.

src/modules/Lan.rb
teclator: If it is not configured and it is bridgeable we will propose it. I could pass current instead of use config["ifcfg"] or we could set the `LanItems.current` before if we consider it is mandatory to have there the log, but probably just adding the current item into the log inside of the comparison should be enough

src/modules/Lan.rb
teclator: The naming is better, and probably what has to be changed is the `IsBridgeable` check adding a convenience method which adds to the `IsBridgeable` checks the `carrier` check and the hwinfo `wlan` check

mchf: these checks are out ```IsBridgeable``` bcs user can select a device without link explicitly if needed.

src/modules/Lan.rb
```diff
@@ -1035,7 +1035,7 @@ def connected_and_bridgeable?(bridge_name, item, config)
       end
 
       hwinfo = config.fetch("hwinfo", {})
-      if hwinfo.fetch("link", false) == true
+      unless hwinfo.fetch("link", false) == true
```
mchf: what about removing ```== true```?

mchf:
lgtm

teclator:
@mchf fix also confirmed by the customer, thnx!

PR #608, opened 2018-02-23
Propose defaults for SuSEFirewall2 also during an autoinstallation (bsc#1080630)

teclator:
- [Trello Card](https://trello.com/c/pKkPMQI4/1332-3-sle-12-sp3-l3-if-firewall-is-enabled-in-autoyast-ssh-is-not-being-allowed)
- https://bugzilla.suse.com/show_bug.cgi?id=1080630

The problem itself is described in the bug and in the Trello Card but as a summary:

After some changes in the systemd dependencies of our `Second Stage.service`, the SuSEFirewall2 configuration defined in the profiles is applied after the SuSEFirewall2 has already started. We avoided to touch the SuSEFirewall2 services during installation because a deadlock we had in the past, which means that he configuration of the profiles is not applied until the service is not restarted.

So this is the solution proposed taking in account if we don't want to bring back the previous systemd `Before` dependencies.

## Proposed Solution

During an auto-installation, if `SuSEFIrewall2` is part of the software selection and installed, we configure the `SuSEFirewall2` with the defaults settings obtained from the profile and/or enable specific access in case of remote installations (`vnc / ssh`) and in case of an `iscsi` installation (`linuxrc option`) we enable full access during the boot of the `Second Stage`.

For finishing the AutoYaST firewall configuration we should also apply the changes (https://github.com/yast/yast-yast2/pull/701), that is bring back the activation of services.

### DIFF
--- a/src/clients/firewall_stage1_finish.rb
+++ b/src/clients/firewall_stage1_finish.rb
@@ -143,16 +143,13 @@ module Yast
     def adjust_ay_configuration
       Builtins.y2milestone("Preparing proposal for autoconfiguration.")
       SuSEFirewall4Network.prepare_proposal unless SuSEFirewallProposal.GetChangedByUser
-      enable_fw = SuSEFirewall4Network.Enabled1stStage
-      return unless enable_fw
       enable_sshd = SuSEFirewall4Network.EnabledSshd
       open_ssh = SuSEFirewall4Network.EnabledSsh1stStage
       open_vnc = SuSEFirewall4Network.EnabledVnc1stStage
 
-      log.info "After installation, firewall will be #{enable_fw ? "enabled" : "disabled"}, " \
-        "SSHD will be #{enable_sshd ? "enabled" : "disabled"}, " \
-        "SSH port will be #{open_ssh_port ? "open" : "closed"}, " \
-        "VNC port will be #{open_vnc_port ? "open" : "closed"}"
+      log.info "After installation, SSHD will be #{enable_sshd ? "enabled" : "disabled"}, " \
+        "SSH port will be #{open_ssh ? "open" : "closed"}, " \
+        "VNC port will be #{open_vnc ? "open" : "closed"}"
 
       # Read the configuration from sysconfig
       # bnc#887406: The file is in inst-sys
--- a/src/clients/firewall_stage1_finish.rb
+++ b/src/clients/firewall_stage1_finish.rb
@@ -102,10 +102,13 @@ module Yast
 
     # Convenience method for opening the ssh port during the first stage when
     # enabled
+    #
+    # If the service-file is not part of the current system an exception will
+    # be raised. For that reason, these files have to be part of the inst-sys
+    #
+    # @param open [Boolean] whether the ssh service has to be opened in the
+    # firewall
     def open_ssh_port(open)
-      # Open or close FW ports depending on user decision
-      # This can raise an exception if requested service-files are not part of the current system
-      # For that reason, these files have to be part of the inst-sys
       if known_firewall_services?(SuSEFirewall4NetworkClass::SSH_SERVICES)
         SuSEFirewall.SetServicesForZones(
           SuSEFirewall4NetworkClass::SSH_SERVICES,
@@ -119,6 +122,11 @@ module Yast
 
     # Convenience method for opening the vnc port during the first stage when
     # enabled
+    #
+    # If the service-file is not part of the current system an exception will
+    # be raised. For that reason, these files have to be part of the inst-sys
+    #
+    # @param open [Boolean]
     def open_vnc_port(open)
       if known_firewall_services?(SuSEFirewall4NetworkClass::VNC_SERVICES)
         SuSEFirewall.SetServicesForZones(
--- a/src/clients/firewall_stage1_finish.rb
+++ b/src/clients/firewall_stage1_finish.rb
@@ -33,9 +33,12 @@ module Yast
     def main
       textdomain "network"
 
+      Yast.import "Mode"
       Yast.import "Service"
       Yast.import "SuSEFirewall"
+      Yast.import "SuSEFirewallServices"
       Yast.import "SuSEFirewall4Network"
+      Yast.import "SuSEFirewallProposal"
 
       @ret = nil
       @func = ""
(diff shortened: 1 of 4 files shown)


src/clients/firewall_stage1_finish.rb
```diff
@@ -66,6 +69,14 @@ def main
           "when"  => [:installation, :autoinst]
         }
       when "Write"
+        # The Autoyast configuration is mainly done during the second stage but
+        # we need to open remote services if enabled by linuxrc (bsc#1080630)
+        if Mode.autoinst && SuSEFirewall4Network.IsInstalled
+          Builtins.y2milestone("Preparing proposal for autoconfiguration.")
+          SuSEFirewall4Network.prepare_proposal unless SuSEFirewallProposal.GetChangedByUser
```
teclator: The prepare proposal can be re used as it does what we need.

src/clients/firewall_stage1_finish.rb
```diff
@@ -66,6 +69,14 @@ def main
...
+          adjust_configuration
```
teclator: adjust_configuration is defined int SuSEFirewallProposal but we don't want to add the package to the software selection as we do for installation. 

https://github.com/yast/yast-network/blob/SLE-12-SP3/src/clients/firewall_stage1_proposal.rb#L401

So this will be AutoYaST specific, and of course will define it as private.

Also, take in account that this client has been completely rewritten for master.

src/clients/firewall_stage1_finish.rb
```diff
@@ -80,6 +91,64 @@ def main
...
+        "VNC port will be #{open_vnc_port ? "open" : "closed"}"
+
+      # Read the configuration from sysconfig
+      # bnc#887406: The file is in inst-sys
+      previous_state = Progress.set(false)
+      SuSEFirewall.Read
+      Progress.set(previous_state)
+
+      SuSEFirewall.SetEnableService(enable_fw)
+      SuSEFirewall.SetStartService(enable_fw)
```
imobachgs: Do we need to execute the rest of the method if `enable_fw == false` ?

src/clients/firewall_stage1_finish.rb
```diff
@@ -80,6 +91,64 @@ def main
...
+
+      # Open or close FW ports depending on user decision
+      # This can raise an exception if requested service-files are not part of the current system
+      # For that reason, these files have to be part of the inst-sys
+      if known_firewall_services?(SuSEFirewall4NetworkClass::SSH_SERVICES)
```
imobachgs: np: You could move this if-else-end to a different method (to keep the current one smaller). Same below.

src/clients/firewall_stage1_finish.rb
teclator: No, we don't need the rest and even we should not take care of the service enablement, in that case we should read the profile definition to know what the user wanted not what is the default.

teclator: But of course it has to fit well with the fact that if the firewall is enabled by default we have to allow the remote installation during the second stage.

src/clients/firewall_stage1_finish.rb
```diff
@@ -80,6 +88,77 @@ def main
...
+    def known_firewall_services?(services)
+      @all_known_services ||= SuSEFirewallServices.all_services.keys
+
+      (services - @all_known_services).empty?
+    end
+
+    # Convenience method for opening the ssh port during the first stage when
+    # enabled
+    def open_ssh_port(open)
+      # Open or close FW ports depending on user decision
```
imobachgs: What about moving this text to the method's description? Additionally, documenting what `open` (a boolean) is would be nice.

test/firewall_stage1_finish_test.rb
```diff
@@ -0,0 +1,127 @@
...
+require_relative "test_helper"
+
+require "yast"
+require_relative "../src/clients/firewall_stage1_finish"
+
+describe Yast::FirewallStage1FinishClient do
+  describe "main" do
+    context "when the client is called with 'Info'" do
+      it "returns a hash with 'steps', 'title' and 'when' keys" do
+
```
imobachgs: np: why an empty line?

test/firewall_stage1_finish_test.rb
```diff
@@ -0,0 +1,127 @@
...
+        expect(res).to include("steps", "title", "when")
+      end
+    end
+
+    context "when the client is called with 'Write'" do
+      before do
+        allow(Yast::WFM).to receive(:Args).and_return("Write")
+      end
+
+      context "in autoninst Mode" do
```
imobachgs: typo: `autoinst` instead of `autoninst`

test/firewall_stage1_finish_test.rb
```diff
@@ -0,0 +1,127 @@
...
+          allow(Yast::SuSEFirewall).to receive(:WriteConfiguration)
+          allow(Yast::Service).to receive(:Enable)
+          allow(Yast::SuSEFirewall4Network).to receive(:IsInstalled).and_return(installed)
+          allow(subject).to receive(:adjust_ay_configuration)
+        end
+
+        context "with SuSEfirewall2 installed" do
+          let(:installed) { true }
+
+          it "allows the remote installations in use in the SuSEFirewall configuration" do
```
imobachgs: np: The description is confusing for me (maybe it is just me). What about *adjust the SuSEFirewall configuration*?

test/firewall_stage1_finish_test.rb
```diff
@@ -0,0 +1,127 @@
...
+      allow(Yast::SuSEFirewall).to receive(:Read)
+      allow(subject).to receive(:open_ssh_port)
+      allow(subject).to receive(:open_vnc_port)
+      allow(Yast::Linuxrc).to receive(:useiscsi).and_return(false)
+      allow(Yast::SuSEFirewall4Network).to receive(:Enabled1stStage).and_return(true)
+    end
+
+    context "when the user has not modified the proposal" do
+      it "obtains the linuxrc options and defauls from the control file" do
+        allow(Yast::SuSEFirewallProposal).to receive(:GetChangedByUser).and_return(false)
```
imobachgs: *Warning: just a matter of preference. Feel free to ignore. :)*

Given that you already have a different context, I would move the `allow` into a before block. Looks more readable to me. Same below.

test/firewall_stage1_finish_test.rb
```diff
@@ -0,0 +1,127 @@
...
+      it "does not modify it" do
+        allow(Yast::SuSEFirewallProposal).to receive(:GetChangedByUser).and_return(true)
+        expect(Yast::SuSEFirewall4Network).to_not receive(:prepare_proposal)
+
+        subject.send(:adjust_ay_configuration)
+      end
+    end
+
+    it "returns if the firewall is not enabled" do
+      expect(Yast::SuSEFirewall4Network).to receive(:Enabled1stStage).and_return(false)
```
imobachgs: expect or allow?

teclator:
@imobachgs I fixed a typo when logging and also during testing realized that we can not just return when SuSEFirewall4Network.Enabled1stStage is `false` as the enablement of services is done by `service-manager` that is almost the last step of the first_stage.

imobachgs:
ok, it makes sense. LGTM.

PR #609, opened 2018-02-23
Really translate firewalld zones (bsc#1082246)

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1082246

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1581,7 +1581,7 @@ module Yast
       zones = [["", _("Automatically Assigned Zone")]]
       if firewalld.installed?
         Y2Firewall::Firewalld::Zone.known_zones.map do |name, full_name|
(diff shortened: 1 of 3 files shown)


src/include/network/lan/address.rb
```diff
@@ -1581,7 +1581,7 @@ def firewall_zones
       zones = [["", _("Automatically Assigned Zone")]]
       if firewalld.installed?
         Y2Firewall::Firewalld::Zone.known_zones.map do |name, full_name|
-          zones << [name, full_name]
+          zones << [name, Builtins.dgettext(full_name)]
```
lslezak: Missing textdomain `"base"`, see https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/builtins.rb#L754

teclator: :see_no_evil:

PR #612, opened 2018-03-01
Copy wicked dhcp

teclator:
- https://trello.com/c/SXTFGogo/2130-1082832-make-sure-that-dhcp-id-keeps-the-same-on-a-running-system
- The bug was blocking @schubi2 so I have created a proposal fix.

## Fix
wickedcopyfiles

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,8 +2,8 @@
 Thu Mar  1 13:58:27 UTC 2018 - knut.anderssen@suse.com
 
 - save_network client copies wicked dhcp files from inst-sys into
-  using the same dhcp id when rebooting after installation.
-  (bsc#1082832)
+  the target using the same dhcp id when rebooting after
+  installation. (bsc#1082832)
 - 4.0.19
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Mar  1 13:58:27 UTC 2018 - knut.anderssen@suse.com
+
+- save_network client copies wicked dhcp files from inst-sys into
+  using the same dhcp id when rebooting after installation.
+  (bsc#1082832)
+- 4.0.19
+
+-------------------------------------------------------------------
 Thu Mar  1 13:56:21 UTC 2018 - lslezak@suse.cz
 
 - Fixed a random build failure (introduced by the previous fix for
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -208,7 +208,7 @@ module Yast
     WICKED_DHCP_PATH = "/var/lib/wicked/".freeze
     WICKED_DHCP_FILES = ["duid.xml", "iaid.xml", "lease*.xml"].freeze
     # For copying dhcp-client leases
-    # FIXME: We probably could ommit the copy of these leases as we are using
+    # FIXME: We probably could omit the copy of these leases as we are using
     # wicked during the installation instead of dhclient.
     DHCPv4_PATH = "/var/lib/dhcp/".freeze
     DHCPv6_PATH = "/var/lib/dhcp6/".freeze
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -31,6 +31,8 @@ module Yast
       nil
     end
 
+  private
+
     def adjust_for_network_disks(file)
       # storage-ng
       # Check if installation is targeted to a remote destination.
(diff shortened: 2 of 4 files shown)


src/lib/network/clients/save_network.rb
```diff
@@ -191,13 +191,37 @@ def copy_from_instsys
...
+      CopyConfiguredNetworkFiles()
+
+      # close and chroot back
+      WFM.SCRSetDefault(old_SCR)
+      WFM.SCRClose(new_SCR)
+
+      nil
+    end
+
+    def copy_dhcp_info
```
imobachgs: What about declaring this methods as `private`?

src/lib/network/clients/save_network.rb
```diff
@@ -191,13 +191,37 @@ def copy_from_instsys
...
+      destdir = String.Quote(Installation.destdir)
+      # Wicked dhcp files
+      wicked_path = "/var/lib/wicked/"
```
imobachgs: constant?

src/lib/network/clients/save_network.rb
```diff
@@ -191,13 +191,37 @@ def copy_from_instsys
...
+      wicked_files = ["duid.xml", "aid.xml", "lease*.xml"]
```
imobachgs: It looks like a constant too.

src/lib/network/clients/save_network.rb
```diff
@@ -191,13 +191,37 @@ def copy_from_instsys
...
+      WFM.Execute(path(".local.bash"), "mkdir -p '#{wicked_path}'")
+      wicked_files.each do |file|
+        WFM.Execute(
```
imobachgs: I may be wrong, but I would say that it is save to use `FileUtils.cp` with the `preserve` option, right? And it would be easier to avoid the loop in that way (as you won't need to worry about quoting the filenames).

teclator:
@imobachgs I did not use `Dir.glob` to expand the list of files so it will fail if not exist, I fixed that and also skipped the copy of the dhcpv6 cache file if not exist.

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Mar  1 13:58:27 UTC 2018 - knut.anderssen@suse.com
+
+- save_network client copies wicked dhcp files from inst-sys into
```
kobliha: into what :)?

src/lib/network/clients/save_network.rb
```diff
@@ -220,6 +204,33 @@ def copy_from_instsys
...
 
+    WICKED_DHCP_PATH = "/var/lib/wicked/".freeze
+    WICKED_DHCP_FILES = ["duid.xml", "iaid.xml", "lease*.xml"].freeze
+    DHCP_CLIENT_PATH = "/var/lib/dhcpcd/".freeze
+    DHCP_CLIENT_CACHE = "dhcpcd-*.cache".freeze
+    DHCPv6_CLIENT_CACHE_PATH = "/var/lib/dhcpv6".freeze
+
+    def copy_dhcp_info
+      # Ensure directories exist
+      wicked_dest_dir = Installation.destdir + WICKED_DHCP_PATH
```
kobliha: We usually use http://ruby-doc.org/core-2.5.0/File.html#method-c-join or similar

src/lib/network/clients/save_network.rb
```diff
@@ -220,6 +204,33 @@ def copy_from_instsys
       nil
     end
 
+    WICKED_DHCP_PATH = "/var/lib/wicked/".freeze
+    WICKED_DHCP_FILES = ["duid.xml", "iaid.xml", "lease*.xml"].freeze
+    DHCP_CLIENT_PATH = "/var/lib/dhcpcd/".freeze
+    DHCP_CLIENT_CACHE = "dhcpcd-*.cache".freeze
+    DHCPv6_CLIENT_CACHE_PATH = "/var/lib/dhcpv6".freeze
```
kobliha: BTW, missing "/" at the end?

teclator: In this case it just a file not a dir

src/lib/network/clients/save_network.rb
teclator: true :+1:

src/lib/network/clients/save_network.rb
kobliha: Just checked my system, there is no /var/lib/dhcpv6 but /var/lib/dhcp6

teclator: Ups, you are right the original code is a directory but it is working at all? Isn't it doing a:

```
/bin/cp -p /var/lib/dhcpv6/ /mnt/var/lib/dhcpv6
```

without recursie option looks strange

teclator: Also the bug reference of copying the dhclient files seems to be wrong

teclator: And this is the man of the client used in SLE11 so probably now /var/lib/dhcp6 should be used instead as you commented

https://www.unix.com/man-page/suse/8/dhcp6c/

src/lib/network/clients/save_network.rb
```diff
@@ -220,6 +204,39 @@ def copy_from_instsys
       nil
     end
 
+    WICKED_DHCP_PATH = "/var/lib/wicked/".freeze
```
kobliha: Just an idea: maybe `#comment` mentioning the bug?

src/lib/network/clients/save_network.rb
```diff
@@ -220,6 +204,39 @@ def copy_from_instsys
...
+    WICKED_DHCP_FILES = ["duid.xml", "iaid.xml", "lease*.xml"].freeze
+    DHCPv4_PATH = "/var/lib/dhcp/".freeze
+    DHCPv6_PATH = "/var/lib/dhcp6/".freeze
+    DHCP_FILES = "*.leases".freeze
+
+    def copy_dhcp_info
```
kobliha: Hmm, the whole function looks like copy&paste. 3x the same.

teclator: Yep it smells so and probably could be handled with loop with dirs and files, but as I was not sure about removing the copy /var/lib/dhcp files I let it like duplicated and clearly separated

src/lib/network/clients/save_network.rb
```diff
@@ -204,37 +204,42 @@ def copy_from_instsys
       nil
     end
 
+    # For copying wicked dhcp files (bsc#1082832)
     WICKED_DHCP_PATH = "/var/lib/wicked/".freeze
     WICKED_DHCP_FILES = ["duid.xml", "iaid.xml", "lease*.xml"].freeze
+    # For copying dhcp-client leases
+    # FIXME: We probably could ommit the copy of these leases as we are using
```
imobachgs: *omit*

PR #617, opened 2018-03-16
Read and write firewalld config

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1085658

A unit test would be nice, but old client is a little pain (the fix has been tested manually by @fcrozat )


teclator:
@imobachgs thnx

PR #619, opened 2018-03-27
Added missing "textdomain" call (bsc#1081466)

lslezak:
- Properly translate the web access check box label
- See https://bugzilla.suse.com/show_bug.cgi?id=1081466
- The second `textdomain` call is actually not needed (no translatable text in that class so far), but fixed as well to avoid similar issue in the future
- 4.0.24

## Original Dialog

remote_broken_translation

## Fixed Translation

remote_fixed_translation

### DIFF
--- a/src/lib/y2remote/widgets/remote.rb
+++ b/src/lib/y2remote/widgets/remote.rb
@@ -157,6 +157,10 @@ module Y2Remote
 
     # Checkbox widget for setting vnc web access as enabled when checked.
     class AllowWeb < CWM::CheckBox
+      def initialize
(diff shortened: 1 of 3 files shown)


src/lib/y2remote/widgets/remote.rb
```diff
@@ -178,6 +182,7 @@ def initialize
           "services"        => ["tigervnc", "tigervnc-https"],
           "display_details" => true
         )
+        textdomain "network"
```
teclator: It does not hurt, but also seems not needed at all

PR #621, opened 2018-04-18
Do not crash with internal error when /etc/hosts is corrupted

mchf:
Bnc1077435

extends https://github.com/yast/yast-network/pull/603 some parts should be backported

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -114,7 +114,7 @@ module Yast
     # @return true if success, raises an exception in case of malformed file
     def Read
       return true if @initialized
-      return true if !load_hosts
+      return false if !load_hosts
 
       Builtins.y2debug("hosts=#{@hosts.inspect}")
 
@@ -358,14 +358,25 @@ module Yast
   def load_hosts
     return false if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-    @hosts = CFA::Hosts.new
-    @hosts.load
+    ret = true
+    begin
+      @hosts = CFA::Hosts.new
+      @hosts.load
 
-    # save hosts to check for changes later
-    @hosts_init = CFA::Hosts.new
-    @hosts_init.load
+      # save hosts to check for changes later
+      @hosts_init = CFA::Hosts.new
+      @hosts_init.load
+    rescue RuntimeError => error
+      log.error("Loading /etc/hosts failed with exception #{error.inspect}")
 
-    true
+      # get clean environment, crashing due to exception is no option here
+      @hosts = CFA::Hosts.new
+      @hosts_init = nil
+
+      ret = false
+    end
+
+    ret
   end
 
   Host = HostClass.new
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -358,7 +358,6 @@ module Yast
   def load_hosts
     return false if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-    ret = true
     begin
       @hosts = CFA::Hosts.new
       @hosts.load
@@ -373,10 +372,11 @@ module Yast
       @hosts = CFA::Hosts.new
       @hosts_init = nil
 
-      ret = false
+      # reraise the exception - let the gui takes care of it
+      raise
     end
 
-    ret
+    true
   end
 
   Host = HostClass.new
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -358,25 +358,26 @@ module Yast
   def load_hosts
     return false if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-    begin
-      @hosts = CFA::Hosts.new
-      @hosts.load
+    @hosts = CFA::Hosts.new
+    @hosts.load
 
-      # save hosts to check for changes later
-      @hosts_init = CFA::Hosts.new
-      @hosts_init.load
-    rescue IOError, SystemCallError, RuntimeError => error
-      log.error("Loading /etc/hosts failed with exception #{error.inspect}")
-
-      # get clean environment, crashing due to exception is no option here
-      @hosts = CFA::Hosts.new
-      @hosts_init = nil
-
-      # reraise the exception - let the gui takes care of it
-      raise
-    end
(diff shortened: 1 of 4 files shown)


src/modules/Host.rb
```diff
@@ -358,12 +358,23 @@ def set_names(address, names)
...
-    @hosts = CFA::Hosts.new
-    @hosts.load
+    begin
+      @hosts = CFA::Hosts.new
+      @hosts.load
+
+      # save hosts to check for changes later
+      @hosts_init = CFA::Hosts.new
+      @hosts_init.load
+    rescue RuntimeError => error
```
jreidinger: this won't help with IOErrors like file missing. I think it makes sense to add it also there.

src/modules/Lan.rb
```diff
@@ -348,38 +348,48 @@ def Read(cache)
...
+        Builtins.sleep(sl)
+
+        return false if Abort()
+        # Progress step 8/8
+        ProgressNextStage(_("Detecting current status...")) if @gui
+        NetworkService.Read
+        Builtins.sleep(sl)
+
+        return false if Abort()
+      rescue RuntimeError => error
```
jreidinger: same IOError can happen here like if you try to write to nfs that is temporary unavailable

src/modules/Host.rb
mchf: This https://github.com/yast/yast-network/pull/621/files#diff-4be40ec54e4c9b91ae249f9931c14eb8R359 should help for missing files

mchf: However adding IOException is definitely good idea

jreidinger: it not just about missing files, but also other temporary IO problems that can happen. and it is really `IOError` not IOException

src/modules/Lan.rb
```diff
@@ -348,38 +345,47 @@ def Read(cache)
...
+
+        return false if Abort()
+        # Progress step 8/8
+        ProgressNextStage(_("Detecting current status...")) if @gui
+        NetworkService.Read
+        Builtins.sleep(sl)
+
+        return false if Abort()
+      rescue IOError, SystemCallError, RuntimeError => error
+        msg = format(_("Network configuration is corrupted.\n"\
```
teclator: FYI there is already a wrap_text helper https://github.com/search?q=org%3Ayast+wrap_text&type=Code that you could use for wrapping text with a specific width.

mchf: thx

src/modules/Host.rb
```diff
@@ -358,12 +358,23 @@ def set_names(address, names)
   def load_hosts
     return false if SCR.Read(path(".target.size"), CFA::Hosts::PATH) <= 0
 
-    @hosts = CFA::Hosts.new
-    @hosts.load
+    begin
```
mvidner: Did you know you can `rescue` the whole method body?
```rb
def foo
  bar
  baz
rescue E
  qux
end
```

It works just as well here, and the diff+history will be clearer.

src/modules/Host.rb
```diff
@@ -358,12 +358,23 @@ def set_names(address, names)
...
+      @hosts = CFA::Hosts.new
+      @hosts.load
+
+      # save hosts to check for changes later
+      @hosts_init = CFA::Hosts.new
+      @hosts_init.load
+    rescue IOError, SystemCallError, RuntimeError => error
```
mvidner: I would prefer rescuing their common superclass, which is StandardError.

I know I said "RuntimeError is too generic" in the original PR, but that was in the context of a global Yast handler for such things where I would introduce a GarbageDataError. And I forgot to do it myself when [changing CFA](https://github.com/config-files-api/config_files_api/pull/23/files#diff-53b7f088c9b3245fd0e594323ab213e8R377)

In this context, enumerating the exceptions is confusing to me as a reader, wondering whether we are letting some other exceptions slip through intentionally. Are we? Add a comment if yes, use StandardError if no.

src/modules/Lan.rb
```diff
@@ -348,38 +348,47 @@ def Read(cache)
...
+        Builtins.sleep(sl)
+
+        return false if Abort()
+        # Progress step 8/8
+        ProgressNextStage(_("Detecting current status...")) if @gui
+        NetworkService.Read
+        Builtins.sleep(sl)
+
+        return false if Abort()
+      rescue IOError, SystemCallError, RuntimeError => error
```
mvidner: Same as in load_hosts

src/modules/Lan.rb
```diff
@@ -348,38 +348,47 @@ def Read(cache)
       return false if Abort()
       # Progress step 4/8
       ProgressNextStage(_("Reading network configuration...")) if @gui
-      NetworkConfig.Read
+      begin
```
mvidner: I wonder why only a part of this long method is worth rescuing. Split off a method with a telling name? Rescue the whole method?

src/modules/Lan.rb
```diff
@@ -348,38 +348,47 @@ def Read(cache)
...
+        NetworkService.Read
+        Builtins.sleep(sl)
+
+        return false if Abort()
+      rescue IOError, SystemCallError, RuntimeError => error
+        msg = format(_("Network configuration is corrupted.\n"\
+                "If you continue resulting configuration can be malformed."\
+                "\n\n%s"), wrap_text(error.message))
+        ret = @gui && Popup.ContinueCancel(msg)
+        return false if !ret
```
mvidner: I find this logic hard to understand, I would write
```rb
  return false unless @gui
  return false unless ContinueCancel...
```
but it's your call.

src/modules/Lan.rb
mchf: ```Lan::Read``` mixes GUI preparation, GUI handling and configuration reading all together. I wanted to catch only exceptions related to configuration reading.

src/modules/Lan.rb
mchf: ok

src/modules/Host.rb
mchf: added comment

src/modules/Host.rb
mchf: ok

PR #622, opened 2018-04-18
Remove some additional network related packages proposal

mchf:
[bsc#1089471](https://bugzilla.suse.com/show_bug.cgi?id=1089471)

vlan contains only vconfig which is never called from y2-network. Moreover it was proposed only when installer had vlan configured

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -239,9 +239,7 @@ module Yast
     end
 
     # Read all network settings from the SCR
-    # @param [Symbol] cache:
-    #  `cache=use cached data,
-    #  `nocache=reread from disk (for reproposal); TODO pass to submodules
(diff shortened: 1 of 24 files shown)


src/modules/Lan.rb
```diff
@@ -975,50 +969,24 @@ def ProposeVirtualized
...
+    # Proposes additional packages when needed by current networking setup
+    #
     # @return [Array] of packages needed when writing the config
     def Packages
-      # various device types require some special packages ...
-      type_requires = {
-        # for wlan require iw instead of wireless-tools (bnc#539669)
-        "wlan" => "iw",
-        "vlan" => "vlan",
-        "tun"  => "tunctl",
```
mvidner: The changelog and the PR title only mention vlan.

mchf: both updated

PR #633, opened 2018-05-09
Fix root filesystem check for adjusting ifcfg config files

teclator:
[Trello Card](https://trello.com/c/siYsznJj/1433-3-help-with-nfs-root-in-mbrgap-bug)

See #555 & #529 & https://github.com/yast/yast-network/pull/553/commits/02412f2134502523b4fa355e3ba0e1474b99e1d7

Probably the original modifications were pushed directly to the storage-ng branch


mchf:
lgtm

PR #635, opened 2018-05-29
Fixed Internal Error on 'yast remote allow set=yes'

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=1094934


mvidner:
Now, this is just the minimal fix, but we should make this properly and add a test.

teclator:
Ouch, and my fault when cleaning up the module, thnx for taking care

jreidinger:
@mvidner what is status here? Fix looks valid from my POV.

PR #636, opened 2018-06-01
Cleanup

mchf:
Dropped some unreferenced methods, routing dialog definition into smaller parts


teclator:
Apart of that, the methods that are left are only used by the inst_do_net_test.rb client, that is only used for examples https://github.com/search?p=1&q=org%3Ayast+do_net_test&type=Code

What about droping it?

mchf:
dropping inst_do_net_test is the goal in this case, I'm still searching for more code to remove -> less code to refactor ;-))

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Jun 20 07:24:24 UTC 2018 - mfilka@suse.com
+
+- dropped useless and / or dead code
```
lslezak: NP: dead code is usually useless, `dropped dead code` could be enough IMHO...

PR #639, opened 2018-06-08
[RFC] Udev rule preparations

mvidner:
This is work in progress.
The goal is to clean up Udev Rule handling in a controlled way so that it keeps working as usual.


doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
+* Udev Rules in yast-network
```
mvidner: @jreidinger Here, I moved the file so that you can comment on it better. Enjoy!

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+The code of yast-network is messy, so Udev Rules is one area to clean it up.
+It is simple enough and well defined.
+
+** Plan
+
+- [x] collect all mentions of "udev" in yast-network
+- [ ] read them all and design a decent API
+- [ ] cover the code with tests
+- [ ] use the new API, possibly using old code underneath
+- [ ] replace old implementation with a better one
```
jreidinger: let me phrase it a bit more officially.
```
- [ ] collect existing requirements
- [ ] propose new API that fullfils this requirements
- [ ] write tests to ensure existing code works as initially expected
- [ ] replace code with new one
```
I do not think we should necessary use old implementation if new one will be straight enough

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+** AY interface
+
+The rnc schema says all elements are optional
+but the SLE12(also 15) docs says name+rule+value are all required
+
+https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#CreateProfile.Network.names
+
+Semantics: the device *name* is identified
+by the field with the key *rule* (sic!) having the value *value*
+(that does not allow some complex identifications)
```
jreidinger: well, example is better then this description. It is not much obvious for me how it works just from text.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+    <rule>
+      <name>eth0</name>
+      <rule>ATTR{address}</rule>
+      <value>00:30:6E:08:FF:80</value>
+    </rule>
+  </net-udev>
+</networking>
+#+END_SRC
+
+** where is "udev" mentioned?
```
jreidinger: is it used also outside of yast2-network? it is important info, especially to collect requirements.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+** API
+*** Naming
+
+Anything named "udev" should be code-clean.
+
+LanItems - keep it for now but remember that "item" is too generic
+  yet it combines info about a "nic" and its (stored) config.
+
+"Udev" is an implementation detail, the upper layer should say
+"NameRule"
```
jreidinger: I am not sure if NameRule is good name. What about PersistentName? or PersistancyRule?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+
+"Udev name" -> "device name", or "persistent device name"
```
jreidinger: It is hard to say without context if it makes sense. Is it rule name? or name of device after applying rule?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+
+In LanItems, keep the general interface of item["udev"]["net"]
+so that existing code works, BUT
+  instead of ["net"] which is_a Array<String>
+  make       ["name_rule"] which is_a NameRule
```
jreidinger: why on this lebev and not simply have object directly in "udev" just with "name_rule" ? or any other name?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+
+*** copying and identity
+#Items are deeply copied values? NO. GetLanItem and getCurrentItem return the
+original data.
+
+*** LanItems item API
+
+This is strictly speaking above our area of focus, but the current naming is
+so awful that at least a glossary is needed to be able to understand what's
+going on.
```
jreidinger: I prefer here to have document in code and here just link to it.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** GetLanItem(item_id) => item(item_id) # surprisingly no collisions with "item" lvar
+**** getCurrentItem      => current_item
+**** item_name_rule(item_id) = item(item_id)["udev"]["name_rule"]
+also considering that #item would be an adaptor object
+that would translate #name_rule to ["udev"]["name_rule"]
+Does it need to exist? Nil? NullRule?
+**** current_item_name_rule  =  current_item["udev"]["name_rule"]
+
+*** target API
+This is a sketch of the new API as emerging from the Usage section below
```
jreidinger: sorry, but usage is bad source for API desing. I would like to see requirements. So what we need from this rules to do. Remember Tell, don't ask and say what actions we need to do on it.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** NameRule
+***** @udev [UdevRule]
+***** #matcher= and #matcher(:bus_id or :mac)
+***** #value (case sensitive??)
+***** #name
+udev["NAME"]
+**** NameRules
```
jreidinger: Names does not look good to me, as it indicate that it is array.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+also considering that #item would be an adaptor object
+that would translate #name_rule to ["udev"]["name_rule"]
+Does it need to exist? Nil? NullRule?
+**** current_item_name_rule  =  current_item["udev"]["name_rule"]
+
+*** target API
+This is a sketch of the new API as emerging from the Usage section below
+**** NameRule
+***** @udev [UdevRule]
+***** #matcher= and #matcher(:bus_id or :mac)
```
jreidinger: why having `matcher=`? this looks like only item that is not immutable.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** item_name_rule(item_id) = item(item_id)["udev"]["name_rule"]
+also considering that #item would be an adaptor object
+that would translate #name_rule to ["udev"]["name_rule"]
+Does it need to exist? Nil? NullRule?
+**** current_item_name_rule  =  current_item["udev"]["name_rule"]
+
+*** target API
+This is a sketch of the new API as emerging from the Usage section below
+**** NameRule
+***** @udev [UdevRule]
```
jreidinger: so Name rule contain UdevRule? That looks wrong for me. If you want to have generic NameRule and Udev only as backend, then Udev should be serializer, that reads NameRule from udev and write it back.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+***** #matcher= and #matcher(:bus_id or :mac)
+***** #value (case sensitive??)
+***** #name
+udev["NAME"]
+**** NameRules
+***** #pathname
+"/etc/udev/rules.d/70-persistent-net.rules"
```
jreidinger: where it is needed? It more looks like udev specific.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+
+*** Usage
+Here I list all the mentions of "udev" in the code
+and sketch out how to write them better.
+**** InstallInfConvertor
+***** AllowUdevModify
+checks if cmdline contains "biosdevname=..."
```
jreidinger: Please try to transform this into requirements. What means when there is biosdevname? we need to write some special rule?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** NetworkLanComplexInclude src/include/network/lan/complex.rb
+***** calls LanItems.update_item_udev_rule!(:bus_id)
```
jreidinger: this is quite useless. What it in fact do? What is required action for udev rules? Then maybe it make sense to have similar method on RuleNames?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** Lan#Export
+calls LanItems#Export
```
jreidinger: relevance to udev?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** lan_auto
+***** ToAY converts the net-udev piece from a hash to an array
```
jreidinger: does it make sense to allow creation of NameRules from this hash and method like `NameRules.load_from_ay` ?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** NetworkLanHardwareInclude
+it's the Hardware tab
+      device_name = LanItems.current_udev_name
```
jreidinger: what is requirement here? just to get persistent name of device?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+let's keep that
+**** save_network
+#copy_udev_rules
+
+s390 51* leave that
+
+the rule file needs to be copied from inst-sys to target:
+  need its fs path
+  NameRules#pathname (and use std ruby dirname+basename)
```
jreidinger: really? Is it really needed? Maybe we can think one level high and it can be NameRules.load and then save with optional prefix? so API can look like with notes from above
```
name_rules = UdevSerializer.load
UdevSerializer.save(name_rules, prefix: "/mnt")
```

Does it make sense? Or maybe there is another requirements that is still hidden?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+BTW the https://bugzilla.suse.com/show_bug.cgi?id=293366#c7 comment means
+a mkdir -p is fine
```
jreidinger: requirement. It should be in own list.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** edit_nic_name EditNicName
+is a freshly rewritten class, yay 2013-09 mchf
+well, it is called like EditNicName.new.run
+and its #initialize uses the ugly LanItemsApi
+so does #run
+and #CheckUdevNicName (sic)
```
jreidinger: what is its goal? Change Persistent name? what is allow to modify? Again, please capture requirements.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+***** initialize
+@old_key = current_item_name_rule.matcher
+***** run
+        LanItems.update_item_udev_rule!(udev_type)
+(watch out, uses the ui symbol directly)
+***** CheckUdevNicName
+uses LanItems#GetCurrentName
+  which is GetDeviceName(@current) ... and it never uses the "udev name" which
+  confuses my naming plan :(
+renamed! to check_new_device_name
```
jreidinger: I do not like new name and I think here you go to non-relevant part and too deep. At first we should propose that new api. Renaming of non-relevant part is out of scope.

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+**** network_autoyast
+renaming logic
```
jreidinger: renaming of what? what it actually do with udev rules?

doc/udev-rules.org
```diff
@@ -0,0 +1,244 @@
...
+***** create_udevs
+"# Creates udev rules according definition from profile"
+rename to create_name_rules_from_profile
```
jreidinger: non-relevant to udev. Lets please focus on new API and how to use it in existing code. If we do not focus, we start again rewritting everything.

src/lib/network/edit_nic_name.rb
```diff
@@ -137,8 +137,8 @@ def close
     #
     # Pops up an explanation if the name is invalid
     #
-    # @return [boolean] false if name is invalid
-    def CheckUdevNicName(name)
+    # @return [Boolean] false if name is invalid
+    def check_new_device_name(name)
```
jreidinger: hmm, I am not fan of renaming in api proposal. Other documentation improvements :+1:

doc/udev-rules.org
mvidner: "Udev" is mentioned a lot in yast-storage-ng, but the situation is quite different:

What network and storage devices have in common is that both kinds of devices have a *bus position*, and some kind of *unique id*, and a *short kernel name*.

Storage devices have a device file that exists in the filesystem, so naturally you can make symlinks pointing to `/dev/sda1` from `/dev/disk/by-path/pci-0000:00:1f.2-scsi-0:0:0:0-part1` or from `/dev/disk/by-id/ata-ST3500418AS_9VMN8X8L-part1`. For which there are udev rules that simply *work* and YaST does not have to care.

Networking devices **suck** because they **do not exist** in the filesystem, so we cannot have multiple symlinks to a "canonical" device name. We must embed the paths and ids in the udev rules, even to choose one of the naming schemes over the others (name NICs by MAC, not by PCI bus position).

mvidner: **yast-country** has one place which needs to write a simple udev rule, which is done by treating it as a string. I think it is better to do it so without adding a udev library dependency:

https://github.com/yast/yast-country/blob/1538404d23c64c5733862abd0ded363f47594732/keyboard/src/modules/Keyboard.rb#L1340

doc/udev-rules.md
```diff
@@ -0,0 +1,163 @@
...
+### Persistent Name Rules
+
+
+#### set the name of a network interface
+
+NIC > Edit > Hardware > Udev Rules > Device Name > Change >
+  Device Name
+
+(But if a name is changeable, what then identifies an interface?
+Between reboots it is the interface index.)
```
jreidinger: here it should be current name, not?

doc/udev-rules.md
```diff
@@ -0,0 +1,163 @@
...
+### Y2Network::NameRule
+
+Ensures that one NIC keeps the name assigned to it
+(using a {::UdevRule udev rule})
+
+- @udev [UdevRule]
+- attr_accessor matcher (:bus_id or :mac)
+- attr_accessor match_value [String]
+  eg. "11:22:33:aa:bb:cc" or "0000:00:03.0"
+- attr_accessor name [String]
```
jreidinger: ugh, do you really want to make so much muttable class? What happens when I do something like
```
rule = Rules.read.first
rule.name = "eth2"
rule.match_value = "abc"
rule.matcher = :mac
```

how do you want to write it?

doc/udev-rules.md
```diff
@@ -0,0 +1,163 @@
...
+
+API
+---
+
+### Y2Network::NameRule
+
+Ensures that one NIC keeps the name assigned to it
+(using a {::UdevRule udev rule})
+
+- @udev [UdevRule]
```
jreidinger: what is benefit of this? Why not having dedicated UdevWriter that gets set of NameRules and write them or read them?

doc/udev-rules.md
```diff
@@ -0,0 +1,163 @@
...
+- .from_ay(Hash{"name", "rule", "value"})
+- #to_ay[Hash{"name", "rule", "value"}]
+
+### Y2Network::NameRules
+
+Is this just an array of NameRule ?
+Mostly, but importantly it deals with their persistence
+
+- #pathname
+"/etc/udev/rules.d/70-persistent-net.rules"
```
jreidinger: is it important to expose pathname? And is it attribute of NameRules? Why? For me it will make more sense to have it in some reader/writer.

src/modules/LanItems.rb
```diff
@@ -2790,6 +2808,8 @@ def find_by_sysconfig
       GetDeviceNames(items)
     end
 
+  public # rubocop:disable Lint/UselessAccessModifier
```
jreidinger: ugh what? why it is needed?

doc/udev-rules.md
```diff
@@ -0,0 +1,163 @@
...
+
+- .from_ay(Array<Hash{"name", "rule", "value"}>)
+- #to_ay [Array<Hash{"name", "rule", "value"}>]
+
```
jreidinger: I expect here method to mass assign that rules or is it hidden in from_ay?

imobachgs:
We have rewritten the [support for udev rules](https://github.com/yast/yast-network/blob/0f141be7bb7e616002875c05f9f0f1f5bac8a295/src/lib/y2network/udev_rule.rb).

We basically use udev rules to do devices renaming and to set the driver to be used. So, instead of exposing the udev rules thingie, we offer `Y2Network::Config` methods to set the [renaming mechanism](https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/interface.rb#L49) and the [driver](https://github.com/yast/yast-network/blob/0f141be7bb7e616002875c05f9f0f1f5bac8a295/src/lib/y2network/physical_interface.rb#L34).

That information will be translated to the corresponding udev rules.

PR #642, opened 2018-06-12
checking empty host name entries in AY

schubi2:


src/modules/Host.rb
```diff
@@ -186,6 +187,8 @@ def Import(settings)
         set_names(ip, names)
       end
 
+      check_profile_for_errors(imported_hosts)
```
teclator: As we discussed, here imported_hosts have been already treated,

First, it will parsed by the host_auto client, modifying the structure

https://github.com/yast/yast-network/blob/master/src/clients/host_auto.rb#L87

and then by this ugly hack conversion for old profiles format 

https://github.com/yast/yast-network/blob/ca7ec09979c4cc7939a4d3ce527680a87fd0921f/src/modules/Host.rb#L179

With the current profile format you should add the profile before the conversion (if not the entries could be joined in one single name), and what should be tested is that the hosts.size for the current ip is less than 2

src/modules/Host.rb
```diff
@@ -343,6 +346,22 @@ def SetModified
 
   private
 
+    # Semantic AutoYaST profile check
+    #
+    # Problems will be stored in AutoInstall.issues_list.
+    # @param [Hash] input autoyast settings
+    def check_profile_for_errors(imported_hosts)
+      # Checking for empty hostnames
+      imported_hosts.each do |ip,hosts|
```
teclator: As commented previously, duplicate ips with the current profiles means, `hosts.size > 2` if I'm not wrong

PR #645, opened 2018-06-14
Do not enforce particular connection mode for IPoIB devices

mchf:
bnc#1086454


src/include/network/widgets.rb
```diff
@@ -212,10 +212,16 @@ def MakeStartmode(ids)
...
+      {
+        # translators: a possible value for: IPoIB device mode
+        "connected" => _("connected"),
+        "datagram"  => _("datagram")
+      }
+    end
 
-      return unless LanItems.ipoib_modes.keys.include?(ipoib_mode)
+    def init_ipoib_mode_widget(key)
+      ipoib_mode = ipoib_modes[LanItems.ipoib_mode] || "default"
```
jreidinger: shouldn't default should be also translated?

mchf: it is ... https://github.com/yast/yast-network/pull/645/files/d35027dc68becfa19b3e80597c81c7d78be45078#diff-d3e502cdeb4efc24d0912f7c88bd2374R243

jreidinger: my problem here is that first part of expression in translated and second part not. I think it is a bug. I actually think that first part should be only `LanItems.ipoib_mode || "default"` as That change widget below want Id and not value.

mchf: ah ... I see ... you're right

PR #647, opened 2018-06-18
Fixed several rpmlint warnings

mchf:
Those two reported in https://bugzilla.opensuse.org/show_bug.cgi?id=1097954 plus 

yast2-network.noarch: W: obsolete-not-provided yast2-network-devel-doc
 If a package is obsoleted by a compatible replacement, the obsoleted package
 should also be provided in order to not cause unnecessary dependency breakage.
 If the obsoleting package is not a compatible replacement for the old one,
 leave out the Provides.

### DIFF
--- a/package/yast2-network.spec
+++ b/package/yast2-network.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.0.33
+Version:        4.0.34
 Release:        0
 BuildArch:      noarch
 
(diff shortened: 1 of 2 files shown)


package/yast2-network.spec
```diff
@@ -64,12 +65,15 @@ Conflicts:      yast2-core < 2.10.6
 
 Requires:       yast2-ruby-bindings >= 1.0.0
 
-Obsoletes:      yast2-network-devel-doc
+Obsoletes:      yast2-network-devel-doc <= 3.1.154
+Provides:       yast2-network-devel-doc > 3.1.154
```
jreidinger: what about `Provides:  yast2-network-devel-doc = %{version}` ?

mchf: good tip, thanks

PR #651, opened 2018-06-22
Activate s390 network devices before applying udev naming rules

mvidner:
https://bugzilla.suse.com/show_bug.cgi?id=1095761


src/lib/network/network_autoyast.rb
```diff
@@ -69,6 +70,9 @@ def create_udevs
       no_rules &&= ex_udev_rules.empty?
       return if no_rules
 
+      s390_devices = ay_networking_section.fetch("s390-devices", {})
+      s390_devices.each { |rule| LanItems.createS390Device(rule) } if Arch.s390
```
jreidinger: What happens when there is no net-udev key? is it fine that this call will be skipped?

mchf: and if there are rules then ```.createS390Device``` is called twice (second one is in ```LanItems.Import```? May be moving it from here somewhere around https://github.com/yast/yast-network/blob/master/src/lib/network/clients/save_network.rb#L188 and dropping it from ```LanItems.Import``` would also work.

jreidinger: @mvidner ok, so please write a comment to this code that it is fine. As it is not obvious from first view.

PR #653, opened 2018-06-26
Cleanup - removed NetworkStorage module

mchf:


package/yast2-network.changes
```diff
@@ -1,3 +1,8 @@
+-------------------------------------------------------------------
+Tue Jun 26 09:18:50 UTC 2018 - mfilka@suse.com
+
+- removed NetworkStorage module, replaced by new storage
```
mvidner: Yay!
But this leaves the reader wondering whether you are replacing it by "new storage" only now.
How about:
> removed NetworkStorage as it is unused (had been replaced by storage-ng)

mchf: ok, updated. thx

PR #661, opened 2018-07-12
Dropped calls to obsolete NetworkInterfaces API

mchf:
bnc#964856

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -3,6 +3,8 @@ Thu Jul 12 11:46:53 UTC 2018 - mfilka@suse.com
 
 - bnc#964856
   - dropped calls to obsolete NetworkInterfaces API
+  - CLI: setting device type explicitly using new type option is
+    mandatory when adding new device
 - 4.1.4
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Jul 12 11:46:53 UTC 2018 - mfilka@suse.com
+
+- bnc#964856
+  - dropped calls to obsolete NetworkInterfaces API
+- 4.1.4
+
+-------------------------------------------------------------------
 Mon Jul  2 16:20:55 CEST 2018 - schubi@suse.de
 
 - lan_test: stubbin Host.GetModified call (bnc#1052042).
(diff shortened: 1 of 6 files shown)


src/include/network/lan/address.rb
```diff
@@ -1430,9 +1424,7 @@ def AddressDialog
       end
 
       # proceed with WLAN settings if appropriate, #42420
-      if ret == :next && LanItems.type == "wlan" && LanItems.alias == ""
```
teclator: So, aliases are not important anymore, could it arise in some problem?

teclator:
Drop code seems to be OK, but should be fine to fix real bug during the cleanup process

src/include/network/lan/address.rb
mchf: ```LanItems::aliases``` is still in game. ```LanItems::alias``` is not. Looks confusing? That happens when someone re-implements a new feature and do not clean the code properly.

mchf:
It is a real bug ... see description ;-) but seriously using obsolete api *is* a bug and one never knows when it bits him. And we do not wait for a customer's report. That it hasn't crashed in this case up to know is partly luck and partly a result of nontrivial effort spent in various other bugs.

teclator:
Code looks good, but I missed like unit tests or some explanations, although taking in account that it is probably leftlover code then it is ok

src/include/network/lan/cmdline.rb
```diff
@@ -194,9 +194,9 @@ def AddHandler(options)
         [LanItems.current, "ifcfg"],
         Ops.get(options, "name", "")
       )
-      LanItems.type = NetworkInterfaces.device_type(
-        Ops.get(options, "name", "")
-      )
+      LanItems.type = options["type"]
+      raise "Device type is mandatory." if !LanItems.type
```
jreidinger: this looks like API change to me. So now in CLI type is mandatory? If so, then please mention it in changes file

mchf: yes. Bcs before it was "detected" from device name which is not possible anymore. In UI it is done by explicitly selecting by user ... so I had to introduce something similar here too.

Will add change log entry.

package/yast2-network.changes
```diff
@@ -3,6 +3,8 @@ Thu Jul 12 11:46:53 UTC 2018 - mfilka@suse.com
 
 - bnc#964856
   - dropped calls to obsolete NetworkInterfaces API
+  - mentioning device type explicitly using new option is mandatory
+    since now when adding device using CLI
```
jreidinger: I would rephrase it a bit: "CLI: for new devices the option type is now mandatory"

PR #663, opened 2018-08-03
Permitted the use of listentry in networking AutoYaST schema list entries

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1103712

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Aug  3 16:42:04 UTC 2018 - knut.anderssen@suse.com
+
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Aug  3 16:42:04 UTC 2018 - knut.anderssen@suse.com
+
+- Fixes to the networking AY schema (bsc#Bug 1103712)
```
imobachgs: The `bsc` format looks wrong to me.

teclator: Yep, a mistake pasting the number

PR #671, opened 2018-09-19
Merged SLE-12-SP3

mchf:
It seems that by accident "someone" did version 3.2.53 for SP3 even SP4 -> minor version bumped

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -17,6 +17,7 @@ module Yast
 
     def initialize
       # import has to be done here, there are some collisions otherwise
+      Yast.import "Arch"
       Yast.import "Lan"
       Yast.import "LanItems"
       Yast.import "Linuxrc"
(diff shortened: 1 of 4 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -69,6 +70,10 @@ def create_udevs
       no_rules &&= ex_udev_rules.empty?
       return if no_rules
 
+      # for the purpose of setting the persistent names, create the devices 1st
+      s390_devices = ay_networking_section.fetch("s390-devices", {})
+      s390_devices.each { |rule| LanItems.createS390Device(rule) } if Arch.s390
```
lslezak: IMHO the whole block should be in the `if Arch.s390` condition. Or does it make sense to fetch the "s390-devices" on non-S390? This is a bit strange...

mchf: well, I'm not author of this change but ... it is expectable that the section is empty on non-s390 ... and if not then it is catched by the second condition.

so, having it encapsulated in a if statement is cleaner, but doesn't hurt when done this way.

lslezak: ok

PR #682, opened 2018-10-16
Added new network configuration proposal (fate#326480)

teclator:
- [Trello Card](https://trello.com/c/DLi2b1mG/338-fate-326480-add-network-to-installation-proposal-overview-page)

## Added network proposal
networkproposal

## Compact the current information

completenetworkproposal

### DIFF
--- a/src/lib/network/lan_items_summary.rb
+++ b/src/lib/network/lan_items_summary.rb
@@ -56,6 +56,29 @@ module Yast
       Summary.DevicesList(items)
     end
 
+    # Generates a summary in RichText format for the configured interfaces
+    #
+    # @example
+    #   LanItemsSummary.new.proposal
+    #   => "<ul><li><p>Configured with DHCP: eth0, eth1<br></p></li>" \
+    #      "<li><p>br0 (Bridge)<br>IP address: 192.168.122.60/24" \
+    #      "<br>Bridge Ports: eth2 eth3</p></li></ul>"
+    #
+    # @see Summary
+    # @return [String] summary in RichText
+    def proposal
+      items = []
+
+      items << "<li>#{dhcp_summary}</li>" unless dhcp_ifaces.empty?
+      static_ifaces.each do |name|
+        items << Summary.Device(summary_label_for(name), summary_details_for(name))
+      end
+
+      return Summary.NotConfigured if items.empty?
+
+      Summary.DevicesList(items)
+    end
+
     # Generates a one line text summary for the configured interfaces.
     #
     # @example with one configured interface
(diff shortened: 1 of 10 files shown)


src/lib/network/lan_items_summary.rb
```diff
@@ -109,5 +132,61 @@ def ifcfg_protocol(ifcfg)
         LanItems.DeviceProtocol(ifcfg)
       end
     end
+
+    # Return the label name for the given interface
+    #
+    # @return [String] label name for the given interfave
+    def summary_label_for(name)
```
mchf: missing ```@param```

src/lib/network/lan_items_summary.rb
```diff
@@ -109,5 +132,61 @@ def ifcfg_protocol(ifcfg)
...
+    end
+
+    # Convenience method that obtains the list of static configured interfaces
+    #
+    # @return [Array<String>] static configured interface names
+    def static_ifaces
+      LanItems.find_static_ifaces
+    end
+
+    # Convenience method that obtains the list of static configured interfaces
```
mchf: no ... all devices which have an ifcfg file

src/lib/network/lan_items_summary.rb
```diff
@@ -109,5 +132,61 @@ def ifcfg_protocol(ifcfg)
...
+    #
+    # @return [Array<String>] static configured interface names
+    def configured_ifaces
```
mchf: seems unused

src/lib/network/lan_items_summary.rb
```diff
@@ -109,5 +132,61 @@ def ifcfg_protocol(ifcfg)
...
+
+    # Return the label name for the given interface
+    #
+    # @return [String] label name for the given interfave
+    def summary_label_for(name)
+      case NetworkInterfaces.GetType(name)
+      when "br"
+        "#{name} (Bridge)"
+      when "bond"
+        "#{name} (Bond Master)"
```
mchf: I'm not sure here but ```_()``` may be needed

mchf:
I was thinking of it one more time and I'd go for even more minimalist solution. I mean something like:
```
- Configured with DHCP: eth1, eth2, ...
- Statically configured: eth0, ...
- Bridges: br0, ...
...
```
for quick overview this is enough in my POV (it answers questions like: Did I forget to configure eth0 or not?) all other details are available when the caption is clicked - why to show static ip in the overview when dynamic one is not shown etc. So, in short I'd make the overview even more compact and high level.

also please notice that there are no empty lines between bullets in other proposals ... it should look consistently.

teclator:
Already realized the extra line between bullets, I thought it was intentional for device list entries and was not sure if remove or not, so for me is completely fine to remove them.

Also fine with a more compact overview, the idea about having the bridges and bondings apart was because the enslaved interfaces (just to be reflected in some section). 

So the main doubt at this point is if we should also collect all of them in the same entry, or maybe just having something like:

`
- Slaves interfaces: eth0, eth1 (Bridge br0) - eth2, eth3 (Bonding bond0)
`

mchf:
Well it depends, I'd prefer:
```
- Bridges: br0 (eth0, eth1), br1 (bond0, bond1)
- Bondings: bond0 (eth4, eth5)
```
bcs you can e.g. have bridge on top of bonded devices and bond device / bridge is more interesting device for me then enslaved ones.

mchf:
and I don't know if the empty line was intentional or not but now is new era ... so remove it ;-)

teclator:
@mchf I buy it ;)

teclator:
Yep, I also realized that, and probably we should use `Bonds` or `Bonding interfaces`, `Bondings` is not very common.

teclator:
@mchf sorted

mchf:
Well I usually use bond device(s) ... You're right that "Bondings" looks strange ;-)

Otherwise LGTM

PR #683, opened 2018-10-24
always propose wpa supplicant for wlan

mchf:
Bnc#1111925 

contains backport of https://github.com/yast/yast-network/pull/684


src/modules/Lan.rb
```diff
@@ -542,6 +542,7 @@ def Write
       # Progress step 3 - multiple devices may be present, really plural
       ProgressNextStage(_("Writing device configuration..."))
       LanItems.write
+      return false if PackagesInstall(Packages()) != :next
```
teclator: Is it really needed?, Isn't it handled by the sequence itsef?

src/modules/Lan.rb
```diff
@@ -888,37 +889,22 @@ def Packages
       # various device types require some special packages ...
       type_requires = {
         # for wlan require iw instead of wireless-tools (bnc#539669)
```
teclator: aparently this depend also in the sles/.. version as @mtomaschewski pointed on IRC, so we should check if it is still valid or we should do some modifications here (wireless-tools is still the package required in some versions

mchf: this is SLE-12-SP4, the method looks completely differently in SLE-15 where dependencies were fixed already

src/modules/Lan.rb
mchf: just wanted to stop Lan::Write earlier ... otherwise it will try to write Routing, DNS, ...

teclator: Then seems it should be like an extra step about checking missed dependencies.

mchf: fixed ... I did a mistake in my tests before. thaks

src/modules/Lan.rb
teclator: aha, I checked it in SLE-12-SP3 and I get:

```bash
$ rpm -qf /usr/sbin/iwlist 
wireless-tools-30.pre9-34.15.x86_64
```

teclator: The need of "iw" was basically because old sysconfig ifup-wireless script but I think it was only for (<= sles11) @mtomaschewski can confirm it, see the original bug (bnc#539669). So seems the removed dependency "wireless-tools" makes more sense for the iwlist issue, but for that special case it should be installed before the first wlan interface configured or when the network scanning is requested.

teclator: BTW as I said it could be handled in a separate PR to not block this fixes. I already confirmed that iw is at least present in a SLE-12-SP4 normal installation and don't have the issue I found on SLE-12-SP3

mchf: I'll answer according to my knowledge here. Requirements were already revised and updated for SLE-15 and iw removal is one of these updates - based on discussion with @mtomaschewski, so hopefully in shape with wicked needs.

I'm currently testing some default proposals for WPA / PEAP, so I'll try to somehow handle even iwlist issue, but I want to backport as less as possible from SLE-15

mchf: and I don't care of sle-12-sp3 ... I don't think that wlan configuration is that crucial thing for sle product line ... at leas no one complained so far for this product ;-)

package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Oct 24 08:26:24 UTC 2018 - mfilka@suse.com
+
+- bnc#1111925
```
teclator: Not mentioned the new WPA EAP defaults, I would add it

mchf: ok

teclator:
LGTM

PR #685, opened 2018-10-26
Require the installation of wireless-tools for scanning (bsc#1112952)

teclator:
Possible fix for https://bugzilla.suse.com/show_bug.cgi?id=1112952

It requires the installation of wireless-tools package only when it tries to configure a new interface in order to enable the scanning functionality.

screenshot from 2018-10-26 09-56-52

## If canceled the installation then the scanning functionality is disabled
screenshot from 2018-10-26 09-57-20

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Fri Jan 25 13:24:03 UTC 2019 - knut.anderssen@suse.com
+
+- bsc#112952
+  - Try to install the wireless-tools package when the package is
+    not installed and the wifi networks are scanned.
+- 4.1.35
+
+-------------------------------------------------------------------
 Wed Jan 23 08:52:03 UTC 2019 - mfilka@suse.com
 
 - bnc#1122387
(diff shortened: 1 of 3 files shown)


mchf:
well I know that my fixes take ages to be finished but that's bcs in this case I'm playing with two approaches and evaluating pros and cons of both ...
1) make yast2-network depend on wireless-tools
2) yast2-network asks for installing of the package when needed

what i dislike on this patch is when i switch to the wireless setup it immediately can raise a popup and then installation which simply slows down the workflow for every user even when e.g. checking for setup and so on.

so for the 2nd approach I'd prefer to install the package when explicitly asked for scanning. 

in fact i currently slightly tend to 1st solution from the above. Simply bcs it makes life simpler in the installer.

teclator:
1. requiring the package only when writing or before it is not enough for the first time you try to configure a wlan interface, that is why I had to add an explicit requirement.
2. I already played with both, in the first attempt already added it in the :scan_for_networks events, so it is also fine with me.

mchf:
I talked (and I'm thinking) about making yast2-network package depend on wireless-tools in rpm way ... so no explicit handling in the code at all

teclator:
Isn't it too much?

mchf:
ok, so let put it into "Scan network" handler pls.

PR #686, opened 2018-10-26
"yast lan add" needs "type" now

mvidner:
https://openqa.suse.de/tests/2096097#step/yast2_cmdline/22 fails,
broken since 9abc2864df02710e30682874c71f98db81f73571
reported by @mloviska


mvidner:
Also, before I forget, only tangentially related, that test failure first reports an Internal Error, but then the test harness reports "ok 2 vlan50: added", which is wrong. It turns out that `y2start` discards the return value of Yast::WFM.CallFunction and always returns 0.

PR #690, opened 2018-11-05
added bridge setting to rnc file

schubi2:


teclator:
Just for record, it backport some changes added in https://github.com/yast/yast-network/pull/365

teclator:
Mmmm maybe we should also backport the changes added in https://github.com/yast/yast-network/pull/676 ?

PR #692, opened 2018-11-09
Cleanup in hostname / dns tab

mchf:
https://trello.com/c/so0amSNp


mchf:
hostname_dns_tab

teclator:
It looks like it was requested, so fine with me

src/include/network/services/dns.rb
```diff
@@ -244,14 +212,13 @@ def initialize_network_services_dns(include_target)
 
       @dns_td = {
         "resolv" => {
-          "header"       => _("Hostname/DNS"),
+          "header"       => _("DHCP/DNS"),
```
teclator: Well, it is still Hostname / DNS configuration although it has the option for setting the hostname via DHCP, so, not sure if this will make the user more confuseD about what is the Tab about.

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Mon Nov 26 11:57:35 UTC 2018 - mfilka@suse.com
+
+- bnc#934055
```
teclator: What about referencing also this bug that is for SLES

src/include/network/services/dns.rb
mchf: i want to suppress any notice of this Hostname stuff when I was unable to get permission to remove it ;-) Moreover I'd like to move some DHCP pieces from global tab here in the future

package/yast2-network.changes
mchf: well i don't have that many bug numbers to choose from.

mchf:
Who reads helps? ;-) OK, will check

package/yast2-network.changes
teclator: Sorry, I forgot to paste the bug number :)

teclator: https://bugzilla.suse.com/show_bug.cgi?id=972463

src/include/network/services/dns.rb
teclator: Then do it when removed / moved as currently it will a point of confusion.

teclator:
@mchf for me it looks OK once we resolve this pending discussion https://github.com/yast/yast-network/pull/692/files#r236228926

src/include/network/services/dns.rb
mchf: reverted

PR #693, opened 2018-11-09
Allow to change network backend from proposal

dgdavid:
Related to https://trello.com/c/4ptXAFVW/441-network-proposal-in-leap-151-networkmanager

Examples of resulting proposals: 

| Using wicked | Using NetworkManager |
|--------------|-----------------------|
| using wicked | using NetworkManager |

### DIFF
--- a/src/lib/network/clients/network_proposal.rb
+++ b/src/lib/network/clients/network_proposal.rb
@@ -6,6 +6,11 @@ module Yast
     include Yast::I18n
     include Yast::Logger
 
+    BACKEND_LINKS = [
+      SWITCH_TO_WICKED = "network--switch-to-wicked".freeze,
+      SWITCH_TO_NETWORK_MANAGER = "network--switch-to-nm".freeze
+    ].freeze
+
     def initialize
       Yast.import "UI"
       Yast.import "Lan"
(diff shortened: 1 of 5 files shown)


src/lib/network/clients/network_proposal.rb
```diff
@@ -24,24 +29,79 @@ def description
...
+      :next
+    end
+
+    def wicked_backend?
+      Yast::NetworkService.wicked?
+    end
+
+    # TODO: move to HTML.ycp
+    def Hyperlink(href, text)
+      Builtins.sformat("<a href=\"%1\">%2</a>", href, text)
```
lslezak: NP: We should HTML escape the text, just to make sure it does not break when using a translated string.

```suggestion
      Builtins.sformat("<a href=\"%1\">%2</a>", href, CGI.escapeHTML(text))
```

and add `require "cgi"` at the top.

src/lib/network/clients/network_proposal.rb
```diff
@@ -24,24 +29,79 @@ def description
...
+      proposal_text.prepend(Yast::Lan.Summary("proposal")) if wicked_backend?
+      proposal_text
+    end
 
-        log.info "Returning from the network configuration with: #{result}"
-      ensure
-        Yast::Wizard.CloseDialog
+    def switch_backend_link
+      if wicked_backend?
+        info     = _("Using Wicked")
```
lslezak: Please add `TRANSLATORS: ` comment with some details, it might be difficult to translate properly without the context.

src/lib/network/clients/network_proposal.rb
```diff
@@ -24,24 +29,79 @@ def description
...
+        switcher = Hyperlink(SWITCH_TO_NETWORK_MANAGER, _("switch to Network Manager"))
```
lslezak: Please add `TRANSLATORS: ` comment with some details, the translators should know that this is a clickable link.  The same problem is also below.

teclator: NP: NetworkManager

lslezak:
BTW a nice trick to use a table in the description to show the screenshots side-by-side! :+1:

test/network_proposal_test.rb
```diff
@@ -20,27 +20,116 @@
...
+      allow(Yast::NetworkService).to receive(:wicked?).and_return(using_wicked)
+    end
+
+    it "returns a hash describing the proposal" do
+      expect(proposal).to include("label_proposal", "preformatted_proposal", "links")
     end
 
-    it "returns a map with 'preformatted_proposal' as an array with the network summary'" do
-      expect(subject.make_proposal({})["preformatted_proposal"]).to eql("rich_text_summary")
+    context "when using a Wicked backend" do
```
teclator: NP: it is not `a` wicked backend but the `wicked` backend, same below with NetworkManager

src/lib/network/clients/network_proposal.rb
```diff
@@ -24,24 +29,79 @@ def description
...
+      else
+        info     = _("Using Network Manager")
```
teclator: NP: Maybe would highlight the backend, but probably having the link after the info it is more than enough

```ruby
info = _("Using <b>%s</b>") % manager
```

teclator:
In general it looks good to me. Maybe we will show also some common configuration in the future, but the most important need is having the link to switch between network services, so :+1:.

src/lib/network/clients/network_proposal.rb
```diff
@@ -24,24 +29,79 @@ def description
...
+        switcher = Hyperlink(SWITCH_TO_WICKED, _("switch to Wicked"))
       end
 
-      { "workflow_sequence" => result }
+      "<ul><li>#{info} (#{switcher})</li></ul>"
+    end
+
+    def launch_network_configuration(args)
```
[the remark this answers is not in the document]
dgdavid: I usually use the extra spaces (empty lines) to ease the reading of method when there is related/unrelated "group of lines". Do you know what I mean? I also like to left an empty line before the return value.

That said, I'll to remove them if you think that it is a better approach, but it would be nice to achieve an agreement about that :)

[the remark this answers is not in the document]
lslezak: BTW the latest Rubocop has some new checks related to the spacing so maybe it will force a decision here. :wink:

PR #694, opened 2018-11-11
Expose the dhcp_ntp_servers method directly in Yast::Lan

teclator:
The method for retrieving `dhcp ntp servers` was initially added into [yast-caasp] repository(https://github.com/yast/yast-caasp) (**see https://github.com/yast/yast-caasp/pull/30**) but it was also needed in [yast-ntp-client](https://github.com/yast/yast-ntp-client) for SLES installation.

This PR adds that method into Yast::Lan following the DRY principle.

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -1024,7 +1024,7 @@ module Yast
       return [] if !NetworkService.isNetworkRunning || Yast::NetworkService.is_network_manager
 
       ReadWithCacheNoGUI()
-      Yast::LanItems.dhcp_ntp_servers.values.reduce(&:concat) || []
+      Yast::LanItems.dhcp_ntp_servers.values.flatten.uniq
     end
 
     publish variable: :ipv6, type: "boolean"
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -1013,6 +1013,20 @@ module Yast
       have_br
     end
 
+    # Provides a list with the NTP servers obtained via any of dhcp aware
+    # interfaces
+    #
+    # @note parsing dhcp ntp servers when NetworkManager is in use is not
+    #   supported yet (bsc#798886)
+    #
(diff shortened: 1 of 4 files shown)


test/lan_test.rb
```diff
@@ -444,3 +444,49 @@ def expect_modification_succeedes(modname, method)
...
+  context "when wicked is in use" do
+    let(:nm_enabled) { false }
+
+    it "reads the current network configuration" do
+      expect(Yast::Lan).to receive(:ReadWithCacheNoGUI)
+      subject.dhcp_ntp_servers
+    end
+
+    it "returns a list of the ntp_servers provided by dhcp " do
+      expect(Yast::LanItems).to receive(:dhcp_ntp_servers).and_return(servers)
```
dgdavid: NP: I prefer to use `allow` here, since it is not an expectation. Actually, you are "mocking" the response, isn't it?

src/modules/Lan.rb
```diff
@@ -1013,6 +1013,20 @@ def HaveXenBridge
...
+    #
+    # @note parsing dhcp ntp servers when NetworkManager is in use is not
+    #   supported yet (bsc#798886)
+    #
+    # @return [Array<String>] list of ntp servers obtained byg DHCP
+    def dhcp_ntp_servers
+      return [] if !NetworkService.isNetworkRunning || Yast::NetworkService.is_network_manager
+
+      ReadWithCacheNoGUI()
+      Yast::LanItems.dhcp_ntp_servers.values.reduce(&:concat) || []
```
dgdavid: NP: why not `...values.flatten` instead of `...values.reduce(&:concat)`?

teclator: I originally just moved the method from where it was implemented, and I'm fine with both and probably would even verify a uniqueness.

test/lan_test.rb
teclator: Yep, it should be an allow, and even could be moved to the before block.

PR #698, opened 2018-11-22
Merged SLE-12-SP3 into Sle 12 sp4

mchf:


package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Nov 22 11:54:59 UTC 2018 - mfilka@suse.com
+
+- bnc#110740
```
schubi2: Bug Nr. is unfortunately wrong

PR #708, opened 2018-12-06
YaST Security Audit Fixes: Calling External Commands and Absolute Paths and Shell-Escaped Arguments

shundhammer:
This is part of the YaST security audit by **@jsegitz**.

From the audit:

> yast2-network/yast2-network-4.1.6/src/include/network/routines.rb:680 
> and
> yast2-network/yast2-network-4.1.6/src/include/network/routines.rb:710
>
> RunAndRead and Run should be reworked to receive an absolut path and arguments that then can be quoted/escaped

## Reworking those Functions Completely for General Use

Completely reworking those functions would have meant to make using them a lot more awkward; right now it's simple and well-readable.

But both those functions are only used locally in that same file anyway, nowhere else, so that generic approach seemed very much overkill. But the _TO DO_ comment was extended as a reminder; if this is ever moved to one of our library packages like _yast-yast2_, this should clearly be more security-hardened, for example by passing the arguments separately and automatically quoting them all. But that makes the code a lot less readable

So the change done in this PR was to add absolute paths to all the commands and to shell-escape arguments.

## Using Absolute Paths

Intentionally **not** using paths like /usr/sbin, but plain /bin or /sbin instead because of certain peculiar things with those commands:

```shell
[shundhammer @ morgul] ~ % ls -l /sbin/if{up,down}
lrwxrwxrwx 1 root root 14 20. Nov 22:37 /sbin/ifdown -> /usr/sbin/ifup
lrwxrwxrwx 1 root root 14 20. Nov 22:37 /sbin/ifup -> /usr/sbin/ifup
```

So they both are symlinked to the same binary; `/sbin/ifdown eth0` can check its program name (`argv[0]`) what should be done, but what if we can only call `/usr/sbin/ifup` instead? Will the device go up or down? And what will somebody read in the logs to see what happened?

So, all commands here use the traditional `/sbin` or `/bin` paths, not their new `/usr/sbin` or `/usr/bin` paths; some for the reason above, the others for symmetry within this code.

## Quoting / Shell-Escaping Arguments

Most arguments to those few commands where those functions are used are fixed strings, so that is perfectly safe. In most cases, the kernel device name of a network device is used. It is very unlikely that such a device name contains characters that might be dangerous to not being quoted (blanks, shell wildcards, whatever), but it's always better to be safe than sorry.

With that Ruby [Shellwords.shellescape](https://docs.ruby-lang.org/en/2.5.0/Shellwords.html) function there is only a change if it actually contains such characters, otherwise the argument remains unchanged; in particular, no quotes are ever added.

### DIFF
--- a/src/include/network/routines.rb
+++ b/src/include/network/routines.rb
@@ -26,6 +26,9 @@
 # Summary:	Miscellaneous routines
 # Authors:	Michal Svec <msvec@suse.cz>
 #
+
+require "shellwords"
+
 module Yast
   module NetworkRoutinesInclude
     include I18n
@@ -674,10 +677,16 @@ module Yast
 
     # TODO: begin:
     # Following functions should be generalized and ported into yast-yast2
+    # and security-hardened in the process by making sure a command always uses
+    # an absolute path and all arguments are properly quoted (e.g. with
+    # shellescape()).
 
     # @param command [String] Shell command to run
     # @return Hash in form $[ "exit": <command-exit-status>, "output": [ <1st line>, <2nd line>, ... ] ]
     def RunAndRead(command)
+      if !command.lstrip.start_with?("/")
+        log.warn("Command does not have an absolute path: #{command}")
+      end
       ret = { "exit" => false, "output" => [] }
       result = Convert.to_map(SCR.Execute(path(".target.bash_output"), command))
       output = Ops.get_string(result, "stdout", "")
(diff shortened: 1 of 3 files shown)


src/include/network/routines.rb
```diff
@@ -674,10 +677,16 @@ def ReadHardware(hwtype)
...
     # TODO: begin:
     # Following functions should be generalized and ported into yast-yast2
+    # and security-hardened in the process by making sure a command always uses
+    # an absolute path and all arguments are properly quoted (e.g. with
+    # shellescape()).
 
     # @param command [String] Shell command to run
     # @return Hash in form $[ "exit": <command-exit-status>, "output": [ <1st line>, <2nd line>, ... ] ]
     def RunAndRead(command)
+      if !command.lstrip.start_with?("/")
```
jreidinger: what happens when ENV is used like `command = "LANG=C /usr/bin/command read"` ?

PR #709, opened 2018-12-07
Hardening

jreidinger:


src/include/network/lan/cmdline.rb
```diff
@@ -134,8 +137,8 @@ def ShowHandler(options)
           if key == Ops.get(options, "id", "0")
             # create plain text from formated HTML
             text = Builtins.sformat(
-              "echo \"%1\"|sed s/'<br>'/'\\n'/g|sed s/'<\\/li>'/'\\n'/g|sed s/'<[/a-z]*>'/''/g",
-              Ops.get_string(value, "rich_descr", "")
+              "/usr/bin/echo %1 | /usr/bin/sed 's/<br>/\\n/g;s/<\\/li>/\\n/g;s/<[/a-z]*>//g'",
```
mvidner: 1. `echo` is a shell builtin. Replacing it with `/usr/bin/echo` may produce slightly different results
2. but anyway, we should just use String#gsub, it will be easier than making the shell+sed magic safe
3. a unit test for "create plain text from formated HTML" would be great. Yes, this PR is getting out of hand already :D

src/lib/network/clients/save_network.rb
```diff
@@ -77,12 +79,12 @@ def CopyConfiguredNetworkFiles
...
         adjust_for_network_disks(file) if file.include?("ifcfg-")
 
-        copy_from = String.Quote(file)
-        copy_to = String.Quote(inst_dir + receipt[:dir])
+        copy_from = file
+        copy_to = inst_dir + receipt[:dir]
 
         log.info("Copying #{copy_from} to #{copy_to}")
 
-        cmd = "cp " << copy_from << " " << copy_to
```
mvidner: BTW the old code was wrong, `String.Quote` needs the caller to provide apostrophes around its result:

src/lib/network/clients/save_network.rb
```diff
@@ -77,12 +79,12 @@ def CopyConfiguredNetworkFiles
...
+        cmd = "cp " << copy_from.shellescape << " " << copy_to.shellescape
```
mvidner: Please use `+` or `#{}` to join strings, mutating with `<<` will break once Ruby enables frozen literals.

src/lib/network/clients/save_network.rb
```diff
@@ -99,7 +101,7 @@ def CopyConfiguredNetworkFiles
         # FIXME: this must be ripped out, refactored and tested
         # In particular, values containing slashes will break the last sed
         command = "\n" \
-          "source_file=#{source_file};dest_file=#{dest_file}\n" \
+          "source_file=#{source_file.shellescape};dest_file=#{dest_file.shellescape}\n" \
```
mvidner: Now `source_file` has the correct value...

src/lib/network/clients/save_network.rb
```diff
@@ -99,7 +101,7 @@ def CopyConfiguredNetworkFiles
...
           "grep -v \"^[[:space:]]*#\" $source_file | grep = | while read option\n" \
```
mvidner: ... but we're using it unquoted which will break. Use `\"$source_file\"` here, and quote *all* other shell variable expansions too ($key, $dest_file).

Proof:
```console
$ F="/tmp/one space"
$ touch "$F"
$ ls -l "$F"
-rw-r--r-- 1 mvidner users 0  7. pro 15.25 /tmp/one space
$ ruby -rshellwords -e 's = "/tmp/one space"; cmd = "f=#{s.shellescape}; ls -l $f; echo Retry; ls -l \"$f\""; system cmd'
ls: cannot access '/tmp/one': No such file or directory
ls: cannot access 'space': No such file or directory
Retry
-rw-r--r-- 1 mvidner users 0 Dec  7 15:25 '/tmp/one space'
```

mvidner: Given the FIXME above, it may be easier to rewrite this in plain Ruby.

src/lib/network/network_autoconfiguration.rb
```diff
@@ -177,7 +179,7 @@ def configured?(devname)
     #
     # active device <=> a device which is reported as "up" by wicked
     def active_config?(devname)
-      wicked_query = "wicked ifstatus --brief #{devname} | grep 'up$'"
+      wicked_query = "/usr/sbin/wicked ifstatus --brief #{devname.shellescape} | /usr/bin/grep 'up$'"
```
mvidner: require "shellwords"
is missing in this file

src/modules/DNS.rb
```diff
@@ -143,14 +144,12 @@ def GetHostnameFromGetent(line)
     # @param [String] ip given IP address
     # @return resolved canonical hostname (FQDN) for given IP or empty string in case of failure.
     def ResolveIP(ip)
-      command = "/usr/bin/getent hosts \"%1\""
-      getent = Convert.to_map(
-        SCR.Execute(path(".target.bash_output"), Builtins.sformat(command, ip))
-      )
+      command = Builtins.sformat("/usr/bin/getent hosts \"%1\"", ip.shellescape)
```
mvidner:
```suggestion
      command = Builtins.sformat("/usr/bin/getent hosts %1", ip.shellescape)
```

src/modules/DNS.rb
```diff
@@ -570,7 +569,7 @@ def update_hostname
       log.info("domain=#{@domain}")
 
       # change the hostname
-      SCR.Execute(path(".target.bash"), Ops.add("/bin/hostname ", @hostname))
+      SCR.Execute(path(".target.bash"), "/bin/hostname #{@hostname.shellwords}")
```
mvidner:
```suggestion
      SCR.Execute(path(".target.bash"), "/bin/hostname #{@hostname.shellescape}")
```

src/modules/Lan.rb
```diff
@@ -186,16 +184,12 @@ def isAnyInterfaceDown
...
-              SCR.Execute(
-                path(".target.bash_output"),
-                Builtins.sformat(
-                  "cat /sys/class/net/%1/address|tr -d '\n'",
-                  devname
-                )
-              ),
-              from: "any",
-              to:   "map <string, any>"
+            SCR.Execute(
```
mvidner: All this can be replaced with 
```rb
mac = File.read("/sys/class/net/#{devname}/address").chomp
```
YCP forever!

src/modules/NetHwDetection.rb
```diff
@@ -197,8 +198,8 @@ def DuplicateIP(ip)
       # floods logs.
       return false if ip.nil? || ip.empty?
 
-      command = "LC_ALL=C ip link show|grep BROADCAST|grep -v NOARP|cut -d: -f2"
-      exe = Convert.to_map(SCR.Execute(path(".target.bash_output"), command))
+      command = "LC_ALL=C /sbin/ip link show | /usr/bin/grep BROADCAST | /usr/bin/grep -v NOARP | /usr/bin/cut -d: -f2"
```
mvidner: `ip` is everywhere but `/usr/sbin/ip` is the real file

src/lib/network/clients/save_network.rb
jreidinger: yes, I already noticed it. That is reason why ideally I would like to use Yast::Execute which prevents it by default, but there is other problems with it ( like exception or popup ) which does not appear with that SCR one. So I worry if I start replacing it in network, I can easily break it quite lot.

src/lib/network/clients/save_network.rb
jreidinger: yep, I just worry it will get easily out of hands.

src/modules/NetHwDetection.rb
jreidinger: on my machine both exists, but I agree we should prefer /usr prefixed variant.

src/modules/Lan.rb
jreidinger: I expect we do not need to care that File is locally and SCR::Execute on target system as sys should be mount binded, right? BTW if we need to have chroot aware file, then this one can be used: https://github.com/yast/yast-yast2/blob/30a9253009a02a8fb0366c372fe82cdb746dfd40/library/system/src/lib/yast2/target_file.rb

src/lib/network/clients/save_network.rb
```diff
@@ -286,12 +288,12 @@ def configure_target
       # set proper network service
       set_network_service
 
-      SCR.Execute(path(".target.bash"), "chkconfig network on")
+      SCR.Execute(path(".target.bash"), "/usr/bin/chkconfig network on")
```
mchf: i hardly doubt that this is needed at all

jreidinger: well, goal of PBI and this PR is not to revisit code. But if you really believe we should remove this line, I can do it.

src/modules/DNS.rb
```diff
@@ -143,14 +144,12 @@ def GetHostnameFromGetent(line)
     # @param [String] ip given IP address
     # @return resolved canonical hostname (FQDN) for given IP or empty string in case of failure.
     def ResolveIP(ip)
-      command = "/usr/bin/getent hosts \"%1\""
-      getent = Convert.to_map(
-        SCR.Execute(path(".target.bash_output"), Builtins.sformat(command, ip))
-      )
+      command = Builtins.sformat("/usr/bin/getent hosts %1", ip.shellescape)
+      getent = SCR.Execute(path(".target.bash_output"), command)
```
mchf: well when already touching it, then modifying to
```
command = "/usr/bin/getent hosts #{ip.shellescape}"
```
makes more sense to me

src/lib/network/clients/save_network.rb
mchf: yes please. It is a left over from SysV init times which has no replacement in systemd era (in case of this service)

src/include/network/lan/cmdline.rb
jreidinger: @mvidner @mchf done at https://github.com/yast/yast-network/pull/709/commits/0bdd17af6bc8e762e2795700d74786bd9ba7a6e8

src/lib/network/clients/save_network.rb
jreidinger: I check what the code do and it in fact read and merge two sysconfig files. Maybe something for CFA, but it will be non-trivial change.

src/modules/LanItems.rb
```diff
@@ -2354,70 +2356,66 @@ def createS390Device(dev_attrs = {})
...
         end
         command1 = Builtins.sformat(
-          "qeth_configure %1 %2 %3 %4 %5 1",
+          "/sbin/qeth_configure %1 %2 %3 %4 %5 1",
           @options_param,
           @qeth_layer2 ? "-l" : "",
           @portname_param,
           @portnumber_param,
           @qeth_chanids
         )
```
mchf: couldn't all that escaping above replaced simply by ```command1 = <formating>.shellescape``` ?

mchf: similarly for the ```command2```

mchf: i mean somewhere after the ```case``` before execution.

src/modules/LanItems.rb
```diff
@@ -2667,16 +2665,12 @@ def export_udevs(devices)
...
-                "driver=$(ls -l /sys/class/net/%1/device/driver);echo ${driver##*/}|tr -d '\n'",
-                device
-              )
-            ),
-            from: "any",
-            to:   "map <string, any>"
+          driver = SCR.Execute(
+            path(".target.bash_output"),
+            Builtins.sformat(
+              "driver=$(/usr/bin/ls -l /sys/class/net/%1/device/driver); /usr/bin/echo ${driver##*/} | /usr/bin/tr -d '\n'",
```
mchf: i think that ```echo``` and ```/usr/bin/echo``` are different things ... and I don't know if the later one supports bashisms like ```${...##*/}```

src/modules/LanItems.rb
```diff
@@ -2717,45 +2711,31 @@ def export_udevs(devices)
...
-            0
+          port_name = SCR.Execute(
+            path(".target.bash_output"),
+            Builtins.sformat(
+              "/usr/bin/cat /sys/class/net/%1/device/portname | /usr/bin/tr -d '\n'",
+              device.shellescape
+            )
           )
+          if !port_name["stdout"].empty?
             portname = String.CutBlanks(Ops.get_string(port_name, "stdout", ""))
```
mchf: when already touching, what about ```strip```?

src/modules/LanItems.rb
```diff
@@ -2770,22 +2750,15 @@ def export_udevs(devices)
...
-                "port0=$(ls -l /sys/class/net/%1/device/cdev0);echo ${port0##*/}|tr -d '\n'",
-                device
-              )
-            ),
-            from: "any",
-            to:   "map <string, any>"
+          port0 = SCR.Execute(
+            path(".target.bash_output"),
+            Builtins.sformat(
+              "port0=$(/usr/bin/ls -l /sys/class/net/%1/device/cdev0); /usr/bin/echo ${port0##*/} | /usr/bin/tr -d '\n'",
```
mchf: again the ```echo``` question

src/modules/LanItems.rb
mchf: well bad example, expansion is done by shell ... but anyway. Is there any known difference between ```echo``` and ```/usr/bin/echo```?

src/modules/NetHwDetection.rb
mchf: in fact ```/bin/ip``` and ```/sbin/ip``` are just symlinks ...

src/modules/LanItems.rb
jreidinger: I worry not, because we no longer know what is param and what is space in original variable.

src/modules/LanItems.rb
jreidinger: I am not much aware of it.

src/modules/LanItems.rb
```diff
@@ -2704,8 +2698,8 @@ def export_udevs(devices)
             SCR.Execute(
               path(".target.bash_output"),
               Builtins.sformat(
-                "for i in $(seq 0 2);do chanid=$(ls -l /sys/class/net/%1/device/cdev$i);echo ${chanid##*/};done|tr '\n' ' '",
-                device
+                "for i in $(seq 0 2); do chanid=$(/usr/bin/ls -l /sys/class/net/%1/device/cdev$i); /usr/bin/echo ${chanid##*/}; done | /usr/bin/tr '\n' ' '",
```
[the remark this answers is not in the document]
jreidinger: again variable evaluated by shell, not echo itself.

src/modules/LanItems.rb
jreidinger: fine for me.

src/modules/LanItems.rb
jreidinger: and same reply. BTW that shell formatting is hell.

src/lib/network/clients/save_network.rb
jreidinger: OK, I address this in https://github.com/yast/yast-network/pull/709/commits/ac10b0988707aae95591d88ef70e1e22f6c69c16 which I hope is easy enough to read and should be much smarter then now.

jreidinger:
I am not sure why in travis CFA failing. on my system rake test:unit and even osc:build works

src/lib/network/clients/save_network.rb
mvidner: Great!

src/include/network/lan/cmdline.rb
```diff
@@ -120,30 +123,23 @@ def getItem(options, config)
...
-              to:   "map <string, any>"
-            )
-            CommandLine.Print(Ops.get_string(descr, "stdout", ""))
-          end
+      return false unless validateId(options, config)
+
+      required_id = options["id"]
+      config.each do |row|
+        row.each_pair do |key, value|
+          next if key != required_id
```
mchf: well, ```select``` might fit better. But that is just a NP ;-)

src/modules/LanItems.rb
mchf: ok

src/modules/LanItems.rb
mchf: I found this: https://askubuntu.com/questions/960822/why-is-there-a-bin-echo-and-why-would-i-want-to-use-it

btw shell formatting is hell but not that deep as used to be formatting in ycp ;-)

mchf:
> I am not sure why in travis CFA failing. on my system rake test:unit and even osc:build works

Augeas parsing error: Get did not match entire input at /usr/src/app/test/data/dhcp.original:12:0, lens /usr/share/augeas/lenses/dist/sysconfig.aug:65.12-.71:

sounds like a lens issue - old augeas in the docker image?

src/modules/LanItems.rb
```diff
@@ -2667,125 +2665,64 @@ def export_udevs(devices)
...
-            from: "any",
-            to:   "map <string, any>"
-          )
+          begin
+            driver = begin
+                       File.readlink("/sys/class/net/#{device}/device/driver")
+                     rescue
+                       nil
+                     end
+          rescue Errno => e
```
mvidner: `rescue Errno` does not work. Please test. But why are you adding two nested rescues??

src/include/network/lan/cmdline.rb
```diff
@@ -120,30 +123,23 @@ def getItem(options, config)
...
+
+          # create plain text from formated HTML
```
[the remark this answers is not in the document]
jreidinger: problem is that it is quite specific code, that is not generic html to plain text. IT handles specifically summary which use only br and list.

src/modules/LanItems.rb
jreidinger: @mvidner my fault. I am testing it in irb, that it throws Errno::<something> and I do not test that Errno is common base.

jreidinger: nested rescue was mistake in approach where are try two possible solutions and does not clean it enough.

jreidinger:
merge waiting to finally solve staging.

jreidinger:
@mvidner PR is waiting till staging is finally merged to not delay it anymore.

PR #711, opened 2018-12-12
/etc/hosts load speed up

mchf:
bnc#1116568

See https://trello.com/c/mbJyRV1V/500-tw-1116568-p2-yast2-hostnames-can-freeze-os

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Jan  4 09:46:06 UTC 2019 - mfilka@suse.com
+
+- bnc#1116568
+  - /etc/hosts load performance improvements
+- 4.1.27
+
+-------------------------------------------------------------------
 Wed Jan  2 17:17:58 CET 2019 - schubi@suse.de
 
 - Showing correct start mode for nfsroot device (bsc#1105692).
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -344,9 +344,9 @@ module Yast
       end
 
       ret = nil
+      host = nil
 
       loop do
-        host = nil
         ret = UI.UserInput
         break if ret != :ok
 
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -175,7 +175,7 @@ module Yast
 
       ret = nil
       modified = false
-      while ![:abort, :cancel, :back, :next].include?(ret) do
+      until [:abort, :cancel, :back, :next].include?(ret)
         UI.ChangeWidget(Id(:edit), :Enabled, table_items.any?)
         UI.ChangeWidget(Id(:delete), :Enabled, table_items.any?)
 
@@ -212,11 +212,6 @@ module Yast
 
           table_items = Builtins.maplist(table_items) do |e|
             if cur == Ops.get_integer(e, [0, 0], -1)
-              oldentry = Builtins.mergestring(
-                [Ops.get_string(olditem, 2, ""), Ops.get_string(olditem, 3, "")],
-                " "
-              )
-
               ip = Ops.get_string(item, 1, "")
               oldip = Ops.get_string(olditem, 1, "")
 
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -88,7 +88,7 @@ module Yast
       hosts = Host.name_map
 
       # make ui items from the hosts list
-      table_items = hosts.map do |host, names|
+      hosts.each do |host, names|
         if names.empty?
           log.error("Invalid host: %1, (%2)", host, names)
           next
@@ -96,12 +96,13 @@ module Yast
 
         name, *aliases = names.first.split(/\s/).delete_if(&:empty?)
 
-        Item(
+        item = Item(
           Id(table_items.size),
           host,
           Punycode.DecodeDomainName(name),
           Punycode.DecodePunycodes([aliases.join(" ")]).first || ""
         )
+        table_items.push(item)
       end
 
       # Hosts dialog contents
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -30,6 +30,8 @@
 # Hosts configuration dialogs
 module Yast
   module NetworkServicesHostInclude
+    include Logger
+
     def initialize_network_services_host(include_target)
       Yast.import "UI"
 
@@ -85,12 +87,10 @@ module Yast
       deleted_items = []
       hosts = Host.name_map
 
-      Builtins.y2debug("hosts=%1", hosts)
-
       # make ui items from the hosts list
       table_items = hosts.map do |host, names|
         if names.empty?
-          Builtins.y2error("Invalid host: %1, (%2)", host, names)
+          log.error("Invalid host: %1, (%2)", host, names)
           next
         end
 
@@ -103,7 +103,6 @@ module Yast
           Punycode.DecodePunycodes([aliases.join(" ")]).first || ""
         )
       end
-      Builtins.y2debug("table_items=%1", table_items)
 
       # Hosts dialog contents
       contents = HBox(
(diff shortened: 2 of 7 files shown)


src/include/network/services/host.rb
```diff
@@ -81,33 +81,29 @@ def HostsMainDialog(standalone)
...
       hosts = Host.name_map
 
       Builtins.y2debug("hosts=%1", hosts)
 
       # make ui items from the hosts list
-      Builtins.maplist(hosts) do |host, names|
-        if Ops.less_than(Builtins.size(names), 1)
+      table_items = hosts.map do |host, names|
+        if names.empty?
           Builtins.y2error("Invalid host: %1, (%2)", host, names)
```
dgdavid: Please, contribute to kill those deprecated `Builtins` using `Yast::Logger` instead :)

src/include/network/services/host.rb
```diff
@@ -175,36 +171,30 @@ def HostsMainDialog(standalone)
       end
 
       UI.ChangeWidget(Id(:table), :Items, table_items)
-      UI.SetFocus(Id(:table)) if Ops.greater_than(Builtins.size(table_items), 0)
+      UI.SetFocus(Id(:table)) if !table_items.empty?
```
dgdavid: Better to use `any?`

```suggestion
      UI.SetFocus(Id(:table)) if table_items.any?
```

src/include/network/services/host.rb
```diff
@@ -175,36 +171,30 @@ def HostsMainDialog(standalone)
...
-          :Enabled,
-          Ops.greater_than(Builtins.size(table_items), 0)
-        )
-        UI.ChangeWidget(
-          Id(:delete),
-          :Enabled,
-          Ops.greater_than(Builtins.size(table_items), 0)
-        )
+        UI.ChangeWidget(Id(:edit), :Enabled, !table_items.empty?)
+        UI.ChangeWidget(Id(:delete), :Enabled, !table_items.empty?)
```
dgdavid: The same as above, use `if ...any?` instead.

src/include/network/services/host.rb
```diff
@@ -175,36 +171,30 @@ def HostsMainDialog(standalone)
...
 
         ret = UI.UserInput
-        Builtins.y2debug("ret=%1", ret)
 
         # abort?
         if ret == :abort || ret == :cancel
           ReallyAbortCond(Host.GetModified) ? break : next
         # add host
         elsif ret == :add
+          max = table_items.size
```
dgdavid: NP: `max` looks as a weird name to me for this variable. What about `index`, `position`, `item_position`...?

src/include/network/services/host.rb
```diff
@@ -175,36 +171,30 @@ def HostsMainDialog(standalone)
...
           item = HostDialog(max, term(:empty))
+
           next if item.nil?
-          table_items = Builtins.add(table_items, item)
+
+          table_items = table_items.push(item)
```
dgdavid: This (re)assignation is not needed ;) `Array#push` directly appends the item to the array.

src/modules/Host.rb
```diff
@@ -362,8 +358,7 @@ def load_hosts
     @hosts.load
 
     # save hosts to check for changes later
-    @hosts_init = CFA::Hosts.new
-    @hosts_init.load
+    @hosts_init = @hosts.clone
```
dgdavid: Ok, if you decide to keep this, what about...

* ... to choose a more meaningful name? `previous_hosts_entries`? `initial_hosts`?
* ... to finally delete the `@modified` occurrences, in order to have a single point to check if `hosts` were modified?

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Dec 12 10:44:17 UTC 2018 - mfilka@suse.com
+
+- bnc#1116568
+  - /etc/hosts load performance improvements
```
dgdavid: NP: Usually our team mates encourage me to write less technical log entries. So, I guess something like `Improve the performance loading the /etc/hosts, avoiding to crash when dealing with quite long files (bsc#...)` could be better. But, it is only a suggestion. Remember that, sadly, English is not one of my strengths :wink:

src/modules/Host.rb
dgdavid: NP: BTW, this change probably is out of the scope of this commit (Refactored Host::Summary), isn't it?

mchf: it is needed. it is used for checking modifications (and only there). If other changes didn't speed things enough I'd move it into ```Modified``` method, but it would slow down this method and it would be only moving the problem elsewhere.

src/include/network/services/host.rb
mchf: ;-) there is so many ```Builtins``` that I never know when to stop ;-)

src/include/network/services/host.rb
dgdavid: I must admit that I say "better to use" meaning "better to read". That said, if you know more about the performance, there is a [bit old discussion about it](https://github.com/JuanitoFatas/fast-ruby/pull/49) in the https://github.com/JuanitoFatas/fast-ruby repo .

Below, you can find three [`compare!`](https://www.rubydoc.info/gems/benchmark-ips/Benchmark/Compare) samples from three different execution of a slightly modified [benchmark of the next-to-last comment](https://github.com/JuanitoFatas/fast-ruby/pull/49#issuecomment-128444354) (to use `ary.any?` instead of `ary.any? { true }`), using `ruby 2.5.3p105`

```
Comparison:
         !ary.empty?:  3408896.0 i/s
       ary.size != 0:  3294235.0 i/s - same-ish: difference falls within error
            ary.any?:  2937220.1 i/s - same-ish: difference falls within error
     !ary.size.zero?:  2871812.4 i/s - same-ish: difference falls within error
           ary != []:  2005702.0 i/s - 1.70x  slower
 !!ary.size.nonzero?:  1305874.0 i/s - 2.61x  slower
```

```
Comparison:
         !ary.empty?:  3346054.4 i/s
       ary.size != 0:  3161546.0 i/s - same-ish: difference falls within error
            ary.any?:  2999996.7 i/s - 1.12x  slower
     !ary.size.zero?:  2823247.6 i/s - 1.19x  slower
           ary != []:  1974117.8 i/s - 1.69x  slower
 !!ary.size.nonzero?:  1278290.9 i/s - 2.62x  slower

```

```
Comparison:
         !ary.empty?:  3366606.0 i/s
       ary.size != 0:  3169765.1 i/s - same-ish: difference falls within error
            ary.any?:  3068372.1 i/s - same-ish: difference falls within error
     !ary.size.zero?:  2870325.6 i/s - 1.17x  slower
           ary != []:  2124715.2 i/s - 1.58x  slower
 !!ary.size.nonzero?:  1350402.9 i/s - 2.49x  slower
```

I hope this information could be useful to help you make a decision.

mchf: huh. very exhaustive, thank you.

src/modules/Host.rb
mchf: refactoring ```Host::Summary``` is not out of scope ... it is used when using yast CLI

mchf:
@jreidinger I did some update in CFA part too. Could you pls review too as you are author of this part? Thx

src/lib/cfa/hosts.rb
```diff
@@ -18,6 +29,11 @@ def initialize(file_handler: nil)
       super(AugeasParser.new("hosts.lns"), PATH, file_handler: file_handler)
     end
 
+    def initialize_clone(source)
+      # Cloning AugeasTree
+      @data = source.data.clone
```
jreidinger: this won't work for more deeply nested trees. In general I recommend to not redefine it here. Also maybe it fits better in CFA itself. And last but not least in general there can be problem if both clones writes, as it shared parser cache. ( but  your usecase is to only compare read-only copy and modified one, so it should be fine ).

btw name clone is a bit confusing as ruby clone is also just one level. So I recommend to name it also deep_clone similar to deep_copy. See https://medium.com/@raycent/ruby-clone-vs-dup-8a49b295f29a

mchf: that is why I've defined ```initialize_clone``` hook even for ```AugeasTree``` [here](https://github.com/yast/yast-network/pull/711/files/9f5243baf40b0e664ec9343bbed971349b071727#diff-2412af142fac00805b68b41fae60a699R10)

regarding renaming. I do not understand here. AFAIK ```initialize_clone``` is standard ruby hook so I can hardly choose its name. Isn't it so?

jreidinger: ah, so it using hook. Then ok. I do not know about that ruby hook for cloning. So fine for me.

src/lib/cfa/hosts.rb
```diff
@@ -6,6 +6,17 @@
 require "cfa/augeas_parser"
 
 module CFA
+  class AugeasTree
```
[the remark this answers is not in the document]
mchf: done

src/lib/cfa/hosts.rb
mchf: thanks

mchf:
@dgdavid I completely agree there is lot of places which can be refactored ... as in the whole yast2-network. I'll most probably add some simple improvements before merging, however I cannot refactor it completely. So, something more next time ;-)

PR #712, opened 2018-12-18
Fixed copying ifcfgs to target

mchf:
bnc#1119723


mvidner:
BTW a ["receipt"](https://en.wiktionary.org/wiki/receipt) means "účtenka". ["recipe"](https://en.wiktionary.org/wiki/recipe) is "recept"

PR #715, opened 2019-01-04
Added needed method GetDHCPHostnameIP again

schubi2:


src/modules/DNS.rb
```diff
@@ -519,6 +519,27 @@ def IsHostLocal(check_host)
 
   private
 
+    # Get current hostname and IP Address
+    # if these are set by DHCP
+    # @return map with ip, hostname_short and hostname_fq keys
+    def GetDHCPHostnameIP
```
dgdavid: Since it is a private method, it could be renamed in a more Ruby style. Something like `dhcp_data` sounds good for me.

test/dns_test.rb
```diff
@@ -106,5 +106,31 @@ module Yast
         end
       end
     end
+
+    describe ".IsHostLocal" do
```
dgdavid: IMHO, this test is a bit incomplete.

See [this gist](https://gist.github.com/dgdavid/451c11d81211d8f98470cc35b08155f7) with a tested suggestion to improve it a little, which probably will need some changes if the chainable `stdout` method is finally used. [Here there is a modified version](https://gist.github.com/dgdavid/791127e83c4c9c7f1ff169c0941efa21).

schubi2:
"I would like to propose a few changes that I leave as comments of this review. Of course, they are completely optional as the main objective of PR is already done and it is not my intention to block your work :) "
You do not have to apologize for your better code :-) I have taken everything. Thanks, for your suggestions !

PR #717, opened 2019-01-10
Write ETHERDEVICE into vlan config

mchf:
bnc#1111483

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -754,7 +754,6 @@ module Yast
 
     def AddNew
       @current = @Items.to_h.size
-      @Items[@current] = { "commited" => false }
       @operation = :add
 
       nil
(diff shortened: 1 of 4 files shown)


src/modules/LanItems.rb
```diff
@@ -1148,19 +1152,19 @@ def Read
...
+        items.each do |key, value|
           match = value.fetch("hwinfo", {}).fetch("dev_name", "") == confname
-          @Items[key]["ifcfg"] = confname if match
+          items[key]["ifcfg"] = confname if match
         end
       end
 
       interfaces.each do |confname|
-        next if @Items.keys.any? { |key| @Items.fetch(key, {}).fetch("ifcfg", "") == confname }
+        next if items.keys.any? { |key| items.fetch(key, {}).fetch("ifcfg", "") == confname }
```
jreidinger: this looks strange to use keys, when in fact you use key only to accept value. What about
```
next if items.values.any? { |item| item && item["ifcfg"] == confname }
```

jreidinger: but just NP

mchf: a consequence of an earlier more less direct rewrite from ycp -> ruby ... going to update ;-)

PR #721, opened 2019-01-22
Fixed internal error in cloning /etc/hosts 

mchf:
bnc#112251


mvidner:
I see, `Host.SetModified` was dropped in #711.

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 22 09:27:44 UTC 2019 - mfilka@suse.com
+
+- bnc#112251
```
mvidner: one digit out of is missing here

PR #722, opened 2019-01-23
Fixed internal error when modifying /etc/hosts

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1122387

### DIFF
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -249,15 +249,14 @@ module Yast
           UI.ChangeWidget(Id(:table), :Items, table_items)
           modified = true
         elsif ret == :next
-          # check_
           next if !modified
           Host.clear
 
           table_items.each do |row|
-            encoded_aliases = Punycode.EncodePunycodes([row.fetch(3, "")])
-            encoded_canonical = Punycode.EncodeDomainName(row.fetch(2, ""))
+            encoded_aliases = Punycode.EncodePunycodes([row[3] || ""])
+            encoded_canonical = Punycode.EncodeDomainName(row[2] || "")
             value = encoded_canonical + " " + encoded_aliases.join(" ")
-            key = row.fetch(1, "")
+            key = row[1] || ""
 
             Host.add_name(key, value)
           end
--- a/src/include/network/services/host.rb
+++ b/src/include/network/services/host.rb
@@ -69,6 +69,23 @@ module Yast
       Popup.AnyQuestion("", text, Label.YesButton, Label.NoButton, :focus_no)
     end
 
+    # Encode canonical name and aliases using puny code
+    #
+    # @param canonical [String] canonical hostname
+    # @param aliases [Array<String>] space separated list of aliases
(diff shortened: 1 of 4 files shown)


src/include/network/services/host.rb
```diff
@@ -249,15 +249,14 @@ def HostsMainDialog(standalone)
...
           modified = true
         elsif ret == :next
-          # check_
           next if !modified
           Host.clear
 
           table_items.each do |row|
-            encoded_aliases = Punycode.EncodePunycodes([row.fetch(3, "")])
-            encoded_canonical = Punycode.EncodeDomainName(row.fetch(2, ""))
+            encoded_aliases = Punycode.EncodePunycodes([row[3] || ""])
```
dgdavid: Change seems to make sense, but I really miss a test. Could you please add a simple one?

BTW, if `Punycode.EncodeXYZ` accepts `nil`, you also could write the code like

```ruby
table_items.each do |key, canonical, alias|
  .  .  .
end
```

or even avoid the `nil` values

```ruby
table_items.each do |key = '', canonical = '', alias = ''|
  .  .  . 
end
```

mchf: well adding simple test  is not that simple here ... the method is spread over 250 lines and is something like all-in-one dialog handler.

test/host_service_test.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      result = subject.encode_hosts_line(canonical, aliases.split)
+
+      expect(result).to eql "xn--ala-qma83eb.xn--jlinka-3mb.go.home xn--ala-qma83eb xn--jlinka-3mb earthworm"
+    end
+
+    it "returns empty string when invalid arguments were passed" do
+      result = subject.encode_hosts_line(nil, nil)
+
+      expect(result).to be_empty
```
dgdavid: According to the example description, why not use the below expectation?

```ruby
expect(result).to eq("")
```

src/include/network/services/host.rb
```diff
@@ -253,12 +270,9 @@ def HostsMainDialog(standalone)
           Host.clear
 
           table_items.each do |row|
-            encoded_aliases = Punycode.EncodePunycodes([row[3] || ""])
-            encoded_canonical = Punycode.EncodeDomainName(row[2] || "")
-            value = encoded_canonical + " " + encoded_aliases.join(" ")
-            key = row[1] || ""
+            hostnames = encode_hosts_line(row[2], (row[3] || "").split)
```
dgdavid: I still finding it difficult to understand.  I am going to come back with my previous suggestion 

```ruby

table_items.each do |key = '', canonical = '', aliases = [] |
  hostnames = encoded_host_line(canonical, aliases)

  Host.add_name(key, )
end
```

What do you think?

mchf: i don't think that this would work without further intervention in the code (```table_items``` is an array of something what used to be Term ```Item``` in YCP times). I think I was intrusive enough for P1 hotfix ;-)

PR #723, opened 2019-01-29
Fixed internal error when no hostname is provided in device static configuration

mchf:
bnc#1123102

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1129,7 +1129,7 @@ module Yast
     end
 
     def address_tab
-      # in such case ifcfg is not existing and /sys might not contain
+      # in case that ifcfg does not exist, /sys might not cointain
       # any device info (especially for virtual devices like vlan)
       # @type variable is already initialized by @see HardwareDialog
       # resp its storage handler @see storeHW
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1129,7 +1129,12 @@ module Yast
     end
 
     def address_tab
-      type = LanItems.GetCurrentType
+      # in such case ifcfg is not existing and /sys might not contain
+      # any device info (especially for virtual devices like vlan)
+      # @type variable is already initialized by @see HardwareDialog
+      # resp its storage handler @see storeHW
+      type = LanItems.type
+
       drvtype = DriverType(type)
       is_ptp = drvtype == "ctc" || drvtype == "iucv"
       # TODO: dynamic for dummy. or add dummy from outside?
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1521,7 +1521,7 @@ module Yast
       names = Host.names(LanItems.ipaddr).first
       Host.remove_ip(LanItems.ipaddr)
 
-      if ip_changed && !hostname_changed
+      if ip_changed && !hostname_changed && !names.nil?
         log.info("Dropping record for #{LanItems.ipaddr} from /etc/hosts")
 
         Host.add_name(ipaddr, names)
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1350,8 +1350,6 @@ module Yast
 
       Builtins.y2milestone("ShowAndRun: %1", ret)
 
-      LanItems.Rollback if ret == :abort
-
       if ret != :back && ret != :abort
         # general tab
         LanItems.startmode = Ops.get_string(@settings, "STARTMODE", "")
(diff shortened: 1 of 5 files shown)


src/include/network/lan/address.rb
```diff
@@ -1515,7 +1518,7 @@ def update_hostname(ipaddr, hostname)
       initial_hostname = initial_hostname(LanItems.ipaddr)
       hostname_changed = initial_hostname != hostname
 
-      return if !(ip_changed || hostname_changed || hostname.empty?)
+      return if !(ip_changed || hostname_changed) || hostname.empty?
```
teclator: It respect the method description? I mean, if you have a previous `hostname` and you clear it, the hostname is not updated / removed from /etc/hosts?

mchf: good point ... I've to make the condition even more strict.

src/modules/LanItems.rb
```diff
@@ -284,11 +284,6 @@ def GetCurrentName
     #
     # @param itemId [Integer] a key for {#Items}
     def GetDeviceType(itemId)
-      # in such case ifcfg is not existing and /sys might not contain
-      # any device info (especially for virtual devices like vlan)
-      # @type variable is already initialized by @see HardwareDialog
-      # resp its storage handler @see storeHW
-      return @type if @operation == :add
```
teclator: Removing / moving this line is safe enough?, there is no some corner case / leftover when adding a new interface which still uses GetDeviceType? If that is the case then it LGTM

mchf: yes. the line was moved closer to place where it is needed to free "generic" method

test/address_test.rb
```diff
@@ -125,6 +125,17 @@ def initialize
 
       expect(Yast::Host.names(new_ip)).to eql original_names
     end
+
+    it "do not crash when ip was changed, no hostnames exist and the hostname is empty" do
```
teclator: does

test/address_test.rb
```diff
@@ -125,6 +125,17 @@ def initialize
...
+      new_ip = "2.2.2.2"
+
+      allow(Yast::Host)
+        .to receive(:names)
+        .and_return([])
+
+      subject.send(:update_hostname, new_ip, "")
+      expect(Yast::Host.names(new_ip)).to eql []
```
teclator: We are expecting the same that we are allowing, so it looks like a wrong unit test :P

mchf: hi hi forgot to save last change ... updating soon

test/address_test.rb
teclator: It is true that it does not crash now, but I would use contexts explaining the different scenarios and what does each one:

```ruby

context "when the ifcfg ipaddr has changed" do
  context "but the associated hostname not" do
   ...
  end
  context "and the hostname as well" do
  end
end
context "when neither the ifcpfg ipaddr nor associated hostnames have changed" do
  it "does not touch the hosts file"
    expect(Host).to not_receive(:names)
  end
end
```
...

src/modules/LanItems.rb
```diff
@@ -2177,6 +2172,8 @@ def Commit
     # Remove a half-configured item.
     # @return [true] so that this can be used for the :abort callback
     def Rollback
+      return true if LanItems.operation != :add
```
teclator: I was testing it, and if you edit a not configured interface and then :abort it currently crash

mchf: hm, that is close to use case which i was fixing. I'll try even this one. thx

teclator: You can split it in different pull requests or, in case you decide to fix also this issue as part of this pull request, then cover also the scenario with unit tests.

mchf: seems that subsequent commit fixed even "my case" so dropping this change is enough. It is nice how is networking code straightforward and robust ;-) thanks for co-testing

src/include/network/lan/address.rb
```diff
@@ -1129,7 +1129,12 @@ def general_tab
     end
 
     def address_tab
-      type = LanItems.GetCurrentType
+      # in such case ifcfg is not existing and /sys might not contain
```
teclator:
```suggestion
      # in case that ifcfg does not exist, /sys might not cointain
```

src/include/network/lan/address.rb
```diff
@@ -1350,8 +1355,6 @@ def AddressDialog
 
       Builtins.y2milestone("ShowAndRun: %1", ret)
 
-      LanItems.Rollback if ret == :abort
```
teclator: So, why it is not needed now but was in the past? Was you trying to fix the removal of a new virtual interface when **editing**  it after has been added or it is a different scenario?.

mchf: i'm not sure why it was needed before. However calling rollback is defined as abort callback and according to the logs it was called twice. So I dropped it.

PR #724, opened 2019-02-06
Fix reading

teclator:
- It seems that in the past some errors were skipped when not able to read some device files, but since #709 some bugs have been reported [1]

[1]: https://bugzilla.suse.com/show_bug.cgi?id=1121087 & https://bugzilla.suse.com/show_bug.cgi?id=1124002

This PR just tries to fix the [bsc#1124002](https://bugzilla.suse.com/show_bug.cgi?id=1124002) initializing to an empty string the device `portname` and `protocol` when not able to read them.

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -2698,9 +2698,18 @@ module Yast
           if !chan_ids["stdout"].empty?
             chanids = String.CutBlanks(Ops.get_string(chan_ids, "stdout", ""))
           end
+
           # we already know that kernel device exist, otherwise next above would apply
-          portname = ::File.read("/sys/class/net/#{device}/device/portname").strip
-          protocol = ::File.read("/sys/class/net/#{device}/device/protocol").strip
+          # FIXME: It seems that it is not always the case (bsc#1124002)
+          begin
+            portname = ::File.read("/sys/class/net/#{device}/device/portname").strip
+            protocol = ::File.read("/sys/class/net/#{device}/device/protocol").strip
+          rescue SystemCallError => e
+            log.error("Failed to read portname or protocol: #{e.inspect}")
+            portname = ""
+            protocol = ""
+          end
+
           layer2_ret = SCR.Execute(
             path(".target.bash"),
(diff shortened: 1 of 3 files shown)


src/modules/LanItems.rb
```diff
@@ -2698,9 +2698,18 @@ def export_udevs(devices)
...
             chanids = String.CutBlanks(Ops.get_string(chan_ids, "stdout", ""))
           end
+
           # we already know that kernel device exist, otherwise next above would apply
-          portname = ::File.read("/sys/class/net/#{device}/device/portname").strip
-          protocol = ::File.read("/sys/class/net/#{device}/device/protocol").strip
+          # FIXME: It seems that it is not always the case (bsc#1124002)
+          begin
+            portname = ::File.read("/sys/class/net/#{device}/device/portname").strip
+            protocol = ::File.read("/sys/class/net/#{device}/device/protocol").strip
```
lslezak: In the bug it fails when reading `/sys/class/net/eth0/device/protocol` file. That means the `/sys/class/net/eth0/device/portname` was read properly. So it seems only some files might be missing, not both at the same time.

So I'd prefer something like 
```ruby
            portname_file = "/sys/class/net/#{device}/device/portname"
            portname = File.read(portname_file).strip if File.exist?(portname_file)

            protocol_file = "/sys/class/net/#{device}/device/protocol"
            protocol = File.read(protocol_file).strip if File.exist?(protocol_file)
```

lslezak: Ah, that would result in `nil`, you need to do
```ruby
portname_file = "/sys/class/net/#{device}/device/portname"
portname = File.exist?(portname_file) ? File.read(portname_file).strip : ""
```

teclator: Ok, so do you prefer to rescue the exception or just check if it exist before reading?

lslezak: For separate checks I'd prefer `File.exist?` as written above (to be consistent with the other YaST code). The exception would be better for the grouped case.

PR #725, opened 2019-02-12
Update device routes when exist and the interface name has changed

teclator:
**NOTE TO REVIEWERS:** Bump version & changelog is pending.

## Problem

When a interface name is modified, the change is not propagated through other objects than reference the previous name.

It will be nice to have a set of callbacks that updates dependent objects when the interface is modified or deleted, although in this PR we will only face the update of Routes when the interface name has changed.

- *https://trello.com/c/zgnOtxLT/657-3-1060207-sle-15-alpha5installer-network-internal-error-while-after-renaming-interface-new-name-is-not-not-used*

**Note** The change of the interface name is done in an atomic way, that is, if we cancel the edition of the interface the change is not reverted and the interface name will be preserved.

## Solution

1. When a interface name is modified through the udev rules dialog, we check if there are some routes referencing the previous name, in that case we offer the user to update them with the new name.

2. The **Routing devices list** are also updated for being selected in the routing dialog, when?:
    - when the interface name change is done
    - when the interface config is deleted
    - when an interface configuration is confirmed (specially for new interfaces without hwinfo)

3. The Rollback will not drop configured items without hwinfo (that is when editing an already added config)

## Testing

- *Unit test in EditNicName class*
- *Tested manually*

## Inform the user that there exist some routes that must be updated

informtheuser

## Update the routes and the routing devices

updateroutesanddevices

## Update the list of devices when a new one is added, a not configured one is configured or an interface is deleted.

| Description | Before | After |
| --------------- | --------- | --------- |
|  Configure Interface that was not configured | devicesnotupdated | devicelistupdated |

*If the fix affects the UI attach some screenshots here.*


src/modules/LanItems.rb
```diff
@@ -276,6 +276,14 @@ def GetDeviceNames(items)
       items.map { |itemId| GetDeviceName(itemId) }.reject(&:empty?)
     end
 
+    def current_name
+      renamed?(@current) ? renamed_to(@current) : GetDeviceName(@current)
+    end
+
+    def current_device_names
+      GetNetcardInterfaces().map { |i| renamed?(i) ? renamed_to(i) : GetDeviceName(i) }.reject(&:empty?)
```
dgdavid: NP: A bit difficult to read a first glance (for me)

Even more, you are using the same logic twice

```ruby
 renamed?(i) ? renamed_to(i) : GetDeviceName(i)
```

Why not 

```ruby

...

private

...

def name_for(device)
  renamed?(device) ? renamed_to(device) : GetDeviceName(device)
end
```

and then

```ruby
...

def current_name
  name_for(@current)
end

...

def current_device_names
  GetNetcardInterfaces().map { |device| name_for(device) }.reject(&:empty?)
end

...
```

test/lib/network/edit_nic_name_test.rb
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env rspec
+
+require_relative "../../test_helper"
+
+require "yast"
+require "network/edit_nic_name"
+
+current_name = "spec0".freeze
```
dgdavid: Leftovers? Or did you want to use them as the value for below `let`s? In such case, `current_name` and `new_name` are not being used.

test/lib/network/edit_nic_name_test.rb
```diff
@@ -0,0 +1,137 @@
...
+new_name = "new1".freeze
+existing_new_name = "existing_new_name".freeze
+
+Yast.import "LanItems"
+
+describe Yast::EditNicName do
+  let(:subject) { described_class.new }
+  let(:current_name) { "spec0" }
```
dgdavid: See the comment above

test/lib/network/edit_nic_name_test.rb
teclator: completely leftovers :)

src/lib/network/edit_nic_name.rb
```diff
@@ -151,5 +162,36 @@ def CheckUdevNicName(name)
...
+    # update the interface name in the related routes or not.
+    #
+    # return [Boolean] whether the routes have to be updated or not
+    def update_routes?(previous_name)
+      return false unless Routing.device_routes?(previous_name)
+
+      Popup.YesNoHeadline(
+        Label.WarningMsg,
+        # TRANSLATORS: Ask for fixing a possible conflict after renaming
+        # an interface, %1 is the previous interface name %2 is the current one
```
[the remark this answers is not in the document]
teclator: Yep, I forgot to update the documentation.

src/include/network/lan/complex.rb
```diff
@@ -571,5 +572,10 @@ def overview_buttons
 
       ret
     end
+
+    # Convenience method to update the {Yast::Routing} devices list
+    def update_routing_devices!
+      Routing.SetDevices(LanItems.current_device_names)
+    end
```
mchf: same method for the second time, it doesn't look as a good idea.

src/lib/network/edit_nic_name.rb
```diff
@@ -151,5 +162,36 @@ def CheckUdevNicName(name)
...
+    # that belongs to the current renamed {LanItem}
+    def update_routes!(previous_name)
+      Routing.device_routes(previous_name).each do |route|
+        route["device"] = LanItems.current_name
+      end
+    end
+
+    def update_routing_devices!
+      Routing.SetDevices(LanItems.current_device_names)
+    end
```
mchf: again?

src/lib/network/edit_nic_name.rb
```diff
@@ -68,11 +74,16 @@ def run
         # FIXME: it changes udev key used for device identification
         #  and / or its value only, name is changed elsewhere
         LanItems.update_item_udev_rule!(udev_type)
+
+        if new_name != old_name
```
mchf: we already have ```LanItems.renamed?```, should be usable here.

src/lib/network/edit_nic_name.rb
teclator: It is a convenience method for update the routing devices, preferred to have it duplicated than do a wrong abstraction.

src/lib/network/edit_nic_name.rb
teclator: It is not enough... Use case:

1. Modify Item name & update routing devices. (renamed? true) eth0 -> em0
2. Modify Item name from em0 -> eth0 (renamed? false, but routes should be updated)

src/modules/LanItems.rb
```diff
@@ -2636,6 +2653,13 @@ def enslaved?(ifcfg_name)
 
   private
 
+    # Return the current name of the {LanItem} given
+    #
+    # @param item_id [Integer] a key for {#Items}
+    def current_name_for?(item_id)
```
mchf: seems like a copy of https://github.com/yast/yast-network/blob/87918f832546e76f124428b7efb9ca6b0789e877/src/lib/network/network_autoyast.rb#L382

teclator: Seems it behaves the same way, yes.

src/include/network/lan/address.rb
```diff
@@ -1436,6 +1441,19 @@ def initialize_address_settings
...
+    # Return wether the {Yast:Routing} devices list needs to be updated or not
+    # to include the current interface name
+    #
+    # @return [Boolean] false if the current interface name is already present
+    def update_routing_devices?
+      !Routing.devices.include?(LanItems.current_name)
+    end
+
+    # Convenience method to update the {Yast::Routing} devices list
+    def update_routing_devices!
```
dgdavid: @teclator, just an idea, without having too much knowledge about this.

What about to move this two methods to `Routing` module, I mean

```ruby
class Routing # actually is something like class RotuinClass < Module ...
  ...

  def update_routing_devices(device_names) # or def update_devices(names)
    return if devices.include?(device_names)

    SetDevices(device_names)
  end
end
```

Then, you can remove those convenience methods and use `Routing.update_routing_devices` or just `Routing.update_devices` instead.

src/lib/network/edit_nic_name.rb
```diff
@@ -151,5 +162,36 @@ def CheckUdevNicName(name)
 
       true
     end
+
+    # When an interface name has changed, it returns whether the user wants to
+    # update the interface name in the related routes or not.
+    #
+    # return [Boolean] whether the routes have to be updated or not
+    def update_routes?(previous_name)
```
dgdavid: Could this also be moved into `Routing` module? Seems to be its responsibility, isn't it?

src/lib/network/edit_nic_name.rb
mchf: well in fact I think that updating routes after each name change is not needed. It should be done when user actually writes the configuration. Not when he fills in new name.

teclator: I originally added also when "Committing" changes, then realized that "Cancel" also doesn't revert name changes, so moved the responsibility close to the action where persistent changes (of course before final write) where done. But permitting the user to decide if update them or not. 

So, about when **Write** the configuration, that is exactly what reported. If you modify the name and goes to the Routing section the references will be still there.

src/lib/network/edit_nic_name.rb
teclator: Probably yes, I didn't moved this method and also the update_routing_devices! because avoided to Import "LanItems"

src/include/network/lan/address.rb
teclator: Yep, I was already evaluating that.

teclator: Mmm doesn't look so different than calling SetDevices(...)

src/lib/network/edit_nic_name.rb
teclator: https://github.com/yast/yast-network/pull/725/commits/40d4aebeb0971aa1a3aff4c5033deabd57684d54#diff-72293d1fba24e0701afaa84e81e1af12R1394

mchf: then we have a bug that "Cancel" do not revert name change, which should be fixed.

teclator: > well in fact I think that updating routes after each name change is not needed. It should be done when user actually writes the configuration. Not when he fills in new name.

So, what do you propose?

teclator: > then we have a bug that "Cancel" do not revert name change, which should be fixed.

Yep, probably Udev name rules are not reverted when canceling, so basically another missing callback.

src/include/network/lan/address.rb
dgdavid: Right. It only avoids to call `Set` when not necessary :)

src/modules/LanItems.rb
```diff
@@ -2176,7 +2191,10 @@ def Commit
     # @return [true] so that this can be used for the :abort callback
     def Rollback
       log.info "rollback item #{@current}"
-      if getCurrentItem.fetch("hwinfo", {}).empty?
+      # Do not delete elements that are :edited but does not contain hwinfo
+      # yet (Add a virtual device and then edit it canceling the process during the
+      # edition)
+      if LanItems.operation == :add && getCurrentItem.fetch("hwinfo", {}).empty?
```
teclator: Fix for some LanItem cancel editing issues:

- See https://bugzilla.suse.com/show_bug.cgi?id=998540

src/lib/network/edit_nic_name.rb
mchf: I propose to fix cancel behavior instead of introducing new exceptional behavior.

teclator: Fine with me, but then propose to do it separately, in other PR as that is another bug.

mchf: no problem, then pls create a bug and link this PR to make future fix of this behavior easier.

package/yast2-network.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Thu Feb 21 17:50:32 UTC 2019 - knut.anderssen@suse.com
+
+- When renaming a interface (bsc#1060207):
```
mchf: ... an interface ...

package/yast2-network.changes
```diff
@@ -1,3 +1,13 @@
...
+  - Offer to the user the option to update device routes that
+    belongs to the previous interface name.
+  - Update the Routing devices list with the new one when needed.
+- Only delete a LanItem when canceling an addition (bsc#998540)
```
mchf: this is more about internals, not that useful for user.

src/lib/network/edit_nic_name.rb
teclator: I could move the logic back to the Address dialog method, but I preferred to have it Isolated in the change nic name dialog until we fix what already discussed.

As I said, it is strange to ask about updating routes when canceling, so, it is not so strange to ask when changing name, then you have the option to not confirm and cancel, of course it is still not perfect but better than crashing.

package/yast2-network.changes
teclator: Well, it fixes a bug

src/lib/network/network_autoyast.rb
```diff
@@ -334,20 +334,20 @@ def assign_udevs_to_devs(udev_rules)
...
         end
         next if !matching_item
 
-        name_from = item_name(item)
+        name_from = LanItems.current_name_for(item_id)
         log.info("Matching device found - renaming <#{name_from}> -> <#{name_to}>")
 
         # rename item in collision
-        rename_lan_item(colliding_item(name_to), name_from)
+        rename_lan_item(LanItems.colliding_name_item_id(name_to), name_from)
```
mchf: NP: in my POV renaming it to ```colliding_name_item_id``` is a bit overkill

test/lan_items_helpers_test.rb
```diff
@@ -590,3 +565,113 @@ def mock_dhcp_setup(ifaces, global)
...
+  let(:renamed_to) { nil }
+  let(:current) { 0 }
+  let(:item_0) do
+    {
+      "ifcfg"      => "eth0",
+      "renamed_to" => renamed_to,
+      "udev"       => {
+        "net" => [
+          "ATTR{address}==\"24:be:05:ce:1e:91\"",
+          "NAME=\"#{renamed_to}\""
```
mchf: I'd prefer to mock relevant methods bcs for ```renamed_to == nil``` this creates an invalid udev rule and also in theory it can be difference if ```renamed_to``` key is not present or ```{renamed_to => nil}``` 

I know that currently it is not such difference, but  you're mocking by defining an internal structure which hopefully will be subject of change *soon* instead of mocking return values of several already well known methods.

src/lib/network/network_autoyast.rb
teclator: Ok, I will ommit rewording it, but just will move

test/lan_items_rollback_test.rb
```diff
@@ -29,8 +29,9 @@
   end
 
   describe "#Rollback" do
-    context "when the current item is committed" do
+    context "when the current item is edited" do
       before do
+        Yast::LanItems.operation = :edited
```
mchf: AFAIK ```LanItems``` operates with ```operation == :edit```

test/lan_items_helpers_test.rb
teclator: That is true, but basically I moved your mock from

https://github.com/yast/yast-network/blame/master/test/network_autoyast_test.rb#L394

package/yast2-network.changes
mchf: this is usually my problem. Seems that you're also to deep already ;-)

e.g. something like "do not loose newly created virtual device when canceling edit action"

test/lan_items_helpers_test.rb
mchf: that doesn't mean that it is correct. My tests are usually .... not so perfect ;-)

src/lib/network/network_autoyast.rb
mchf: I completely agree but we have to survive it as long as ```LanItems::Items``` is here with us.

PR #726, opened 2019-02-12
Do not reset /etc/hosts

schubi2:
I have tested:
- Host changes in the UI and running system
- No Host changes with AY installation.
- Host changes with AY installtion.
- Running AY in an installed system without host changes.
- Running AY in an installed system with host changes.

### DIFF
--- a/src/modules/Host.rb
+++ b/src/modules/Host.rb
@@ -45,6 +45,8 @@ module Yast
       @initialized = false
 
       @hosts = CFA::Hosts.new
+
+      @configuration_imported = false
     end
 
     # Remove all entries from the host table.
@@ -52,6 +54,7 @@ module Yast
       @hosts.hosts.keys.each do |ip|
         @hosts.delete_by_ip(ip)
       end
+      @configuration_imported = false
     end
 
     # @return [hash] address->list of names
@@ -142,6 +145,12 @@ module Yast
 
       @hosts.save
 
+      # Reset that the configuration has been taken from AY file because the settings
+      # are now on the target system.
+      @configuration_imported = false
+      # Syncing after the settings have been written.
+      @initial_hosts = @hosts.clone
+
       Progress.NextStage if gui
 
       true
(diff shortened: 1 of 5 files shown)


src/modules/Host.rb
```diff
@@ -156,6 +165,7 @@ def Write(gui: false)
     # @return true if success
     def Import(settings)
```
teclator: Have not checked all the code and the bug comments yet, but wasn't enough just returning false before marking it as initialized in case of no "hosts" section?
```suggestion
     def Import(settings)
       return false unless settings.key("hosts")
```

schubi2: That has been the first try but it has failed. The import will not be called at all if the section "host" is not in AY configuration file. And with this call the user would not be able to delete all host entries. Not regarding if it is useful or not :-)

teclator: If that it is the case then the next check is like not necesary
```ruby
       @configuration_imported = true # if settings.key?("hosts")
```

src/modules/Host.rb
```diff
@@ -298,10 +308,8 @@ def ResolveHostnameToStaticIPs
     # Function which returns if the settings were modified
     # @return [Boolean]  settings were modified
     def GetModified
-      # when importing AY profile @initial_hosts is not initialized
-      # bcs it would be the same as @hosts and modification would not be detected
-      return true if @hosts && @initial_hosts.nil?
```
teclator: So we marked it as modified even when there was no `hosts' section

test/host_test.rb
```diff
@@ -167,6 +174,21 @@
...
     end
+
+    context "import an emtpy host section" do
+      let(:org_etx_hosts) { file.content }
+
+      before do
+        Yast::Host.Import("hosts" => {})
+      end
+
+      it "does not reset /etc/hosts" do
```
teclator: Mm it does not reset it ? or it does?

teclator:
LGTM

PR #727, opened 2019-02-18
Permit to choose between NetworkManger and wicked correctly

teclator:
## Problem

Recently, we have added back the network proposal to the installation proposal overview page ([related PBI](https://trello.com/c/DLi2b1mG/338-fate-326480-add-network-to-installation-proposal-overview-page)). 

As part of the network summary it shows the **network backend** in use, permitting the user to choose between **wicked** or **Network Manager**.

The first problem  to solve arises when the package for the selected backend is not available. In that case the backend is selected and in case the dependency was added, it does not permit to continue with the installation.

But there was another problem when the dependencies were satisfied. 

The selection in the proposal is overwritten when saving the network at the end of the installation with the value determined by the control file [**network_manager**](https://github.com/yast/skelcd-control-SLED/blob/master/control/installation.SLED.xml#L172) option. (see [save_network](https://github.com/yast/yast-network/blob/master/src/lib/network/clients/save_network.rb#L299))

- https://bugzilla.suse.com/show_bug.cgi?id=1124478
- https://bugzilla.suse.com/show_bug.cgi?id=1124498
- Related PBI https://trello.com/c/eBiqtDo7/706-2-bug-1124478-and-1124498-build-1584-switching-from-nm-to-wicked-and-back-and-forth

## Solution

- A **ProposalSettings** class has been added under the Y2Network namespace, it will determine at initialization the backend to use and also offers a method to check if **Network Manager** is available at all during the installation.

- When saving the network, it is checked which network backend has been selected and the Network service is set accordingly to it avoiding the override of any change during the proposal.

- As part of the PR we have removed also the summary about not written the hostname into /etc/hosts as that option was already [removed](https://trello.com/c/so0amSNp/342-sle-15-sp1-yast2-network-clarify-hostname-setup) from the GUI.

## Testing

- *Added unit tests*
- *Tested manually*

### Without registering NM package is not available, thus the switch is not offered.

screenshot_testing_2019-02-18_08 59 48

### After going back and registering the system we are able to choose the backend to be used

screenshot_testing_2019-02-18_09 06 02
screenshot_testing_2019-02-18_09 06 15

### Backend default defined by the control file is respected in the proposal

screenshot_testing_2019-02-18_11 12 14

### The selected backend is correctly enabled at the end of the installation

wickedselected

### DIFF
--- a/src/lib/y2network/proposal_settings.rb
+++ b/src/lib/y2network/proposal_settings.rb
@@ -33,11 +33,16 @@ module Y2Network
 
     # Constructor
     def initialize
+      Yast.import "Arch"
       Yast.import "ProductFeatures"
       Yast.import "PackagesProposal"
       Yast.import "Lan"
 
       @backend = use_network_manager? ? :network_manager : :wicked
+
+      log.info("The default proposed network backend is: #{@backend}")
+
+      @backend
     end
 
     # Adds the NetworkManager package to the {Yast::PackagesProposal} and sets
@@ -66,7 +71,10 @@ module Y2Network
     # @return [Boolean] false if no package available, true otherwise
     def network_manager_available?
       p = Y2Packager::Package.find("NetworkManager").first
-      return false if p.nil?
+      if p.nil?
+        log.info("The NetworkManager package is not available")
+        return false
+      end
       log.info("The NetworkManager package status: #{p.status}")
       true
     end
@@ -112,7 +120,7 @@ module Y2Network
       when "always"
         true
       when "laptop"
-        laptop = Arch.is_laptop
+        laptop = Yast::Arch.is_laptop
         log.info("Is a laptop: #{laptop}")
         laptop
       end
--- a/src/lib/y2network/proposal_settings.rb
+++ b/src/lib/y2network/proposal_settings.rb
@@ -28,7 +28,7 @@ module Y2Network
     include Yast::Logger
     include Yast::I18n
 
-    # [Boolean] network service to be used after the installation
+    # @return [Boolean] network service to be used after the installation
     attr_accessor :backend
 
     # Constructor
@@ -39,10 +39,8 @@ module Y2Network
       @backend = use_network_manager? ? :network_manager : :wicked
     end
 
-    # Services
-
-    # Add the NetworkManager package to be installed and sets NetworkManager as
-    # the backend to be used
+    # Adds the NetworkManager package to the {Yast::PackagesProposal} and sets
+    # NetworkManager as the backend to be used
     def enable_network_manager!
       Yast::PackagesProposal.AddResolvables("network", :package, ["NetworkManager"])
       Yast::PackagesProposal.RemoveResolvables("network", :package, ["wicked"])
@@ -51,6 +49,8 @@ module Y2Network
       self.backend = :network_manager
     end
 
+    # Add the wicked package to the {Yast::PackagesProposal} and sets wicked
+    # as the backend to be used
     def enable_wicked!
       Yast::PackagesProposal.AddResolvables("network", :package, ["wicked"])
       Yast::PackagesProposal.RemoveResolvables("network", :package, ["NetworkManager"])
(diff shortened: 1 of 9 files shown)


src/lib/y2network/proposal_settings.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    # Constructor
+    def initialize
+      Yast.import "PackagesProposal"
+      Yast.import "Lan"
+
+      @backend = use_network_manager? ? :network_manager : :wicked
+    end
+
+    # Services
```
dgdavid: Leftover?

src/lib/y2network/proposal_settings.rb
```diff
@@ -0,0 +1,94 @@
...
+      Yast::PackagesProposal.RemoveResolvables("network", :package, ["NetworkManager"])
+
+      log.info "Enabling Wicked"
+      self.backend = :wicked
+    end
+
+    class << self
+      # Singleton instance
+      def instance
+        create_instance unless @instance
```
dgdavid: Why not `@instance ||= create_instance`

dgdavid: BTW, why do you need the "additional" `create_instance` method here? Where is being used or necessary?

teclator: Originally I just included the Singleton module, but the create instance was useful for testing purposes. Ensuring a re-initialization of the instance reading the default  from the control file.

src/lib/y2network/proposal_settings.rb
teclator: Yep, it is not related to services anymore

src/lib/y2network/proposal_settings.rb
```diff
@@ -95,7 +96,26 @@ def create_instance
     def use_network_manager?
       return false unless network_manager_available?
 
-      Yast::Lan.UseNetworkManager
```
teclator: When the proposal settings are instantiated the package is still not installed so we cannot rely on that method.

https://github.com/yast/yast-network/blob/master/src/modules/Lan.rb#L861

src/lib/y2network/proposal_settings.rb
```diff
@@ -0,0 +1,121 @@
...
+      case Yast::ProductFeatures.GetStringFeature("network", "network_manager")
+      when ""
+        # compatibility: use the boolean feature
+        # (defaults to false)
+        Yast::ProductFeatures.GetBooleanFeature("network", "network_manager_is_default")
+      when "always"
+        true
+      when "laptop"
+        laptop = Arch.is_laptop
+        log.info("Is a laptop: #{laptop}")
```
dgdavid: I'd use a more meaningful message here, something similar to

`NetworkManager is selected by default because the installation is running in a laptop`

However, I am wondering why it worth to log specific information for this case and not for the rest,

`NetworkManager is not selected by default`
`Product control file determines to select always NetworkManager as default network backend`,

What am I missing?

src/lib/y2network/proposal_settings.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      log.info "Enabling Wicked"
+      self.backend = :wicked
+    end
+
+    # Convenience method to obtain whether the NetworkManager package is
+    # available or not.
+    #
+    # @return [Boolean] false if no package available, true otherwise
+    def network_manager_available?
```
dgdavid: It could be a private method, isn't it?

test/network_proposal_test.rb
```diff
@@ -20,11 +20,13 @@
   end
 
   describe "#make_proposal" do
+    let(:settings) { Y2Network::ProposalSettings.create_instance }
     let(:using_wicked) { true }
```
dgdavid: Why not call it `use_network_manager`? You're using it to mock the return value of `Yat::Lan::UseNetworkManager`.

src/lib/y2network/proposal_settings.rb
teclator: That only logs if it is a laptop or not, when the control file "network_manager" option is "laptop", so basically when it is a laptop it will be enabled by default but in case of a server or workstation it will not.

src/lib/y2network/proposal_settings.rb
teclator: Well, I actually exposed it

https://github.com/yast/yast-network/pull/727/files#diff-b206be758f56a39bbb81b3b2aebd6580R57

src/lib/y2network/proposal_settings.rb
dgdavid: Thank you for the explanation, @teclator 

To be honest, I did understand that, but I do not know why this case is more special that the other ones. Additionally, I think it will weird to find `Is a laptop: true` in the logs, without more contextual information (apart from the file and line number :P)

teclator: Well, the logs are closed together, so when you are debugging it will be easy to identity as we are currently logging if NM will be used by default or not.

dgdavid:
@teclator 

The PR looks good to me :) Thank you so much for addressing it :)

PR #728, opened 2019-02-22
[RFC] Create a backup of the modified interface and restore if canceled

teclator:
I have created a separate PR, so it could be ignored by now or not.

Basically with this PR any modification done during the edit process related to the current LanItem is restored in case of aborted or canceled.


teclator:
This should not be needed anymore, so closing it by now.

PR #729, opened 2019-02-25
[RFC] First impressions of the current network configuration status

teclator:
- https://trello.com/c/MGUrYrXo/699-3-research-identify-the-worst-places-in-yast-network-for-refactoring

### DIFF
--- a/doc/status-current-code.md
+++ b/doc/status-current-code.md
@@ -1,0 +1,63 @@
+## Current Network Code Overview
+
+The main responsabilities of Network configuration resides between yast2-network 
+and yast2 (NetworkInterfaces) and them are by sure the more complex parts of
+the code.
+
+- NetworkInterfaces (yast2): represent mainly the configured interfaces and is in
+  charge of writing the ifcfg-giles. The numbers of bugs in this area are not
+  so big, we have been dealing mostly with some normalization of the config or
+  better defaults like for wireless configuration.
+
+- LanItems (yast2-network): It was introduced in 2007 [1] for unifying all the
+  maps in a single list.
+
+  So basically each LanItems Item has the information about its (hwinfo, ifcfg
+  file, [udev rules](udev-rules-implementation.org), and table description 
+  (the overview))
+
+  The match between these different objects / hashes is based on the interface 
+  or device name, and as we can see it has been one of me most common point of 
+  failures in network configuration.
+
+  Of course, naming devices is a known problems in Networking an that was the
+  reason of the new predictable network interfaces naming. Something we are
+  not prepared yet but which is already used in TW (we still offer to rename
+  the interfaces based on [udev rules](udev_rules.md) even when does not work
+  on this scenario)
+
+
+Some of the lacks of the current network configuration module:
+
+1. Callbacks (specially in case of renaming or deleting an interface)
+
+    - Update hosts not only when editing static ip addresses but also when
+      removing an interface or changing it to DHCP [2]
+    - Update firewalld configuration
+    - Update routes
+
+2. Code separation / organization: It is not new and inherit from YCP era, but
+   we could start splitting at least the presentation part from the different
+   models, and move some responsabilities that currently are part of the
+   dialogs to other place. (See for example controllers / actions in
+   storage-ng)
+
+3. Duplication of code and bad methods naming, making it difficult to reuse or
+   understand if you do not know too deep about it. (It is also described in
+   the udev status document) [3]
+  
+4. Some configuration workflow or interfaces should be revised or redesigned
+   as it is not clear enough how to set some of the most common options like
+   in wireless config. Not so far hostname setup has been already modified [4].
+
+Note: During the installation, there is not separation between the currenct
+configuration and the proposal (except the backend choose). Network is very
+fragile and specially when remote installations are in use we do not want to
+break current connections (for that maybe having a separation could help, but
+this is only an idea that needs to be ellaborated / discussed than a real
+problem).
+
+[1] https://lists.opensuse.org/yast-devel/2007-06/msg00001.html
+[2] https://trello.com/c/iAKsqifd
+[3] https://github.com/yast/yast-network/blob/da0f35221e89f86fd2d998026cb1a0db32379ba8/doc/udev-rules-implementation.org
+[4] https://github.com/yast/yast-network/pull/692
--- a/doc/status-current-code.md
+++ b/doc/status-current-code.md
@@ -1,63 +1,50 @@
 ## Current Network Code Overview
 
-The main responsabilities of Network configuration resides between yast2-network 
-and yast2 (NetworkInterfaces) and them are by sure the more complex parts of
-the code.
+The main responsabilities of Network configuration resides between yast2-network and yast2 (NetworkInterfaces, NetworkService) and them are by sure the more complex parts of the code.
 
-- NetworkInterfaces (yast2): represent mainly the configured interfaces and is in
-  charge of writing the ifcfg-giles. The numbers of bugs in this area are not
-  so big, we have been dealing mostly with some normalization of the config or
-  better defaults like for wireless configuration.
+### NetworkService (yast2)
 
-- LanItems (yast2-network): It was introduced in 2007 [1] for unifying all the
-  maps in a single list.
+Its responsibilities are well defined, and although the current test covegage could be improved it does not need a redesign / refactorization at all.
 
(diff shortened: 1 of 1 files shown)


doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+configuration and the proposal (except the backend choose). Network is very
+fragile and specially when remote installations are in use we do not want to
+break current connections (for that maybe having a separation could help, but
+this is only an idea that needs to be ellaborated / discussed than a real
+problem).
+
+[1] https://lists.opensuse.org/yast-devel/2007-06/msg00001.html
+[2] https://trello.com/c/iAKsqifd
+[3] https://github.com/yast/yast-network/blob/da0f35221e89f86fd2d998026cb1a0db32379ba8/doc/udev-rules-implementation.org
+[4] https://github.com/yast/yast-network/pull/692
```
ancorgs: Please check [the rendered version](https://github.com/yast/yast-network/blob/d8c2536758ee66b7ffb76967ed2c8785a421e348/doc/status-current-code.md). These newlines are not visible there.

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+    - Update hosts not only when editing static ip addresses but also when
+      removing an interface or changing it to DHCP [2]
+    - Update firewalld configuration
+    - Update routes
+
+2. Code separation / organization: It is not new and inherit from YCP era, but
+   we could start splitting at least the presentation part from the different
+   models, and move some responsabilities that currently are part of the
+   dialogs to other place. (See for example controllers / actions in
+   storage-ng)
```
ancorgs: There are many nice examples on how storage-ng separates responsibilities and layers: models / dialogs / clients in the `Y2Storage` part or actions / dialogs / widgets in the `Y2Partitioner` part to name a couple... But the not-so-obvious relationships between actions and controllers is definitely not the best example of well engineered separation. :wink:

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+- LanItems (yast2-network): It was introduced in 2007 [1] for unifying all the
+  maps in a single list.
+
+  So basically each LanItems Item has the information about its (hwinfo, ifcfg
+  file, [udev rules](udev-rules-implementation.org), and table description 
+  (the overview))
+
+  The match between these different objects / hashes is based on the interface 
+  or device name, and as we can see it has been one of me most common point of 
+  failures in network configuration.
```
jreidinger: can you please add links to some well known bugs in network so we know which use cases we need to have in mind? Also one issue which is not mentioned here is problem that not all hwinfo need to have ifcfg file and so one, which sometimes causes problem.

teclator: Sure, that is the main difference between virtual interfaces and phisical ones, and already described in Michal Zugec email link. So, will try to expand here in a way that the problems behind that are easy to expose and understand.

I will add also a section with the installation overview (not having a separation between the inst-sys network configuration and the proposal for the system being installed).

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
+## Current Network Code Overview
+
+The main responsabilities of Network configuration resides between yast2-network 
+and yast2 (NetworkInterfaces) and them are by sure the more complex parts of
+the code.
```
mchf: and NetworkService

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+
+- NetworkInterfaces (yast2): represent mainly the configured interfaces and is in
+  charge of writing the ifcfg-giles. The numbers of bugs in this area are not
+  so big, we have been dealing mostly with some normalization of the config or
+  better defaults like for wireless configuration.
```
mchf: Since fixing device renaming, NetworkInterfaces is calm place ... however it needs some love. Some methods should be revised if it should live in an upper layer. Some features should be redesigned (filtering devices according to name when filtering to device groups)

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+
+- LanItems (yast2-network): It was introduced in 2007 [1] for unifying all the
+  maps in a single list.
+
+  So basically each LanItems Item has the information about its (hwinfo, ifcfg
+  file, [udev rules](udev-rules-implementation.org), and table description 
+  (the overview))
```
mchf: and for example storing complete hwinfo for each interface is a bit overkill and causes a huge footprint. It is sad bcs in reality only two or three items are actively used.

doc/status-current-code.md
mchf: moreover Items use a useless layer of abstraction (item_id). Designing a structure where devices would be more easily accessible according its names would have bigger benefit (now we search such device with O(n) complexity)

doc/status-current-code.md
mchf: On the other hand this module will become less important in the future when we should support exporting device configuration in wicked's native xml.

mchf: So in the end we will need a support for sysconfig way of configuration (NetworkInterfaces), wicked's xml, ... so we can even think of at least partly supporting NetworkManager

mchf:
Currently a big pain of networking code in my POV is that it mixes backend (controller) and UI (view) code together. It leads to methods with hundred lines with multiple exceptional handling like "if we need only backend functionality and no GUI, then don't show this popup"

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+
+  The match between these different objects / hashes is based on the interface 
+  or device name, and as we can see it has been one of me most common point of 
+  failures in network configuration.
+
+  Of course, naming devices is a known problems in Networking an that was the
+  reason of the new predictable network interfaces naming. Something we are
+  not prepared yet but which is already used in TW (we still offer to rename
+  the interfaces based on [udev rules](udev_rules.md) even when does not work
+  on this scenario)
```
mchf: well not prepared ... there is only one problematic part ... and that is transition (or creating) of configuration from instsys to installed target.

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+
+2. Code separation / organization: It is not new and inherit from YCP era, but
+   we could start splitting at least the presentation part from the different
+   models, and move some responsabilities that currently are part of the
+   dialogs to other place. (See for example controllers / actions in
+   storage-ng)
+
+3. Duplication of code and bad methods naming, making it difficult to reuse or
+   understand if you do not know too deep about it. (It is also described in
+   the udev status document) [3]
```
mchf: naming is not that big problem for me, however some methods are definitely too complex and some still have undocumented side effects.

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+  
+4. Some configuration workflow or interfaces should be revised or redesigned
+   as it is not clear enough how to set some of the most common options like
+   in wireless config. Not so far hostname setup has been already modified [4].
```
mchf: I'd vote for revising methods like ReadHardware. This beasts are really huge and just works ... somehow, but no one knows why ... and if it is still needed to do that way.

doc/status-current-code.md
```diff
@@ -0,0 +1,63 @@
...
+
+Note: During the installation, there is not separation between the currenct
+configuration and the proposal (except the backend choose). Network is very
+fragile and specially when remote installations are in use we do not want to
+break current connections (for that maybe having a separation could help, but
+this is only an idea that needs to be ellaborated / discussed than a real
+problem).
```
mchf: a lot of configuration is stored in strange "global" variables which are wrongly maintained, for example it already happened that some variables was not cleared when configuration of device was canceled ... and configuration got (partly) written when whole module was closed by clicking [OK]

doc/status-current-code.md
```diff
@@ -1,63 +1,50 @@
...
 
-  So basically each LanItems Item has the information about its (hwinfo, ifcfg
-  file, [udev rules](udev-rules-implementation.org), and table description 
-  (the overview))
+### NetworkInterfaces (yast2)
 
-  The match between these different objects / hashes is based on the interface 
-  or device name, and as we can see it has been one of me most common point of 
-  failures in network configuration.
+Represent mainly the configured interfaces and is in charge of writing the ifcfg-giles. The numbers of bugs in this area are not so big, we have been dealing mostly with some normalization of the config or better defaults like for wireless configuration.
```
mchf: typo giles -> files

imobachgs:
Outdated and not relevant anymore.

PR #730, opened 2019-02-27
Network Proposal: Enhanced dependencies handling (bsc#1127228)

teclator:
## Problem

In #727 we did some fixes for SLES improving the switch handling between network backends. But once the network proposal was added into openSUSE (https://github.com/yast/skelcd-control-openSUSE/pull/170 & https://github.com/yast/skelcd-control-openSUSE/pull/169), the selection of the generic desktop role shown that the initialization of the packages requirement was not correct at all (although the backend was correctly selected)

So, it worked fine when switching between backends, but the packages were not added when entering the proposal the first time or when the role selected was modified going back & forth. Which means a not network installation in case **NetworkManager** was not installed by other pattern dependencies. 

- https://bugzilla.opensuse.org/show_bug.cgi?id=1127228

## Solution

- We have modified the proposal network client to refresh the list of resolvables 
 packages at initialization, so the software selection reflects them, and in case of removed a red warning / error will be shown.
- The **backend** used will be the `default` defined by the **control** file `until manual selection`, that is, a click switching between backends.
- At the end of the initialization (**save_network**), if for any reason (like removing the proposal) the package for the **backend selected is not available**, then, we will enable **wicked** as we did in the past (having at least a network configuration).
 
## Testing

- *Added unit tests*
- *Tested manually*

## Selection of role (default backend NM) and package removed later in the proposal

roleselection

warningafterremoval |

### DIFF
--- a/src/lib/y2network/proposal_settings.rb
+++ b/src/lib/y2network/proposal_settings.rb
@@ -29,7 +29,7 @@ module Y2Network
     include Yast::I18n
 
     # @return [Boolean] network service to be used after the installation
-    attr_accessor :backend
+    attr_accessor :selected_backend
 
     # Constructor
     def initialize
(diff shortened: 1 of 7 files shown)


src/lib/y2network/proposal_settings.rb
```diff
@@ -107,6 +136,13 @@ def use_network_manager?
       network_manager_default?
     end
 
+    # Convenience method to determine if the NM package is installed or not
+    #
+    # @return [Boolean] true if NetworkManager is installed; false otherwise
+    def network_manager_installed?
+      Yast::Package.Installed("NetworkManager")
```
mchf: I've still stronger feeling that this code duplicates anything we already have. 

For example this seems to be duplicate of ```NetworkService::backend_available?```

teclator: Well, at the end, both are convenience method for checking if the package is installed. NetworkService make sense to me in a running system, but during installation, there is no separation between .. the proposal and the instsys network configuration That is something I do not like at all, if I just change the service at that point, the configuration module is also affected.

So tried to not depend on NetworkService and now even importing Lan can be removed.

test/lib/y2network/proposal_settings_test.rb
```diff
@@ -100,34 +106,61 @@ def stub_features(features)
...
-      expect(logger).to receive(:info).with(/backend is: network_manager/)
-      described_class.create_instance
+      expect(logger).to receive(:info).with(/backend is: wicked/)
+      subject.default_backend
+    end
+  end
+
+  describe "#current_backend" do
+    let(:selected_backend) { :wicked }
+    let(:default_backend) { "wicked_or_nm" }
```
dgdavid: This seems to be a leftover, isn't it?

test/lib/y2network/proposal_settings_test.rb
```diff
@@ -140,26 +173,62 @@ def stub_features(features)
...
+          .with("network", :package, ["wicked"])
+        subject.refresh_packages
+      end
+    end
+  end
+
+  describe "#enable_network_manager!" do
+    before do
+      subject
+    end
```
dgdavid: You can use a single `before do` :wink:

test/lib/y2network/proposal_settings_test.rb
teclator: Not really, but it should be used in the line below :)

test/lib/y2network/proposal_settings_test.rb
teclator: These are really leftovers :+1: updated!

PR #732, opened 2019-03-14
Cleanup: Remove dead code and small fixes

teclator:
- https://trello.com/c/fTLpPNpb/848-clean-of-dead-code-and-small-fixes
- https://bugzilla.suse.com/show_bug.cgi?id=1129552


src/modules/LanItems.rb
```diff
@@ -2154,7 +2154,7 @@ def Commit
       # FIXME: INTERFACETYPE confuses sysconfig, bnc#458412
       # Only test when newdev has enough info for GetTypeFromIfcfg to work.
       implied_type = NetworkInterfaces.GetTypeFromIfcfg(newdev)
-      if !implied_type.nil? && implied_type != @type
+      if @type == "dummy" || !implied_type.nil? && implied_type != @type
```
mchf: I'd silently remove that ```implied_type``` when already touching. AFAIK dummy and lo types are only reasonable cases we want to support with ```INTERFACETYPE```

PR #735, opened 2019-03-19
add initial object proposal for dialog routes and gateway cwm widget

jreidinger:
original sources at https://github.com/yast/yast-network/blob/master/src/include/network/services/routing.rb


src/lib/y2network/widgets/gateway.rb
```diff
@@ -0,0 +1,44 @@
...
+        [:hstretch]
+      end
+
+      def init
+        Yast::UI.ChangeWidget(Id(:gateway), :ValidChars, Yast::IP.ValidChars +"-")
+        # TODO: init from route object
+      end
+
+      def validate
+        return true if value == "-"
```
mchf: hm. "-" is special char for sysconfig file ... but i don't think we need to support it in the UI.

jreidinger: well, I do not plan to do backend or code modification with functionality. I just plan to modify UI and how it use backend. And now it still uses "-" in UI. I tested it also manually. In fact it is default value there.

jreidinger: for current code see:

https://github.com/yast/yast-network/blob/master/src/include/network/services/routing.rb#L191
https://github.com/yast/yast-network/blob/master/src/include/network/services/routing.rb#L220
https://github.com/yast/yast-network/blob/master/src/include/network/services/routing.rb#L152

mchf: i know it uses it ... but its usual meaning is "unspecified". We can use some user more friendly way how to say that.

jreidinger: yes, fine for me. We will see also how our model define it.

jreidinger: and hopefully also writting help text which mentions special values can help.

src/lib/y2network/dialogs/route.rb
```diff
@@ -0,0 +1,87 @@
+require "ipaddr"
+
+require "cwm/popup"
+require "y2network/route"
+require "y2network/widgets/destination"
+require "y2network/widgets/gateway"
+require "y2network/widgets/devices"
+require "y2network/widgets/gateway"
```
imobachgs: This line is duplicated.

src/lib/y2network/dialogs/route.rb
```diff
@@ -0,0 +1,87 @@
...
+
+module Y2Network
+  module Dialogs
+    # Dialog to create or edit route.
+    class Route < CWM::Popup
+      # @param route is now term from table, but it should be object for single route
+      def initialize(route, available_devices)
+        log.info "route dialog with route: #{route.inspect} " \
+          "and devices #{available_devices.inspect}"
+        params = route.params
```
imobachgs: Ok, this is a temporary solution (as `route` should be an object). But perhaps you could extract the elements to something meaningful:

```ruby
_, destination, gateway, _, interface, options = route.params
```

On the other hand, if it is going to change soon, maybe it does not worth.

src/lib/y2network/widgets/destination.rb
```diff
@@ -30,7 +31,7 @@ def opt
       def init
         Yast::UI.ChangeWidget(Id(widget_id), :ValidChars, Yast::IP.ValidChars + "-/")
         val = @route.to
-        self.value = val == :default ? "-" : val.to_s
+        self.value = val == :default ? "-" : ( val.to_s + "/" + val.prefix.to_s )
```
teclator: If this operatio is common we should include it in helpers.

test/y2network/widgets/destination_test.rb
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  describe "#init" do
+    it "sets valid characters for widget" do
+      expect(Yast::UI).to receive(:ChangeWidget).with(anything, :ValidChars, anything)
+
+      subject.init
+    end
+
+    context "route.to is :default" do
```
teclator: NP: something more descriptive wouldn't hurt :P

src/lib/y2network/widgets/destination.rb
jreidinger: yep, we will see how common it is. I think we should now design public API and then check how to DRY sources, where it makes sense.

test/y2network/widgets/destination_test.rb
jreidinger: any idea? For me this is the the simplest and the most direct way how to say fact that `route attribute to is symbol :default`

src/lib/y2network/dialogs/route.rb
jreidinger: Well, my plan is probably object for route table, then can convert table Item to Route object and back. So ideally this dialog should in future receive simple Route object and responsibility for table conversion be on Table object ( not exist yet ).

mchf:
I don't have any comments for know, we will see what turns out later ;-)

PR #736, opened 2019-03-19
[RFC] New data model for routing (and more)

imobachgs:
After some discussion, we are proposing this alternative data model for the routing stuff. Please,
beware that it is just a PoC but we are interested in comments about the overall approach.  We took
some inspiration from the storage-ng case (actually this case is a lot simpler) but avoiding the
`probed` and `staging` fixed devicegraphs.

These are the key ideas:

* There is a `Y2Network::Config` object (I know that the name sucks badly) which keeps the network
  configuration (interfaces, routes, etc). Of course, those element are separate objects (no nested
  hashes anymore).
* The configuration can be read from different sources, like *wicked* or (in the future) *network
  manager*. See `Y2Network::ConfigReader` module.

So you could do something like this:

```ruby
  config = Y2Network::Config.from(:wicked)
  config.interfaces #=> ["lo", "eth0"]
  config.routes #=> []
  # and so on
```

Extending this idea, we would like to have:

* Classes to write the configuration to the underlying system (something like `Y2Network::ConfigWriter`).
* An API to query the config object. Using `Config#interfaces` or `Config#routes` is ok, but we
  might be interested in something more powerful (check for interfaces of a given type, etc.).
* Some validation logic (for instance the `RoutingTable` might check whether there is only one default route).

```ruby
  routing_table = config.routing_tables.first
  routing_table.valid?
  config.save
```

Finally, I would like to make clear that we are avoiding the typical `Y2Network::Route.all` kind
of interface because:

* We would like to (potentially) handle different configurations. E.g., for 1st and 2nd stage,
  installation and final, etc.
* Testing is easier and we won't need to keep mocking a lot of modules (`LanItems`, `Routing`, etc.).
* It looks like a more flexible approach. And if we miss the other one, is rather easy to put this API behind a `Y2Network::Route.all`.

So, please, have a look and write down your opinions and suggestions.

Thanks a lot!

### DIFF
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -28,6 +28,8 @@ module Y2Network
   class Config
     # @return [Array<Interface>]
     attr_reader :interfaces
+    # @return [Array<RoutingTable>]
+    attr_reader :routing_tables
 
     # @param source [Symbol] Source to read the configuration from
     class << self
@@ -38,8 +40,13 @@ module Y2Network
     end
 
     # @param devices [Array<Device>]
-    def initialize(interfaces:)
+    def initialize(interfaces:, routing_tables:)
       @interfaces = interfaces
+      @routing_tables = routing_tables
+    end
+
+    def routes
+      routing_tables.flat_map(&:routes)
     end
   end
 end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -26,6 +26,8 @@ module Y2Network
   #   config = Y2Network::Config.from(:sysconfig)
   #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
   class Config
+    # @return [Symbol] Configuration ID
+    attr_reader :id
     # @return [Array<Interface>]
     attr_reader :interfaces
     # @return [Array<RoutingTable>]
@@ -39,8 +41,13 @@ module Y2Network
       end
     end
 
-    # @param devices [Array<Device>]
-    def initialize(interfaces:, routing_tables:)
+    # Constructor
+    #
+    # @param id             [Symbol] Configuration ID
+    # @param interfaces     [Array<Interface>] List of interfaces
+    # @param routing_tables [Array<RoutingTable>] List of routing tables
+    def initialize(id: :system, interfaces:, routing_tables:)
+      @id = id
       @interfaces = interfaces
       @routing_tables = routing_tables
     end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -45,8 +45,13 @@ module Y2Network
       @routing_tables = routing_tables
     end
 
+    # Routes in the configuration
+    #
+    # Convenience method to iterate through the routes in all routing tables.
+    #
+    # @return [Array<Route>] List of routes which are defined in the configuration
     def routes
-      routing_tables.flat_map(&:routes)
+      routing_tables.flat_map(&:to_a)
     end
   end
 end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -1,0 +1,45 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "y2network/interface"
+
+module Y2Network
+  # This class represents the current network configuration including interfaces,
+  # routes, etc.
+  #
+  # @example Reading from wicked
+  #   config = Y2Network::Config.from(:wicked)
+  #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
+  class Config
+    # @return [Array<Interface>]
+    attr_reader :interfaces
+
+    # @param source [Symbol] Source to read the configuration from
+    class << self
+      def from(source = :wicked)
+        builder = ConfigReader.for(source)
+        builder.network_config
+      end
+    end
+
+    # @param devices [Array<Device>]
+    def initialize(interfaces:)
+      @interfaces = interfaces
+    end
+  end
+end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -23,7 +23,7 @@ module Y2Network
   # routes, etc.
   #
   # @example Reading from wicked
-  #   config = Y2Network::Config.from(:wicked)
+  #   config = Y2Network::Config.from(:sysconfig)
   #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
   class Config
     # @return [Array<Interface>]
@@ -31,7 +31,7 @@ module Y2Network
 
     # @param source [Symbol] Source to read the configuration from
     class << self
-      def from(source = :wicked)
+      def from(source)
         builder = ConfigReader.for(source)
         builder.network_config
       end
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -25,7 +25,7 @@ module Y2Network
     def self.for(source = :wicked)
       require "y2network/config_reader/#{source}"
       name = source.to_s.split("_").map(&:capitalize).join
-      klass = self.const_get(name)
+      klass = const_get(name)
       klass.new
     end
   end
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -19,10 +19,14 @@
 module Y2Network
   # This module contains a set of classes to read the network configuration from the system
   #
-  # For the time being, only the wicked reader ({Y2Network::ConfigReader::Wicked}) reader is
+  # For the time being, only the wicked reader ({Y2Network::ConfigReader::Sysconfig}) reader is
   # available.
   module ConfigReader
-    def self.for(source = :wicked)
+    # Config reader for a given source
+    #
+    # @param source [Symbol] Source name (e.g., :sysconfig)
+    # @return [#config] Configuration reader from {Y2Network::ConfigReader}
+    def self.for(source)
       require "y2network/config_reader/#{source}"
       name = source.to_s.split("_").map(&:capitalize).join
       klass = const_get(name)
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -1,0 +1,32 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  # This module contains a set of classes to read the network configuration from the system
+  #
+  # For the time being, only the wicked reader ({Y2Network::ConfigReader::Wicked}) reader is
+  # available.
+  module ConfigReader
+    def self.for(source = :wicked)
+      require "y2network/config_reader/#{source}"
+      name = source.to_s.split("_").map(&:capitalize).join
+      klass = self.const_get(name)
+      klass.new
+    end
+  end
+end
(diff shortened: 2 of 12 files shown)


src/lib/y2network/config_reader/wicked.rb
```diff
@@ -0,0 +1,41 @@
...
+    class Wicked
+      # @return [Y2Network::Config] Network configuration
+      def config
+        Yast::LanItems.Read
+        Config.new(interfaces: interfaces)
+      end
+
+      def interfaces
+        # TODO: for the time being, we are just relying in the underlying stuff.
+        Yast::LanItems.current_device_names.map do |name|
```
mchf: I think that ```Wicked``` reader should be build on top of ```NetworkInterfaces``` bcs ```LanItems``` is higher level. At least it collects configured even unconfigured devices together with udev rules and some more things. This can be replaced e.g. by ```NetworkInterfaces::List``` or ```NetworkInterfaces::Filter``` depending of what you need.

src/lib/y2network/route.rb
```diff
@@ -0,0 +1,49 @@
...
+    # @return [IPAddr]
+    attr_reader :to
+    # @return [Symbol] :local, :multicast, etc.
+    attr_reader :type
+    # @return [Device]
+    attr_reader :device
+    # @return [IPAddr,nil]
+    attr_reader :source
+    # @return [IPAddr,nil]
+    attr_reader :via
```
mchf: NP: I'd stick with common naming like gateway

src/lib/y2network/route.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  # This class represents a network route
+  class Route
+    # @return [IPAddr]
+    attr_reader :to
+    # @return [Symbol] :local, :multicast, etc.
+    attr_reader :type
```
mchf: NP: i don't think we need to distinguish that.

src/lib/y2network/routing_table.rb
```diff
@@ -0,0 +1,35 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  class RouteTable
```
mchf: i think that common name for this in system is "RoutingTable"

mchf: at least something like```RouteTable#default_gw``` would be useful

src/lib/y2network/config.rb
```diff
@@ -0,0 +1,45 @@
...
+  # @example Reading from wicked
+  #   config = Y2Network::Config.from(:wicked)
+  #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
+  class Config
+    # @return [Array<Interface>]
+    attr_reader :interfaces
+
+    # @param source [Symbol] Source to read the configuration from
+    class << self
+      def from(source = :wicked)
```
mchf: I'd use e.g. ```:sysconfig``` for what we currently use and keep ```:wicked``` for xml based configuration if we ever implement that

src/lib/y2network/routing_table.rb
imobachgs: Yes, you are right. I just forgot to rename the class (the file is already called `routing_table`) :)

src/lib/y2network/config.rb
```diff
@@ -0,0 +1,52 @@
...
+  #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
+  class Config
+    # @return [Array<Interface>]
+    attr_reader :interfaces
+    # @return [Array<RoutingTable>]
+    attr_reader :routing_tables
+
+    # @param source [Symbol] Source to read the configuration from
+    class << self
+      def from(source)
```
jreidinger: this is a bit confusing for me. Why we need to pass source? I expect that network config recognize itself what network daemon it uses. from is fine, but I expect here also `system` which reads what system currently use. Of course it should not be static variable, but instead read what is really on system.

src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -0,0 +1,60 @@
...
+require "y2network/interface"
+require "y2network/routing_table"
+
+Yast.import "NetworkInterfaces"
+Yast.import "Routing"
+
+module Y2Network
+  module ConfigReader
+    # This class reads the current configuration from the system
+    class Sysconfig
```
jreidinger: I am confused now. This is wicked reader?

src/lib/y2network/route.rb
```diff
@@ -0,0 +1,46 @@
...
+    attr_reader :source
+    # @return [IPAddr,nil]
+    attr_reader :gateway
+
+    def initialize(to, interface, gateway: nil, source: nil, preference: nil)
+      @to = to
+      @interface = interface
+      @gateway = gateway
+      @source = source
+      @preference = preference
```
jreidinger: preferences is current options?

src/lib/y2network/route.rb
```diff
@@ -0,0 +1,46 @@
...
+    # @return [IPAddr]
+    attr_reader :to
+    # @return [Interface]
+    attr_reader :interface
+    # @return [IPAddr,nil]
+    attr_reader :source
+    # @return [IPAddr,nil]
+    attr_reader :gateway
+
+    def initialize(to, interface, gateway: nil, source: nil, preference: nil)
```
jreidinger: so for default route is has to look like `Route.new(nil, "eth0")`? And whats more how will look static route without interface? like `Route.new(nil, "-")`? (see https://bugzilla.suse.com/show_bug.cgi?id=93996 ) This does not look nice.
Can we at least have ability to pass ":default` as to and only internally keep it as nil and for interface having it optional or allowing to pass symbol `:any` ?

jreidinger: Also to have easier review can you add explanation what means nil in gateway, source and preference, to see if it makes sense? Thanks

src/lib/y2network/routing_table.rb
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  class RoutingTable
+    attr_reader :routes
+
+    # MAIN_TABLE_ID = 254
```
jreidinger: ???

src/lib/y2network/routing_table.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def initialize(routes = [])
+      # @id = MAIN_TABLE_ID
+      @routes = routes
+    end
+
+    # @param route [Y2Network::Route] Route to add
```
jreidinger: just curious why shortcut for this method and not others in enumerable for routes?

src/lib/y2network/config.rb
```diff
@@ -0,0 +1,52 @@
...
+  # routes, etc.
+  #
+  # @example Reading from wicked
+  #   config = Y2Network::Config.from(:sysconfig)
+  #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
+  class Config
+    # @return [Array<Interface>]
+    attr_reader :interfaces
+    # @return [Array<RoutingTable>]
+    attr_reader :routing_tables
```
teclator: By now we are not handling policy routing but only the dealing with static routing and the main routing table, so maybe we just use :routes here as already described above.

src/lib/y2network/config.rb
```diff
@@ -0,0 +1,52 @@
...
+require "y2network/interface"
+
+module Y2Network
+  # This class represents the current network configuration including interfaces,
+  # routes, etc.
+  #
+  # @example Reading from wicked
+  #   config = Y2Network::Config.from(:sysconfig)
+  #   config.interfaces.map(&:name) #=> ["lo", eth0", "wlan0"]
+  class Config
```
teclator: I know you can use the variable name for that but I would define an `:id` or `:name` attribute in case we wanted something like in firewalld.

That is, the network instance could be a singleton one, which could have different configs ("running", "autoyast", "staging", "permanent" .. or whatever), being able to merge between them.

src/lib/y2network/config.rb
teclator: For wicked we have different sources, and that is another question, in that case that probably mean read from sysconfig + read from wicked XML and merge them. So, as you said, I'm also fine with `from`method and it is useful, we should think about where to put the responsibility about handling the configs. And not only that, we could want to merge the wicked config with the network manager config.

src/lib/y2network/config_reader/sysconfig.rb
teclator: Wicked reads the sysconfig configuration, and then his own configuration in XML merging all together in his internal state. In YaST we do not deal with the XML configuration but maybe we would like to support that in the future, that is why probably it was suggested to use `Sysconfig` instead of `Wicked`.

src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      def find_interfaces
+        Yast::NetworkInterfaces.Read
+        # TODO: for the time being, we are just relying in the underlying stuff.
+        Yast::NetworkInterfaces.List("").map do |name|
+          Y2Network::Interface.new(name)
+        end
+      end
+
+      def find_routing_tables(interfaces)
```
teclator: If I get this method correctly, it is basically for create a new routing table with all the routes that correspond to some of the interfaces given. Naming method is confusing here, because what we look for are routes and what we define is a routing table.

src/lib/y2network/route.rb
imobachgs: No, it is the priority (as it is called in the `ip route` man page). To be honest, we do not need it right  now, I was just sketching how the API could look like if it was needed.

src/lib/y2network/routing_table.rb
imobachgs: You can have several routing tables in a system an the main one has 254 as ID. I will remove the comment as it is not useful right now.

src/lib/y2network/route.rb
imobachgs: That's open for discussion: we could do the `to` parameter optional. And we could even use the same approach for static routes without and interface. Please, beware that we are not passing strings as interfaces anymore (we have a proper `Y2Network::Interface` class for that.

As a side note, I would say that, at low level, all routes have a device. But it is true that in the `sysconfig` files you can leave them out.

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: Yes, that's it. Initially I created a `Wicked` class but @mchf suggested to rename it to `Sysconfig` and reserve the `Wicked` name for future uses (like using the wicked's XML based interface).

src/lib/y2network/routing_table.rb
imobachgs: For no particular reason. But yes, I guess we could convert this class into a `Enumerable`.

src/lib/y2network/config.rb
imobachgs: Yes, I was wondering whether it makes sense to have a RoutingTable after all.

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: Not really: the interfaces list is just to allow the method to associate the correct interface with the route. I will try to find a better name anyway.

src/lib/y2network/route.rb
teclator:
```suggestion
    def initialize(to: :default, interface: :any, gateway: nil, netmask: nil, options: nil)
```

src/lib/y2network/route.rb
teclator: Probably one of the options, so suggested a rename

jreidinger: so I worry we will really need generic options here. I know it is expert only, but especially in autoyast many customer can depend on it to configure specific routing parameter which we are not even aware they use.

src/lib/y2network/routing_table.rb
teclator: It could be useful if we add support for policy routing and `ip rules`, but that is not supported in YaST so far. So by now we should probably let it out of the scope.

src/lib/y2network/routing_table.rb
jreidinger: then I think it will be only enhanced array. So the API stops to make sense for me. If we really want multiple routing tables, then I suggest to simple have routing table class that do mapping of ID and routes array and maybe some helper for default one, but keep just routes array and do not move it up to routing table itself.

src/lib/y2network/route.rb
mchf: i vote not to use ```netmask``` but ```prefix```

teclator: It is deprecated, and mostly not used, but still attached to IPv4 netmask, and how routes defined it. So.. let see what uses we will give to it. Saying that, for me it is ok with both.

src/lib/y2network/route.rb
```diff
@@ -0,0 +1,46 @@
...
+  # This class represents a network route
+  class Route
+    # @return [IPAddr]
+    attr_reader :to
+    # @return [Interface]
+    attr_reader :interface
+    # @return [IPAddr,nil]
+    attr_reader :source
+    # @return [IPAddr,nil]
+    attr_reader :gateway
```
jreidinger: JFYI: current UI setup this things for specific route (not default one): 

- Destination ( to here)
- Netmask
- gateway
- device ( aka interface here )
- options ( for any other possible options ).

src/lib/y2network/config_reader.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  # This module contains a set of classes to read the network configuration from the system
+  #
+  # For the time being, only the wicked reader ({Y2Network::ConfigReader::Wicked}) reader is
+  # available.
+  module ConfigReader
+    def self.for(source = :wicked)
```
jreidinger: will it work with Sysconfig class name?

imobachgs: Oops, no, you are right.

PR #739, opened 2019-03-20
Fetch also the parent busid when using a virtio netcard

teclator:
- https://trello.com/c/sFSmetoI/883-sles12-sp4-p3-1129012-sles-12-sp4-when-installing-sles-12-sp4-with-kvm-virt-install-autoyast-net-udev-is-not-renaming-the-networ
- https://bugzilla.suse.com/show_bug.cgi?id=1129012

## Problem

The bug appears in a **KVM** autoinstallation with **virtio** netcards. The user defines two **udev rules** using the **busid**, but we are not able to **match** the current rules, because while udev write_net_rules uses the pci busid, hwinfo used the virtio netcard busid.

Screenshot_testing_2019-03-18_09_31_42

## Solution

There are different solutions at different layers that could be used as workarounds, but maybe the less risky would be adding the parent_bus_id and use it also during the match, when using a virtio card that is what has been implemented.

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -284,10 +284,11 @@ module Yast
         item_id, matching_item = LanItems.Items.find do |_, i|
           next unless i["hwinfo"]
           busid = i["hwinfo"]["busid"]
+          # Match also parent busid if exist (bsc#1129012)
           parent_busid = i["hwinfo"]["parent_busid"] || busid
           mac = i["hwinfo"]["mac"]
 
-          [busid, parent_busid, mac].any? { |v| v.downcase == key }
+          [busid, parent_busid, mac].any? { |v| v.casecmp?(key) }
         end
         next if !matching_item
 
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -288,7 +288,7 @@ module Yast
           parent_busid = i["hwinfo"]["parent_busid"] || busid
           mac = i["hwinfo"]["mac"]
 
-          [busid, parent_busid, mac].any? { |v| v.casecmp?(key) }
+          [busid, parent_busid, mac].any? { |v| v.casecmp(key).zero? }
         end
         next if !matching_item
 
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -281,20 +281,24 @@ module Yast
         key.downcase!
 
         # find item which matches to the given rule definition
-        item, matching_item = LanItems.Items.find do |_, i|
-          i["hwinfo"] &&
-            (i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key)
+        item_id, matching_item = LanItems.Items.find do |_, i|
+          next unless i["hwinfo"]
+          busid = i["hwinfo"]["busid"]
+          parent_busid = i["hwinfo"]["parent_busid"] || busid
+          mac = i["hwinfo"]["mac"]
+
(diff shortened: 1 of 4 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -281,20 +281,24 @@ def assign_udevs_to_devs(udev_rules)
         key.downcase!
 
         # find item which matches to the given rule definition
-        item, matching_item = LanItems.Items.find do |_, i|
-          i["hwinfo"] &&
-            (i["hwinfo"]["busid"].downcase == key || i["hwinfo"]["mac"].downcase == key)
+        item_id, matching_item = LanItems.Items.find do |_, i|
+          next unless i["hwinfo"]
+          busid = i["hwinfo"]["busid"]
+          parent_busid = i["hwinfo"]["parent_busid"] || busid
```
mchf: this is eq to ```parent_busid = i["hwinfo"]["parent_busid"] || i["hwinfo"]["busid"]``` so you don't need to assign and compare ```busid``` bellow at all

teclator: If it have a parent busid it will be the parent_bus_id not the busid. so the busid need to be compared.

teclator: Basically:

1) No parent_busid, it becomes the busid, so it is like a duplicate matching that does not hurt

`["mac", "busid","busid"]`

2) parent_busid is present, it becomes the parent

`["mac", "parent_busid","busid"]`

mchf: i wasn't that clear how i should be.

what i wanted to say that it doesn't hurt but is useless ... in the end you compare either ```busid``` or ```parent_busid```. if i write differently what I wrote above then you can define ```busid_to_use_in_comparison = i["hwinfo"]["parent_busid"] || i["hwinfo"]["busid"]``` and then
````[busid_to_use_in_comparison, mac].any? { |v| v.downcase == key }``

Comment (teclator) [reply src/lib/network/network_autoyast.rb]: But we want to compare both

File: src/lib/network/network_autoyast.rb [hunk 2]
```diff
@@ -281,20 +281,24 @@ def assign_udevs_to_devs(udev_rules)
...
+          mac = i["hwinfo"]["mac"]
+
+          [busid, parent_busid, mac].any? { |v| v.downcase == key }
```
mchf: when touching it, then ```casecmp``` seems to be a better choice here.

src/lib/network/network_autoyast.rb
mchf: Ok, seems I'm getting lost.

Do they want to match device either according to busid or its parent busid? That's not a bug but 
feature and should be implemented as such by extending the AY rule to accept not only BUSID and MAC but even parent BUSID. 

Moreover, when matched by parent id I don't think it will generate udev rule as expected.

teclator: Have you seen the trello card then? I should update the description :)

teclator: > Ok, seems I'm getting lost.
> 
> Do they want to match device either according to busid or its parent busid? That's not a bug but
> feature and should be implemented as such by extending the AY rule to accept not only BUSID and MAC but even parent BUSID.
> 
> Moreover, when matched by parent id I don't think it will generate udev rule as expected.

We will replace the udev rule by the one defined in the profile. I already tested it by hand and the user also verified the patch

mchf: i just want to understand it as much as possible ... bcs i dislike it ;-)

pls add some explaining comment or reference. Mainly expressing that virtio is sometime identified by parentid and sometime by virtio id

PR #740, opened 2019-03-20
Writing changes

imobachgs:


src/lib/y2network/config_writer.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  module ConfigWriter
+    # Config writer for a given source
+    #
+    # @param source [Symbol] Source name (e.g., :sysconfig)
+    # @return [#config] Configuration writer from {Y2Network::ConfigReader}
+    def self.for(source)
```
jreidinger: should be part of config be also type of network daemon? So we do not need to call stuff like `for(:wicked).write(config)` This looks a bit strange and especially as from my exp always there is few backend specific options in configuration ( e.g. in bootloader it is always the case ). So I think part of configuration should be also target backend.

jreidinger: btw if you are interested how I did it in bootloader. I create bootloader factory that can read current backend, propose new backend or simply get config for given backend. See https://github.com/yast/yast-bootloader/blob/master/src/lib/bootloader/bootloader_factory.rb

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -0,0 +1,54 @@
...
+        routes = config.routes.map { |r| route_to_hash(r) }
+        Yast::Routing.Import("routes" => routes)
+      end
+
+    private
+
+      def route_to_hash(route)
+        hash =
+          if route.default?
+            { "destination" => "-", "netmask" => "-" }
```
jreidinger: Hmm, I also have in UI so many times this `"-"`. I am not sure how to handle it, but I do not like it much, that it is spread everywhere.

imobachgs: I do not like it either, but for the time being it is how `Yast::Routing` works and this writer basically imports the configuration into `Yast::Routing`.

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -0,0 +1,70 @@
...
+            { "destination" => route.to.to_s, "netmask" => netmask(route.to) }
+          end
+        hash.merge(
+          "gateway" => route.gateway ? route.gateway.to_s : "-",
+          "device"  => route.interface == :any ? "-" : route.interface.name
+        )
+      end
+
+      IPV4_MASK = "255.255.255.255".freeze
+      IPV6_MASK = "fffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff".freeze
```
mchf: i still think that we should use prefixes instead of mask whenever possible. It is more common to write  1.2.3.4/16 then using mask. Moreover in IPv6 prefix is used exclusively ... here you see why.

imobachgs: Yes, I agree :) In this case I am just using it to get the mask string representation which is expected by the underlying `Yast::Routing` module.

src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -81,12 +81,12 @@ def build_route(interfaces, hash)
         )
       end
 
-      # Given an IP and a netmaks, returns a valid IPAddr objecto
+      # Given an IP and a netmask, returns a valid IPAddr object
       #
       # @param ip_str      [String] IP address; {MISSING_VALUE} means that the IP is not defined
       # @param netmask_str [String] Netmask; {MISSING_VALUE} means than no netmaks was specified
```
teclator:
```suggestion
      # @param netmask_str [String] Netmask; {MISSING_VALUE} means that no netmask was specified
```

PR #745, opened 2019-03-26
Add an API to access configurations through Yast::Lan

imobachgs:
Small API to access to network configurations.


teclator:
LGTM

PR #747, opened 2019-03-27
Routing ui rework

jreidinger:
there is a bunch of TODO. Initial review

only qt images. ncurses fit, but my ncurses is broken and screenshots looks bad.
Snímek obrazovky_2019-03-27_12-10-08
Snímek obrazovky_2019-03-27_12-08-14


src/lib/y2network/widgets/routing_buttons.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def label
+        _("&Edit")
+      end
+
+      def handle
+        return nil unless @table.selected_route
+
+        route = @table.selected_route.dup
```
imobachgs: It might cause problem with nested objects. However, I've implemented a method to properly clone the config objects (not ready yet).

jreidinger: as I assign new objects to route it should not be problem. As I do not modify IPAddr or Interface and instead create new object. So I just switching references, not modifying nested objects. But if in future we need to modify object that is part of table, we need to do it.

src/lib/y2network/widgets/destination.rb
```diff
@@ -24,8 +24,11 @@ def contents
       end
 
       def help
-        # TODO: original also does not have help, so write new one
-        ""
+        _(
+          "<p><b>Default Route</b> matches all destination for given IP " \
```
imobachgs: `a given IP`

src/lib/y2network/widgets/destination.rb
```diff
@@ -24,8 +24,11 @@ def contents
...
+            "version as long as no specific route matches. <b>Destination</b>" \
+            "specifies IP address with prefix for which route applies.</p>"
```
imobachgs: what about *specifies the IP address (in CIDR format) for which the route applies* ? Using *prefix* instead of *CIDR*  is fine for me. Just not the *the*.

src/lib/y2network/widgets/gateway.rb
```diff
@@ -21,8 +21,12 @@ def label
       end
 
       def help
-        # TODO: original also does not have help
-        ""
+        _(
+          "<p><b>Gateway</b> defines gateway for given route. Special value here is '-' " \
+            "which means do not use gateway for given destination which for non direct " \
+            "connections result in reject of connection. For direct destination route is " \
```
mchf: an sometime is used when creating so called device routes. Networking is full of exceptions and corner cases which are not interesting for common user and which we may be should try to hide under hood. In my POV this is what yast should help with. We can simply allow user to define rejecting route (or device route) in a some way, but advanced user should know what is he doing and basic user is not interested in this '-' stuff and only complicates things for him.

src/lib/y2network/widgets/route_options.rb
```diff
@@ -18,8 +18,10 @@ def label
       end
 
       def help
-        # TODO: original also does not have help
-        ""
+        _(
+          "<p><b>Options</b> specifies additional options for route. It is directly passed " \
+            "to <i>ip route add</i> with exception of <i>to</i>,<i>via</i> and <i>dev</i>."
```
mchf: NP: i'd wirte something like "passed to routing" ... we don't need to expose particular command used - what if it changes in the future again ;-)

src/lib/y2network/widgets/destination.rb
```diff
@@ -24,8 +24,11 @@ def contents
       end
 
       def help
-        # TODO: original also does not have help, so write new one
-        ""
+        _(
+          "<p><b>Default Route</b> matches all destination for given IP " \
+            "version as long as no specific route matches. <b>Destination</b>" \
```
mchf: NP: we usually say IP address family

src/lib/y2network/widgets/destination.rb
mchf: destination specifies host or subnet

src/lib/y2network/widgets/gateway.rb
```diff
@@ -21,8 +21,12 @@ def label
       end
 
       def help
-        # TODO: original also does not have help
-        ""
+        _(
+          "<p><b>Gateway</b> defines gateway for given route. Special value here is '-' " \
```
teclator: defines the IP address of a host which routes the packets to a remote host or network. It can be blank for rejection or device routes.

src/lib/y2network/widgets/gateway.rb
jreidinger: @mchf but we need have at least in help in which format they can pass it. '-' is known only for users familiar with ifroute. It is not general rejecting route keyword, not?

src/lib/y2network/widgets/route_options.rb
jreidinger: well, that is useless from my POV. User need to know which options he can use. If we use generic route options, then we need to define what we support. If we specify command they know what they can use, if we do not specify it, they will find it themself and if we change it, it stop working for them, which is pretty annoying and not much friendly.

PR #748, opened 2019-03-28
(WIP) Autoyast Routing

teclator:


src/lib/y2network/autoyast_routing.rb
```diff
@@ -0,0 +1,57 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/routing_helpers"
+
+Yast.import "Lan"
+
+module Y2Network
+  # This class is responsible of importing and exporting AutoYast routing
+  # configuration
+  class AutoyastRouting
```
teclator: @jreidinger @imobachgs @mchf 

`RoutingAutoyast` probably should be better, although what about using another adding an extra namespace level here? Something like `Y2Network::Routing::Autoyast` ...`Y2Network::Routing::Config` `Y2Network::Routing::Route`, I know it is not a loved pattern so ... what do you think?

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+    # @param route [Y2Network::Route]
+    # @return [Hash]
+    def route_to_hash(route)
+      hash =
+        if route.default?
+          { "destination" => DEFAULT_DEST, "netmask" => "-" }
+        else
+          { "destination" => route.to.to_s, "netmask" => netmask(route.to) }
+        end
+      hash.merge("options" => route.options) unless route.options.to_s.empty?
```
teclator: `extrapara` was used in the past, so it will break the current AY schema.

src/lib/y2network/autoyast_routing.rb
jreidinger: Well, I am not so big fun of that deep namespaces, as it makes names resolution quite tricky. Consider this example:
```
# we already have this classes
Y2Network::Route
Y2Network::Dialogs::Route

# and consider this code:
module Y2Network
  module Dialogs
    class A
       Route.new # which class it will be? It depends on require and if you forget to require dialogs/route then it will be backend Route.
```

So to sum it, I prefer less namespaces unless it is really logical separation.

src/lib/y2network/routing_helpers.rb
jreidinger: so change that schema?

src/lib/y2network/autoyast_routing.rb
```diff
@@ -0,0 +1,57 @@
...
+    include Y2Network::RoutingHelpers
```
jreidinger: To be honest, always I see *Helpers object it rings bell in my head like it is next utils, complex or routines include that holds mess that does not fit elsewhere. I prefer more focused mixins name.

src/lib/y2network/clients/routing_auto.rb
```diff
@@ -0,0 +1,102 @@
...
+      def initialize
+        textdomain "network"
+      end
+
+      # Import routing configuration
+      #
+      # @param profile [Hash] routing profile section to be imported
+      # @return [Boolean]
+      def import(profile = {})
+        return unless config
```
jreidinger: `unless config` looks like typo.

src/lib/y2network/clients/routing_auto.rb
```diff
@@ -0,0 +1,102 @@
...
+
+        config.routing = ay_routing.import(profile)
+        modified!
+        config.routing
```
jreidinger: it should return boolean according to doc, not?

src/lib/y2network/clients/routing_auto.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      # Convenience method to obtain current network configuration
+      #
+      # @return [Config] current network config
+      def config
+        Yast::Lan.find_config
+      end
+
+      # @return [AutoyastRouting]
+      def ay_routing
```
jreidinger: maybe call it just `autoyast`?

jreidinger: or something like `serializer`

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+    DEFAULT_DEST = "default".freeze
+    private_constant :DEFAULT_DEST
+
+    # Find configured network interfaces
+    #
+    # Configured interfaces has a configuration (ifcfg file) assigned.
+    #
+    # @return [Array<Interface>] Detected interfaces
+    # @see Yast::NetworkInterfaces.Read
+    def find_interfaces
```
jreidinger: maybe have this method in `Y2Network::Interface`?

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      ip = IPAddr.new(ip_str)
+      netmask_str == MISSING_VALUE ? ip : ip.mask(netmask_str)
+    end
+
+    # Build a route given a hash from the SCR agent
+    #
+    # @param hash [Hash] Hash from the `.routes` SCR agent
+    # @return Route
+    def build_route(hash)
```
jreidinger: something for `Y2Network::Route` ?

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        Y2Network::Interface.new(name)
+      end
+    end
+
+    # Given an IP and a netmask, returns a valid IPAddr object
+    #
+    # @param ip_str      [String] IP address; {MISSING_VALUE} means that the IP is not defined
+    # @param netmask_str [String] Netmask; {MISSING_VALUE} means that no netmask was specified
+    # @return [IPAddr,nil] The IP address or `nil` if the IP is missing
+    def build_ip(ip_str, netmask_str = MISSING_VALUE)
```
jreidinger: when seeing this, can we have Y2Network::IP that is child of IPAddr and provide additional methods like this?

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+        gateway:   build_ip(hash["gateway"], MISSING_VALUE),
+        options:   hash["extrapara"]
+      )
+    end
+
+    # Returns a hash containing the route information to be imported into {Yast::Routing}
+    #
+    # @param route [Y2Network::Route]
+    # @return [Hash]
+    def route_to_hash(route)
```
jreidinger: something for Y2Network::Route?

src/lib/y2network/routing_helpers.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    IPV4_MASK = "255.255.255.255".freeze
+    IPV6_MASK = "fffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff".freeze
+
+    # Returns the netmask
+    #
+    # @param ip [IPAddr]
+    # @return [IPAddr]
+    def netmask(ip)
```
jreidinger: something for Y2Network::IP as mentioned above?

src/lib/y2network/routing_helpers.rb
teclator: @imobachgs is working in a presenter, we should maintain it as it is, we could also export as "options" if decided but then we will need to import using both keys

src/lib/y2network/autoyast_routing.rb
teclator: :) I though it was focused enough, but sure, we could think in other options for sharing these common methods

src/lib/y2network/clients/routing_auto.rb
teclator: Let me check, config was a convenience method which retrieved the current Y2Network::Config, that is Yast::Lan.find_config, so if there was no config then import was not done.

Of course, that is still [WiP], in that case we probably should just create a new one. So open to discussions

src/lib/y2network/clients/routing_auto.rb
teclator: Yes, I just returned routing for testing, but will modify it. The ay_routing will return A [Routing] object

src/lib/y2network/clients/routing_auto.rb
teclator: yep, was not sure about naming, so suggestions are welcomed :)

src/lib/y2network/routing_helpers.rb
teclator: Maybe, and this basically is configured interfaces. BTW, for autoyast It should use the config.interfaces as them should be them already imported  or read by Lan

src/lib/y2network/routing_helpers.rb
teclator: Take in account that this methods are auxiliar for routing, that is, for reading and writing configuration but the missing an default values probably does not make sense in other scenarios. And that is why them are part of routing helpers, basically I moved from the sysconfig routing readers because Autoyast basically needed to do create the same structure than the readers.

Autoyast could be seen as a reader from that POV and instead of a mixin a base class could be even used, but prefered to have a serializer with export and import instead of two separate classes for reading and writing.

src/lib/y2network/routing_helpers.rb
teclator: MM maybe, but then it should be something like build_from() ... I think will just let the current methods in the original readers

src/lib/y2network/clients/routing_auto.rb
jreidinger: ah, I though that you want to check that import in not nil. My bad.

teclator:
We have introduced `AutoinstProfile` #751 and with them, the logic has changed a little bit, so most of the comments are out of date, for that reason will create a complete new and clean PR.

PR #749, opened 2019-03-28
Network ng add presenters

imobachgs:
As part of the effort to drop the `Routing` module, I have added some presenter classes to convert the routing information into a section for the 1) AutoYaST profile and the 2) network summary. The latter is a WIP which will be finished as part of dropping the `Routing` module.

### DIFF
--- a/src/lib/y2network/presenters/route_profile.rb
+++ b/src/lib/y2network/presenters/route_profile.rb
@@ -1,0 +1,83 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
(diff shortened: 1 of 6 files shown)


src/lib/y2network/presenters/route_profile.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      # @return [Y2Network::Route]
+      attr_reader :route
+
+      # Returns the destination for the given route
+      #
+      # @param route [Route] Route to get the destination from
+      # @return [String] Route destination
+      def destination(route)
+        return "default" if route.to == :default
```
jreidinger: when I see how common is this comparison, can we add helper like `default?`

imobachgs: oh, sorry, the helper is already there (I just forgot about it) :)

imobachgs:
As discussed in IRC, we will change the approach. So I am closing the PR.

PR #750, opened 2019-03-29
Reading from ifroute-*

mchf:


src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -33,9 +33,14 @@ class Sysconfig
...
         # load /etc/sysconfig/network/routes
-        routing = Y2Network::Routing.new(tables: [SysconfigRoutesReader.new.config])
-        # TODO: SysconfigRoutesReader(s) for ifroute-* files
-        Config.new(interfaces: interfaces, routing: routing, source: :sysconfig)
+        routes = SysconfigRoutesReader.new.config
+        # load /etc/sysconfig/network/ifroute-*
+        dev_routes = interfaces.map do |iface|
+          SysconfigRoutesReader.new(routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}").config
+        end
+        routing = dev_routes.inject(routes) { |memo, r| memo.concat(r.routes) }.uniq
```
imobachgs: I would not call this variable `routing` as it might be confused with a `Y2Network::Routing` instance.

imobachgs: Additionally, I would probably move this logic to a `find_routes` (sure you can come up with a better name) method to hide the complexity a bit (analogous to the `find_interfaces` one). Tomorrow we will need to read DNS info, etc. and this module will grow pretty quickly.

src/lib/y2network/config_reader/sysconfig_routes_reader.rb
```diff
@@ -114,9 +118,73 @@ def build_route(hash)
...
           to:        hash["destination"] != DEFAULT_DEST ? build_ip(hash["destination"], mask) : :default,
           interface: iface,
-          gateway:   build_ip(hash["gateway"], MISSING_VALUE)
+          gateway:   build_ip(hash["gateway"], MISSING_VALUE),
+          options:   hash["extrapara"] || ""
         )
       end
+
+      # SCR agent for routes files definition
+      def ifroute_term(path)
```
imobachgs: Not for now, but in the future we might want to use CFA (or something like that).

src/lib/y2network/config_reader/sysconfig_routes_reader.rb
```diff
@@ -114,9 +118,73 @@ def build_route(hash)
...
+        raise ArgumentError if path.nil? || path.empty?
+
+        non_empty_str_term = Yast::term(:String, "^ \t\n")
```
imobachgs: Aren't these terms just constants?

src/lib/y2network/config_reader/sysconfig_routes_reader.rb
```diff
@@ -114,9 +118,73 @@ def build_route(hash)
...
+      end
+
+      # Registers SCR agent which is used for accessing particular ifroute-device
+      # file
+      #
+      # @param device [String] full path to a file in routes format (e.g. /etc/sysconfig/network/ifroute-eth0)
+      # @return [Path] SCR path of the agent
+      # @raise  [RuntimeError] if it fails
+      def register_ifroute_agent_for_path(path)
+        # /etc/sysconfig/network/ifroute-eth0 define .ifroute-eth0 agent
```
imobachgs: Please, add a `TODO` or `FIXME`.

src/lib/y2network/config_reader/sysconfig_routes_reader.rb
```diff
@@ -114,9 +118,73 @@ def build_route(hash)
...
+        # collisions not handled
+        scr_path = Yast::Path.new(".#{File.basename(path)}")
+        Yast::SCR.RegisterAgent(scr_path, ifroute_term(path)) ||
```
imobachgs: Almost four years in the team and I am still learning new things ;-)

mchf: yeah ... isn't our job beautiful? ;-)

src/lib/y2network/config_reader/sysconfig_routes_reader.rb
mchf: yes, but this is tightly coupled with SCR and old good yast agents. Not useful elsewhere.

PR #751, opened 2019-03-29
Network ng autoinst profile

imobachgs:
This classes implements the same approach followed in storage-ng to handle the conversion to/from an AutoYaST profile. The idea is:

* Model the networking section of the AutoYaST profile using OOP (stop thinking about AutoYaST as a big hash where possible).
* Keeping the import/export logic in the same place. It can do conversions if needed.
* Building real Y2Network::Config objects is out of the scope of these classes. It should belong to `ConfigReader+  (which makes me think that `ConfigBuilder` was a better name).

I've decided to keep the method names as similar as possible to the storage-ng implementation. On the other hand, the `Y2Network::AutoinstProfile::SectionWithAttributes` is basically copied&pasted from storage-ng and, in the future, it should be moved to somewhere else (probably to `autoyast2`).

## Using the OOP layer

Converting from an AutoYaST profile:

```ruby
def Import(settings)
  networking = Y2Network::AutoinstProfile::NetworkingSection.new_from_hashes(settings)
  # Now you can do whatever you want with the information in the profile. For instance, building
  # a Y2Network::Config object using a ConfigReader class.
end
```

And exporting:

```ruby
def Export(settings)
  config = find_config(id: :yast)
  networking = Y2Network::AutoinstProfile::NetworkingSection.new_from_network(config)
  networking.to_hashes
end
```

### DIFF
--- a/src/lib/y2network/autoinst_profile/networking_section.rb
+++ b/src/lib/y2network/autoinst_profile/networking_section.rb
@@ -1,0 +1,42 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Network
+  module AutoinstProfile
+    class NetworkingSection
+      # @return [RoutingSection]
+      attr_accessor :routing
+
+      # Creates an instance based on the profile representation used by the AutoYaST modules
+      # (hash with nested hashes and arrays).
+      #
+      # @param hash [Hash] Networking section from an AutoYaST profile
+      # @return [NetworkingSection]
+      def self.new_from_hashes(hash)
+        result = new
+        result.routing = RoutingSection.new_from_hashes(hash["routing"]) if hash["routing"]
+        result
+      end
+
+      def to_hashes
+        { "routing" => routing.to_hashes }
+      end
+    end
+  end
+end
--- a/src/lib/y2network/autoinst_profile/networking_section.rb
+++ b/src/lib/y2network/autoinst_profile/networking_section.rb
@@ -19,6 +19,15 @@
 
 module Y2Network
   module AutoinstProfile
+    # This class represents an AutoYaST <networking> section
+    #
+    #  <networking>
+    #    <routing>
+    #      <!-- the routing configuration -->
+    #    </routing>
+    #  </networking>
+    #
+    # @see RoutingSection
     class NetworkingSection
       # @return [RoutingSection]
       attr_accessor :routing
@@ -44,6 +53,9 @@ module Y2Network
         result
       end
 
+      # Export the section to a hash so it might be used when cloning the system
+      #
+      # @return [Hash]
       def to_hashes
         { "routing" => routing.to_hashes }
       end
--- a/src/lib/y2network/autoinst_profile/networking_section.rb
+++ b/src/lib/y2network/autoinst_profile/networking_section.rb
@@ -17,6 +17,8 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "y2network/autoinst_profile/routing_section"
+
 module Y2Network
   module AutoinstProfile
     # This class represents an AutoYaST <networking> section
--- a/src/lib/y2network/autoinst_profile/networking_section.rb
+++ b/src/lib/y2network/autoinst_profile/networking_section.rb
@@ -34,6 +34,16 @@ module Y2Network
         result
       end
 
+      # Creates an instance based on the network configuration representation
+      #
+      # @param config [Y2Network::Config]
+      # @return [NetworkingSection]
+      def self.new_from_network(config)
+        result = new
+        result.routing = RoutingSection.new_from_network(config.routing)
+        result
+      end
+
       def to_hashes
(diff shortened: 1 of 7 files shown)


src/lib/y2network/autoinst_profile/networking_section.rb
```diff
@@ -0,0 +1,64 @@
...
+        result.routing = RoutingSection.new_from_network(config.routing)
+        result
+      end
+
+      # Export the section to a hash so it might be used when cloning the system
+      #
+      # @return [Hash]
+      def to_hashes
+        { "routing" => routing.to_hashes }
+      end
```
jreidinger: I just notice here asymmetric that here is missing `to_network` which will return new Y2Network::Config.

src/lib/y2network/autoinst_profile/routing_section.rb
```diff
@@ -0,0 +1,112 @@
...
+      # @param hash [Hash] see {.new_from_hashes}
+      def init_from_hashes(hash)
+        super
+        @routes = routes_from_hash(hash)
+      end
+
+      # Method used by {.new_from_network} to populate the attributes when cloning routing settings
+      #
+      # @param routing [Y2Network::Routing] Network settings
+      # @return [Boolean]
```
jreidinger: what means boolean here?

src/lib/y2network/autoinst_profile/networking_section.rb
imobachgs: This classes only handles the AutoŸaST profile (importing/exporting/converting). The `ConfigReader::Autoyast` that @teclator is implementing should be responsible for that.

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+          result = new(parent)
+          result.init_from_hashes(hash)
+          result
+        end
+
+      protected
+
+        # Macro used in the subclasses to define accessors for all the
+        # attributes defined by {.attributes}
+        def define_attr_accessors
```
jreidinger: I play a bit with `inherited` callbacks so not all class need to define it themself, but it is sadly called before class is defined, so cannot be used. Maybe we found nicer solution then to call it in all subclasses? Maybe method missing or OpenStruct?

src/lib/y2network/autoinst_profile/routing_section.rb
```diff
@@ -0,0 +1,112 @@
...
+      # @param routing [Y2Network::Routing] Routing settings
+      # @return [RoutingSection]
+      def self.new_from_network(routing)
+        result = new
+        initialized = result.init_from_network(routing)
+        initialized ? result : nil
+      end
+
+      # Constructor
+      def initialize(_parent = nil)
```
jreidinger: just note when you just need to pass all arguments to parent, you call use simple `*args` and pass it. So in future you do not need to modify this method if parent need to change arguments.

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+    protected
+
+      def attributes
+        self.class.attributes
+      end
+
+      # Whether an attribute must be skipped during import/export.
+      #
+      # @return [Boolean] true is the value is blank
+      def attrib_skip?(value)
```
jreidinger: I have to say that this accronym is a bit strange. I expect `attr` or full `attribute` this is quite long and still not full.

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+        value.nil? || value == [] || value == ""
```
jreidinger: maybe const and include?
```
VALUES_TO_SKIP = [ nil, [], ""]

def attr_skip?(v)
  VALUES_TO_SKIP.include?(v)
end
```

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+      end
+
+      def attrib_key(attrib)
+        (attrib[:xml_name] || attrib[:name]).to_s
+      end
+
+      def attrib_value(attrib)
+        value = send(attrib[:name])
```
jreidinger: public_send ?

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+        if value.is_a?(Array)
+          value.map { |v| attrib_scalar(v) }
+        else
+          attrib_scalar(value)
+        end
+      end
+
+      def attrib_scalar(element)
+        element.respond_to?(:to_hashes) ? element.to_hashes : element
```
jreidinger: I am not against smaller methods, but why it is separated and not part of attrib_key?

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,189 @@
...
+
+          # This method only reads scalar values
+          next if value.is_a?(Array) || value.is_a?(Hash)
+
+          if attrib_skip?(value)
+            log.debug "Ignored blank value (#{value}) for #{key}"
+            next
+          end
+
+          send(:"#{name}=", value)
```
jreidinger: public_send ?

src/lib/y2network/autoinst_profile/section_with_attributes.rb
imobachgs: I am usually fine with calling the method explicitly and avoiding magic :) However, I agree that we could find a better way.

jreidinger: calling method if you need something is fine for me. But having to call it always in each child looks like unnecessary syntax glue for me.

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -66,8 +66,8 @@ def new_from_hashes(hash, parent = nil)
         # Macro used in the subclasses to define accessors for all the
         # attributes defined by {.attributes}
         def define_attr_accessors
-          attributes.each do |attrib|
-            attr_accessor attrib[:name]
+          attributes.each do |attribute|
```
jreidinger: NP: btw it can one line `attr_accessor *attributes.map{ |a| a[:name] }`

PR #752, opened 2019-03-29
Drop Routing module from Yast::Lan

imobachgs:
This PR removes Yast::Routing references from Yast::Lan. Ideally, we should get rid of this module in other parts of the code as soon as possible, but it is elsewhere and I guess that's something that will take some time.

### DIFF
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -43,8 +43,9 @@ module Y2Network
 
     class << self
       # @param source [Symbol] Source to read the configuration from
-      def from(source)
-        reader = ConfigReader.for(source)
+      # @param opts   [Hash]   Reader options
+      def from(source, opts = {})
+        reader = ConfigReader.for(source, opts)
         reader.config
       end
     end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -69,5 +69,12 @@ module Y2Network
     def write
       Y2Network::ConfigWriter.for(source).write(self)
     end
+
+    # Returns a deep-copy of the configuration
+    #
+    # @return [Config]
+    def copy
+      Marshal.load(Marshal.dump(self))
+    end
   end
 end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -43,7 +43,8 @@ module Y2Network
 
     class << self
       # @param source [Symbol] Source to read the configuration from
-      # @param opts   [Hash]   Reader options
+      # @param opts   [Hash]   Reader options. Check readers documentation to find out
+      #                        supported options.
       def from(source, opts = {})
         reader = ConfigReader.for(source, opts)
         reader.config
@@ -79,7 +80,8 @@ module Y2Network
     #
     # @return [Boolean] true if both configurations are equal; false otherwise
     def ==(other)
-      source == other.source && (interfaces - other.interfaces).empty? &&
+      source == other.source &&
+        ((interfaces - other.interfaces) | (other.interfaces - interfaces)).empty? &&
         routing == other.routing
     end
 
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -35,9 +35,9 @@ module Y2Network
   #   config.write
   class Config
     # @return [Symbol] Configuration ID
-    attr_reader :id
+    attr_accessor :id
     # @return [Array<Interface>]
-    attr_reader :interfaces
+    attr_accessor :interfaces
     # @return [Routing]
     attr_reader :routing
     # @return [Symbol] Information source (see {Y2Network::Reader} and {Y2Network::Writer})
@@ -76,5 +76,15 @@ module Y2Network
     def copy
       Marshal.load(Marshal.dump(self))
     end
+
+    # Determines whether two configurations are equal
+    #
+    # @return [Boolean] true if both configurations are equal; false otherwise
+    def ==(other)
+      source == other.source && (interfaces - other.interfaces).empty? &&
+        routing == other.routing
+    end
+
+    alias_method :eql?, :==
   end
 end
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -34,8 +34,6 @@ module Y2Network
   #   config.routing.tables.first << route
   #   config.write
   class Config
-    # @return [Symbol] Configuration ID
-    attr_accessor :id
     # @return [Array<Interface>]
     attr_accessor :interfaces
     # @return [Routing]
@@ -53,11 +51,10 @@ module Y2Network
 
     # Constructor
     #
-    # @param id         [Symbol] Configuration ID
     # @param interfaces [Array<Interface>] List of interfaces
     # @param routing    [Routing] Object with routing configuration
-    def initialize(id: :system, interfaces:, routing:, source:)
-      @id = id
+    # @param source     [Symbol] Configuration source
+    def initialize(interfaces:, routing:, source:)
       @interfaces = interfaces
       @routing = routing
       @source = source
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -31,7 +31,7 @@ module Y2Network
       require "y2network/config_reader/#{source}"
       name = source.to_s.split("_").map(&:capitalize).join
       klass = const_get(name)
-      klass.new
+      klass.new(opts)
     end
   end
 end
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -25,8 +25,9 @@ module Y2Network
     # Config reader for a given source
     #
     # @param source [Symbol] Source name (e.g., :sysconfig)
+    # @param opts   [Hash] Reader options
     # @return [#config] Configuration reader from {Y2Network::ConfigReader}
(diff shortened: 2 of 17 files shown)


src/lib/y2network/config.rb
```diff
@@ -34,30 +34,28 @@ module Y2Network
...
     # @return [Routing]
     attr_reader :routing
     # @return [Symbol] Information source (see {Y2Network::Reader} and {Y2Network::Writer})
     attr_reader :source
 
     class << self
       # @param source [Symbol] Source to read the configuration from
-      def from(source)
-        reader = ConfigReader.for(source)
+      # @param opts   [Hash]   Reader options
```
jreidinger: I think this will need a bit more description, like what options is supported and so on. Now I have now idea what it is good for.

teclator:
for me it looks fine

src/lib/y2network/config.rb
```diff
@@ -69,5 +67,22 @@ def initialize(id: :system, interfaces:, routing:, source:)
...
+    # @return [Config]
+    def copy
+      Marshal.load(Marshal.dump(self))
+    end
+
+    # Determines whether two configurations are equal
+    #
+    # @return [Boolean] true if both configurations are equal; false otherwise
+    def ==(other)
+      source == other.source && (interfaces - other.interfaces).empty? &&
```
jreidinger: is `interfaces - other.interfaces` correct? it can lead to situation that `==` is order dependent, which is very unexpected. I would rather use `interfaces.sort == other.interfaces.sort` or maybe even with uniq if you want to be sure that duplicates does not break it.

src/lib/y2network/config_reader.rb
```diff
@@ -25,8 +25,9 @@ module ConfigReader
     # Config reader for a given source
     #
     # @param source [Symbol] Source name (e.g., :sysconfig)
+    # @param opts   [Hash] Reader options
     # @return [#config] Configuration reader from {Y2Network::ConfigReader}
-    def self.for(source)
+    def self.for(source, opts = {})
```
jreidinger: opts is not passed here. Is it correct?

src/lib/y2network/interface.rb
```diff
@@ -22,8 +22,19 @@ class Interface
...
+    # Constructor
+    #
+    # @param name [String] Interface name (e.g., "eth0")
     def initialize(name)
       @name = name
     end
+
+    # Determines whether two interfaces are equal
+    def ==(other)
+      return false unless other.respond_to?(:name)
```
jreidinger: hmm, I am not sure that this is expected. What about simple `other.is_a? Interface` ? This way it can be == for object that define name, which is not so uncommon and produce strange results. Or better what is intention of this lazy type check?

src/lib/y2network/presenters/routing_summary.rb
```diff
@@ -0,0 +1,44 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Network
+  module Presenters
+    # This class converts a routing configuration object into a hash to be used
+    # in an AutoYaST summary
+    class RoutingSummary
+      # @return [Y2Network::Config]
+      attr_reader :config
```
jreidinger: Is it needed? If so, then why?

src/lib/y2network/routing.rb
```diff
@@ -38,5 +38,16 @@ def initialize(tables:)
...
       tables.flat_map(&:to_a)
     end
+
+    # Determines whether two set of routing settings are equal
+    #
+    # @param other [Routing] Routing settings to compare with
+    # @return [Boolean]
+    def ==(other)
+      forward_ipv4 == other.forward_ipv4 && forward_ipv6 == other.forward_ipv6 &&
+        (tables - other.tables).empty?
```
jreidinger: same here. I really do not like assymetric `==` which can result that `a == b` but `b != a`

jreidinger:
```
irb(main):001:0> a = [1,2]
=> [1, 2]
irb(main):002:0> b = [2]
=> [2]
irb(main):003:0> (a-b).empty?
=> false
irb(main):004:0> (b-a).empty?
=> true
```

src/lib/y2network/routing_table.rb
```diff
@@ -42,5 +42,15 @@ class RoutingTable
     def initialize(routes = [])
       @routes = routes
     end
+
+    # Determines whether two routing tables are equal
+    #
+    # @param other [RoutingTable] Routing table to compare with
+    # @return [Boolean]
+    def ==(other)
+      routes == other.routes
```
jreidinger: and here check is order dependant for array. I found this a bit confusing. Do we want order dependent comparison or not?

src/modules/Lan.rb
```diff
@@ -1106,6 +1111,34 @@ def refresh_lan_items
...
+      presenter = Y2Network::Presenters::RoutingSummary.new(config.routing)
+      presenter.text(mode: mode.to_sym)
+    end
+
+    # Returns the system configuration
+    #
+    # Just a convenience method.
+    #
+    # @return [Y2Network::Config]
+    def running_config
```
jreidinger: why not `system_config` ? running indecates for me runtime configuration but we read only system one, not actually running one.

src/modules/Lan.rb
```diff
@@ -1106,6 +1111,34 @@ def refresh_lan_items
...
+      find_config(:system)
+    end
+
+    # Returns YaST configuration
+    #
+    # Just a convenience method.
+    #
+    # @return [Y2Network::Config]
+    def yast_config
```
jreidinger: why is this and above method private? It is exactly methods I will need in UI to read and modify configuration.

src/modules/Lan.rb
```diff
@@ -968,15 +972,16 @@ def dhcp_ntp_servers
     #
     # @param id [Symbol] Network configuration ID
     # @return [Y2Network::Config,nil] Network configuration with the given ID or nil if not found
-    def find_config(id: :system)
-      configs.find { |c| c.id == id }
+    def find_config(id)
+      configs[id]
```
jreidinger: Is possible to have here something like lazy loading here? I think it will be easier for routing module to simply ask for config and if it is not yet read, read it. And same with yast_config. If not available, copy system one.

src/lib/y2network/routing_table.rb
imobachgs: In this case, yes, I would like to have the same order (as it may affect how routes are applied).

src/modules/Lan.rb
imobachgs: For the time being, I have decided to read the current configuration with the rest of network related values in `#Import` or `#Read`.

src/lib/y2network/config.rb
```diff
@@ -79,7 +80,8 @@ def copy
     #
     # @return [Boolean] true if both configurations are equal; false otherwise
     def ==(other)
-      source == other.source && (interfaces - other.interfaces).empty? &&
+      source == other.source &&
+        ((interfaces - other.interfaces) | (other.interfaces - interfaces)).empty? &&
```
jreidinger: ok, this will work now, even I am not sure if it is more readable then `interfaces.sort.uniq == other.interfaces.sort.uniq`

src/lib/y2network/interface.rb
```diff
@@ -30,8 +30,11 @@ def initialize(name)
     end
 
     # Determines whether two interfaces are equal
+    #
+    # @param other [Interface,:any] Interface to compare with
+    # @return [Boolean]
     def ==(other)
-      return false unless other.respond_to?(:name)
+      return false if other == :any
```
jreidinger: much better now as it is more explicit.

src/lib/y2network/routing_table.rb
mchf: I'd say no. Routing is not firewall and system should find "best matching" route - routes are applied according to longest matching prefix.

src/modules/Lan.rb
```diff
@@ -110,7 +110,7 @@ def main
     def Modified
       return true if LanItems.GetModified
       return true if DNS.modified
-      return true if Routing.Modified
+      return true unless system_config == yast_config
```
mchf: NP: well ... for me ```if ... != ...``` is more readable ;-)

src/modules/Lan.rb
```diff
@@ -265,6 +265,10 @@ def Read(cache)
         return true
       end
 
+      system_config = Y2Network::Config.from(:sysconfig)
+      add_config(:system, system_config)
+      add_config(:yast, system_config.copy)
```
mchf: can we think about a COW approach? As Config object will grow this will slow module loading as it already happened in Hosts module (and lead to system crash or unusable system with /etc/hosts with > 1000 lines) which used similar approach.

jreidinger: I worry that if we want to do COW, we need something like https://evilmartians.com/chronicles/clowne-clone-ruby-models-with-a-smile as I expect there a lot of exceptions what can be dupped and what not.

src/lib/y2network/config.rb
imobachgs: I did not use `#sort` because I am not sure about the sorting logic... by name? by type? Does it really make sense to sort devices?

jreidinger: I think by name should be fine. Just to ensure stable ordering for comparison.

PR #753, opened 2019-04-01
Implemented reading of ip forwarding setup

mchf:


src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -32,13 +33,21 @@ class Sysconfig
...
       end
 
     private
 
+      # sysctl keys, used as *single* SCR path components below
+      IPV4_SYSCTL = "net.ipv4.ip_forward".freeze
+      IPV6_SYSCTL = "net.ipv6.conf.all.forwarding".freeze
+      # SCR paths
+      SYSCTL_AGENT_PATH = ".etc.sysctl_conf".freeze
+      SYSCTL_IPV4_PATH = SYSCTL_AGENT_PATH + ".\"#{IPV4_SYSCTL}\""
```
imobachgs: You should freeze `SYSCTL_AGENT_PATH` and `SYSCTL_IPV4_PATH` too.

test/y2network/config_reader/sysconfig_test.rb
```diff
@@ -109,4 +109,16 @@
...
   end
+
+  describe "#forward_ipv4?" do
+    it "returns true when IPv4 forwarding is allowed" do
+      expect(reader.send(:forward_ipv4?)).to be_truthy
+    end
+  end
+
+  describe "#forward_ipv6?" do
+    it "returns false when IPv6 forwarding is disabled" do
```
imobachgs: You should add a test when the forwarding is enabled.

test/y2network/config_reader/sysconfig_test.rb
```diff
@@ -109,4 +109,16 @@
       end
     end
   end
+
+  describe "#forward_ipv4?" do
+    it "returns true when IPv4 forwarding is allowed" do
```
imobachgs: Yo should add a test when the forwarding is disabled.

test/y2network/config_reader/sysconfig_test.rb
```diff
@@ -109,4 +109,16 @@
...
+      expect(reader.send(:forward_ipv4?)).to be_truthy
```
imobachgs: Why truthy? Please, if you expect `true`, then check for `true` (use `eq true` or `be true`). Otherwise, any value which is different to `false` and `nil` will be OK.

test/y2network/config_reader/sysconfig_test.rb
```diff
@@ -109,4 +109,16 @@
...
+    end
+  end
+
+  describe "#forward_ipv6?" do
+    it "returns false when IPv6 forwarding is disabled" do
+      expect(reader.send(:forward_ipv6?)).to be_falsy
```
imobachgs: Same than above: if you expect `false`, just check for false.

```ruby
expect(nil).to be_falsy # Passes
```

src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -32,13 +33,21 @@ class Sysconfig
...
+      SYSCTL_IPV6_PATH = SYSCTL_AGENT_PATH + ".\"#{IPV6_SYSCTL}\""
```
jreidinger: why not having it already as Yast::Path? It allows even + operand https://github.com/yast/yast-ruby-bindings/blob/master/src/ruby/yast/path.rb#L31

mchf: just copy-pasted ... I'll rework it

test/y2network/config_reader/sysconfig_test.rb
mchf: well, I'd need two sysctl files for that ... and i was lazy

test/y2network/config_reader/sysconfig_test.rb
```diff
@@ -112,13 +112,23 @@
...
+    it "returns false when IPv4 forwarding is disabled" do
+      allow(Yast::SCR).to receive(:Read).and_return("0")
+      expect(reader.send(:forward_ipv4?)).to be false
     end
   end
 
   describe "#forward_ipv6?" do
+    it "returns true when IPv6 forwarding is allowed" do
+      allow(Yast::SCR).to receive(:Read).and_return("1")
+      expect(reader.send(:forward_ipv4?)).to be true
```
imobachgs: I guess you meant `forward_ipv6`. On the other hand, you are testing a private method and that's wrong. I think that you should check whether the value returned by `#config` contains the right values. What do you think?

PR #754, opened 2019-04-01
use docker image from OBS

aschnell:
For https://trello.com/c/sVVOXYbx/878-5-build-the-docker-images-for-travis-in-obs.


aschnell:
The complain by codacy is a false positive: We *do want* to use the latest image.

PR #755, opened 2019-04-01
Added AutoYaST config reader / builder

teclator:
- Added AutoYaST config readers.
- Added new routing auto client (will be removed as seems to not be used at all)
- Relax routing schema attributes (do not export the attribute instead of export as a dash)


src/lib/y2network/clients/routing_auto.rb
```diff
@@ -0,0 +1,111 @@
...
+require "y2network/routing"
+require "y2network/autoinst_profile/networking_section"
+require "y2network/config_reader/autoyast"
+
+Yast.import "NetworkInterfaces"
+Yast.import "Lan"
+
+module Y2Network
+  module Clients
+    # This client is reponsible of the autoinstallation routing configuration.
```
teclator: It will be completely dropped

src/lib/y2network/config_reader/autoyast.rb
```diff
@@ -0,0 +1,66 @@
...
+      #
+      # @param section [AutoinstProfile::NetworkingSection]
+      def initialize(section)
+        @section = section
+      end
+
+      # @return [Y2Network::Config] Network configuration
+      def config
+        interfaces = find_interfaces
+        routing = section.routing ? AutoyastRouting.new(section.routing).config : nil
```
teclator: @imobachgs @jreidinger we started a discussion about the routing config specially when importing in Autoyast whithout reading the system. If the routing section is empty shall we instantiate a Y2Network::Routing config with defaults or just use `nil` as no routing section has been imported. In other case when exporting it will be different but somehow expected.

teclator: Adding the `nil` initialization means we will need to handle it correctly when writing, exporting .. etc

imobachgs: Perhaps I would rename this class to `AutoinstRouting` (like `AutoinstProfile`, `AutoinstProposal`, etc.).

src/lib/y2network/autoinst_profile/routing_section.rb
```diff
@@ -80,6 +80,9 @@ def initialize(*_args)
       # @param hash [Hash] see {.new_from_hashes}
       def init_from_hashes(hash)
         super
+        ip_forward = hash["ip_forward"]
+        @ipv4_forward = hash["ipv4_forward"] || ip_forward
```
imobachgs: You could do `@ipv4_forward = hash["ipv4_forward"] || hash["ip_forward"]` directly.

src/lib/y2network/config_reader/autoyast.rb
jreidinger: good question as I think we face it in many modules. Sadly I do not see yet good solution for this issue. I vote for defaults to make our code easier and result more predictable.

src/lib/y2network/clients/routing_auto.rb
imobachgs: Then I would say that it does not make sense to merge this.

src/lib/y2network/route.rb
```diff
@@ -33,13 +31,11 @@ class Route
     # @param to        [IPAddr,:default] Destination
     # @param interface [Interface,:any] Interface to associate the root to
     # @param gateway   [IPAddr,nil] Gateway IP
-    # @param source    [IPAddr,nil] Source IP
     # @param options   [String] Additional options
-    def initialize(to: :default, interface: :any, gateway: nil, source: nil, options: "")
```
teclator: source is not used by YaST, so get rid of it by now

src/lib/y2network/config_reader/autoyast_routing.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+    private
+
+      # Build a route given a route section
+      #
+      # @return [Array<Route>]
+      def build_routes
+        section.routes.map do |route_section|
+          Y2Network::Route.new(to:        route_section.destination || :default,
```
jreidinger: how looks default route? It use nil there or "-" or "default" ?

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -32,7 +36,11 @@ class Sysconfig
       # @param config [Y2Network::Config] Configuration to write
       def write(config)
         routes = config.routing.routes.map { |r| route_to_hash(r) }
-        Yast::Routing.Import("routes" => routes)
+        Yast::Routing.Import(
+          "ipv4_forward" => config.routing.forward_ipv4,
```
imobachgs: Can `config.routing` be `nil?`?

test/y2network/autoinst_profile/routing_section_test.rb
```diff
@@ -92,5 +92,33 @@
         expect(section.routes).to eq([])
       end
     end
+
+    context "when ip_forward is enabled" do
+      let(:hash) { { "ip_forward" => true } }
+
+      it "enables ip4_forward" do
```
imobachgs: `ipv4_forward`

test/y2network/autoinst_profile/routing_section_test.rb
```diff
@@ -92,5 +92,33 @@
...
+        section = described_class.new_from_hashes(hash)
+        expect(section.ipv4_forward).to eq(true)
+      end
+
+      it "enables ip6_forward" do
```
imobachgs: `ipv6_forward`

src/lib/y2network/config_writer/sysconfig.rb
teclator: That is what asked above, if we should handle situations like that.

src/lib/y2network/config_reader/autoyast_routing.rb
teclator: In the profile it should be "default" or IP[/prefix]:

```xml
 <route>
        <destination>default</destination>
        <device>eth0</device>
        <gateway>192.168.1.1</gateway>
        <netmask>-</netmask>
 </route>
```

So, the init_from_hashes for RouteSection is still pending. Will add it

teclator: I prepared it for being able to do

```xml
  <route>
    <destination>default</destination>
    <gateway>192.168.1.1</gateway>
  </route>
```
https://github.com/yast/yast-network/pull/755/files#diff-89a18cae535af72c8ff5a12849265230R268

But destination has to be always declared

jreidinger: ok, so in this line "to" will be `"default"`, which is not intended, it should be `:default`

src/lib/y2network/config_writer/sysconfig.rb
jreidinger: Well, I do not like much to have nil checks everywhere. If we need something like this, then at least use NullObject pattern that behave reasonably for usual cases and handle only specific ones.

src/lib/y2network/config_reader/autoyast_routing.rb
teclator: Yep, in the `RouteSection` object it is "default" and should be transformed to a symbol or to an `IPAddr` in case it is not "default"

teclator: I have seen in different places that an assumption of "-" for default was used and that is not correct.

https://github.com/yast/yast-network/pull/755/files#diff-2072ed0d82b988333aa0f3e3e802570cR55 (fixed)

See original validation:

https://github.com/yast/yast-network/blob/master/src/include/network/services/routing.rb#L189

So, take it into account in widgets

https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/widgets/destination.rb#L34
https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/widgets/destination.rb#L54

I will propose the fix as part of the PBI if you want although probably it will use route.default? instead of handling the content for that time.

jreidinger: widgets are already modified in my branch

src/lib/y2network/config_reader/autoyast.rb
teclator: Used `Autoinst` and `AutoinstRouting`

src/autoyast-rnc/networking.rnc
```diff
@@ -266,9 +266,9 @@ routes =
 route =
   element (route | listentry) {
     destination &
-    element netmask { text } &	#overloaded
-    element device { text } &	#overloaded
-    gateway
```
teclator: Allowed to avoid the declaration of `netmask`. `device` and `gateway`. So something like:
```xml
<route>
  <destination>default</destination>
  <device>eth0</device>
</route>
```

src/lib/y2network/config_reader/autoinst.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    private
+
+      # Find configured network interfaces
+      #
+      # Configured interfaces have a configuration (ifcfg file) assigned.
+      #
+      # @return [Array<Interface>] Detected interfaces
+      # @see Yast::NetworkInterfaces.Read
+      def find_interfaces
```
mchf: it looks same like what we already have in ConfigReader::Sysconfig ... is it time to create e.g. a mixin?

src/lib/y2network/autoinst_profile/route_section.rb
```diff
@@ -70,6 +70,20 @@ def self.new_from_network(route)
...
+      # Method used by {.new_from_hashes} to populate the attributes.
+      #
+      # @parm hash [Hash] see {.new_from_hashes}
+      # @return [Boolean]
+      def init_from_hashes(hash)
+        @destination = destination_from_hash(hash)
+        @gateway = gateway_from_hash(hash)
+        @netmask = netmask_from_hash(hash)
+        @device = device_from_hash(hash)
+        @extrapara = hash["extrapara"]
```
mchf: why not even extrapara_from_hash when everything else is done that way?

mchf:
in general, as this reads what user produces, i miss tests for malformed configuration (missing destination, typo in ip, incorrect netmask / prefix, ...)

src/lib/y2network/config_reader/autoinst.rb
teclator: @jreidinger @imobachgs 

I already created the `routing_helpers` mixin, that could be renamed to reader_helpers. There where some discussion about the current include abuse and a fear of finishing with the same. So, I decided to add duplication of code before adding a wrong abstraction.

So maybe it should be handled by `Y2Network::Interface` and a method like `Y2Network::Interface.configured`, what do you think?

src/lib/y2network/autoinst_profile/route_section.rb
teclator: Basically because extrapara is ommited while the rest of the params declare it of "-" when no present. So extrapara does not need a special handling although we could add it if needed.

src/lib/y2network/config_reader/autoinst.rb
jreidinger: yes, much better for me to have it as class method of Yast2::Interface.

PR #756, opened 2019-04-01
Fix RoutingTable#concat

imobachgs:
The problem lies in forwarding the `concat` method to the `@routes` array. Bear in mind that `#concat` does not belong to `Enumerable`.

### DIFF
--- a/src/lib/y2network/config_reader/sysconfig.rb
+++ b/src/lib/y2network/config_reader/sysconfig.rb
@@ -64,13 +64,19 @@ module Y2Network
       # return [RoutingTable] an object with routes
       def load_routes
         # load /etc/sysconfig/network/routes
-        routes = SysconfigRoutesReader.new.config
+        routing_table = SysconfigRoutesReader.new.config
         # load /etc/sysconfig/network/ifroute-*
-        dev_routes = find_interfaces.map do |iface|
-          SysconfigRoutesReader.new(routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}").config
+        dev_routing_tables = find_interfaces.map do |iface|
+          SysconfigRoutesReader.new(
+            routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}"
+          ).config
         end
 
-        dev_routes.inject(routes) { |a, e| a.concat(e.routes) }.uniq
+        dev_routing_tables.reduce(routing_table) do |rt, dev_rt|
+          rt.concat(dev_rt.routes)
+        end
+        routing_table.routes.uniq!
+        routing_table
       end
     end
   end
--- a/src/lib/y2network/routing_table.rb
+++ b/src/lib/y2network/routing_table.rb
@@ -37,10 +37,16 @@ module Y2Network
     attr_reader :routes
 
     def_delegator :@routes, :each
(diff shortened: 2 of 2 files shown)


src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -64,13 +64,19 @@ def find_interfaces
...
-        routes = SysconfigRoutesReader.new.config
+        routing_table = SysconfigRoutesReader.new.config
         # load /etc/sysconfig/network/ifroute-*
-        dev_routes = find_interfaces.map do |iface|
-          SysconfigRoutesReader.new(routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}").config
+        dev_routing_tables = find_interfaces.map do |iface|
+          SysconfigRoutesReader.new(
+            routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}"
+          ).config
         end
```
mchf: strictly speaking all this is for reading one routing table from different sources, so this "_table(s)" is misleading

src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -64,13 +64,19 @@ def find_interfaces
...
 
-        dev_routes.inject(routes) { |a, e| a.concat(e.routes) }.uniq
+        dev_routing_tables.reduce(routing_table) do |rt, dev_rt|
+          rt.concat(dev_rt.routes)
+        end
+        routing_table.routes.uniq!
+        routing_table
```
mchf: this is not needed, isn't it?

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: I would say that `SysconfigRoutesReader#config` returns a`RoutingTable`, right? So dev_routing_tables is basically an array of tables, right?

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: yes, if you want to return a `RoutingTable`. Another approach might to filter the routes when calling to `#concat` to be sure that they are not repeated.

src/lib/y2network/config_reader/sysconfig.rb
mchf: no. SysconfigRoutesReader returns one routing table. It is reading one routing table which is spread over multiple files. It read "routes" (a routing table entries) and puts them together into RoutingTable.

src/lib/y2network/config_reader/sysconfig.rb
mchf: ok

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: I might be missing something, but `find_interfaces.map` will return an array, although `SysconfigRoutes` returns a single table.

mchf: from system POV ```SysconfigRoutes``` returns a part of complete routing table. Real routing table is build from ifroute-* files + /etc/.../routes.

PR #758, opened 2019-04-03
Improve tests

jreidinger:


test/y2network/widgets/routing_table_test.rb
```diff
@@ -0,0 +1,113 @@
...
+require "cwm/rspec"
+
+require "y2network/route"
+require "y2network/widgets/routing_table"
+require "y2network/routing_table"
+
+describe Y2Network::Widgets::RoutingTable do
+  let(:routing_table) do
+    Y2Network::RoutingTable.new([
+                                  Y2Network::Route.new(to: IPAddr.new("127.0.0.1/24")),
```
imobachgs: np: This indentation looks weird.

test/y2network/widgets/routing_table_test.rb
```diff
@@ -0,0 +1,113 @@
...
+    it "returns Route object according to selected row in table" do
+      allow(Yast::UI).to receive(:QueryWidget)
+        .with(Id(subject.widget_id), :SelectedItems).and_return(Id(1))
+
+      expect(subject.selected_route).to eq Y2Network::Route.new(to: :default)
+    end
+  end
+
+  describe "#add_route" do
+    it "appends to routing table new route" do
```
imobachgs: "appends the new route to the routing table"
or at least "appends to routing table the new route".

test/y2network/widgets/routing_table_test.rb
jreidinger: maybe weird, but according to our rubocop style, as it indent next to `[`

imobachgs: In those cases I prefer to do something like:

```ruby
Y2Network::RoutingTable.new(
  [
     Y2Network::Route.new(to: IPAddr.new("127.0.0.1/24")),
     Y2Network::Route.new(to: IPAddr.new("127.0.0.1/24"))
  ]
)```

In this way, you avoid getting too long lines. But, as I've said, it is just a np.

PR #759, opened 2019-04-03
Y2Network::SysconfigRoutesFile

imobachgs:
While I was adjusting `ConfigWriter::Sysconfig` to not rely on `Yast::Routing`, I realized that I needed part of the logic which was present in `ConfigReader::SysconfigRoutesReader`. At that point, I decided to replace that class with a more object-oriented `SysconfigRoutesFile` class. It implements a CFA-like style API, although it is using agents and not CFA.

Additionally, this PR fixes a bug that caused that YaST" left uneeded "ifroute-*" files in the filesystem.


src/lib/y2network/config_reader/sysconfig.rb
```diff
@@ -69,21 +74,20 @@ def find_interfaces
...
-          SysconfigRoutesReader.new(
-            routes_file: "/etc/sysconfig/network/ifroute-#{iface.name}"
-          ).config
+      def find_routing_tables
+        main_routes = load_routes_from
+        iface_routes = find_interfaces.flat_map do |iface|
+          load_routes_from("/etc/sysconfig/network/ifroute-#{iface.name}")
         end
+        all_routes = main_routes + iface_routes
+        [Y2Network::RoutingTable.new(all_routes.uniq)]
```
jreidinger: is uniq really needed? is duplicates expected, working?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,195 @@
...
+  #   file.routes.size #=> 1
+  class SysconfigRoutesFile
+    DEFAULT_ROUTES_FILE = "/etc/sysconfig/network/routes".freeze
+
+    # @return [Array<Route>] Routes
+    attr_reader :routes
+
+    # @param path [String] File path
+    def initialize(path = nil)
+      path ||= DEFAULT_ROUTES_FILE
```
jreidinger: this looks strange, why not having DEFAULT_ROUTES_FILE as default parameter?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,195 @@
...
+      @file =
```
jreidinger: hmm, names are quite confusing here as `@file` is Path and `path` is string.

src/lib/y2network/sysconfig_routes_file.rb
imobachgs: In the beginning, I needed to distinguish between *no parameters* and the default value. But I've adjusted the code, and now it is not required anymore.

src/lib/y2network/config_reader/sysconfig.rb
imobachgs: Not sure, it was in the original code, so I just kept it.

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+      netmask_str == MISSING_VALUE ? ip : ip.mask(netmask_str)
+    end
+
+    # Build a route given a hash from the SCR agent
+    #
+    # @param hash [Hash] Hash from the `.routes` SCR agent
+    # @return Route
+    def build_route(hash)
+      # TODO: check whether the iface is configured in the system
+      iface = Interface.new(hash["device"])
```
teclator: TODO: It loads absence of device as an Interface with "-" name, so we should fix that initialization. When reading from routes it should be :any while when loading from `ifroute` it should be the interface name inferred from the file name

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+      # normalized SCR output contains either subnet mask or /<prefix length> under
+      # "netmask" key
+      # TODO: this should be improved in normalize_routes
+      mask = hash["netmask"] =~ /\/[0-9]+/ ? hash["netmask"][1..-1] : hash["netmask"]
```
jreidinger: hmm, why not simple `hash["netmask"].delete("/")`? It should be same as it is ip or `/prefix`

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+          Yast.term(
+            :Optional,
+            Yast.term(
+              :extrapara,
+              Yast.term(:String, "^\n")
+            )
+          )
+        ),
+        "\n"
+      )
```
jreidinger: up to this line all is just constant. Only the rest is path based. So maybe having it in constant or at least separated?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+
+      Yast.term(
+        :ag_anyagent,
+        Yast.term(
+          :Description,
+          Yast.term(:File, path),
+          "#\n",
+          false,
```
jreidinger: please add comment what this false means.

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+        ),
+        "\n"
+      )
+
+      Yast.term(
+        :ag_anyagent,
+        Yast.term(
+          :Description,
+          Yast.term(:File, path),
+          "#\n",
```
jreidinger: also add comment what is this param or make it clear with proper constant name.

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    # Registers SCR agent which is used for accessing particular ifroute-device
+    # file
+    #
+    # @param device [String] full path to a file in routes format (e.g. /etc/sysconfig/network/ifroute-eth0)
+    # @return [Path] SCR path of the agent
+    # @raise  [RuntimeError] if it fails
+    def register_ifroute_agent_for_path(path)
+      # /etc/sysconfig/network/ifroute-eth0 define .ifroute-eth0 agent
+      # TODO: collisions not handled
```
jreidinger: well, you can also unregister scr path. So in reality you can use some temporary scr path just for loading and then simply unregister it again. See real world example at https://github.com/yast/yast-bootloader/blob/78f31a9f30e42641c1147d199cac103aa1a4380f/src/lib/bootloader/sysconfig.rb#L144

test/y2network/sysconfig_routes_file_test.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    context "when there is no netmask" do
+      let(:netmask) { "-" }
+
+      before(:each) do
+        allow(Yast::SCR).to receive(:Read).and_return(nil)
+        allow(Yast::SCR).to receive(:Read).with(Yast::Path.new(".routes")).and_return([scr_route])
+      end
+
+      it "does not set destination netmask" do
```
jreidinger: hmm, please test case when destination contain prefix ( e.g. ipv6 supported way ) and netmask is empty. So something like `127.0.0.0/16 - - -`

jreidinger: in this case it should set properly destination with prefix.

src/lib/y2network/sysconfig_routes_file.rb
imobachgs: This code was just moved from the reader, but I will have a look anyway.

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -35,20 +35,19 @@ def write(config)
...
 
-          writer = if dev == :any
-            ConfigWriter::SysconfigRoutesWriter.new
+        # list of devices used by routes
+        devices = config.interfaces + [:any]
+
+        devices.each do |dev|
+          routes = find_routes_for(dev, config.routing.routes)
+          file = routes_file_for(dev)
+          if dev != :any && routes.empty?
```
jreidinger: please explain in comment why we do not remove file for any

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -97,6 +96,43 @@ def write_ipv6_forwarding(forward_ipv6)
...
+      #
+      # @see #find_routes_for_any
+      # @see #find_routes_for_iface
+      def find_routes_for(iface, routes)
+        iface == :any ? find_routes_for_any(routes) : find_routes_for_iface(iface, routes)
+      end
+
+      # Finds routes which are not tied to an specific interface
+      def find_routes_for_any(routes)
+        routes.select { |r| r.interface == :any }
```
jreidinger: hmm, more and more I see this `:any` everywhere for route, maybe we can have specific Interface instance for any with its method?

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -97,6 +96,43 @@ def write_ipv6_forwarding(forward_ipv6)
...
+        end
+      end
+
+      # Returns the routes file for a given interace
+      #
+      # @param iface  [Interface,:any] Interface to search routes for
+      # @return [Y2Network::SysconfigRoutesFile]
+      def routes_file_for(iface)
+        return Y2Network::SysconfigRoutesFile.new if iface == :any
+        Y2Network::SysconfigRoutesFile.new("/etc/sysconfig/network/ifroute-#{iface.name}")
```
jreidinger: not, sure, but can it make sense to have such path in Interface object? like iface.routing_file_path?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,270 @@
...
+
+    # Writes configured routes
+    #
+    # @return [Boolean] true on success
+    def save
+      # create if not exists, otherwise backup
+      if Yast::FileUtils.Exists(file_path)
+        Yast::SCR.Execute(
+          Yast::Path.new(".target.bash"),
+          "/bin/cp #{file_path} #{file_path}.YaST2save"
```
jreidinger: shell escaping? Or even better using cheetah here?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,270 @@
...
+        )
+      else
+        Yast::SCR.Write(Yast::Path.new(".target.string"), file_path, "")
+      end
+
+      with_registered_ifroute_agent(file_path) do |scr|
+        # work around bnc#19476
+        Yast::SCR.Write(Yast::Path.new(".target.string"), file_path, "")
```
jreidinger: that is strange, you already did it above, so why here second time?

src/lib/y2network/sysconfig_routes_file.rb
```diff
@@ -0,0 +1,270 @@
...
+    # SCR agent for routes files definition
+    def ifroute_term(path)
+      raise ArgumentError if path.nil? || path.empty?
+
+      Yast.term(
+        :ag_anyagent,
+        Yast.term(
+          :Description,
+          Yast.term(:File, path),
+          "#\n", # TODO: document these arguments
```
jreidinger: hint: https://github.com/yast/yast-core/blob/master/agent-any/doc/ag_anyagent.html#L49

src/lib/y2network/sysconfig_routes_file.rb
jreidinger: still true?

src/lib/y2network/config_writer/sysconfig.rb
imobachgs: Probably yes. We should come up with a solution for the ":any everywhere" problem and this one together.

src/lib/y2network/config_writer/sysconfig.rb
jreidinger: hmm, comment explaining what code did, but not why. Why we do not want to remove that `route` file? Is it mandatory to be present on system?

imobachgs: I would expect it to be present. At least, we are keeping it currently so I decided to not change the behavior. But it might deserve some documentation.

PR #760, opened 2019-04-04
Drop Yast::Routing from the routing client

teclator:
Move the routing client below the new namespace and drop Yast::Routing

Screenshot from 2019-04-08 08-07-28


src/include/network/services/routing.rb
```diff
@@ -78,7 +78,7 @@ def RoutingMainDialog
       Wizard.HideBackButton
 
       CWM.ShowAndRun(
-        "widget_descr"       => @wd_routing,
+        "widget_descr"       => widgets,
```
jreidinger: thanks :+1:

src/lib/y2network/clients/routing.rb
```diff
@@ -0,0 +1,466 @@
...
+        log.info("Routing module started")
+        ret = block.call
+        log.debug("ret=#{ret}")
+        # Finish
+        log.info("Routing module finished")
+        log.info("----------------------------------------")
+        ret
+      end
+
+      def import_modules
```
jreidinger: this is not needed to be done in instance. Just use it same as require at top of file.

src/lib/y2network/clients/routing.rb
```diff
@@ -0,0 +1,466 @@
...
+
+      def modified?
+        yast_config.routing != Yast::Lan.system_config.routing
+      end
+
+      def yast_config
+        Yast::Lan.yast_config
+      end
+
+      def yast_routing_table
```
jreidinger: now I see that collision. I agree that include for such big think as routines.rb was bad idea and hopefully will die soon.

src/lib/y2network/clients/routing.rb
```diff
@@ -0,0 +1,466 @@
...
+      end
+
+      # Command line definition
+      def cmdline_definition
+        {
+          # Commandline help title
+          "help"       => _("Routing Configuration"),
+          "id"         => "routing",
+          "guihandler" => fun_ref(method(:RoutingGUI), "any ()"),
+          "initialize" => fun_ref(method(:read), "boolean ()"), # FIXME
```
jreidinger: I expect rubocop will complain about that FIXME without `:` and note

src/lib/y2network/clients/routing.rb
```diff
@@ -0,0 +1,494 @@
...
+        Yast::Lan.add_config(:system, Y2Network::Config.from(:sysconfig))
+
+        true
+      end
+
+      # Write down the :yast {Y2Network::Config} in case of modified
+      #
+      # @return [Boolean]
+      def write
+        # TODO: Do not write if not modified
```
jreidinger: looks like already implemented?

src/lib/y2network/clients/routing.rb
```diff
@@ -0,0 +1,494 @@
...
+        end
+
+        !!route
+      end
+
+      # Initialize the yast and system {Y2Network::Config}s from the sysconfig
+      #
+      # @return [Boolean]
+      def read
+        Yast::Lan.add_config(:yast, Y2Network::Config.from(:sysconfig))
```
jreidinger: hmm, this basically means we are reading it twice. What about just copying data there?

src/lib/y2network/serializer/route_sysconfig.rb
```diff
@@ -0,0 +1,90 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "y2network/route"
+
+module Y2Network
+  module Serializer
+    class RouteSysconfig
```
imobachgs: Please, try to document classes (especially in terms of intention and adding some example). I do not block this PR because of that, but please add it to your TODO list :)

test/y2network/serializer/route_sysconfig_test.rb
```diff
@@ -0,0 +1,71 @@
...
+        "destination" => "default",
+        "gateway"     => "192.168.1.1",
+        "netmask"     => "-",
+        "device"      => "em1",
+        "extrapara"   => "table one"
+      }
+    end
+
+    it "instantiates a Y2Network::Route from the given hash" do
+
```
imobachgs: Please, remove the `do...end` block so the test will appear as pending. If it is not needed, just delete it entirely.

test/y2network/clients/routing_test.rb
```diff
@@ -34,5 +58,157 @@
...
+          expect(Yast::CommandLine).to receive(:Print).with(/No entry for/)
+          subject.main
+        end
+      end
+
+      context "with an existent route destination" do
+        let(:args) { ["delete", "dest=default"] }
+
+        it "deletes the routes with the given destination" do
+          expect(routing.routes.size).to eq(1)
```
imobachgs: This "expect" call is not necessary as you already know the value. If you want to express the change, you can use the `expect {}.change {}.from().to()` form. Check below.

PR #761, opened 2019-04-04
Write routing configuration using scr directly

mchf:
TODO: unit tests

### DIFF
--- a/src/lib/y2network/sysconfig_routes.rb
+++ b/src/lib/y2network/sysconfig_routes.rb
@@ -23,67 +23,67 @@ module Y2Network
   module SysconfigRoutes
     DEFAULT_ROUTES_FILE = "/etc/sysconfig/network/routes".freeze
 
-      # SCR agent for routes files definition
-      def ifroute_term(path)
-        raise ArgumentError if path.nil? || path.empty?
+    # SCR agent for routes files definition
+    def ifroute_term(path)
+      raise ArgumentError if path.nil? || path.empty?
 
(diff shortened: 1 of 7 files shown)


mchf:
Fixed remote testsuite failures
Some cleanup and squashed commits a bit

src/lib/y2network/sysconfig_routes.rb
```diff
@@ -0,0 +1,89 @@
...
+      )
+    end
+
+    # Registers SCR agent which is used for accessing particular ifroute-device
+    # file
+    #
+    # @param device [String] full path to a file in routes format (e.g. /etc/sysconfig/network/ifroute-eth0)
+    # @return [Path] SCR path of the agent
+    # @raise  [RuntimeError] if it fails
+    def register_ifroute_agent_for_path(path)
```
jreidinger: hmm, this looks a bit duplicite to @imobachgs pull request, where I already comment this part of code.

mchf: in fact it is copy of what already was in y2-network and copied into new routing here: https://github.com/yast/yast-network/pull/750

imobachgs: @jreidinger We discussed yesterday that we would merge this PR (which contains the tested logic to write changes) and I will rebase my changes on top to move this logic into `Y2Network::SysconfigRoutesFile` class.

jreidinger: ok, it is fine then.

PR #769, opened 2019-04-05
add missing space

jreidinger:
I am not sure if I should also add help to routing table, which will be quite duplicite to route dialog?


teclator:
@jreidinger well, checking it, maybe we should add some general explanation as currently only IP Forwarding is described in the help when running the routing dialog.

It is true that in the past there were default route inputs and now all are handled by the Routing Table, so not sure yet what we should explain. Maybe, that the routing table loads the configuration from sysconfig files and write down the routes where correspond or something like that, but not sure.

jreidinger:
@teclator me neither. I am not sure what to write to that table, but maybe some quick introduction can be enough? I simply do not want to duplicate info from dialog.

PR #773, opened 2019-04-09
Remove Yast::Routing module

imobachgs:
This PR drops the `Yast::Routing` module.

## TODO

* [ ] Consider using a `Set` to hold interfaces.
* [ ] Adapt the summary
* [ ] Remove `ifroute-X` file when not needed (this bug was already present in the old version)


src/include/network/lan/complex.rb
```diff
@@ -464,7 +463,7 @@ def handleOverview(_key, event)
           end
 
           LanItems.DeleteItem
-          LanItems.update_routing_devices!
+          LanItems.remove_current_device_from_routing
```
jreidinger: can this be part of `LanItems.DeleteItem` so no extra call is needed?

src/lib/y2network/presenters/routing_summary.rb
```diff
@@ -17,29 +17,72 @@
...
+
+        "<ul>#{summary}</ul>"
+      end
+
+    private
+
+      # Returns a text representation of the gateway
+      #
+      # @param routing [Y2Network::Routing] Routing configuration
+      # @return [String]
```
jreidinger: please document also nil.

src/modules/LanItems.rb
```diff
@@ -2667,18 +2667,43 @@ def colliding_item(name)
...
-    def update_routes!(previous_name)
-      Routing.device_routes(previous_name).each { |r| r["device"] = current_name }
+    # Renames an interface
+    #
+    # @todo This method exists just to keep some compatibility during
+    #       the migration to network-ng.
+
+    # @param old_name [String] Old device name
+    def rename_current_device_in_routing(old_name)
+      interface = yast_config.interfaces.find { |i| i.name == old_name }
```
jreidinger: well, this find usage expect that we have always interfaces unique. But I am not sure if it is always true. Maybe interfaces should be Set instead of Array?

src/modules/LanItems.rb
```diff
@@ -2667,18 +2667,43 @@ def colliding_item(name)
...
+    # Removes the interface with the given name
+    #
+    # @todo This method exists just to keep some compatibility during
+    #       the migration to network-ng.
+    # @fixme It does not check orphan routes.
+    # @param name [String] Device name
+    def remove_current_device_from_routing
+      name = current_name
+      return if name.empty?
+      yast_config.interfaces.reject! { |i| i.name == name }
```
jreidinger: is it enough? What if there is route that uses this interface?

test/lan_items_helpers_test.rb
```diff
@@ -578,7 +580,7 @@ def mock_dhcp_setup(ifaces, global)
 
   before do
     allow(Yast::LanItems).to receive(:Items).and_return(0 => item_0)
-    Yast::Routing.SetDevices(["eth0"])
+    # Yast::Routing.SetDevices(["eth0"])
```
jreidinger: if it is not needed, just kill it :)

test/lan_items_helpers_test.rb
```diff
@@ -658,12 +734,12 @@ def mock_dhcp_setup(ifaces, global)
...
 
     before do
-      Yast::Routing.Routes = original_routes
+      # Yast::Routing.Routes = original_routes
     end
 
     it "modifies all existent device routes with the current device name" do
       Yast::LanItems.update_routes!("eth0")
-      routes = Yast::Routing.Routes().select { |r| r["device"] == renamed_to }
+      # routes = Yast::Routing.Routes().select { |r| r["device"] == renamed_to }
```
jreidinger: hmm, use either xdescribe or commented out code, but not both.

test/lan_test.rb
```diff
@@ -12,6 +12,10 @@
   subject { Yast::Lan }
   let(:system_config) { instance_double(Y2Network::Config, "System") }
 
+  before do
+    Yast::Lan.clear_configs
```
jreidinger: maybe even hook it to general test_helper? As no test should depend on previous configs.

test/y2network/routing_test.rb
```diff
@@ -64,4 +65,28 @@
...
+
+  describe "#default_route" do
+    let(:routes) { [no_default, route1] }
+    let(:no_default) { double("Y2Network::Route", default?: false) }
+
+    it "returns the default route" do
+      expect(routing.default_route).to eq(route1)
+    end
+
+    context "when there are not routes" do
```
jreidinger: typo `not` instead of `no`?

test/y2network/routing_test.rb
```diff
@@ -23,8 +23,9 @@
 
 describe Y2Network::Routing do
   subject(:routing) { described_class.new(tables: [table1]) }
-  let(:table1) { Y2Network::RoutingTable.new([route1]) }
-  let(:route1) { double("Y2Network::Route") }
+  let(:table1) { Y2Network::RoutingTable.new(routes) }
+  let(:route1) { double("Y2Network::Route", default?: true) }
```
jreidinger: well, to my taste I would expect real Route object here. I think it is better to mock on edges of API or if there is strong reason like hard to use API for testing. But as it is new API, I think it will be better to use API directly with `Y2Network::Route.new`

src/modules/LanItems.rb
imobachgs: It is enough because the Route object is linked to the Interface one. There was a bug but it is solved now.

src/modules/LanItems.rb
imobachgs: Yes, I would expect the interfaces to be unique. But using a `Set` is fine for me. I will add a TODO item in the card.

PR #775, opened 2019-04-10
Autoyast Fixes: Part One

teclator:


src/modules/Lan.rb
```diff
@@ -1138,7 +1138,9 @@ def routing_summary
       config = find_config(:yast)
       return "" unless config && config.routing
       presenter = Y2Network::Presenters::RoutingSummary.new(config.routing)
-      presenter.text
+      summary = "<ul><li><b>" + _("Routing:") + "</b></li><li>"
+      summary += presenter.text
+      summary += "</li></ul>"
```
jreidinger: this looks really strange. If you need specific summary for autoinst routing, please create a child of RoutingSummary and redefine text there.

teclator: I just wanted to check how it looked like in the AutoYaST summary, as it was suggested during the network installation proposal review.

teclator: Proposal

src/lib/y2network/config_reader/autoinst.rb
```diff
@@ -43,7 +43,7 @@ def initialize(section)
       def config
         interfaces = find_interfaces
         routing = section.routing ? AutoinstRouting.new(section.routing).config : nil
-        Y2Network::Config.new(interfaces: interfaces, routing: routing, source: :autoyast)
```
teclator: The `source: :sysconfig` instead of `:autoinst` is a workaround we should discuss. The writer used is based on the source, and that is not the case for AutoYaST. That is, the target writer should be decided based on the current network backend and not based on the source from where the config was read. So basically I used :sysconfig to use its writer.

src/modules/Lan.rb
```diff
@@ -751,7 +751,7 @@ def Import(settings)
       settings = {} if settings.nil?
 
       profile = Y2Network::AutoinstProfile::NetworkingSection.new_from_hashes(settings)
-      config = Y2Network::Config.from(:autoyast, profile)
+      config = Y2Network::Config.from(:autoinst, profile)
```
teclator: leftover: we renamed the config_reader to `Autoinst`

src/modules/Lan.rb
jreidinger: hmm, this is really big. I think there is ways to compress like having together IP Forwarding or somehow compress it with less details. This way it is a lot of info that has problem to fit and I worry in ncurses it will be even worse.

src/lib/y2network/config_reader/autoinst.rb
jreidinger: makes sense for me to just have backend value in Config.

src/modules/Lan.rb
teclator: Although this concretely presentation should be responsibility of `Yast::Lan` Summary 

https://github.com/yast/yast-network/blob/master/src/modules/Lan.rb#L813

teclator: Yep, and get bigger with complex configuration

PR #776, opened 2019-04-10
network-ng: fix gateways summary and interfaces linking

imobachgs:
* All the gateways are now included in the summary.
* Fixes a problem when linking routes and interfaces.

### DIFF
--- a/test/y2network/presenters/routing_summary_test.rb
+++ b/test/y2network/presenters/routing_summary_test.rb
@@ -27,12 +27,13 @@ describe Y2Network::Presenters::RoutingSummary do
   subject(:presenter) { described_class.new(routing) }
   let(:routing) do
     instance_double(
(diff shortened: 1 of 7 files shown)


test/y2network/presenters/routing_summary_test.rb
```diff
@@ -27,12 +27,13 @@
   subject(:presenter) { described_class.new(routing) }
   let(:routing) do
     instance_double(
-      Y2Network::Routing, default_route: default_route, forward_ipv4: true, forward_ipv6: false,
+      Y2Network::Routing, default_routes: default_routes, forward_ipv4: true, forward_ipv6: false,
       routes: [double("Y2Network::Route")]
```
jreidinger: still I am not much fan of that instance doubles when we can use real objects.

PR #779, opened 2019-04-10
fix coverage

jreidinger:
Reason is that Lan is loaded before coverage, so it is ignored and also all classes that it loads.


teclator:
LGTM

PR #781, opened 2019-04-11
Autoyast fixes two

teclator:
AutoYaST Fixes:

- [X] Export the section
- [X] Import when no config has been read yet (Autoyast configuration dialog)

Pending:

- [x] Autoyast configuration: Routing dialog when empty crashes.

### DIFF
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -91,19 +91,19 @@ module Yast
 
     def config
       # TODO: get it from some config holder
-      @routing_config = Yast::Lan.yast_config
+      @routing_config ||= Yast::Lan.yast_config
     end
 
     def routing_table
-      @routing_table_widget = Y2Network::Widgets::RoutingTable.new(config.routing.tables.first)
+      @routing_table_widget ||= Y2Network::Widgets::RoutingTable.new(config.routing.tables.first)
     end
 
     def ip4_forwarding
-      @ip4_forwarding_widget = Y2Network::Widgets::IP4Forwarding.new(config)
+      @ip4_forwarding_widget ||= Y2Network::Widgets::IP4Forwarding.new(config)
     end
 
     def ip6_forwarding
-      @ip6_forwarding_widget = Y2Network::Widgets::IP6Forwarding.new(config)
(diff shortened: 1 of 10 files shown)


src/modules/Lan.rb
```diff
@@ -788,7 +788,6 @@ def Export
         "config"               => NetworkConfig.Export,
         "devices"              => devices,
         "ipv6"                 => @ipv6,
-        "routing"              => profile.routing,
```
jreidinger: hmm, I am not sure if we should skill it completely or just use empty hash instead.

src/include/network/services/routing.rb
```diff
@@ -91,19 +91,19 @@ def RoutingMainDialog
 
     def config
       # TODO: get it from some config holder
-      @routing_config ||= Yast::Lan.yast_config
+      @routing_config = Yast::Lan.yast_config
```
jreidinger: variable `@routing_config` no longer make sense. It can be also removed.

src/include/network/services/routing.rb
```diff
@@ -91,19 +91,19 @@ def RoutingMainDialog
...
     end
 
     def routing_table
-      @routing_table_widget ||= Y2Network::Widgets::RoutingTable.new(config.routing.tables.first)
+      @routing_table_widget = Y2Network::Widgets::RoutingTable.new(config.routing.tables.first)
```
jreidinger: hmm, I am not sure I like it as it basically mean that you always create new instance of widget. So if widget store itself any value, it will be forgotten.

teclator: Do not like neither, so, what about caching the widget and just renew the cache in case the table changed?

jreidinger: well, on one side we need to have widget same during dialog run. On other hand when dialog is shown again with different values, we need to reset it. That is a problem of includes in module which uses instance variable. I am fighting it a lot now when trying to abstract some widget outside. Not sure about proper solution now intermediate. Maybe add TODO or FIXME and use what you did in this PR with note in head that it cannot be the final solution.

PR #782, opened 2019-04-11
Proof of concept

mchf:
Purpose was to replace the LanItems::Items beast with something like objects and then start the yast lan module and see devices listed as expected.

This was (with one minor issue in name column) achieved with this patch.

Next steps:
1) Objects from the patch need some love now.
2) ~~Enable as much tests as possible~~ (now only tests tightly related to ```BuildLanOverview``` are disabled)


mchf:
replaced_lanitems

jreidinger:
hmm, at first, why you kill all tests? I will check code.

src/lib/y2network/hwinfo.rb
```diff
@@ -16,14 +16,36 @@
 #
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
-require_relative "../test_helper"
-require "y2network/config_writer"
+module Y2Network
+  # FIXME: should read hwinfo for a network device and store only necessary info
+  class Hwinfo
```
jreidinger: so goal is to have HWInfo per interface?

src/modules/LanItems.rb
```diff
@@ -1193,7 +1201,8 @@ def Read
     # 3) variables which keeps (some of) attributes of the current item (= item
     # which is being pointed by the iterator)
     def reset_cache
-      LanItems.Items = {}
+# FIXME: what is suitable replacement with new config storage?
```
jreidinger: probably something like `Yast::Config.yast = Yast::Config.system`

src/modules/LanItems.rb
```diff
@@ -1232,37 +1241,17 @@ def Import(settings)
...
 
-    def GetDescr
+    def GetDescr(overviews)
       descr = []
-      Builtins.foreach(
-        Convert.convert(
-          @Items,
-          from: "map <integer, any>",
-          to:   "map <integer, map <string, any>>"
+      overviews.each_pair do |key, value|
```
jreidinger: `map` is your friend.

src/modules/LanItems.rb
```diff
@@ -1580,19 +1565,37 @@ def BuildLanOverview
...
+          "table_descr" => [descr, DeviceProtocol(iface.config), iface.name, note]
         }
       end
-      [Summary.DevicesList(overview), links]
+
+      {
+        overview: overview,
+        overviews: overviews,
+        links: links
+      }
```
jreidinger: This should be presenter for sure.

src/modules/LanItems.rb
```diff
@@ -2683,6 +2686,10 @@ def update_routes!(previous_name)
 
   private
 
+    def yast_config
```
jreidinger: I suggest to simply use `config` as method.

src/lib/y2network/hwinfo.rb
imobachgs: We already had a reader in storage-ng: https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/hwinfo_reader.rb

src/lib/y2network/hwinfo.rb
```diff
@@ -16,14 +16,36 @@
...
+    def busid
+      @hwinfo ? @hwinfo.fetch("busid", "") : ""
+    end
 
-describe Y2Network::ConfigWriter do
-  describe ".for" do
-    it "returns the writer for the given source" do
-      reader = Y2Network::ConfigWriter.for(:sysconfig)
-      expect(reader).to be_a(Y2Network::ConfigWriter::Sysconfig)
+    def name
```
imobachgs: Duplicated.

src/lib/y2network/interfaces.rb
```diff
@@ -16,36 +16,33 @@
...
-  describe "#==" do
-    context "given two interfaces with the same name" do
-      let(:other) { Y2Network::Interface.new(interface.name) }
+module Y2Network
+  # A container for network devices. In the end should carry methods for mass operations
+  # over network interfaces like old LanItems::find_dhcp_ifaces or so.
+  #
+  # FIXME: Intended for LanItems::Items separation
+  # proper cleanup is must
+  class Interfaces
```
imobachgs: I would rather prefer a different name like `InterfacesList`, `InterfacesCollection`, `InterfacesSet`, or whatever you prefer.

We have something similar [in storage-ng](https://github.com/yast/yast-storage-ng/blob/master/src/lib/y2storage/planned/devices_collection.rb). You could also extend it to support queries like this:

```ruby
interfaces.configured.type(:eth)
```
Note: actually, the yast2-storage counterpart should return a collection itself so we can chain methods.

src/lib/y2network/interfaces.rb
```diff
@@ -16,36 +16,33 @@
...
+    extend Forwardable
 
-    context "given two interfaces with a different name" do
-      let(:other) { Y2Network::Interface.new("eth1") }
+    def_delegator :@old_items, :each
 
-      it "returns false" do
-        expect(interface).to_not eq(other)
-      end
+    def initialize(old_items:)
```
imobachgs: If this is a container, it should receive a list of `Y2Network::Interface` objects. Converting from a list of `old_items` should be done in a different constructor IMHO.

```ruby
Y2Network::Interfaces.from_old_items(old_items)
```

src/modules/LanItems.rb
mchf: this would reset everything (even e.g. Routing). And hidden question is "Find out whether it is still needed"

src/lib/y2network/hwinfo.rb
mchf: @jreidinger:
purpose is to store only what is needed. Hw info output is large and we need only couple of values. Also some values are used e.g. once other multiple times. It can turn out that using a singleton as pointed by @imobachgs. We will see ... but we definitely need some api for accessing it otherwise than touching the hash map - that's the point of this class for now.

src/modules/LanItems.rb
mchf: definitely agree ... the future is bright ;-)

src/lib/y2network/hwinfo.rb
jreidinger: ok, so goal of class is to have network friendly layer on top of hwinfo hash (so no direct access to hwinfo should be done). Then please document it in class description.

src/modules/LanItems.rb
```diff
@@ -1471,7 +1471,7 @@ def FormatOverview
         overviews[iface.name] = {}
         rich = ""
 
-        descr = iface.name
+        descr = iface.hardware.description
```
teclator: Virtual interfaces do not have have harward info, ins't it?

mchf: well ... depends. As I see it ```hwinfo``` provides more a kind of "driver" information provided by kernel. So, if you use ```hwinfo --network``` instead of ```hwinfo --netcard``` you'll see that (e.g.) vlan device can be found like on one of my test VM:
```
17: None 00.0: 10701 Ethernet
  [Created at net.126]
  Unique ID: +HKJ.ndpeucax6V1
  Parent ID: rBUF.GkSpXDNdaaB
  SysFS ID: /class/net/vlan0
  Hardware Class: network interface
  Model: "Ethernet network interface"
  Driver: "802.1Q VLAN Support"
  Device File: vlan0
  HW Address: 08:00:27:57:ed:fe
  Permanent HW Address: 08:00:27:57:ed:fe
  Link detected: yes
  Config Status: cfg=new, avail=yes, need=no, active=unknown
  Attached to: #5 (Ethernet controller
```

On the other hand yast2-network currently queries only for ```-netcard``` which do not show the vlan device in this case. 

I only try to separate access to ```LanItems::Items``` and wrap it to some objects here. It definitely is not perfect so far and needs some love. Moreover the piece of code you refer to is a direct "translation" of old code into new - in original code it simply put empty string into ```descr``` when corresponding hwitem field (```hwinfo["name"]``` - a textual description of device like "Intel 82546 controller") was not available for the device.

mchf:
Rebased this patch on top of updated network-ng branch -> reopening as a new pull request

PR #783, opened 2019-04-11
More Object CWM + dead code removal

jreidinger:


src/include/network/widgets.rb
```diff
@@ -240,7 +216,11 @@ def ipoib_mode_widget
         "widget" => :radio_buttons,
         # ipoib_modes contains known IPoIB modes, "default" is place holder for
         # "do not set anything explicitly -> driver will choose"
```
mchf: missing note for translators which was in ```ipoib_modes```

mchf:
so far looks nice. It always surprises me how much code is still possible to remove.

src/lib/y2network/widgets/bond_slave.rb
```diff
@@ -0,0 +1,217 @@
...
+      end
+
+      def help
+        # TODO: write it
+        ""
+      end
+
+      # Default function to init the value of slave devices box for bonding.
+      def init
+        # TODO: why? it should respect previous @settings, not?
```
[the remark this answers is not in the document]
teclator: https://github.com/teclator/yast-network/blob/db55eb4f166b251b0f035c21963ad274152ba7cc/src/modules/LanItems.rb#L128

## Previous related work in the area:

https://github.com/yast/yast-network/pull/403 (Bonding)
https://github.com/yast/yast-network/pull/448 (Bridging)

src/lib/y2network/widgets/bond_slave.rb
```diff
@@ -0,0 +1,217 @@
...
+      # Default function to store the value of slave devices box.
+      def store
+        configured_slaves = @settings["SLAVES"] || []
+
+        selected_slaves = Yast::UI.QueryWidget(:bond_slaves_items, :SelectedItems) || []
+
+        @settings["SLAVES"] = selected_slaves
+
+        # XXX: Hmm, it stores different widget? This is dark cwm hack and hidden dependency
+        # that both widget have to be used together
```
teclator: Well, it is a CustomWidget with a ComboBox and a Custom MultiSelectionBox with order, so both can store their state separetly and are not dependend each other although can be seen like a Bonding widget which will initialize and store both widget. so it is so black or dark cwm hack after all?

src/lib/y2network/widgets/bond_slave.rb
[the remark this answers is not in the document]
jreidinger: yep, but it is very inconsistent to rest of dialog. Other widgets use `@settings` as controller which is initialized and then stored, but this one not.

src/lib/y2network/widgets/bond_slave.rb
jreidinger: @teclator problem is that BONDOPTION widget does not have store and init and instead it is done in BONDSLAVES widget. So it works only when both CWM widgets is used together, so completely against CWM idea of reusable widgets.

src/lib/y2network/widgets/bond_slave.rb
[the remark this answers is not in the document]
teclator: @jreidinger yep, in general it should be unified with the rest.

src/lib/y2network/widgets/bond_slave.rb
teclator: @jreidinger mmm revising the code you are right which means the bond_options widget should be added, or it should be a `Bonding` widget with both.

src/lib/y2network/widgets/bond_slave.rb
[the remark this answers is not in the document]
teclator: And I would say that bridge_ports are similar. That is, the initialization is done through @settings, the store modifies the object itself.

src/lib/y2network/widgets/bond_options.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module Y2Network
+  module Widgets
+    class BondOptions < CWM::ComboBox
+      def initialize(settings)
+        textdomain "network"
+        @settings = settings
+      end
+
+      def items
```
imobachgs: Why not using a constant?

jreidinger: good note, it can be constant as we do not refresh it.

src/lib/y2network/widgets/bridge_ports.rb
```diff
@@ -0,0 +1,80 @@
...
+      def label
+        _("Bridged Devices")
+      end
+
+      def help
+        # TODO: write it
+        ""
+      end
+
+      # Default function to init the value of slave devices box for bonding.
```
teclator:
```suggestion
      # Default function to init the value of slave devices box for bridging.
```

src/lib/y2network/widgets/bridge_ports.rb
```diff
@@ -0,0 +1,80 @@
...
+        # it is list of Items, so cannot use `change_items` helper
+        Yast::UI.ChangeWidget(Id(widget_id), :Items, items)
+      end
+
+      # Default function to store the value of slave devices box.
+      def store
+        @settings["BRIDGE_PORTS"] = value
+      end
+
+      # Validates created bonding. Currently just prevent the user to create a
```
teclator:
```suggestion
      # Validates created bridge. Currently just prevent the user to create a
```

src/lib/y2network/widgets/bridge_ports.rb
```diff
@@ -0,0 +1,80 @@
...
+      # bond with more than one interface sharing the same physical port id
```
teclator:
```suggestion
      # bridge with already configured interfaces
```

test/y2network/widgets/bridge_ports_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require_relative "../../test_helper"
+require "cwm/rspec"
+
+require "y2network/widgets/bridge_ports"
+
+describe Y2Network::Widgets::BridgePorts do
+  subject { described_class.new({}) }
+
+  include_examples "CWM::MultiSelectionBox"
```
teclator: What about testing also the #validate method like with bonding?

test/y2network/widgets/bridge_ports_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  describe "#validate" do
+    context "when all devices are not yet configured" do
+      it "returns true" do
+
+        expect(subject.validate).to eql(true)
+      end
+    end
+
+    context "when there is more than one physical port id per interface" do
```
teclator: The context looks like a bonding test instead of a bridge one.

teclator:
For me it looks good.

PR #784, opened 2019-04-12
Display a popup when dropping static default route

mchf:
bnc#1135188

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1473,7 +1473,15 @@ module Yast
           #		 "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
           if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
-            Routing.RemoveDefaultGw
+            remove_gw = Routing.default_gw? && Popup.YesNo(
+              _(
+                "A static default gateway is defined.\n" \
+                "It is suggested to remove static gateway definition,\n" \
+                "if one can be obtained also via DHCP.\n" \
+                "Do you want to remove static default gateway?"
+              )
+            )
+            Routing.RemoveDefaultGw if remove_gw
           end
         end
       end
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -1475,10 +1475,10 @@ module Yast
           if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
             remove_gw = Routing.default_gw? && Popup.YesNo(
               _(
-                "A static default gateway is defined.\n" \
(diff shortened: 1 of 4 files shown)


src/include/network/lan/address.rb
```diff
@@ -1473,7 +1473,15 @@ def AddressDialog
           #		 "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
           if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
-            Routing.RemoveDefaultGw
+            remove_gw = Routing.default_gw? && Popup.YesNo(
+              _(
+                "A static default gateway is defined.\n" \
```
teclator: Just not sure about the wording.

teclator:
```
            _(
                "A static default route is defined.\n" \
                "It is suggested to remove the static default route \n" \
                "when it can be set also via DHCP.\n" \
                "Do you want to remove it now?"
             )
```

src/include/network/lan/address.rb
```diff
@@ -1473,7 +1473,15 @@ def AddressDialog
           #		 "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
           if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
-            Routing.RemoveDefaultGw
+            remove_gw = Routing.default_gw? && Popup.YesNo(
+              _(
+                "A static default route is defined.\n" \
+                "It is suggested to remove static default route definition,\n" \
```
mvidner:
```suggestion
                "It is suggested to remove the static default route definition,\n" \
```

src/include/network/lan/address.rb
```diff
@@ -1473,7 +1473,15 @@ def AddressDialog
...
+                "if one can be obtained also via DHCP.\n" \
+                "Do you want to remove static default route?"
```
mvidner:
```suggestion
                "Do you want to remove the static default route?"
```

src/include/network/lan/address.rb
```diff
@@ -1473,7 +1473,15 @@ def AddressDialog
           #		 "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
           if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
-            Routing.RemoveDefaultGw
+            remove_gw = Routing.default_gw? && Popup.YesNo(
+              _(
+                "A static default route is defined.\n" \
+                "It is suggested to remove the static default route definition\n" \
```
teclator:
```suggestion
                "It is suggested to remove the static default route definition \n" \
```

PR #785, opened 2019-04-15
Remove Y2Network::Config read which is already read by Yast::Lan.Read

teclator:


teclator:
This PR is completely out of date, so closing.

PR #786, opened 2019-04-16
Merge network-ng into master

teclator:
It mainly re-implement the routing module adding already some classes that will be part of the core.


mvidner:
Design question: when we're dealing with configuration, some frameworks make explicit distinction between transient and persistent config (kernel vs config files, see eg. libvirt's [virsh](https://libvirt.org/sources/virshcmdref/html-single/), [CIM](https://en.wikipedia.org/wiki/Common_Information_Model_(computing)) takes it to an extreme). It looks like currently the config classes (Y2N::Config, Y2N::Routing, Y2N::Interface) deal only with persistent config, right?

jreidinger:
@mvidner yep, only persistent. Same as with firewall or bootloader. I think whole yast deals mainly with transient partially is services.

teclator:
@jreidinger @mvidner currently we initialize from persistent configuration, but the idea is that we could add other configuration later. 

So currently we initialize a `system` configuration and then a `yast` configuration where we apply the changes done by YaST, which are compared with the system one before doing changes. But, nothing limits us to add a new config from running in the future if needed.

src/lib/y2network/autoinst_profile/networking_section.rb
```diff
@@ -0,0 +1,67 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/autoinst_profile/routing_section"
+
+module Y2Network
+  module AutoinstProfile
+    # This class represents an AutoYaST <networking> section
```
mvidner:
```suggestion
    # This class represents an AutoYaST \<networking> section
```

this fixes the yard rendering

mvidner: How about:

```rb
# TODO so far this contains only RoutingSection.
# The other aspects (interfaces, DNS resolver) are done in legacy code,
# see {Yast::LanClass#Import}.
```

src/lib/y2network/autoinst_profile/section_with_attributes.rb
```diff
@@ -0,0 +1,192 @@
...
+    # @todo This class is supposed to live in autoyast2 or yast2. When done,
+    #   please, adapt yast2-storage-ng accordingly.
+    class SectionWithAttributes
+      include Yast::Logger
+
+      class << self
+        # Description of the attributes in the section.
+        #
+        # To be defined by each subclass. Each entry contains a hash with the
+        # mandatory key :name and an optional key :xml_name.
```
mvidner:
```suggestion
        # mandatory key :name and an optional key :xml_name; the values are symbols.
```

mvidner: So far I see no use of `:xml_name`, and the instantiated `attributes` methods are seemingly redundant with `name:` all over. Is this needed?

src/lib/y2network/interface.rb
```diff
@@ -0,0 +1,43 @@
...
+    def initialize(name)
+      @name = name
+    end
+
+    # Determines whether two interfaces are equal
+    #
+    # @param other [Interface,:any] Interface to compare with
+    # @return [Boolean]
+    def ==(other)
+      return false if other == :any
```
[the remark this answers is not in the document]
jreidinger: @mvidner well `==` is not `include?`. So `==` should be order independent. Just imagine that `interface1 == interface2` but `interface2 != interface1`.

src/lib/y2network/autoinst_profile/networking_section.rb
jreidinger: well, I hope it is just intermediate state and we move forward with this changes.

src/lib/y2network/autoinst_profile/section_with_attributes.rb
jreidinger: well, for now it is just copy from yast-storage-ng and hopefully this will be moved to autoyast or yast2 itself as it is generic class.

mvidner: Oh! Then add a comment "Copied from yast-storage-ng, TODO deduplicate" otherwise they will diverge.

src/lib/y2network/interface.rb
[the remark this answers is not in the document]
mvidner: Uh, OK, this is a sign of of a bigger problem which I'll comment in the main PR body

teclator:
@mvidner, it is true that the special case for :any was not correct at all, and handling it in the ```==``` evidenced it even more. 

The `Interface` class will be a replacement of the current Lant::Items Item, although by the moment it is only used in routing as it is the first component that has been rewritten.

A `Y2Network::Route` can be associated with an interface or not. The association is done through the interface `name`. We could use a `InterfaceMatcher` or even the interface `name` as we had in the past. In that case, we need to take care of interface renaming as we have to update the associated routes while it is not necessary in case we work with a `Interface` object directly.

I think that removing the concept of :any is enough (specially when we only write the file and do not do any check / comparison with the ifcfg-files), and just model it as a one-to-many association allowing null.

What do you think?

I have created this PR which adapt the branch: https://github.com/yast/yast-network/pull/791

src/lib/y2network/config_reader/autoinst_routing.rb
```diff
@@ -0,0 +1,96 @@
...
+        netmask = route_section.netmask
+        return IPAddr.new(destination) unless netmask
+        netmask.delete!("/")
+        IPAddr.new("#{destination}/#{netmask}")
+      end
+
+      # Return the IPAddr of de host defined as the gateway.
+      #
+      # @param route_section [AutoinstProfile::RouteSection]
+      # @return [IPAddr, :default]
```
mvidner: The implementation returns `nil` instead

mvidner:
@teclator OK. I am not entirely convinced, but it can work that way. Protecting against interface renaming is a strong argument.

PR #787, opened 2019-04-16
exterimental move of settings that does not work

jreidinger:


src/lib/y2network/widgets/bond_slave.rb
```diff
@@ -65,9 +65,6 @@ def help
 
       # Default function to init the value of slave devices box for bonding.
       def init
-        # TODO: why? it should respect previous @settings, not?
-        @settings["SLAVES"] = Yast::LanItems.bond_slaves || []
-
         items = slave_items_from(
           Yast::LanItems.GetBondableInterfaces(Yast::LanItems.GetCurrentName),
           Yast::LanItems.bond_slaves
```
teclator: Still references bond_slaves instead of `@settings["SLAVES"]`

teclator: BTW, for me it works in both cases...

src/include/network/lan/address.rb
```diff
@@ -1157,6 +1157,7 @@ def AddressDialog
           end
         end
 
+
```
teclator: not needed

src/include/network/lan/address.rb
```diff
@@ -1212,6 +1219,9 @@ def initialize_address_settings
         @settings["VLAN_ID"]     = LanItems.vlan_id.to_i
       end
 
+      @settings["BONDOPTION"] = Yast::LanItems.bond_option
+      @settings["SLAVES"] = Yast::LanItems.bond_slaves || []
```
teclator: The initialization should be done only when working with a bond.

teclator:
LGTM

PR #789, opened 2019-04-22
First version of network-ng-kickstart.md

ancorgs:
See https://trello.com/c/ZCwLjVeM/839-document-the-outcome-of-the-meeting

### DIFF
--- a/doc/network-ng-kickstart.md
+++ b/doc/network-ng-kickstart.md
@@ -1,9 +1,9 @@
 # Refactoring YaST Network - Kickstart Meeting
 
-Since a long time, the YaST Team has been discussing about the possibility of
+Since a long time, the YaST Team has been discussing the possibility of
 starting a sustained effort to refactor and rewrite several parts of the YaST
 Network module. On 14th March 2019 a meeting was finally held to define the
-goals of such effort and, to some extend, the strategy to follow.
+goals of such effort and, to some extent, the strategy to follow.
 
 Since the relationship between YaST and `wicked` was considered a central point
 for the topic, both Marius and Rubén (developers of Wicked at that point in
@@ -16,7 +16,9 @@ files. That means it does not communicate directly with `wicked`. They work
 together somehow only because both understand those "legacy" formats and
 commands.
 
-YaST does not communicate with Network Manager in any way.
+YaST does not communicate with Network Manager in any significant way. YaST only
+starts, stops or restarts the Network Manager service as needed. But it doesn't
+read or modify the Network Manager configuration.
 
 Network Manager offers a DBUS interface. `wicked` also offers a DBUS interface,
 but it's far from being as powerful as directly reading and writing some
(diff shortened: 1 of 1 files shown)


doc/network-ng-kickstart.md
```diff
@@ -0,0 +1,60 @@
...
+time) attended the meeting.
+
+## Status
+
+Currently YaST manages configurations using traditional Linux commands and
+files. That means it does not communicate directly with `wicked`. They work
+together somehow only because both understand those "legacy" formats and
+commands.
+
+YaST does not communicate with Network Manager in any way.
```
jreidinger: well, maybe I would say here, that yast only start/restart service, but does not configure it in any way and neither read anything from network manager beside systemd status.

doc/network-ng-kickstart.md
```diff
@@ -0,0 +1,60 @@
...
+We want YaST to be able to better cooperate with `wicked`, but without coupling
+both components. Quite the opposite, we want YaST to be able to interoperate
+with a legacy system, with `wicked`, with Network Manager or with whatever other
+system that could be adopted by (open)SUSE in the future.
+
+We want to have an object oriented data model for YaST Network. For the reason
+explained above, we will not adopt directly the `wicked` model, but we will
+define our own model incrementally as we refactor the current pieces.
+
+We will start refactoring the current code little by little based on a series
```
jreidinger: maybe I will call it `refine` instead of refactoring, because managers do not like that term and also because when we touching code we also check whenever it still makes sense and adapt it if needed ( and same for bugs we found), so it is not strict refactoring that does not affect any functionality.

jreidinger:
in general nice meeting minutes :+1:

doc/network-ng-kickstart.md
```diff
@@ -0,0 +1,60 @@
...
+## Conclusions
+
+Although we usually call this initiative "Network NG" as a reference to "Storage
+NG", this is not intended as a full rewrite like it was done in the YaST storage
+area. Quite the opposite, the goal is to change things in a way in which the new
+code can be integrated into Tumbleweed often and can be included in subsequent
+releases of Leap and SLE.
+
+We want YaST to be able to better cooperate with `wicked`, but without coupling
+both components. Quite the opposite, we want YaST to be able to interoperate
```
imobachgs: What about something like *On the contrary* to avoid repetition?

doc/network-ng-kickstart.md
```diff
@@ -0,0 +1,60 @@
+# Refactoring YaST Network - Kickstart Meeting
+
+Since a long time, the YaST Team has been discussing about the possibility of
+starting a sustained effort to refactor and rewrite several parts of the YaST
+Network module. On 14th March 2019 a meeting was finally held to define the
+goals of such effort and, to some extend, the strategy to follow.
```
imobachgs: s/extend/extent

doc/network-ng-kickstart.md
```diff
@@ -0,0 +1,60 @@
+# Refactoring YaST Network - Kickstart Meeting
+
+Since a long time, the YaST Team has been discussing about the possibility of
```
imobachgs: np: you can get rid of "about"

PR #790, opened 2019-04-22
Initializes mode/second_stage element properly

imobachgs:
It fixes [bsc#1132057](https://bugzilla.suse.com/show_bug.cgi?id=1132057).


teclator:
@imobachgs quite better now covering it with unit tests :+1:

PR #791, opened 2019-04-23
Do not use :any for no device routes

teclator:


src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -101,38 +101,32 @@ def write_ipv6_forwarding(forward_ipv6)
 
       # Finds routes for a given interface
       #
-      # @param iface  [Interface,:any] Interface to search routes for
+      # @param iface  [Interface] Interface to search routes for
```
jreidinger: document that interface can be nil and what it means.

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -101,38 +101,32 @@ def write_ipv6_forwarding(forward_ipv6)
...
         routes.select do |route|
-          route.interface != :any && route.interface.name == iface.name
+          route.interface && route.interface.name == iface.name
         end
       end
 
       # Returns the routes file for a given interace
       #
-      # @param iface  [Interface,:any] Interface to search routes for
+      # @param iface  [Interface,nil] Interface to search routes for
```
jreidinger: also document here what means nil ( like global routes file ).

src/lib/y2network/interface.rb
```diff
@@ -31,10 +31,9 @@ def initialize(name)
 
     # Determines whether two interfaces are equal
     #
-    # @param other [Interface,:any] Interface to compare with
+    # @param other [Interface] Interface to compare with
     # @return [Boolean]
     def ==(other)
-      return false if other == :any
       name == other.name
```
jreidinger: `nil.name` crash?

jreidinger: maybe have something like `return false unless other.is_a?(Interface)` ?

src/lib/y2network/route.rb
```diff
@@ -21,18 +21,18 @@ module Y2Network
   class Route
     # @return [IPAddr,:default] Destination; :default if it is the default route
     attr_accessor :to
-    # @return [Interface,:any] Interface to associate the route to; :any if no interface is given
+    # @return [Interface,nil] Interface to associate the route to
```
jreidinger: document what nil means.

src/lib/y2network/route.rb
```diff
@@ -21,18 +21,18 @@ module Y2Network
...
     attr_accessor :interface
     # @return [IPAddr,nil] Gateway IP address ('via' in ip route)
     attr_accessor :gateway
     # @return [String] Additional options
     attr_accessor :options
 
     # @param to        [IPAddr,:default] Destination
-    # @param interface [Interface,:any] Interface to associate the root to
+    # @param interface [Interface, nil] Interface to associate the root to
```
jreidinger: and same here.

src/lib/y2network/serializer/route_sysconfig.rb
```diff
@@ -73,9 +73,9 @@ def build_ip(ip_str, netmask_str = MISSING_VALUE)
         netmask_str == MISSING_VALUE ? ip : ip.mask(netmask_str)
       end
 
-      # @return [Y2Network::Interface, :any]
+      # @return [Y2Network::Interface, nil]
       def interface_from(hash)
-        return :any if hash.fetch("device", "-") == MISSING_VALUE
+        return if hash.fetch("device", "-") == MISSING_VALUE
```
jreidinger: please add explicit nil there so `return nil`

src/lib/y2network/interface.rb
teclator: An interface is expected, we can return if classes differs of in case of other absence, I just unified with the rest.

jreidinger: this won't work. Here it will crash or modify Route to not do this comparison for interface if it is nil https://github.com/yast/yast-network/pull/791/files/66ed1669abc644fee168701832c3701310c5dc09#diff-6901639be33a5a1baf300f012c547db3R54

teclator: True, so, adding explicit return.

src/lib/y2network/route.rb
teclator: In this case, I think it is self explanatory, it basically allows null, so not associated with any interface explicitly.

src/lib/y2network/interface.rb
```diff
@@ -34,6 +34,7 @@ def initialize(name)
     # @param other [Interface] Interface to compare with
     # @return [Boolean]
     def ==(other)
+      return false if self.class != other.class
```
jreidinger: well, I still like more my code as this one basically means that you kill inheritance. using `is_a?` allows to have child of interface that can be also compared.

teclator: Sorry, somehow I overlooked the suggestion. Applied.

src/lib/y2network/route.rb
teclator: Same than above.

PR #793, opened 2019-04-24
Drop unused Network module

imobachgs:
It is empty, and it looks like it is not used anymore (not even in Perl).


teclator:
LGTM

PR #795, opened 2019-04-29
Add a basic DNSConfig class

imobachgs:
This PR adds:

* a Y2Network::DNS class
* a reader for DNS configuration
* a presenter for the DNS summary

### DIFF
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -19,6 +19,7 @@
 require "y2network/interface"
 require "y2network/config_writer"
 require "y2network/config_reader"
+require "y2network/dns"
 
 module Y2Network
   # This class represents the current network configuration including interfaces,
@@ -36,8 +37,10 @@ module Y2Network
   class Config
     # @return [Array<Interface>]
     attr_accessor :interfaces
-    # @return [Routing]
+    # @return [Routing] Routing configuration
     attr_accessor :routing
+    # @return [DNS] DNS configuration
+    attr_accessor :dns
     # @return [Symbol] Information source (see {Y2Network::Reader} and {Y2Network::Writer})
     attr_accessor :source
 
@@ -83,10 +86,12 @@ module Y2Network
     #
     # @param interfaces [Array<Interface>] List of interfaces
     # @param routing    [Routing] Object with routing configuration
+    # @param dns        [DNS] Object with DNS configuration
     # @param source     [Symbol] Configuration source
-    def initialize(interfaces:, routing:, source:)
+    def initialize(interfaces:, routing:, dns: nil, source:)
       @interfaces = interfaces
       @routing = routing
+      @dns = dns || Y2Network::DNS.new
       @source = source
     end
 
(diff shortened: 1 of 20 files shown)


src/lib/y2network/config_reader/sysconfig_dns.rb
```diff
@@ -0,0 +1,104 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "yast"
+require "y2network/dns_config"
+
+module Y2Network
+  module ConfigReader
+    # Reads DNS configuration from sysconfig files
+    class SysconfigDNS
```
jreidinger: Hmm, to be honest, I do not like much this double name. For me it a bit indicate that maybe we need namespace for ConfigReader::Syconfig and there having stuff like DNS. What do you think?

imobachgs: Yes, I prefer that approach too. But I have found sometimes some resistance to have more than 2 levels :) If you are fine, I will move it as I prefer that approach too.

jreidinger: well, I am not fan of more then 2 levels, but in this case it makes some sense to me. So as with any rule, use common sense :) Having three levels for few classes is too much for me, but having it for set of class as separator if there are more of such classes, makes sense for me. Another approach can be `Y2Network::ConfigReader::DNS::Sysconfig` where you basically need to use only `ConfigReader::DNS.from_system` to read what is currently used.

src/lib/y2network/dns_config.rb
```diff
@@ -0,0 +1,59 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Network
+  # DNS configuration (hostname, name servers, etc.).
+  class DNSConfig
```
jreidinger: well, when I see all that separation. Maybe it makes more sense to have `Y2Network::DNS` or `Y2DNS` and inside config reader, config and others? Still thinking about reasonable naming and structure to be

1. intuitive
2. easy to read
3. easy to write
4. as short as possible while still above is valid.

teclator: I also had the same doubts, and liked Y2Routing, Y2DNS but it was argued that they where below y2network namespace, so Y2Network::Routing && Y2Network::DNS looks reasonable to me.

src/lib/y2network/config_reader/sysconfig_dns.rb
```diff
@@ -74,6 +79,39 @@ def resolv_conf_policy
       #
       # @return [String]
       def hostname
```
teclator: In yast/yast-yast2#921 I added a `Hostname` class and a reader only for the `Hostname`, just for cases where the complete `DNS` config is not needed at all but only the `hostname`. 

It is committed as [RFC].

src/lib/y2network/config.rb
```diff
@@ -83,10 +86,12 @@ def configs
     #
     # @param interfaces [Array<Interface>] List of interfaces
     # @param routing    [Routing] Object with routing configuration
+    # @param dns        [DNS] Object with DNS configuration
```
jreidinger: well, nil is not documented. So I suggest to simply use as default value Y2Network::DNS.new and not have nil anywhere.

PR #796, opened 2019-04-30
Wrapping LanItems::Items into objects

mchf:
Currently you can somehow work with "Add" and "Edit" interface workflows. It is only partially working in sense that it do not crash and handles configuration / user input somehow. Definitely needs some more love but should contain all crucial updates which should be needed for switching to new backend

Follow up of https://github.com/yast/yast-network/pull/782


src/lib/y2network/interfaces.rb
```diff
@@ -0,0 +1,72 @@
...
+
+require "forwardable"
+
+module Y2Network
+  # A container for network devices. In the end should carry methods for mass operations
+  # over network interfaces like old LanItems::find_dhcp_ifaces or so.
+  #
+  # FIXME: Intended for LanItems::Items separation
+  # proper cleanup is must
+  class Interfaces
```
imobachgs: I like the idea of having a container for interfaces (as we have in [storage-ng](https://github.com/yast/yast-storage-ng/blob/5a8963f0c4b5a242ac8d3122ca5225e7f216e6a4/src/lib/y2storage/planned/devices_collection.rb)), although I would call it `InterfacesCollection` or `InterfacesList`.

However, this class is responsible for converting the information from sysconfig into `Interface` objects too (it receives the result of calling [Yast::LanItems.Items](https://github.com/yast/yast-network/pull/796/files#diff-66d00977c16e82a3b5f1046c2e5feb60R39)) which looks wrong to me. What if tomorrow we want to support any other technology? I think it should receive an array of `Interface` objects.

Then, who is responsible for building the `Interface` objects? I think `ConfigReader::Sysconfig` (or any ancillary class) is.

src/include/network/lan/complex.rb
```diff
@@ -149,10 +172,19 @@ def Commit
...
+      # itself yet
       ret = Lan.Read(:cache)
-      # Currently just a smoketest for new config storage - something what should replace Lan module in the bright future
-      # TODO: find a suitable place for this config storage
-      Y2Network::Config.from(:sysconfig)
+
+      # Load sysconfig into new backend
+      # It is done here bcs this method is currently called during "yast2 lan"
+      # initialization, so it makes configuration available for all submodules
+      system_config = Y2Network::Config.from(:sysconfig)
```
imobachgs: Why here and not in `Lan.Read`?

teclator:
I basically shared the same opinion.

imobachgs:
I've been checking some internals and I am not sure whether it is possible to follow the top-bottom approach we were talking about :(

src/include/network/lan/complex.rb
mchf: 1) ```Lan``` module should be replaced by ```Y2Network::Config``` in the future
2) place it properly in ```Lan::Read``` is a bit tricky
3) I don't see any benefit from placing it into ```Lan::Read```

mchf:
well originally I started with this to sort out my thoughts bcs I was completely unsure where should I start from when turning old internals into shining new ones.

However, somewhere around 25th commit I started to be slightly more optimistic - and it doesn't happen that often to me in y2-network ;-)

In the end I think that, as we are at the very beginning of 15sp2 development, we should be able to fix x86 parts easily and quickly when we go by way of accepting this patch. For example we can keep it living a side of master for while and ask QA team to run this version against current integration tests, and so on.  What I'm a bit more scared of is s390, but on the other hand I don't see anything that supper complex what should break. There are some additional dialogs which need to be tested and a difference in device initialization. A part of that everything else should be the same. So again nothing what should not be fixed after one or two sprints of intensive work when we have an s390 machine for playing.

What benefits I see? Better test coverable code (currently some parts are very hardly testable bcs of nasty side effects, spaghetti code, ... you now ;-) and cleaned code. For example during work on this patch I did a grep for ```LanItems.type``` to replace this calls and surprisingly found a whole dialog and some other related code which was not used at all and the relevant feature was removed in 2007. And I don't talk about other benefits like clear separation of responsibilities and more readable code.

And in the end I don't see this as a (clean) top to down rewriting. In the end I only took the ```LanItems::Items``` hash and replaced it with something else (originally I thought I should put it into a class, but in the end I decided to use newly created ```Y2Network::Interface``` and go more deeply) and injected this new structure into old workflow. All other changes then replacing ```GetType(Items[current]["ifcfg"])``` by ```iface.type```, were caused a messy code. For example add workflow in the old code was partly editing ```LanItems::Items``` directly, partly storing user input in a separate hash ... simply as it turned to be useful during past years of y2-network evolution.

So to summarize that ... I think that doing "the major" cut is best now (I mean now == before all that alfa, beta, GC phases) otherwise we can simply end with nicely rewritten UI into CWM but still working on top of sucking internals. I also don't see it doable and / or useful to rewrite it for example on per interface type basis (I mean create an object e.g. for wlan and rewrite parts of workflows related to this type into objects).

src/lib/y2network/interfaces.rb
mchf: we can separate this conversion method to another kind of reader for example. I'm not against that.

Renaming from ```Interfaces``` to e.g. ```InterfacesCollection``` also makes sense.

mchf:
And if you'd ask me how to continue ... Up to now I have been (manually) testing only eth workflows, so now I'd continue with other types. This should create a hints for building a ```Interface``` specialization hierarchy. 

Some other task which remains and should be done independently on the future of this patch is rewriting udev handling into objects.

imobachgs:
I am still thinking that this PR does too much :), but I can see your points and I think we could work from here. I have started a deeper review, but I would like to set a target and finish (and merges it) as soon as possible so we can start working in parallel. I want this PR to stop growing :)

mchf:
> I want this PR to stop growing :)

me too. Only think I plan to add is tests and necessary fixes, no more functionality.

src/modules/LanItems.rb
```diff
@@ -1990,29 +1932,32 @@ def setup_bonding(devmap, slaves, options)
...
+      newdev = builder.device_sysconfig
 
       # #104494 - always write IPADDR+NETMASK, even empty
-      newdev["IPADDR"] = @ipaddr
-      if !@prefix.empty?
-        newdev["PREFIXLEN"] = @prefix
-      else
-        newdev["NETMASK"] = @netmask
-      end
+      # newdev["IPADDR"] = @ipaddr
```
imobachgs: why not remove the code instead of just commeting it?

src/modules/LanItems.rb
```diff
@@ -2154,14 +2100,15 @@ def Commit
 
       # ZONE uses an empty string as the default ZONE which means that is not
       # the same than not defining the attribute
-      current_map = (GetCurrentMap() || {}).select { |k, v| !v.nil? && (k == "ZONE" || !v.empty?) }
+      iface = config.interfaces.find(builder.name)
+      current_map = (!iface.nil? && iface.configured ? iface.config : {}).select { |k, v| !v.nil? && (k == "ZONE" || !v.empty?) }
```
imobachgs: This code is pretty hard to read. What about splitting it in different lines?

imobachgs:
Some relevant things to fix:

- [ ] LanItems.isCurrentDHCP && LanItems.isCurrentHotplug
- [ ] The cmdline will be broken
- [ ] LanItems.current_udev_name
- [ ] Drop LanItems.Rollback calls
- [ ] Udev handling
- [ ] Move Interfaces to ConfigReader::Sysconfig
- [ ] https://github.com/yast/yast-network/pull/796/files#diff-cb38af52173f54fb28e2fa7a934f3557R1419

src/lib/y2network/interface.rb
```diff
@@ -16,17 +16,75 @@
...
     def initialize(name)
-      @name = name
+      # TODO: move reading hwinfo into Hwinfo class
+
+      # @hardware and @name should not change during life of the object
+      @hardware = Hwinfo.new(name: name)
+
+      if !(name.nil? || name.empty?)
+        @name = name
+      elsif hwinfo.nil?
```
teclator:
```suggestion
      elsif hardware.nil?
```

src/lib/y2network/interface_defaults.rb
```diff
@@ -0,0 +1,90 @@
...
+      Yast.import "ProductFeatures"
+
+      product_startmode = Yast::ProductFeatures.GetStringFeature(
+        "network",
+        "startmode"
+      )
+
+      startmode = case product_startmode
+      when "ifplugd"
+        if replace_ifplugd?
```
teclator: Where is this method defined?

src/lib/y2network/interface_defaults.rb
```diff
@@ -0,0 +1,90 @@
...
+          hotplug_usable? ? "hotplug" : "auto"
```
teclator: Same with this one.

src/include/network/lan/complex.rb
```diff
@@ -139,8 +141,29 @@ def tabs_descr
 
     # Commit changes to internal structures
     # @return always `next
-    def Commit
-      LanItems.Commit
+    def Commit(builder: nil)
+      # 1) update NetworkInterfaces with corresponding devmap
+      # FIXME: new item in NetworkInterfaces was created from handleOverview by
```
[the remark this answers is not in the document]
mchf: yes, currently testing it as a final change for this patch

src/include/network/lan/complex.rb
```diff
@@ -149,10 +172,19 @@ def Commit
...
-      # Currently just a smoketest for new config storage - something what should replace Lan module in the bright future
-      # TODO: find a suitable place for this config storage
-      Y2Network::Config.from(:sysconfig)
+
+      # Load sysconfig into new backend
+      # It is done here bcs this method is currently called during "yast2 lan"
+      # initialization, so it makes configuration available for all submodules
+      system_config = Y2Network::Config.from(:sysconfig)
+      Y2Network::Config.add(:system, system_config)
+      Y2Network::Config.add(:yast, system_config.copy)
```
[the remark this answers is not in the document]
mchf: 1) this is not regular dialog but that one which is showing lines like "Detect network devices" at the very beginning when starting yast2 lan UI. So something what is here exactly for the purpose of initialization of the module internals. I agree that this particular logic should be moved at another place (ideally the one which can be shared together with command line interface - but that could cause unacceptable lags from user's POV between starting module and getting some information from it)
2) Yast::Lan module is initialized a few lines above

src/include/network/lan/complex.rb
```diff
@@ -376,14 +422,21 @@ def initOverview(_key)
 
     def handleOverview(_key, event)
       if !disable_unconfigureable_items([:_hw_items, :_hw_sum] + overview_buttons.keys, false)
+        # nasty side effect: this method also sets LanItems::current
```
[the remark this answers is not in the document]
mchf: yes - just delete it, bcs setting LanItems::current should not be needed anymore. This was a note for me when I started to modify the workflows.

src/include/network/lan/complex.rb
```diff
@@ -376,14 +422,21 @@ def initOverview(_key)
...
 
+      # name is currently used as item id
+      iface_name = UI.QueryWidget(Id(:_hw_items), :CurrentItem)
+      config = Y2Network::Config.find(:yast)
+
       if Ops.get_string(event, "EventReason", "") == "Activated"
         case Ops.get_symbol(event, "ID")
         when :add
-          LanItems.AddNew
+          # FIXME: replace
```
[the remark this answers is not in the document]
mchf: I'm not sure yet. But definitely don't want to keep ```LanItems::operation = ``` calls. 

In fact I've an idea. Currently the overview dialog do not use artificial ids but existing device names. So we can detect if we work on an existing device (-> :edit operation) or a new one (-> :add operation).

src/lib/y2network/interface_defaults.rb
mchf: mostprobably in LanItems. I haven't tested this startmode yet.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,90 @@
...
+  #
+  # FIXME: it should be in charge of generating reasonable defaults too
+  #        @see LanItems::new_item_default_options & co
+  class InterfaceConfigBuilder
+    # @return [String] Device name (eth0, wlan0, etc.)
+    attr_accessor :name
+    # @return [String] type which is intended to be build
+    attr_accessor :type
+
+    include Y2Network::InterfaceDefaults
```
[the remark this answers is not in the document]
mchf: no. It is just bunch of methods moved from old code which I haven't check in depth yet.

src/lib/y2network/config.rb
```diff
@@ -109,7 +109,7 @@ def copy
     # @return [Boolean] true if both configurations are equal; false otherwise
     def ==(other)
       source == other.source &&
-        ((interfaces - other.interfaces) | (other.interfaces - interfaces)).empty? &&
+        interfaces == other.interfaces &&
```
imobachgs: So the order matters?

src/lib/y2network/config_writer/sysconfig.rb
```diff
@@ -36,21 +36,22 @@ def write(config)
...
-          end
+          # Remove ifroutes-* if empty
+          file.remove if  routes.empty?
+
+          file.routes = routes
+          file.save
         end
+
+        # update /etc/sysconfig/network/routes file
+        file = routes_file_for(nil)
```
imobachgs: using `nil` here is not much expressive. What about a symbol or something like that?

src/lib/y2network/config.rb
imobachgs: Oh, we replaced the arrays for `Y2Network::InterfacesCollection`. It looks good.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -58,5 +58,14 @@ def find(name)
     def add(name)
       interfaces.push(Interface.new(name))
     end
+
+    # Compares InterfacesCollections
+    #
+    # @return [Boolean] true when both collections are equal, false otherwise
+    def ==(other)
+      @interfaces - other.interfaces && other.interfaces == @interfaces
```
imobachgs: Sorry, but I do not get the conditions.

src/include/network/lan/address.rb
```diff
@@ -956,8 +910,9 @@ def wireless_tab
 
     # Dialog for setting up IP address
     # @return dialog result
-    def AddressDialog
-      initialize_address_settings
+    def AddressDialog(builder: nil)
```
jreidinger: what is a goal of nil builder? does not it fail latter in builder.type?

src/include/network/lan/address.rb
```diff
@@ -1005,12 +960,12 @@ def AddressDialog
         ]
       )
 
-      wd["IPOIB_MODE"] = ipoib_mode_widget if LanItems.GetCurrentType == "ib"
+      wd["IPOIB_MODE"] = ipoib_mode_widget if builder.type == "ib"
```
jreidinger: this start smelling to me. This type should be enum or own class. But something for future. As plain text comparison can be silently failing for long type. ( I made types often ).

src/include/network/lan/address.rb
```diff
@@ -1080,9 +1035,9 @@ def AddressDialog
 
       if ret != :back && ret != :abort
         # general tab
-        LanItems.startmode = Ops.get_string(@settings, "STARTMODE", "")
-        LanItems.mtu = Ops.get_string(@settings, "MTU", "")
-        LanItems.firewall_zone = firewall_zone.store_permanent if firewalld.installed?
+        builder.set(option: "STARTMODE", value: Ops.get_string(@settings, "STARTMODE", ""))
+        builder.set(option: "MTU", value: Ops.get_string(@settings, "MTU", ""))
+        builder.set(option: "ZONE", value: firewall_zone.store_permanent) if firewalld.installed?
```
jreidinger: hmm, I am still not much fan of this set which is just hidden `builder["STARTMODE"] = Ops.get_string(@settings, "STARTMODE", "")` so basically hash like API. I still prefer something like `builder.startmode = Ops.get_string(@settings, "STARTMODE", "")`

src/include/network/lan/address.rb
```diff
@@ -1117,82 +1073,87 @@ def AddressDialog
...
+        # it updates list of available interfaces which is done elsewhere already
+        # LanItems.add_current_device_to_routing if LanItems.update_routing_devices?
       end
 
-      if LanItems.type == "vlan"
-        LanItems.vlan_etherdevice = Ops.get_string(@settings, "ETHERDEVICE", "")
-        LanItems.vlan_id = Builtins.tostring(
-          Ops.get_integer(@settings, "VLAN_ID", 0)
+      if builder.type == "vlan"
+        builder.set(option: "ETHERDEVICE", value: Ops.get_string(@settings, "ETHERDEVICE", ""))
```
jreidinger: do we plan to keep this `@settings` if so, then I would prefer something like `builder.load_settings(@settings)` which will do this transformation inside.

src/include/network/lan/address.rb
```diff
@@ -1117,82 +1073,87 @@ def AddressDialog
...
 
   private
 
     # Initializes the Address Dialog @settings with the corresponding LanItems values
-    def initialize_address_settings
-      @settings.replace( # general tab:
-        "STARTMODE"        => LanItems.startmode,
-        "IFPLUGD_PRIORITY" => LanItems.ifplugd_priority,
+    def initialize_address_settings(builder)
+      @settings.replace(
```
jreidinger: so in future we will pass builder instead of @settings to CWM widgets, right?

src/include/network/lan/complex.rb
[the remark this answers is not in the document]
jreidinger: ok, so what I see as problem in these three lines:

1. it should be one line that say read from system
2. working copy of system during initial read is config detail that should be hidden
3. it hardcode reading from sysconfig, which looks wrong for me. "The read" should detect what kind of configuration it uses and read it. not something for dialog, even for read dialog. Main goal of read dialog is to provide reasonable feedback and steps. So anything that is more then one line read looks suspicious for me.

src/include/network/lan/complex.rb
```diff
@@ -393,14 +446,23 @@ def handleOverview(_key, event)
...
-          if LanItems.IsCurrentConfigured
-            LanItems.SetItem
-
-            if LanItems.startmode == "managed"
+          iface = config.interfaces.find(iface_name)
+          # FIXME: turn following lines into something like builder.load(iface) ?
+          @builder.name = iface.name
+          @builder.type = iface.type
+          if iface.configured
+            @builder.load_sysconfig(iface.config)
```
jreidinger: again too much tight to sysconfig. I prefer here generic load. In general I do not like naming here, as it is more like load current config to builder. Not anything about sysconfig.

src/include/network/lan/complex.rb
```diff
@@ -393,14 +446,23 @@ def handleOverview(_key, event)
...
+            @builder.load_s390_config(LanItems.s390_ReadQethConfig(iface.hardware.name)) if iface.hardware
```
jreidinger: I hope that it should be hidden in that generic load.not job of UI to handle different internals for different arch. What is its job if there is specific option for s390, then we need to have specialized UI, but not for loading on s390.

src/include/network/lan/complex.rb
```diff
@@ -393,14 +446,23 @@ def handleOverview(_key, event)
...
+            # FIXME: move to more appropriate place
+            LanItems.operation = :edit
```
jreidinger: I hope config itself can recognize if it already existed before or not. Maybe something trivial like system_name ( so name before all renames? ) and if it is nil, then it means new and of course nice method for it.

src/include/network/lan/complex.rb
```diff
@@ -566,6 +628,8 @@ def MainDialog(init_tab)
         break if input_done?(ret)
       end
 
+      @builder = nil
```
jreidinger: is it needed?

src/include/network/lan/complex.rb
```diff
@@ -515,7 +575,9 @@ def input_done?(ret)
       true
     end
 
-    def MainDialog(init_tab)
+    def MainDialog(init_tab, builder: nil)
```
jreidinger: and again what nil means here? New item and new builder?

src/include/network/lan/hardware.rb
```diff
@@ -848,16 +720,16 @@ def storeHW(_key, _event)
 
     # S/390 devices configuration dialog
     # @return dialog result
-    def S390Dialog
+    def S390Dialog(builder: nil)
```
jreidinger: nil crash here, not?

src/include/network/lan/wizards.rb
```diff
@@ -116,12 +118,13 @@ def LanAutoSequence(mode)
...
-        "global"    => -> { MainDialog("global") },
-        "overview"  => -> { MainDialog("overview") },
-        "add"       => [-> { NetworkCardSequence("add") }, true],
-        "edit"      => [-> { NetworkCardSequence("edit") }, true],
-        "init_s390" => [-> { NetworkCardSequence("init_s390") }, true]
+        "global"    => -> { MainDialog("global", builder: iface_builder) },
+        "overview"  => -> { MainDialog("overview", builder: iface_builder) },
+        "add"       => [-> { NetworkCardSequence("add", builder: iface_builder) }, true],
+        "edit"      => [-> { NetworkCardSequence("edit", builder: iface_builder) }, true],
+        "init_s390" => [-> { NetworkCardSequence("init_s390", builder: iface_builder) }, true]
```
jreidinger: sharing builder accross sequence looks reasonable approach for me.

src/lib/y2network/hwinfo.rb
```diff
@@ -0,0 +1,75 @@
...
+      { name: "driver", default: "" },
+      { name: "drivers", default: [] },
+      { name: "requires", default: [] },
+      { name: "hotplug", default: false },
+      { name: "wl_auth_modes", default: "" },
+      { name: "wl_enc_modes", default: nil },
+      { name: "wl_channels", default: nil },
+      { name: "wl_bitrates", default: nil }
+    ].each do |hwinfo_item|
+      define_method hwinfo_item[:name].downcase do
```
jreidinger: why downcase when everything is upcase? or there is keys in hwinfo with upper case? Because this can produce strange looking result like `DevName` -> `devname` where you expect `dev_name`.

src/lib/y2network/hwinfo.rb
```diff
@@ -0,0 +1,75 @@
...
+        @hwinfo ? @hwinfo.fetch(hwinfo_item[:name], hwinfo_item[:default]) : hwinfo_item[:default]
+      end
+    end
+    alias_method :name, :dev_name
+
+    def exists?
+      !@hwinfo.nil?
```
jreidinger: to be honest I do not like much this approach with checking everywhere nil. What about raising exception when hwinfo not found and simply return constant with defaults that is defined above? And does it make sense at all having this hwinfo full of nils, false and empty arrays?

src/lib/y2network/interface.rb
```diff
@@ -16,17 +16,75 @@
...
+  # In other words - we have two configurations 1) system configuration - the one
+  # which is currently used by running system and 2) yast configuration - the one which
+  # is being modified an will be applied to the system on user's request. name and
+  # hardware should always reflect the first configuration bcs it is the only connection /
+  # identifier we can use when touching the system (e.g. when setting interfaces down, etc.)
   class Interface
     # @return [String] Device name (eth0, wlan0, etc.)
-    attr_accessor :name
+    attr_reader :name # TODO: when implementing renaming over new backend modifying name has to be checked
+    attr_accessor :renamed_to # TODO: use in renaming workflow
```
jreidinger: I prefer something like system_name and name. This way you can also identify newly created interface.

src/lib/y2network/interface.rb
```diff
@@ -16,17 +16,75 @@
...
+    attr_reader :configured
+    attr_reader :hardware
+
+    extend Forwardable
+
+    def_delegator :@hardware, :exists?, :hardware?
+
+    # Shortcuts for accessing interfaces' ifcfg options
+    ["STARTMODE"].each do |ifcfg_option|
+      define_method ifcfg_option.downcase do
```
jreidinger: btw do not forget to properly document this dynamically created methods with yardoc `@method`.

src/lib/y2network/interface.rb
```diff
@@ -35,11 +93,36 @@ def initialize(name)
...
     alias_method :eql?, :==
+
+  private
+
+    def system_config(name)
+      Yast::NetworkInterfaces.devmap(name)
+    end
+
+    # Initializates depending variables
+    def init(name)
```
jreidinger: this is strange construct, as you basically have two initialize methods. Either add it directly to initialize or rename it like `read_configured` so it is clear what it does.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    # Returns currently stored option value
+    def option(option)
+      @config.fetch(option, "")
+    end
+
+    # Provides stored configuration in sysconfig format
+    #
+    # @return [Hash<String, String>] where key is sysconfig option and value is the option's value
+    def device_sysconfig
```
jreidinger: this is a bit strange for me. On one hand we are creating config builder, that can based on type do some validation. Proposing defaults and also hand partially defined config.
But here we tight it heavily with sysconfig and its keys. This looks wrong for me. We should try to have abstraction layer in builder and only as result take this abstract config and convert it into some sysconfig values. Ideally in different class that is sysconfig writer?

src/lib/y2network/interface_defaults.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    # Initialiates device configuration map with default values when needed
+    #
+    # @param devmap [Hash<String, String>] current device configuration
+    #
+    # @return device configuration map where unspecified values were set
+    #                to reasonable defaults
+    def init_device_config(devmap)
+      Yast.import "Arch"
```
jreidinger: wrong line? Should be lower in s390, not?

src/lib/y2network/interface_defaults.rb
```diff
@@ -0,0 +1,89 @@
...
+      # (as opposed to what a new interface gets, in {#Select)}
+      defaults = YAML.load_file(Yast::Directory.find_data_file("network/sysconfig_defaults.yml"))
+      defaults.merge(devmap)
+    end
+
+    def init_device_s390_config(devmap)
+      return {} if !Yast::Arch.s390
+
+      # Default values used when creating an emulated NIC for physical s390 hardware.
+      s390_defaults = YAML.load_file(Directory.find_data_file("network/s390_defaults.yml")) if Arch.s390
```
jreidinger: double test on Arch and here it will fail due to missing `Yast::`

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,71 @@
...
+    # Add an interface with the given name
+    #
+    # @param name [String] Interface's name
+    def add(name)
+      interfaces.push(Interface.new(name))
+    end
+
+    # Compares InterfacesCollections
+    #
+    # @return [Boolean] true when both collections are equal, false otherwise
```
jreidinger: a bit confusing, you probably want to say something like true when contain same elements?

src/lib/y2network/interfaces_collection.rb
jreidinger: yep, probably something like `other.interfaces.sort == @interfaces.sort`

src/modules/LanItems.rb
```diff
@@ -1193,7 +1186,8 @@ def Read
     # 3) variables which keeps (some of) attributes of the current item (= item
     # which is being pointed by the iterator)
     def reset_cache
-      LanItems.Items = {}
+      # FIXME: what is suitable replacement with new config storage?
```
jreidinger: probably something like reassign `:yast` config again with system one? but here it makes config basically completely empty. What is usecase?

src/modules/LanItems.rb
```diff
@@ -1232,95 +1226,35 @@ def Import(settings)
...
-    def needFirmwareCurrentItem
-      need = false
-      if IsNotEmpty(Ops.get_string(@Items, [@current, "hwinfo", "driver"], ""))
-        if Builtins.haskey(
-          @request_firmware,
-          Ops.get_string(@Items, [@current, "hwinfo", "driver"], "")
-        )
-          need = true
-        end
+    # Checks whether device need (known) firmware to be loaded
```
jreidinger: ok, please document that it returns driver or nil. Name of method is already horrible.

src/modules/LanItems.rb
```diff
@@ -1232,95 +1226,35 @@ def Import(settings)
...
+    def needFirmwareCurrentItem(iface)
+      need = if iface.hardware.driver
```
jreidinger: can you rename variable to something like driver?

src/modules/LanItems.rb
```diff
@@ -1386,8 +1320,8 @@ def bridge_slaves(master)
 
     # Creates item's startmode human description
     #
-    # @param item_id [Integer] a key for {#Items}
-    def startmode_overview(item_id)
+    # @param devmap [Hash] Interface sysconfig configuration
```
jreidinger: I expect here TODO as overview should not be tight to sysconfig only.

test/bridge_test.rb
```diff
@@ -84,6 +84,11 @@
       end
     end
 
+    # mock for network-ng
+    hwinfo_items.each do |i|
+      allow(Yast::NetworkInterfaces).to receive(:devmap).with(i["dev_name"]).and_return(nil)
+    end
```
jreidinger: if you need it at more places, you can do it in test helper and unstub only for specific tests.

src/include/network/lan/address.rb
mchf: right, the parameter should be mandatory and defined

src/include/network/lan/address.rb
mchf: definitely. But we don't have type hierarchy yet. Just one simple Interface class.

src/include/network/lan/address.rb
mchf: no problem with that, already done something similar for the Interface and HwInfo classes and make sense even here. I just wanted isolate all cases where the builder should be used instead of (e.g.) direct access to LanItems internals. For that purpose was hashlike api enough

src/include/network/lan/address.rb
mchf: I currently don't see any benefit in keeping it.

src/include/network/lan/address.rb
mchf: I'd like to. Original idea was to keep "open" interface in it and either submit or drop it as the user wishes (confirms Add / Edit or cancel it)

src/include/network/lan/complex.rb
mchf: yes it is. Bcs I'm currently injecting new things into old workflow. And as i currently use sysconfig option names as identifiers in the builder I've picked this name. Definitely something for improving

src/include/network/lan/complex.rb
mchf: Whole Add / Edit / ... workflow handling needs some polishing as it is somehow spread over the code. We currently need to isolate what is needed to keep ... so thanks for any hint ;-)

src/include/network/lan/complex.rb
mchf: yes ... already planed and described somewhere in this PR. This operation can be identified according to configuration name (I've introduced interface.name to be "persistent" and reflects current state of the system) name exists -> edit action, name do not exist -> add action

src/include/network/lan/complex.rb
mchf: most probably not. Just wanted to catch some unexpected accesses.

src/include/network/lan/complex.rb
mchf: nothing. Should be ```(builder:)``` and mandatory

src/include/network/lan/hardware.rb
mchf: same as above

src/lib/y2network/hwinfo.rb
mchf: only readability ... i prefer ```obect.method``` over ```object.METHOD``` or so. And I'm also not sure if there is any exception in naming in the hash provided by libhd

src/lib/y2network/hwinfo.rb
mchf: i planned to keep only necessary things or load on demand in the future.

src/lib/y2network/interface.rb
mchf: yes, as already described ... we differ only in variable naming ;-)

src/lib/y2network/interface.rb
mchf: this has already changed several times. Need to polish once it is stable. Taking note.

src/lib/y2network/interface_config_builder.rb
mchf: completely agree

src/modules/LanItems.rb
mchf: leftover from the old code when it handled cases when some variables were left initialized to some value and not properly refreshed. Hopefully will not be needed in the bright future.

src/modules/LanItems.rb
mchf: ok

src/modules/LanItems.rb
mchf: ok

src/modules/LanItems.rb
mchf: ok

src/lib/y2network/interfaces_collection.rb
mchf: ok

src/lib/y2network/interface_defaults.rb
mchf: ok

src/lib/y2network/interface_defaults.rb
mchf: true

src/lib/y2network/interface.rb
mchf: ok

src/include/network/lan/complex.rb
mchf: replaced

src/include/network/lan/hardware.rb
mchf: replaced

src/include/network/lan/address.rb
mchf: replaced

mchf:
Rebased on top of current network-ng

src/lib/y2network/interface_config_builder.rb
```diff
@@ -91,5 +87,70 @@ def load_sysconfig(devmap)
...
+      # FIXME: NetHwDetection is done in Lan.Read
+      Yast.import "NetHwDetection"
+
+      # #104494 - always write IPADDR+NETMASK, even empty
+      # #50955 omit computable fields
+      config["BROADCAST"] = ""
+      config["NETWORK"] = ""
+
+      if !config["NETMASK"] || config["NETMASK"].empty?
+        config["NETMASK"] = Yast::NetHwDetection.result["NETMASK"] || "255.255.255.0"
```
jreidinger: how can fallback happen and is it nice behavior to user to write something that they do not specify?

src/lib/y2network/interface_config_builder.rb
```diff
@@ -91,5 +87,70 @@ def load_sysconfig(devmap)
...
+      startmode
+    end
+
+    def replace_ifplugd?
+      Yast.import "Arch"
+      Yast.import "NetworkService"
+
+      return true if !Yast::Arch.is_laptop
+      return true if Yast::NetworkService.is_network_manager
+      return true if ["bond", "vlan", "br"].include? type
```
jreidinger: can you add comment, why? I see what it does, but missing rationale. It can help in future if we need to extend list of types for replacing it.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -91,5 +87,70 @@ def load_sysconfig(devmap)
...
+
+      false
+    end
+
+    def hotplug_usable?
```
jreidinger: I am not sure if name is not misleading. From code it looks like if device is hotplug of not, right?

src/lib/y2network/interface_config_builder.rb
mchf: well I can only add my opinion but I cannot add real reason. I've only moved what already had been in LanItems for long time

src/lib/y2network/interface_config_builder.rb
mchf: yes

src/lib/y2network/interface_config_builder.rb
jreidinger: ok, so please add fixme to find that reason or that we do not know it. It at least make us more brave when we want to touch it, as we have confirmed that we do not know :)

src/lib/y2network/interface_config_builder.rb
jreidinger: so what about `hotplug_interface?` as method name? that "usable" is quite confusing for me.

src/lib/y2network/hwinfo.rb~HEAD
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) [2019] SUSE LLC
+#
```
mvidner: This file is a mistake

mchf: true

mchf:
Some parts were already merged in other PRs, remaining parts were postponed for now.

PR #797, opened 2019-04-30
Use nameservers, added section documentation and Autoinst reader

teclator:


src/lib/y2network/config_reader/autoinst_dns.rb
```diff
@@ -0,0 +1,49 @@
...
+      def initialize(section)
+        @section = section
+      end
+
+      # Creates a new {DNS} config from the imported profile dns section
+      #
+      # @return [DNS] the imported {DNS} config
+      def config
+        options = {}
+        section.class.attributes.each do |attr_entry|
```
imobachgs: I would use `each_with_object`.

PR #800, opened 2019-05-03
network-ng DNS module cleanup

imobachgs:
This PR contains some fixes to make the DNS UI work. Additionally, it removes some unused code.


src/include/network/services/dns.rb
```diff
@@ -275,14 +275,18 @@ def StoreSettings(settings)
       )
 
       DNS.hostname = Ops.get_string(settings, "HOSTNAME", "")
-      DNS.nameservers = NonEmpty(nameservers)
+      valid_nameservers = NonEmpty(nameservers).each_with_object([]) do |ip_str, all|
+        begin
+          ip = IPAddr.new(ip_str)
+          all << ip
```
mchf: we already have IP validators in yast - see ```Yast::IP.Check```

src/lib/network/network_autoconfiguration.rb
```diff
@@ -90,15 +90,7 @@ def configure_virtuals
     # Propose DNS and Hostname setup
     def configure_dns
```
mchf: it seems to be completely useless now ;-) However I'm not sure if removing ```ProposeHostname``` is correct

src/lib/y2network/dns.rb
```diff
@@ -21,19 +21,19 @@ module Y2Network
   # DNS configuration (hostname, nameservers, etc.).
   class DNS
     # @return [String] Hostname (local part)
-    attr_reader :hostname
+    attr_accessor :hostname
```
mchf: I don't think that hostname belongs here. I know that in old module it was in DNS module, however this hostname (/etc/hostname) is not much related to networking and DNS.

imobachgs: We already had a discussion about that and @teclator and I did not find a better place (perhaps directly under `Y2Network::Config`), so we decided to keep it here meanwhile. But I do not have a strong opinion about moving it to a different place.

src/lib/network/network_autoconfiguration.rb
imobachgs: Why?

src/lib/y2network/dns.rb
teclator: @mchf If I'm not wrong it was proposed to remove the handling of the `/etc/hostname` but not sure if it was really discussed or not and the conclusions as at least that option should be moved to other component.

So maybe the tab should just take care of DNS, and the `dhcp_hostname` option should be moved elsewhere (maybe in the General tab with the **DHCLIENT_SET_DEFAULT_ROUTE**)
I'f

teclator:
For me looks good.

Maybe I would modify the validation of the `Hostname` Input in include/network/services/dns.rb. It validates it is a short Hostname, but in case the /etc/hostname is a `FQDN` it will complain and will obligate the user to modify it to apply changes in the DNS section. 

But from other POV that is also what was somehow requested in https://bugzilla.suse.com/show_bug.cgi?id=972463#c25

src/lib/network/network_autoconfiguration.rb
imobachgs: I've restored the method (now it is called #propose_hostname) so we are hopefully on the safe side :)

src/lib/y2network/dns.rb
mchf: I was not talking about UI but the class responsibilities. If the class is named DNS then it should take care of DNS and not implement everything what was in the same module in the old code.

mchf:
it lgtm except one np comment

imobachgs:
> it lgtm except one np comment

It is a valid concern. We can address it when the hostname handling is finally clarified (@teclator si working on that).

PR #807, opened 2019-05-13
more CWM

jreidinger:


src/include/network/widgets.rb
```diff
@@ -152,6 +152,7 @@ def ValidateIP(key, _event)
     end
 
     def handleStartmode(_key, _event)
+      # FIXME: not working with object CWM
```
jreidinger: or better, it can work, but it need to store ifplugd_priority widget and operate on them and not direcly on UI ChangeWidget level.

src/lib/y2network/widgets/startmode.rb
```diff
@@ -0,0 +1,99 @@
...
+              "<b>Never</b> does not start the device.\n" \
+              "%1</p>\n"
+          ),
+          items_help.join(" ")
+        )
+      end
+
+      def items
+        [
+          # onboot, on and boot are aliases for auto
```
mchf: seems that "on" is uselees

jreidinger: yes, I also removed if you notice that NetworkManager item, as it is not used nor displayed. If we need dynamic approach in future, We should handle it inside this class.

src/lib/y2network/widgets/ifplugd_priority.rb
```diff
@@ -0,0 +1,46 @@
...
+    class IfplugdPriority < CWM::IntField
+      def initialize(config)
+        textdomain "network"
+
+        @config = config
+      end
+
+      def label
+        _(
+          "Ifplugd Priority"
```
jreidinger: Btw I hate this label. We should not disclosure some ifplugd where noone knows what it is. It is something like "Apply Priority" or stuff like this. Basically if more on cable connection is  there, then this number decide which to apply ( the one which is available and has the lowest number ).

teclator:
LGTM

PR #808, opened 2019-05-13
Add ConnectionConfig classes

teclator:
We have added a new `ConnectionConfig`class in order to be able to configure multiple connections in general but also multiple connections for the same interface, something that currently is not supported by YaST. 

Wicked uses `nanny` policies, and NetworkManager already uses the `connection` terminology.

It will permit to have for example multiple wireless configurations or have different configuration for the same interface depending on a set of rules ... (that is, if the connection is from home, the office.. or whatever)


src/lib/y2network/connection/connection.rb
```diff
@@ -0,0 +1,20 @@
+require "y2network/interface"
+
+module Y2Network
+  module Connection
+    class Connection
+      # @return [Interface]
+      attr_accessor :interface
+      # @return [String] Bootproto (static, dhcp, none)
```
jreidinger: does it make sense for all connection or only physical ones?

src/lib/y2network/connection/connection.rb
```diff
@@ -0,0 +1,20 @@
...
+      attr_accessor :bootproto
+      # @return [IPAddr]
+      attr_accessor :ip_address
+      # @return [Array<IPAddr>]
+      attr_accessor :secondary_ip_addresses
+      # @return [Integer]
+      attr_accessor :mtu
```
jreidinger: same here, does mtu makes sense for all connections?

src/lib/y2network/connection/connection.rb
```diff
@@ -0,0 +1,20 @@
...
+      # @return [IPAddr]
+      attr_accessor :remoteip
```
jreidinger: what is this? next hop when communicate?

src/lib/y2network/connection/ethernet.rb
```diff
@@ -0,0 +1,8 @@
+require "y2network/connection/connection"
+
+module Y2Network
+  module Connection
+    class Ethernet < Connection
```
jreidinger: this for me looks like that connection is too much ethernet, I think there are connections that has differences to Ethernet config.

src/lib/y2network/interface.rb
```diff
@@ -21,6 +21,8 @@ module Y2Network
   class Interface
     # @return [String] Device name (eth0, wlan0, etc.)
     attr_accessor :name
+    # @return [String]
+    attr_accessor :ethtool_options
```
jreidinger: really all interfaces have to have this options?

src/lib/y2network/connection/connection.rb
teclator: Why only for physical? BTW, I have not checked all possible scenarios, I started moving the attributes from `LanItems` to the `Interface` and then as we discussed about adding the connection abstraction class .. I moved most of the attributes that seems to be generic to the connection class and the ones specific for some type to the specialized class but some of there probably need to be revised.

src/lib/y2network/connection/connection.rb
teclator: While the `bootproto` is or seems to be clear a connection attribute, I was not so sure about the MTU. Probably it should be moved to the specialized classes that permits to establish the MTU

src/lib/y2network/connection/connection.rb
teclator: It should be specialized attribute. It is the remote IP address of a point to point connection, basically them are in connection but definitely should add a comment to the ones which are unclear yet or not specialized class defined.

src/lib/y2network/interface.rb
teclator: No, they haven't. We have to look a better place to add it, but so far we only have specialized classes for connection apart of the virtual and physical Interface subclasses. 

I will remove this option by now or will just update the comments, first adding a FIXME or TODO label, and permitting nil in the ones that are not generic / mandatory.

src/lib/y2network/connection/connection.rb
mchf: I didn't see startmode attr. The simplest possible configuration is
```
STARTMODE=auto
BOOTPROTO=dhcp
```

src/lib/y2network/connection/connection.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module Y2Network
+  module Connection
+    class Connection
+      # @return [Interface]
+      attr_accessor :interface
+      # @return [String] Bootproto (static, dhcp, none)
+      attr_accessor :bootproto
+      # @return [IPAddr]
+      attr_accessor :ip_address
```
mchf: i don't think this is basic attribute. The most often you need to distinguish whether the interface was configured dynamically or statically and nothing more. But the connection object can provide a kind of configuration descriptor which can contain some details (like ip address, ...)

src/lib/y2network/connection/connection.rb
mchf: MTU is related to physical layer (eth, ib, tokenring ;-) and do not make sense e.g. for vlans

src/lib/y2network/connection/connection.rb
```diff
@@ -0,0 +1,20 @@
+require "y2network/interface"
+
+module Y2Network
+  module Connection
+    class Connection
```
mchf: i don't understand the name. What the "Connection" means here? I'd expect e.g. PTP connection but not a class carrying interface configuration.

imobachgs: We could change the name to "Config" or whatever (we used the network manager term). This class basically holds the configuration (and for the time being is related to an interface).

teclator: It is based somehow on the idea of permiting having different connection profiles. That could belongs to an specific interface or not, wicked also support nanny policies that would be nice to support in the future. So we are trying to create a model that permits us to handle that.

src/lib/y2network/connection/connection.rb
teclator: In nm-connection-editor.. the MTU is allowed per VLAN... BTW, currently we offer it in all the interfaces configuration, isn't it?

mchf: yes, in general tab ... but it doesn't mean that it is correct.

vlan is defined at physical layer (you can see it as extension of physical layer) - especially tagged vlan's which extends (e.g.) ethernet frame with vlan id.

src/lib/y2network/connection/connection.rb
mchf: I understand what it tries to do. Just wanted to point that "Connection" is a bit misleading term in networking. It could be more precise, like "ProfileConnection" or something even better.

src/lib/y2network/connection/connection.rb
teclator: It basically let the responsibility into the user

src/lib/y2network/physical_interface.rb
```diff
@@ -1,8 +1,9 @@
 require "y2network/interface"
+require "y2network/hardware_info"
 
 module Y2Network
   module Interfaces
-    class Physical < Interface
+    class PhysicalInterface < Interface
```
jreidinger: I think it should no longer live in Interfaces module, not?

src/lib/y2network/connection/connection.rb
teclator: BTW: STARTMODE is defined in interface in your opened PR

src/lib/y2network/connection/connection.rb
teclator: Used "ConnectionConfig" namespace

examples/connection.rb
```diff
@@ -0,0 +1,27 @@
+require "yast"
+require "y2network/config"
+require "y2network/virtual_interface"
+require "y2network/connection_config/ethernet"
+require "y2network/connection_config/bond"
+require "y2network/connection_config/wireless"
+
+# given an interfaces list (interfaces)
+
+config = Y2Network::Config.from(:sysconfig)
```
jreidinger: to be honest I still do not like this initialization. Ideally it should be `Y2Network::Config.read` and it will read configuration whatever it want and however. I am not interested if it use sysconfig, wicked xml or network manager udev service. Just read current configuration.

imobachgs: It is fine for me to have such a method, but you need to detect which backend you are using. So, for me, it makes sense to have a `Y2Network::Config.read` which finds out which is the backend and then it just simply call to `Config.from` with the proper value.

examples/connection.rb
```diff
@@ -0,0 +1,27 @@
...
+
+eth0 = config.interfaces.find("eth0").first
+eth_conn = Y2Network::ConnectionConfig::Ethernet.new
+eth_conn.interface = eth0
+
+eth1 = config.interfaces.find("eth1").first
```
jreidinger: I do not get why find returns array and not single element. That is strange. Also it can find multiple interfaces with name "eth1" - it is strange.

examples/connection.rb
```diff
@@ -0,0 +1,27 @@
...
+eth2 = config.interfaces.find("eth2").first
+bond = Y2Network::ConnectionConfig::Bond.new
+bond.slaves = [eth1, eth2]
+bond.options = "mode=active-backup miimon=100"
+bond.interface = Y2Network::VirtualInterface.new("bond0")
+config.interfaces << bond
+
+wlan0 = config.interfaces.find("wlan0").first
+wlan_conn = Y2Network::ConnectionConfig::Wireless.new
```
jreidinger: hmm, and how can I find current config for given interface? or even multiple confiuration if we allow profiles?

src/lib/y2network/connection_config/base.rb
```diff
@@ -0,0 +1,38 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/interface"
+
+module Y2Network
+  module ConnectionConfig
+    # This class is reponsible of a connection configuration
+    class Base
+      # @return [Interface, nil]
```
jreidinger: what means if Interface is nil? is it valid config or not?

examples/connection.rb
imobachgs: `interfaces.find { |i| i.name == "eth1" }`

src/lib/y2network/connection_config/base.rb
```diff
@@ -0,0 +1,38 @@
...
+      attr_accessor :interface
+      # @return [String] Bootproto (static, dhcp, none)
+      attr_accessor :bootproto
```
jreidinger: I play now with bootproto and I think that ideally this should not be just writter, but be a bit smarter. E.g. for dhcp explicitelly clear ip address and netmask nor hostname and so on.

src/lib/y2network/connection_config/base.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require "y2network/interface"
+
+module Y2Network
+  module ConnectionConfig
+    # This class is reponsible of a connection configuration
+    class Base
+      # @return [Interface, nil]
+      attr_accessor :interface
+      # @return [String] Bootproto (static, dhcp, none)
```
jreidinger: ibft is also possible value. And what about zeroconf ( aka autoip ) and autoip combination with dhcp and also dhcp version specific ( dhcp4 really helps with speed on ipv4 only network ).

src/lib/y2network/connection_config/wireless.rb
```diff
@@ -0,0 +1,63 @@
...
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for wireless connections
+    class Wireless < Base
+      # wireless options
+      # @return [String]
+      attr_accessor :mode
```
jreidinger: maybe enum instead of string? at least add fixme.

src/lib/y2network/connection_config/wireless.rb
```diff
@@ -0,0 +1,63 @@
...
+      attr_accessor :channel
+      # @return [Integer]
+      attr_accessor :frequency
+      # @return [Integer]
+      attr_accessor :bitrate
+      # @return [String]
+      attr_accessor :accesspoint
+      # @return [Boolean]
+      attr_accessor :power
+      # @return [String]
```
jreidinger: maybe also enum?

examples/connection.rb
imobachgs: Good question. I guess we should set up a relationship between them. But that's something that, most probably, the config reader should do (and we will add the relation when needed).

examples/connection.rb
teclator: I updated the example

src/lib/y2network/connection_config/base.rb
teclator: It is open to discuss. You could define a connection configuration not attached to any specific interface, so the first resource / device available will apply the configuration

src/lib/y2network/connection_config/base.rb
teclator: Sure, I could put all the options

static*|dhcp|dhcp4|dhcp6|autoip|dhcp+autoip|auto6|6to4|none

src/lib/y2network/connection_config/wireless.rb
teclator: I will add the FIXME, in general I just went from LanItems variables moving here the wireless specific, but not going too much into details.

So when we start to add some real behavior, or specific use of this class we will be able to rethink the current internal state.

examples/connection.rb
jreidinger: yes Config.from makes sense for me, but in general I think Config.read should be preferred way to read configuration.

src/lib/y2network/connection_config/base.rb
jreidinger: ok, so please document it what it stands for. As at first look it looks strange, but if nil works as * wildcard, it is fine for me.

src/lib/y2network/connection_config/base.rb
jreidinger: or point to possible values to enum, which is ideal solution as then we can have storage like `to_human_string` method, that returns human friendly name.

src/lib/y2network/connection_config/base.rb
teclator: We could think on it when adding the new dialogs / workflow. Currently its just associated with bootproto ifcfg attribute but we could add another abstraction also the configuration associated with dhcp itself...

src/lib/y2network/connection_config/base.rb
```diff
@@ -25,7 +25,8 @@ module ConnectionConfig
     class Base
       # @return [Interface, nil]
       attr_accessor :interface
-      # @return [String] Bootproto (static, dhcp, none)
+      # @return [String] Bootproto (static, dhcp, ,dhcp4, dhcp6, autoip,
+      #   dhcp+autoip, auto6, 6to4, none)
```
jreidinger: interesting, we do not support in ui 6to4 and it will fail there. Probably something to fix?

teclator: And there are more interfaces that are supported by wicked but not by YaST yet, so would be nice to have but step by step.

src/lib/y2network/connection_config/base.rb
teclator: We will not apply it yet, I mean, by now, with wicked as the target, all the connections will have an interface associated.

src/lib/y2network/connection_config/base.rb
jreidinger: well, issue is that now if user has this configuration it will crash. We probably need to introduce something like bootloader UnsupportedConfiguration exception that allow user to repropose configuration or keep it.

teclator: Have you tried?, If that is the case then definitely.

PR #809, opened 2019-05-14
Edit src/include/network/services/routing.rb  for accept "0.0.0.0" and "/0"like netmask in the routing table 

SuseJose:
https://bugzilla.opensuse.org/show_bug.cgi?id=1100432


SuseJose:
> https://bugzilla.opensuse.org/show_bug.cgi?id=1100432

Rakefile
```diff
@@ -1,7 +1,5 @@
 require "yast/rake"
 
-Yast::Tasks.submit_to :sle12sp3
```
jreidinger: looks like you use the wrong target branch. Do you really want to fix it in SLE12-SP3 ( aka leap 42.3 ) ?

jreidinger: also some other changes looks wrong in this branch as it is from newer branch.

PR #810, opened 2019-05-14
Routing default zero

SuseJose:
This PR fix the issue reported in this [bug](https://bugzilla.opensuse.org/show_bug.cgi?id=1100432) and replaces #809 where I did some mistakes merging master into SLE-15-SP1 and other issues so preferred to close it opening a new one.

It just permits to use a 0.0.0.0 ip address with a /0 netmask to defined a default route from the table route addition dialog as reported / requested in the bug.

bug img
[fix img]()

### DIFF
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -472,7 +472,7 @@ module Yast
       end
 
       route_dest = "#{dest}/#{cidr}"
-      route["destination"] = route_dest.start_with?("0.0.0.0" ) ? "default" : route_dest
+      route["destination"] = route_dest.start_with?("0.0.0.0") ? "default" : route_dest
       route["netmask"] = "-"
 
       route
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -152,7 +152,7 @@ module Yast
       return true if Netmask.Check4(netmask)
 
       if netmask.start_with?("/")
-        return true if netmask[1..-1].to_i.between?(1, 128)
+        return true if netmask[1..-1].to_i.between?(0, 128)
       end
 
       false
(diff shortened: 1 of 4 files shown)


src/include/network/services/routing.rb
```diff
@@ -152,7 +152,7 @@ def valid_netmask?(netmask)
       return true if Netmask.Check4(netmask)
 
       if netmask.start_with?("/")
-        return true if netmask[1..-1].to_i.between?(1, 128)
+        return true if netmask[1..-1].to_i.between?(0, 128)
```
teclator: Probably only valid for "default" route.

test/routing_helpers_test.rb
```diff
@@ -69,13 +69,11 @@ def initialize
       expect(routing.valid_netmask?("/255/")).to be false
       expect(routing.valid_netmask?("/255")).to be false
       expect(routing.valid_netmask?("/-255")).to be false
-      expect(routing.valid_netmask?("/0")).to be false
```
teclator: only valid when defined with a `default` route.

test/routing_helpers_test.rb
```diff
@@ -69,13 +69,11 @@ def initialize
...
     end
 
     it "declines malformed IPv4 netmask" do
       expect(routing.valid_netmask?("/255.0.0.0")).to be false
       expect(routing.valid_netmask?("255.0.255.0")).to be false
-      expect(routing.valid_netmask?("0.0.0.0")).to be false
```
teclator: Same than above

teclator:
@SuseJose it needs to fix some conflicts and some minor changes. 

Closing it by now as we agreed recently.

PR #812, opened 2019-05-15
More cwm objects

jreidinger:


src/lib/y2network/widgets/boot_protocol.rb
```diff
@@ -0,0 +1,364 @@
...
+      def valid_netmask(ip, mask)
+        valid_mask = false
+        mask = mask[1..-1] if mask.start_with?("/")
+
+        if Yast::IP.Check4(ip) && (Yast::Netmask.Check4(mask) || Yast::Netmask.CheckPrefix4(mask))
+          valid_mask = true
+        elsif Yast::IP.Check6(ip) && Yast::Netmask.Check6(mask)
+          valid_mask = true
+        else
+          log.warn "IP address #{ip} is not valid"
```
mchf: a bit strange when method is named ```valid_netmask``` ;-) May be it should be ```valid_static_ip_setup``` or something like that

src/lib/y2network/widgets/boot_protocol.rb
```diff
@@ -0,0 +1,364 @@
...
+
+        ipa = Yast::UI.QueryWidget(:bootproto_ipaddr, :Value)
+        if !ipa.empty? && !Yast::IP.Check(ipa)
+          Yast::Popup.Error(_("No valid IP address."))
+          Yast::UI.SetFocus(:bootproto_ipaddr)
+          return false
+        end
+
+        mask = Yast::UI.QueryWidget(:bootproto_netmask, :Value)
+        if ipa != "" && mask != "" && !valid_netmask(ipa, mask)
```
mchf: NP: these checks for empty values can be removed. Validating functions should catch that ... or should be fixed to catch that.

src/lib/y2network/widgets/boot_protocol.rb
```diff
@@ -0,0 +1,364 @@
...
+        else
+          raise "Unexpected boot protocol value #{Yast::UI.QueryWidget(:bootproto, :Value).inspect}"
+        end
+      end
+
+      def validate
+        return true if value != :bootproto_static
+
+        ipa = Yast::UI.QueryWidget(:bootproto_ipaddr, :Value)
+        if !ipa.empty? && !Yast::IP.Check(ipa)
```
mchf: ```Yast::IP.Check("")``` -> false => ```ipa.empty?``` can be removed

src/lib/y2network/widgets/boot_protocol.rb
```diff
@@ -0,0 +1,364 @@
...
+                  [
+                    Item(Id(:bootproto_dhcp_both), _("DHCP both version 4 and 6")),
+                    Item(Id(:bootproto_dhcp_v4), _("DHCP version 4 only")),
+                    Item(Id(:bootproto_dhcp_v6), _("DHCP version 6 only"))
+                  ]
+                )
+              )
+            ),
+            VBox(
+              # TODO : Stat ... Assigned
```
mchf: ?

jreidinger: I have no idea what it means, it was in original code.

src/lib/y2network/widgets/boot_protocol.rb
jreidinger: good tip.

src/lib/y2network/widgets/boot_protocol.rb
```diff
@@ -0,0 +1,364 @@
+require "cwm/custom_widget"
```
mchf: NP: I'd expect ```require "yast"``` too

src/lib/y2network/widgets/boot_protocol.rb
jreidinger: that is a bit problematic, as it means that we cannot leave tab without specifying it.

There are two possible approaches:

1. do not allow to leave tab until everything is correct
2. builder that validates its content and write error popup when user try to finish creating such device

Not such which one is better as both have advantages and disadvantages.

src/lib/y2network/widgets/boot_protocol.rb
jreidinger: yes, true

src/lib/y2network/widgets/boot_protocol.rb
jreidinger: well, original name is validPrefixOrNetmask(ipa, mask) so I just try to shorten it :)

src/lib/y2network/widgets/boot_protocol.rb
jreidinger: ah, now I see. `ipa` does not need to be checked. and mask is first time checked here and here is validating function and really catched.

src/lib/y2network/widgets/boot_protocol.rb
mchf: that was the point .... doesn't make sense to keep comments which cannot be understand ;-)

PR #813, opened 2019-05-16
Add an InterfacesCollection class

imobachgs:
* Taken from https://github.com/yast/yast-network/pull/796.
* Not intended to be merged until `Y2Network::Interface`provides a `#hardware` (or `#hwinfo`) method.

### DIFF
--- a/src/lib/y2network/interfaces_collection.rb
+++ b/src/lib/y2network/interfaces_collection.rb
@@ -51,6 +51,8 @@ module Y2Network
 
     # Returns an interface with the given name if present
     #
+    # @todo It uses the hardware's name as a fallback if interface's name is not set
+    #
     # @param name [String] Returns the interface with the given name
     # @return [Interface,nil] Interface with the given name or nil if not found
     def by_name(name)
--- a/src/lib/y2network/interfaces_collection.rb
+++ b/src/lib/y2network/interfaces_collection.rb
@@ -65,7 +65,7 @@ module Y2Network
     # @return [Boolean] true when both collections contain only equal interfaces,
     #                   false otherwise
     def ==(other)
-      @interfaces - other.interfaces && other.interfaces == @interfaces
+      ((interfaces - other.interfaces) | (other.interfaces - interfaces)).empty?
     end
 
     alias_method :eql?, :==
--- a/src/lib/y2network/interfaces_collection.rb
+++ b/src/lib/y2network/interfaces_collection.rb
@@ -1,0 +1,73 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/interface"
+require "forwardable"
+
+module Y2Network
+  # A container for network devices. In the end should implement methods for mass operations over
+  # network interfaces like old LanItems::find_dhcp_ifaces.
+  #
+  # @example Create a new collection
+  #   eth0 = Y2Network::Interface.new("eth0")
+  #   collection = Y2Network::InterfacesCollection.new(eth0)
+  #
+  # @example Find an interface using its name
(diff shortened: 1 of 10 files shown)


src/lib/y2network/interfaces_collection.rb
```diff
@@ -65,7 +65,7 @@ def add(name)
     # @return [Boolean] true when both collections contain only equal interfaces,
     #                   false otherwise
     def ==(other)
-      @interfaces - other.interfaces && other.interfaces == @interfaces
+      ((interfaces - other.interfaces) | (other.interfaces - interfaces)).empty?
```
jreidinger: well, `|` is quite expensive and complex operation. I think simple `+` is more understandable and also works.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,77 @@
...
+    def_delegators :@interfaces, :each, :push, :<<, :reject!
+
+    # Constructor
+    #
+    # @param interfaces [Array<Interface>] List of interfaces
+    def initialize(interfaces = [])
+      @interfaces = interfaces
+    end
+
+    # Returns a interface with the given name if present
```
jreidinger: `an interface`

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,77 @@
...
+    # @param name [String] Returns the interface with the given name
+    # @return [Interface,nil] Interface with the given name or nil if not found
+    def by_name(name)
+      find { |i| i.name ? i.name == name : i.hardware.name }
+    end
+
+    # Add an interface with the given name
+    #
+    # @param name [String] Interface's name
+    def add(name)
```
jreidinger: this looks very strange for me. What is purpose of this method?

src/lib/y2network/interfaces_collection.rb
jreidinger: and please also mention not so obvious thing, that it also search for hardware name if it is not configured.

teclator:
For me looks good once updated according to @jreidinger comment.

src/lib/y2network/interfaces_collection.rb
imobachgs: It was basically inherited from the previous implementation, but now with the new Physical/Virtual implementation, it should be removed. I will drop it.

mchf: used here https://github.com/mchf/yast-network/blob/replace-items-rebased/src/include/network/lan/complex.rb#L158 (when adding a configuration for nonexistent interface)

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,88 @@
...
+module Y2Network
+  # A container for network devices. In the end should implement methods for mass operations over
+  # network interfaces like old LanItems::find_dhcp_ifaces.
+  #
+  # @example Create a new collection
+  #   eth0 = Y2Network::Interface.new("eth0")
+  #   collection = Y2Network::InterfacesCollection.new(eth0)
+  #
+  # @example Find an interface using its name
+  #   iface = collection.find("eth0") #=> #<Y2Network::Interface:0x...>
```
jreidinger: this does not look correct. probably `by_name`

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,88 @@
...
+  class InterfacesCollection
+    # Objects of this class are able to keep a list of interfaces and perform simple queries
+    # on such a list.
+    #
+    # @example Finding an interface by its name
+    #   interfaces = Y2Network::InterfacesCollection.new([eth0, wlan0])
+    #   interfaces.find("wlan0") # => wlan0
```
jreidinger: same here

src/lib/y2network/interfaces_collection.rb
```diff
@@ -0,0 +1,88 @@
...
+    end
+
+    # Returns an interface with the given name if present
+    #
+    # @todo It uses the hardware's name as a fallback if interface's name is not set
+    #
+    # @param name [String] Returns the interface with the given name
+    # @return [Interface,nil] Interface with the given name or nil if not found
+    def by_name(name)
+      interfaces.find { |i| i.name ? i.name == name : i.hardware.name }
```
jreidinger: is it correct? it will return the first interface that has defined `hardware.name` . Probably comparison to name is missing?

src/lib/y2network/interfaces_collection.rb
```diff
@@ -64,7 +64,10 @@ def initialize(interfaces = [])
     # @param name [String] Returns the interface with the given name
     # @return [Interface,nil] Interface with the given name or nil if not found
     def by_name(name)
-      interfaces.find { |i| i.name ? i.name == name : i.hardware.name }
+      interfaces.find do |iface|
+        iface_name = iface.name ? iface.name : iface.hwinfo.name
```
jreidinger: note, why not simple `iface.name || iface.hwinfo.name`

PR #814, opened 2019-05-21
Immutable type

jreidinger:


src/lib/y2network/widgets/interface_type.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      def label
+        _("&Device Type")
+      end
+
+      def help
+        # FIXME: help is not helpful
+        _(
+          "<p><b>Device Type</b>. Various device types are available, select \n" \
+            "one according your needs.</p>"
```
teclator:
```suggestion
            "one according to your needs.</p>"
```

src/lib/y2network/widgets/vlan_id.rb
```diff
@@ -0,0 +1,37 @@
...
+      def label
+        _("VLAN ID")
+      end
+
+      def help
+        # TODO: previously not exist, so write it
+      end
+
+      def init
+        self.value = (@config["VLAN_ID"] || "0").to_i
```
teclator: It seems to not be based on the proposed name anymore

src/lib/y2network/widgets/interface_type.rb
jreidinger: This is basically just copy to not break translation. As you can see in fixme, we really really should write there something useful. E.g. about types of cards.

src/lib/y2network/widgets/interface_name.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Widgets
+    class InterfaceName < CWM::ComboBox
+      def initialize(settings)
+        textdomain "network"
+
+        @settings = settings
+      end
+
+      def label
+        _("&Configuration Name")
```
teclator: Currently it is the interface name and not the configuration one.

NP: It would be nice the widget to be inside a Frame just for be consistent with the rest of the layout.

src/lib/y2network/widgets/vlan_id.rb
jreidinger: yes, it is true as name is not defined before vlan id anymore.

src/lib/y2network/widgets/interface_name.rb
jreidinger: Well, not necessary, some are inside frame and some not. I think we probably need to unify it.

about interface versus configuration name. I just copy label from previous one. I plan to revisit all texts if it makes sense later when we can easily change it in each widget.

teclator:
In general it looks good to me

src/lib/y2network/widgets/interface_name.rb
teclator: Ok, then. Maybe being just above the "Device Activation", using "Connection" and then "Device" instead of using same term was somehow strange.

But we were already discussing about use a matching rule for matching the connection configuration with an interface so.. let it as it is.

teclator: :+1: about revisit all the texts. I think you are doing good improvements with each CWM object that you are currently rewritten.

teclator: BTW: I was thinking that we could use always "Device Name", and when it was named with udev, just show that with the change button and maybe show the description the match rule key ("bus_id" or "mac address")... but that is just and idea

src/lib/y2network/widgets/vlan_id.rb
teclator: Yep, but at least would be nice to be the same than the proposed name. So currently if you have more than one in order starting with 0... the ID will be always 0. But that is something not critical, just something nice to have.

src/include/network/lan/wireless.rb
```diff
@@ -1123,27 +1123,31 @@ def WirelessKeysDialog
 
     # -------------------- WPA EAP --------------------
 
+    # `RadioButtonGroup uses CurrentButton instead of Value, grrr
```
imobachgs: :)

src/lib/y2network/widgets/additional_addresses.rb
```diff
@@ -0,0 +1,274 @@
...
+                  # Table header label
+                  _("IP Address"),
+                  # Table header label
+                  _("Netmask")
+                ),
+                []
+              ),
+              Left(
+                HBox(
+                  # PushButton label
```
imobachgs: np: I would write `PushButton labels` once.

src/include/network/lan/wireless.rb
jreidinger: Actually not my comment. Just Copy and paste :) object CWM has it already handled - https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/common_widgets.rb#L290

src/lib/y2network/widgets/additional_addresses.rb
```diff
@@ -0,0 +1,274 @@
...
+            alias_["PREFIXLEN"] = mask[1..-1]
+          else
+            param = Yast::Netmask.Check6(mask) ? "PREFIXLEN" : "NETMASK"
+            alias_[param] = mask
+          end
+          Yast::LanItems.aliases[i.to_s] = alias_
+        end
+        # TODO: this should not be in UI and also deleting looks strange as it remove all old
+        aliases_to_delete.each_pair do |a, v|
+          Yast::NetworkInterfaces.DeleteAlias(Yast::NetworkInterfaces.Name, a) if !v.nil?
```
imobachgs: I know it is old code, but what about using just `if v`?

src/lib/y2network/widgets/additional_addresses.rb
jreidinger: It sadly won't work (I think). I think it need to be before each translation to be properly extracted to potfile.

PR #815, opened 2019-05-28
Connection configuration reader

imobachgs:
This PR adds the concept of _connection configuration_ as discussed within our squad. However, it is a WIP, and it is far from being complete.

* [ ] Add missing connection types
* [ ] Use the interface name (or even better some a matcher class) to reference the `Interface` from the `ConnectionConfig` object.
* [ ] Read interfaces/configuration without relying on old modules (we will not implement it by now).


src/lib/y2network/config_reader/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,154 @@
...
+    # * Physical interfaces are read from the hardware.
+    # * Virtual interfaces + Connections are read from sysconfig.
+    #
+    # @see Y2Network::Interface
+    # @see Y2Network::Connection::Connection
+    class SysconfigInterfaces
+      # Returns the interfaces and connections configuration
+      #
+      # @todo Should we use different readers?
+      # @todo Are virtual interfaces coming from connections only?
```
jreidinger: well, if it is existing one, then it also exists on system e.g. hwinfo can return it, but not much info about it.

src/lib/y2network/config_reader/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,154 @@
...
+        @virtual_interfaces ||= connections.map(&:interface).compact
+      end
+
+      # @return [Array<ConnectionConfig>]
+      def connections
+        configured_devices.map do |name|
+          interface = physical_interfaces.find { |i| i.name == name }
+          conn_type = interface ? interface.type : find_type_for(name)
+          # TODO: it may happen that the interface does not exist yet (hotplug, usb, or whatever)
+          # How should we handle those cases?
```
jreidinger: well, in such cases it should be fake and exists only in configuration.

src/lib/y2network/config_reader/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      # @return [Regex] expression to filter out invalid ifcfg-* files
+      IGNORE_IFCFG_REGEX = /(\.bak|\.orig|\.rpmnew|\.rpmorig|-range|~|\.old|\.scpmbackup)$/
+
+      # List of devices which has a configuration file (ifcfg-*)
+      #
+      # @return [Array<String>] List of configured devices
+      def configured_devices
+        files = Yast::SCR.Dir(Yast::Path.new(".network.section"))
+        files.reject { |f| IGNORE_IFCFG_REGEX =~ f || f == "lo" }
```
jreidinger: interesting, I do not know that regex also has `=~` operand. I always use `string =~ /regexp/`

src/lib/y2network/config_reader/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      # Returns a connection configuration
+      #
+      # @param [Symbol] Connection type (e.g., :eth)
+      # @param [String] Interface name
+      # @return [ConnectionConfig::Base] an object representing the configuration
+      def connection_config_for(type, name)
+        # find type (from associated interface, name or attributes)
+        # create connection from detected type
+        send("#{type}_connection_config_for", name)
```
jreidinger: this looks a bit strange for me. Why not having separate class ( even nested ) for each type and matching based on class name?

jreidinger:
in general approach looks reasonable.

src/lib/y2network/config_reader/sysconfig_interfaces.rb
imobachgs: I've added som PoC having separate classes, but I am not happy at all with the result. I've decided to not use nested classes, but put them into a module. See commit 12eabd5.

src/lib/y2network/config_reader/connection_config/sysconfig_handlers/wlan.rb
```diff
@@ -24,7 +24,7 @@ module ConfigReader
     module ConnectionConfig
       # This class is able to build a ConnectionConfig::Wireless object given a
       # SysconfigInterfaceFile object.
-      class SysconfigWireless
+      class Wlan
```
jreidinger: missing sysconfig handler namespace?

Otherwise this change looks reasonable for me.

jreidinger: btw I expect in future we will share some parts in mixin or common ancestor for things that is used for every kind of device.

src/lib/y2network/config_reader/sysconfig_interfaces.rb
imobachgs: I will handle this scenario in the next PR.

src/lib/y2network/config_reader/connection_config/sysconfig.rb
```diff
@@ -0,0 +1,68 @@
...
+      class Sysconfig
+        # Constructor
+        #
+        # @param interface [String] Interface name
+        # @return [Y2Network::SysconfigInterfaceFile]
+        def read(interface)
+          type = find_type_for(interface)
+          handler_class = find_handler_class(type)
+          if handler_class.nil?
+            log.info "Unknown connection type: '#{type}'"
```
jreidinger: why not exception?

src/lib/y2network/config_writer/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,47 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "yast"
+
+module Y2Network
+  module ConfigWriter
+    class SysconfigInterfaces
+      COMMON_KEYS = [
```
jreidinger: what is goal of this constant? No documentation and no usage.

src/lib/y2network/config_writer/sysconfig_interfaces.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    private
+
+      def write_interface(iface)
+        interface_to_hash(iface)
+      end
+
+      def interface_to_hash(iface)
+        {
+          "NAME" => iface.name
```
jreidinger: so it writes only name now, right?

src/lib/y2network/interface.rb
```diff
@@ -21,12 +21,17 @@ module Y2Network
   class Interface
     # @return [String] Device name (eth0, wlan0, etc.)
     attr_accessor :name
+    # @return [String]
+    attr_accessor :ethtool_options
```
jreidinger: this is a kinda strange as only eth can have this options.

jreidinger: btw there is also related to interface kernel module and its options. ( that is used for given interface )

src/lib/y2network/config_writer/sysconfig_interfaces.rb
imobachgs: You are right, that class is basically a placeholder. I am removing it by now.

src/lib/y2network/interface.rb
```diff
@@ -19,11 +19,11 @@
 module Y2Network
   # Network interface.
   class Interface
-    # @return [String] Device name (eth0, wlan0, etc.)
+    # @return [String] Device name ('eth0', 'wlan0', etc.)
     attr_accessor :name
-    # @return [String]
-    attr_accessor :ethtool_options
+    # @return [String] Interface descripion
```
jreidinger: type descrip*t*ion

test/y2network/config_reader/connection_config/sysconfig_handlers/wlan_test.rb
```diff
@@ -0,0 +1,170 @@
...
+        eap_mode:     "PEAP",
+        eap_auth:     "mschapv2",
+        wpa_password: "example_passwd"
+      )
+    end
+  end
+
+  context "WPA-PSK network configuration" do
+    let(:wireless_attributes) do
+      COMMON_PARAMETERS.merge(
```
mvidner: @imobachgs Sorbet has told me that this constant is not defined. What kind of magic is this??

mvidner: OK, this `let` block is never executed, that's why the bug does not trigger ;-)

PR #816, opened 2019-05-29
Extracted and adapted interface builder

mchf:
A part of https://github.com/yast/yast-network/pull/796


src/lib/y2network/dialogs/add_interface.rb
```diff
@@ -23,18 +24,13 @@ def run
         ret = super
         log.info "AddInterface result #{ret}"
         ret = :back if ret == :abort
-        # TODO: replace with builder initialization
-        if ret == :back
```
jreidinger: for now you cannot remove it unless you also remove that LanItems.AddItem otherwise it crash. Just try it yourself - click on AddDevice and click on Cancel.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+      @config[option] = value
+    end
+
+    # Returns currently stored option value
+    # FIXME: obsolete by []
+    def option(option)
+      @config.fetch(option, nil)
+    end
+
+    def []=(key, value)
```
jreidinger: BTW this should be also in long term obsolete as it is very unreliable interface just acting like hash. But good compatible starting point.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+      # #104494 - always write IPADDR+NETMASK, even empty
+      # #50955 omit computable fields
+      @config["BROADCAST"] = ""
+      @config["NETWORK"] = ""
+
+      if !@config["NETMASK"] || @config["NETMASK"].empty?
+        @config["NETMASK"] = Yast::NetHwDetection.result["NETMASK"] || "255.255.255.0"
+      end
+
+      @config["STARTMODE"] = new_device_startmode if !@config["STARTMODE"] || @config["STARTMODE"].empty?
```
jreidinger: I do not get this. Why is if needed? It should be always nil, not?

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+
+    # returns default startmode for a new device
+    #
+    # startmode is returned according product, Arch and current device type
+    def new_device_startmode
```
jreidinger: I do not get why it is rewritten here from InterfaceDefaults.

src/lib/y2network/interface_defaults.rb
```diff
@@ -0,0 +1,89 @@
...
+        "auto"
+      else
+        hotplug_usable? ? "hotplug" : "auto"
+      end
+
+      startmode
+    end
+
+    # returns a map with device options for newly created item
+    def new_item_default_options
```
jreidinger: Hmmm, I am not sure this is good design to have defaults in mixin. I prefer to have separated object that produce new builder as output and can do more reasonable proposal ( like propose device name, be device type specific, check other stuffs and so on). But I do not want to delay this PR more then needed.

src/lib/y2network/interface_config_builder.rb
mchf: yes, but for now it made integration of your widgets much easier.

src/lib/y2network/interface_config_builder.rb
mchf: in fact it should be never nil bcs STARTMODE is preloaded from yaml with some default. I'd like to deal with these preloaded defaulds somehow - if you check the yaml you'll see that a lot of them is empty. But I don't want to do it now bcs it is not that straight forward (in some cases missing option behaves differently then empty option) so I kept the code as it was in the old code and did only some polishing.

src/lib/y2network/interface_config_builder.rb
mchf: see above it needs some polishing. Most probably it used to be initialized to a default originally and later someone designed this product based proposal.

src/lib/y2network/interface_defaults.rb
mchf: why not. But currently I was focused on isolating what should be kept and what can be deleted ;-) Especially this initialization to some defaults / proposals is really messy and done at least on three major places so moving the initialization at one place is not that straightforward. I think we will touch this area many times when network-ng will get sharper edges ;-)

src/lib/y2network/dialogs/add_interface.rb
mchf: ah I still live in the other PR where Items are not used.

Reverted

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+# find current contact information at www.suse.com.
+require "yast"
+require "y2network/interface_defaults"
+
+module Y2Network
+  # Stores what's needed when creating a new configuration for an interface
+  class InterfaceConfigBuilder
+    # @return [String] Device name (eth0, wlan0, etc.)
+    attr_accessor :name
+    # @return [String] type which is intended to be build
```
mvidner: Can you be more specific? A "type" which is a String is awfully vague.

Is network-ng using some kind of enums, where you have a clearly defined set of allowed values?

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+      # all keys / values has to be strings, agent won't write it otherwise
+      config.map { |k, v| [k, v.to_s] }.to_h
+    end
+
+    # Updates itself according to the given sysconfig configuration
+    #
+    # @param devmap [Hash<String, String>] a key, value map where key is sysconfig option
+    #                                      and corresponding value is the option value
+    def load_sysconfig(devmap)
+      @config = !devmap.nil? ? @config.merge(devmap) : @config
```
mvidner: This is an implementation detail but SCNR
```suggestion
      @config.merge!(devmap || {})
```

Also, make up your mind whether devmap is allowed to be `nil` and update the `@param` or the code.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+
+    def load_s390_config(devmap)
+      @s390_config = !devmap.nil? ? @s390_config.merge(devmap) : @s390_config
```
mvidner: Same problem as in `load_sysconfig`

src/lib/y2network/interface_config_builder.rb
mchf: not yet.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,181 @@
...
+    end
+
+    # Stores one option value tuple
+    # FIXME: obsolete by []=
+    def set(option: nil, value: nil)
+      @config[option] = value
+    end
+
+    # Returns currently stored option value
+    # FIXME: obsolete by []
```
mvidner: Should we add a comment here about why the new code keeps using this method even if it's obsolete?

src/lib/y2network/hwinfo.rb
```diff
@@ -17,14 +17,59 @@
...
+    alias_method :name, :dev_name
+
+    def exists?
+      !@hwinfo.nil?
+    end
+
+    # Device type description
+    # FIXME: collision with alias for dev_name
+    def description
+      @hwinfo ? @hwinfo.fetch("name", "") : ""
```
mvidner:
```suggestion
      name
```
or alias it (but I can never remember the correct syntax for that)

mchf: i cannot ... I've already aliased dev_name to name

src/lib/y2network/interface.rb
```diff
@@ -18,6 +18,13 @@
 # find current contact information at www.suse.com.
 module Y2Network
   # Network interface.
+  #
+  # name and hardware attributes should be imutable for the whole life of the object.
```
mvidner: If it's immutable, then change to `attr_reader :name` or explain what prevents it.

What hardware attributes are we talking about? There are none here.

BTW, i*mm*utable

src/lib/y2network/interface_config_builder.rb
mchf: bcs I was lazy to write a regex to replace it ;-) - fixed

src/lib/y2network/interface.rb
mchf: it has leaked from https://github.com/yast/yast-network/pull/796, I'll remove the comment

src/lib/y2network/hwinfo.rb
mvidner: `alias` and `alias_method` **add** a name, they don't move/destroy the old one. `alias_method :description, :dev_name` will just work. But don't trust me, just test it.

mchf: I don't get it so I'll describe what i did and ask you if you can explain your requirement again

in hwinfo we have:
"dev_name" -> contains device name (eth0)
"name" -> textual device description like "Inter Ethernet Controller 82546"

I created a "name" alias for "dev_name" (```alias_method :name, :dev_name```)
and now what do you propose? Create an alias "description" for hwinfo's name item (alias_method :description, :name) ? I really don't know if it would behave correctly ...

mvidner: > in hwinfo we have:
> "dev_name" -> contains device name (eth0)
> "name" -> textual device description like "Inter Ethernet Controller 82546"

OK

> I created a "name" alias for "dev_name" (alias_method :name, :dev_name)

So now `name` no longer returns "Inter Ethernet Controller 82546", it returns `eth0`.

Anyway, your current implementation of `description` can be replaced by `alias_method :description, :name`.

mchf: yes but is it correct? What if user uses ```hwinfo.name```? What is expected result in such case?

mvidner: First, let's forget about aliases, that's an implementation detail that has carried me away.

The important point IMHO is: For each element in `@hwinfo`, is its method name in `Hwinfo` the same? If not, there should be a comment with an explicit explanation. `Hwinfo` can have more than one name for the same thing, that's OK.

mvidner: | Hwinfo | `@hwinfo` |
|--|--|
| Hwinfo#dev_name | `@hwinfo["dev_name"]` |
| Hwinfo#name | `@hwinfo["dev_name"]` |
| Hwinfo#description | `@hwinfo["name"]` |

src/lib/y2network/hwinfo.rb
```diff
@@ -17,14 +17,69 @@
...
+      { name: "requires", default: [] },
+      { name: "hotplug", default: false },
+      { name: "wl_auth_modes", default: "" },
+      { name: "wl_enc_modes", default: nil },
+      { name: "wl_channels", default: nil },
+      { name: "wl_bitrates", default: nil }
+    ].each do |hwinfo_item|
+      method_name = hwinfo_item[:name].downcase
+
+      # @!method #{method_name} read value of #{hwinfo_item[:name]} from hwinfo or
```
mvidner: Unfortunately, YARD does not execute the code, and the reader will be just confused

Compare, `to_integer` is rendered as
https://www.rubydoc.info/github/yast/yast-ruby-bindings/Yast/Convert#to_integer-class_method
and documented as
https://github.com/yast/yast-ruby-bindings/blob/4288ac455902ac43aeec0175523657a6a538e251/src/ruby/yast/convert.rb#L19

src/lib/y2network/hwinfo.rb
```diff
@@ -17,14 +17,69 @@
...
+    def initialize(name:)
+      # FIXME: store only what's needed.
+      @hwinfo = load_hwinfo(name)
+    end
+
+    # Shortcuts for accessing hwinfo items. Each hwinfo item has own method for reading
+    # its value. There are two exceptions however. First exception is hwinfo["name"] item
+    # which carries device model human friendly description. This item is accessible via
+    # {Hwinfo::description}. Second exception is hwinfo["dev_name"] item which can be read
+    # through {Hwinfo::dev_name} or its alias {Hwinfo::name}.
```
mvidner: For comparison, in the Salt Widgets feature, I have explicitly removed a `name` attribute, because it was unclear whether it referred to a machine readable form (`id`) or a human readable form (`label`).

I will leave this to @jreidinger to judge, because I don't see the big picture here.

PR #817, opened 2019-05-30
convert old CWM to new CWM for edit interface

jreidinger:


src/lib/y2network/widgets/address_tab.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      # For s390 hwinfo gives us a multitude of types but some are handled
+      # the same, mostly acording to the driver which is used. So let's group
+      # them under the name Driver Type.
+      # TODO: move outside
+      # @param [String] type a type, as in Lan::type
+      # @return driver type, like formerly type2 for s390
+      def driver_type(type)
+        drvtype = type
+        # handle HSI like qeth, S#40692#c15
```
mvidner:
```suggestion
        # handle HSI like qeth, bsc#55692 #c15
```

src/lib/y2network/dialogs/edit_interface.rb
```diff
@@ -0,0 +1,58 @@
...
+require "y2network/widgets/bridge_slaves_tab.rb"
+require "y2network/widgets/general_tab.rb"
+require "y2network/widgets/hardware_tab.rb"
+
+Yast.import "LanItems"
+Yast.import "Label"
+
+module Y2Network
+  module Dialogs
+    # Dialog to create or edit route.
```
teclator: It is not a route but an interface

src/lib/y2network/dialogs/edit_interface.rb
```diff
@@ -0,0 +1,58 @@
...
+    class EditInterface < CWM::Dialog
```
teclator: In general, would be nice to have new classes proper documented. In this case, at least the `settings` param.

src/lib/y2network/widgets/address_tab.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      def label
+        _("&Address")
+      end
+
+      def contents
+        type = @settings.type
+
+        drvtype = driver_type(type)
+        is_ptp = drvtype == "ctc" || drvtype == "iucv"
```
teclator: Add a #FIXME or comment to check if we can already remove this interfaces.

See:
https://www.suse.com/documentation/sles-15/book_sle_deployment/data/sec_zseries_requirements.html
https://www.suse.com/documentation/sles-15/book_sle_admin/data/sec_network_yast.html#sec_network_yast_s390_ctc

src/lib/y2network/widgets/general_tab.rb
```diff
@@ -0,0 +1,65 @@
...
+        ifplugd_widget = IfplugdPriority.new(@settings)
+        MarginBox(
+          1,
+          0,
+          VBox(
+            MarginBox(
+              1,
+              0,
+              VBox(
+                # FIXME: udev rules for anything without hwinfo is wrong
```
teclator: # Partially true. 

A udev rule for configuring interfaces could exist without hwinfo. That is, when it is a hotplug interface or think in a damaged one that needs to be replaced. In that last case, if the rule was based on mac address instead of busid, the new card will appear as a new interface ("and this case propably is not well handled anyway").

So, if a rule is present without hardware information, maybe it should be shown but in disabled mode.

teclator:
In general looks good to me, just added few notes, nothing so important to block a merge into network-ng.

PR #818, opened 2019-06-03
Backported patch for bnc#1123102

mchf:
to fix l3 bnc#1134784


teclator:
JFR: The fix is backported from https://github.com/yast/yast-network/pull/723

PR #822, opened 2019-06-03
Merged SLE-15-GA

mchf:


package/yast2-network.changes
```diff
@@ -1,3 +1,14 @@
+-------------------------------------------------------------------
+Mon May 20 14:04:47 UTC 2019 - Michal Filka <mfilka@suse.com>
+
+- bnc#1123102, bnc#1134784, bnc#1136103
+  - do not crash with internal error when no hostname is set when
```
teclator: Isn't this section already present in the changelog? At least, there is no code added related to what described

teclator: https://github.com/yast/yast-network/blob/12ff6379469ebab0c797848c54f0d88a9d003321/package/yast2-network.changes#L145

mchf: hm, the fix is already in this branch. I kept the changelog entry only bcs of all those bncs ... but you're right that it is strange. i'll remove it.

PR #823, opened 2019-06-04
Builder improvements

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -43,6 +50,27 @@ def [](key)
...
+      !(value != /^[[:alnum:]._:-]{1,15}\z/)
+    end
+
+    def already_exists?(name)
+      Yast::NetworkInterfaces.List("").include?(name)
+    end
+
+    def name_valid_characters
+      Yast::NetworkInterfaces.ValidCharsIfcfg
+    end
```
jreidinger: seeing all those method, maybe having object for name makes sense?

mchf: I think that this method is not correct. It is used only at one place for validating ifcfg name but it can validate even characters which are not allowed in ifcfg (respectively network device) names - see https://github.com/yast/yast-yast2/blob/master/library/types/src/modules/String.rb#L591 and https://github.com/yast/yast-yast2/blob/master/library/types/src/modules/String.rb#L42
Correct validator is a few lines above - valid_name?

jreidinger: so is it bug in ValidCharsIfcfg? I just need to get array with chars that is allowed.

mchf: i think so

src/lib/y2network/interface_config_builder.rb
```diff
@@ -83,6 +83,19 @@ def driver=(value)
       Yast::LanItems.setDriver(value)
     end
 
+    def driver_options
+      target_driver = driver
+      if target_driver.empty?
+        target_driver = Yast::Ops.get_string(Yast::LanItems.getCurrentItem, ["hwinfo", "module"], "")
+      end
```
mchf: we have Hwinfo class so we can relax dependency on LanItems here

```
Hwinfo.new(name).module
````
(module method is currently missing in the Hwinfo class but it is very easy to add)

PR #824, opened 2019-06-05
Fixed ethtool widget to work with the new builder

mchf:


package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.2.2
+Version:        4.2.3
```
jreidinger: do we need bump as it is only in network-ng?

PR #825, opened 2019-06-05
Do not create duplicate udev attributes

teclator:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1137324

When editing an interface name. 

If no previous udev rule exist and the udev rule to be written is based on the interface :mac, then it generates a duplicate attribute:

edit_nic_name

Something like:

```ruby
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="9c:eb:e8:1a:d2:2f", ATTR{dev_id}=="0x0", ATTR{type}=="1", ATTR{address}=="9c:eb:e8:1a:d2:2f", NAME="eth1"
```

That also happens with an existent rule and pressing 'OK' without modifying the key attribute.

In combination with this another [bug](https://bugzilla.suse.com/show_bug.cgi?id=1136929) the situation is even worworst because there could be two different mac addresses, the current and the permanent one.

## Solution

When editing an interface name:

1.- The rule only will be updated if the matching attribute or value from the original differs.
2.- The update_item_udev_rule! method will check if a rule already exist and if not will initialize it. As the default rule is based on :mac, we will return after initialization in case the update rule is based on :mac.

## Testing

- *Added a new unit test for the new current udev rule method*
- *Tested manually*

### DIFF
--- a/src/lib/network/edit_nic_name.rb
+++ b/src/lib/network/edit_nic_name.rb
@@ -15,9 +15,9 @@ module Yast
     include I18n
 
     # @return [String] current udev name before modifying it
-    attr_accessor :old_name
+    attr_reader :old_name
     # @return [String] current udev match criteria
-    attr_accessor :old_key
+    attr_reader :old_key
 
     # udev rule attribute for MAC address
     MAC_UDEV_ATTR   = "ATTR{address}".freeze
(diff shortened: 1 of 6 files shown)


src/lib/network/edit_nic_name.rb
```diff
@@ -14,6 +14,11 @@ class EditNicName
     include UIShortcuts
     include I18n
 
+    # @return [String] current udev name before modifying it
+    attr_accessor :old_name
+    # @return [String] current udev match criteria
+    attr_accessor :old_key
```
jreidinger: accessor or reader?

teclator: Should be a reader. Updated

PR #826, opened 2019-06-06
adapt firewall zone widget

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -19,6 +19,8 @@
 require "yast"
 
 require "y2network/hwinfo"
+require "y2firewall/firewalld"
+require "y2firewall/firewalld/interface"
```
teclator: Maybe some methods defined in https://github.com/yast/yast-yast2/blob/master/library/network/src/lib/y2firewall/helpers/interfaces.rb could be useful.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -53,10 +55,20 @@ def [](key)
...
+      if !driver.empty?
+        Yast::LanItems.setDriver(driver)
+        Yast::LanItems.driver_options[driver] = driver_options
+      end
+
+      # create new instance as name can change
+      firewall_interface = Y2Firewall::Firewalld::Interface.new(name)
+      if Y2Firewall::Firewalld.instance.installed?
+        Yast::LanItems.firewall_zone = firewall_zone
+        # TODO: should change only if different, but maybe firewall_interface responsibility?
```
teclator: In fact it already takes that responsibility.

https://github.com/yast/yast-yast2/blob/master/library/network/src/lib/y2firewall/firewalld/relations.rb#L194
https://github.com/yast/yast-yast2/blob/master/library/network/src/lib/y2firewall/firewalld/interface.rb#L105

So, maybe not needed at all.

src/lib/y2network/interface_config_builder.rb
jreidinger: I see that, but I worry that this class is already quite big and if I include mixin, it will be even better.

src/lib/y2network/interface_config_builder.rb
jreidinger: not sure, as that zone=  at first delete it and then add it again. Not sure if it is problem or not.

src/lib/y2network/interface_config_builder.rb
jreidinger: I use from it at least that installed.

teclator:
LGTM

PR #827, opened 2019-06-07
Builder lifetime

jreidinger:


src/include/network/lan/address.rb
```diff
@@ -97,6 +97,7 @@ def AddressDialog(builder:)
       # proceed with WLAN settings if appropriate, #42420
       ret = :wire if ret == :next && builder.type == "wlan"
 
+      Builtins.y2milestone("AddressDialog res: %1", ret)
```
teclator: `log.info is already used in the file, so will continue using it`

jreidinger: ah, ok. I add it mainly for debugging, but I think it makes sense to keep it.

src/lib/y2network/dialogs/add_interface.rb
```diff
@@ -20,24 +20,48 @@ def contents
...
+        Yast::LanItems.AddNew
+        # FIXME: can be partly deleted and partly moved
+        Yast::Lan.Add
+
+        # FIXME: This is for backward compatibility only
+        # dhclient needs to set just one dhcp enabled interface to
+        # DHCLIENT_SET_DEFAULT_ROUTE=yes. Otherwise interface is selected more
+        # or less randomly (bnc#868187). However, UI is not ready for such change yet.
+        # As it could easily happen that all interfaces are set to "no" (and
+        # default route is unrecheable in such case) this explicite setup was
```
teclator:
```suggestion
        # default route is unreachable in such case) this explicit setup was
```

teclator:
In general looks good to me.

PR #830, opened 2019-06-10
add ipoib to builder and own types

jreidinger:


test/lan_test.rb
```diff
@@ -357,8 +357,8 @@ def expect_modification_succeedes(modname, method)
       end
 
       it "does not configure the interface if it is not connected" do
-        allow(Yast::Lan).to receive(:connected_and_bridgeable?).with(anything).and_return(false)
-        expect(Yast::Lan).not_to receive(:ProposeItem)
+        allow(Yast::Lan).to receive(:connected_and_bridgeable?).with(anything, anything, anything).and_return(false)
```
mchf: ```any_args``` doesn't fit?

jreidinger: anything match single param, not list of params. In fact maybe better to remove whole "with".

teclator:
For me it looks good

test/lan_test.rb
mchf: well I thought that having any_args is shorter than having anything 3 times. But removing with completely makes the same service in this case.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -29,6 +29,18 @@ module Y2Network
   # Collects data from the UI until we have enough of it to create a {Y2Network::Interface}.
   # {Yast::LanItemsClass#Commit Yast::LanItems.Commit(builder)} uses it.
   class InterfaceConfigBuilder
+    include Yast::Logger
+
+    def self.for(type)
```
imobachgs: Just add some documentation here, please.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -29,6 +29,18 @@ module Y2Network
...
+      require "y2network/interface_config_builders/#{type}"
+      InterfaceConfigBuilders.const_get(type.to_s.capitalize).new
+    rescue LoadError => e
+      log.info "Specialed builder for #{type} not found. Fallbacking to default. #{e.inspect}"
+      res = new
```
imobachgs: Why not to pass the `type` in the constructor?

PR #835, opened 2019-06-13
Merge SLE-15-SP1 into master

teclator:
It merges #832 & #829 & #825


src/modules/Lan.rb
```diff
@@ -909,7 +909,6 @@ def ProposeVirtualized
         bridge_name = LanItems.new_type_device("br")
         next unless connected_and_bridgeable?(bridge_name, current, config)
         LanItems.current = current
-        LanItems.add_current_device_to_routing
```
teclator: Using the current name does not help here, so adapted to permit passing the name explicitly.

PR #836, opened 2019-06-14
Fixed default gw check which still used the ould Routing module.

teclator:


src/include/network/lan/address.rb
```diff
@@ -1366,17 +1366,18 @@ def AddressDialog
...
-                "A static default route is defined.\n" \
-                "It is suggested to remove the static default route definition \n" \
-                "if one can be obtained also via DHCP.\n" \
-                "Do you want to remove the static default route?"
-              )
-            )
-
             yast_config = Y2Network::Config.find(:yast)
-            yast_config.routing.remove_default_routes if yast_config && remove_gw
+            if yast_config && yast_config.routing && yast_config.routing.default_route
```
jreidinger: btw in newer ruby which is in TW and also in SLE15 you can write it as `if yast_config&.routing&.default_route`

PR #838, opened 2019-06-14
bridges and bonding without lanitems

jreidinger:


src/lib/y2network/interface_config_builders/bridge.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      def already_configured?(devices)
+        configurations = Yast::NetworkInterfaces.FilterDevices("netcard")
+        netcard_types = (Yast::NetworkInterfaces.CardRegex["netcard"] || "").split("|")
+
+        confs = netcard_types.reduce([]) do |res, devtype|
+          res.concat((configurations[devtype] || {}).keys)
+        end
+
+        devices.each do |device|
```
mchf: ```NetworkInterfaces::List("netcard")``` and 8 lines above can be dropped

src/lib/y2network/interface_config_builders/bridge.rb
```diff
@@ -0,0 +1,40 @@
...
+          next if !confs.include?(device)
+
+          dev_type = Yast::NetworkInterfaces.GetType(device)
+          ifcfg_conf = configurations[dev_type][device]
+
+          next if ifcfg_conf["BOOTPROTO"] == "none"
```
mchf: ```next if [nil, "none"].include? NetworkInterfaces.devmap(device)``` and 5 lines above can be dropped

src/lib/y2network/interface_config_builders/bridge.rb
jreidinger: any reason why? I think it is needed to detect if device is already configured, not?

src/lib/y2network/interface_config_builders/bridge.rb
jreidinger: so this one line replace whole that part above?

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,24 @@
+require "yast"
+require "y2network/interface_config_builder"
+
+Yast.import "NetworkInterfaces"
+
+module Y2Network
+  module InterfaceConfigBuilders
+    class Br < InterfaceConfigBuilder
```
teclator: why not Bridge?

src/lib/y2network/widgets/bond_slave.rb
```diff
@@ -69,10 +68,8 @@ def help
...
       def init
-        items = slave_items_from(
-          Yast::LanItems.GetBondableInterfaces(Yast::LanItems.GetCurrentName),
-          @settings["SLAVES"]
-        )
+        # TODO: use def items, but problem now is that slave_items returns term and not array
+        items = slave_items_from(@settings.bondable_interfaces, @settings["SLAVES"])
 
         # reorder the items
         l1, l2 = items.partition { |t| @settings["SLAVES"].include? t[0][0] }
```
teclator: it seems `@settings["SLAVES"]` is not properly initialized, so faced a nil.include? exception

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,24 @@
...
+        super(type: "br")
+      end
+
+      def bridgable_interfaces
+        Yast::LanItems.GetBridgeableInterfaces(name)
+      end
+
+      def already_configured?(devices)
+        devices.any? do |device|
+          ![nil, "none"].include?(Yast::NetworkInterfaces.devmap(device))
```
mchf: this is wrong (my proposal was wrong) ... you need ```...devmap(device)[BOOTPROTO]``` bcs ```devmap``` returns the device's configuration (if any) and you need to check for BOOTPROTO option value.

src/lib/y2network/interface_config_builders/bridge.rb
teclator: In case we permit to add it to the bond when the BOOTPROTO is not defined, then we should at least adapt the BOOTPROTO when the interface is added to the bond.

teclator:
In general looks good, there is some fixes that will be needed, like the overview and configuration of enslaved interfaces. But looks good so far.

mchf:
No more comments from my side

PR #843, opened 2019-06-18
Read not present interfaces

imobachgs:
This PR adds the ability to read interfaces that, for some reason, are not present in the system (USB, hotplug, ...) but the configuration is still present. In such a case, a `FakeInterface` object is created.

### DIFF
--- a/src/lib/y2network/config_reader/connection_config/sysconfig.rb
+++ b/src/lib/y2network/config_reader/connection_config/sysconfig.rb
@@ -45,8 +45,9 @@ module Y2Network
         def find_handler_class(type)
           require "y2network/config_reader/connection_config/sysconfig_handlers/#{type}"
           SysconfigHandlers.const_get(type.to_s.capitalize)
-        rescue LoadError, NameError
-          log.info "Unknown connection type: '#{type}'"
+        rescue LoadError, NameError => e
+          log.info "Unknown connection type: '#{type}'. " \
+                   "Connection handler could not be loaded: #{e.message}"
           nil
         end
       end
--- a/src/lib/y2network/config_reader/connection_config/sysconfig.rb
+++ b/src/lib/y2network/config_reader/connection_config/sysconfig.rb
@@ -18,49 +18,32 @@
 # find current contact information at www.suse.com.
 
 require "y2network/sysconfig_interface_file"
-Yast.import "NetworkInterfaces"
 
 module Y2Network
   module ConfigReader
     module ConnectionConfig
       # Reads a connection configuration for a given interface
       class Sysconfig
+        include Yast::Logger
+
         # Constructor
         #
         # @param interface [String] Interface
         # @return [Y2Network::SysconfigInterfaceFile]
         def read(interface)
-          type = find_type_for(interface)
-          handler_class = find_handler_class(type)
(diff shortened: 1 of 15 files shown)


src/lib/y2network/config_reader/connection_config/sysconfig.rb
```diff
@@ -18,49 +18,35 @@
...
-          type = Yast::NetworkInterfaces.GetTypeFromSysfs(interface.name)
-          type.nil? ? :eth : type.to_sym
-        end
-
         def find_handler_class(type)
           require "y2network/config_reader/connection_config/sysconfig_handlers/#{type}"
           SysconfigHandlers.const_get(type.to_s.capitalize)
-        rescue LoadError
+        rescue LoadError, NameError
+          log.info "Unknown connection type: '#{type}'"
```
jreidinger: can you add at least as debug also exception message? it can help e.g. if required class contain some syntax error

src/lib/y2network/config_reader/sysconfig_interfaces.rb
```diff
@@ -39,51 +41,75 @@ module ConfigReader
...
-      # @todo Are virtual interfaces coming from connections only?
+      # @return [Hash<Symbol,Object>] Returns a hash containing
+      #   an interfaces collection (with the key `:interfaces`)
+      #   and an array of connection config objects.
       def config
-        [interfaces, connections]
+        return @config if @config
+        find_physical_interfaces
+        find_connections
+        { interfaces: @interfaces, connections: @connections }
```
jreidinger: hmmm, I do not like much this magic with find and find. Can we simply load it when first called and have methods like interfaces and connections? Here it is localized at one place, so not big deal for me. BTW you also probably miss to assign `@config` here

src/lib/y2network/interface_type.rb
```diff
@@ -0,0 +1,60 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+
+module Y2Network
+  # This class represents the interface types which are supported.
+  #
+  # Constants may be defined using the {define_type} method.
+  class InterfaceType
```
jreidinger: good, I already have TODO for having such ENUM. So what is missing here? for sure old type string that is used in sysconfig like "eth", "wlan" and so on, so we can easily get from old type this new type. Also ideally if you can add all types now, so we can use it in other parts easily. Idea is to start using this new objects as soon as possible, so it won't be so separated.

src/lib/y2network/config_reader/sysconfig_interfaces.rb
imobachgs: The trick is that we need to execute both methods in order to get a correct configuration. For sure, `find_physical_interfaces` should be called first, but that interfaces list could be augmented while reading the connection configurations (with fake interfaces).

So I just wanted to be sure that, if you call `#interfaces` (but not `config`), you get all the information read. Having a "#read" method (which is basically what this #config method does) solves the issue. Another approach might be to have separate methods for fake and physical interfaces (so getting fake interfaces implies reading connections but not the other way around), but I was simply not sure about this separation.

PR #844, opened 2019-06-18
Bonding / brindging backend in interfaces collection

mchf:


src/lib/y2network/interface.rb
```diff
@@ -25,13 +25,39 @@ class Interface
...
+    attr_reader :hardware
+
+    # Shortcuts for accessing interfaces' ifcfg options
+    ["STARTMODE", "BOOTPROTO"].each do |ifcfg_option|
+      method_name = ifcfg_option.downcase
+
+      define_method method_name do
+        # when switching to new backend we need as much guards as possible
+        if !configured || config.nil? || config.empty?
+          raise "Trying to read configuration of an unconfigured interface"
```
jreidinger: please specify more details in exception at least with "name" of interface.

src/lib/y2network/interface.rb
```diff
@@ -25,13 +25,39 @@ class Interface
...
-      @name = name
       @description = ""
+      # @hardware and @name should not change during life of the object
+      @hardware = Hwinfo.new(name: name)
+
+      if !(name.nil? || name.empty?)
+        @name = name
+      elsif @hardware.nil?
+        # the interface has to be either configured (ifcfg) or known to hwinfo
+        raise "Attempting to create representation of nonexistent interface"
```
jreidinger: please add to documentation of method that name can be empty or nil and what happen in such case.

src/lib/y2network/interface.rb
```diff
@@ -46,5 +72,23 @@ def ==(other)
     # eql? (hash key equality) should alias ==, see also
     # https://ruby-doc.org/core-2.3.3/Object.html#method-i-eql-3F
     alias_method :eql?, :==
+
+    def type
```
jreidinger: maybe sync with @imobachgs about using smarter object as type.

src/lib/y2network/interface.rb
```diff
@@ -46,5 +72,23 @@ def ==(other)
...
+    end
+
+    def config
+      system_config(name)
+    end
+
+  private
+
+    def system_config(name)
+      Yast::NetworkInterfaces.devmap(name)
```
jreidinger: so basically hash, so we do not plan yet to integrate new config object, right?

src/lib/y2network/interfaces_collection.rb
```diff
@@ -87,5 +104,129 @@ def ==(other)
     end
 
     alias_method :eql?, :==
+
+  private
+
+    # FIXME: this is only helper when coexisting with old LanItems module
+    # can be used in new API of network-ng for read-only methods. It converts
+    # old LanItems::Items into new Interface objects
+    def lan_items
```
jreidinger: name is quite misleading. It is not! lan_items, it is `interfaces`. And if you read it using new object or LanItems should not matter.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -87,5 +104,129 @@ def ==(other)
...
+    # Checks whether an interface can be enslaved in particular bond interface
+    #
+    # @param bond_iface [Interface]
+    # @param iface [Interface] an interface to be validated as bond_iface slave
+    # TODO: Check for valid configurations. E.g. bond device over vlan
+    # is nonsense and is not supported by netconfig.
+    # Also devices enslaved in a bridge should be excluded too.
+    def bondable?(bond_iface, iface)
+      Yast.import "Arch"
+      Yast.include self, "network/lan/s390.rb"
```
jreidinger: this is not nice, as it basically modify this collection depending on method. So nasty side effect. If you need reuse existing include, I prefer to separate it to own object, so no accidental collision of instance variables - https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/widgets/slave_items.rb#L85

src/lib/y2network/interfaces_collection.rb
```diff
@@ -57,6 +58,22 @@ def initialize(interfaces = [])
...
+      lan_items.select { |i| bridgeable?(bridge_iface, i) }
+    end
+
+    # @param bond_iface [Interface] an interface of bond type
+    #
+    # @return [Array<Interface>] list of interfaces usable in bridge_iface
+    def select_bondable(bond_iface)
+      raise "Bond interface expected" unless bond_iface.type == "bond"
+      lan_items.select { |i| bondable?(bond_iface, i) }
+    end
```
jreidinger: For me this is very strange location. Why it is not method on Interface, VirtualInterface or ideally BondInterface class? Having it in collection where you need to pass name of element of collection is strange for me.

jreidinger: BTW with this change object start looking for me quite big and also I start getting impression it can be soon another LanItems

src/lib/y2network/interface.rb
mchf: basically what we use now. If I start playing with config objects then we get super big PR2 which is not what I want ;-)

src/lib/network/lan_items_summary.rb
```diff
@@ -153,28 +154,20 @@ def static_summary
     #
     # @return [String] bridge configured interfaces summary
     def bridge_summary
-      _("Bridges: %s") % bridges.map { |n| "#{n} (#{LanItems.bridge_slaves(n).sort.join(", ")})" }
+      config = Y2Network::Config.find(:yast)
+      _("Bridges: %s") % config.interfaces.by_type("br").map do |n|
+        "#{n} (#{config.interfaces.bridge_slaves(n.name).sort.join(", ")})"
+      end
```
jreidinger: are you sure it works when you have more bridges?

```
irb(main):003:0> "a %s" % [1, 2].map { |i| i }
=> "a 1"
```

And even if it works, it is a bit too much magic for me. So please use join on result of map and ideally prepare string in advance.

```
so something likes
```
bridges = config.interfaces.by_type("br").map do |n|
  "#{n} (#{config.interfaces.bridge_slaves(n.name).sort.join(", ")})"
end
_("Bridges: %s") % bridges.join(", ")
```

File: src/lib/network/lan_items_summary.rb [hunk 9]
```diff
@@ -153,28 +154,20 @@ def static_summary
...
-      LanItems.getNetworkInterfaces("bond").sort
-    end
-
-    # Convenience method that obtains the list of bridge configured interfaces
-    #
-    # @return [Array<String>] bridge configured interface names
-    def bridges
-      LanItems.getNetworkInterfaces("br").sort
+      config = Y2Network::Config.find(:yast)
+      _("Bonds: %s") % config.interfaces.by_type("bond").map do |n|
```
jreidinger: same here

src/lib/y2network/interfaces_collection.rb
```diff
@@ -105,6 +113,26 @@ def ==(other)
 
     alias_method :eql?, :==
 
+    # Returns the interfaces that are enslaved in the given bridge
+    #
+    # @param master [String] bridge name
+    # @return [Array<String>] a list of interface names
+    # TODO: move to class of interface type br
```
jreidinger: :+1:

src/lib/y2network/widgets/bridge_ports.rb
```diff
@@ -31,7 +31,7 @@ def help
       def init
         br_ports = @settings["BRIDGE_PORTS"]
         items = slave_items_from(
-          Yast::LanItems.GetBondableInterfaces(Yast::LanItems.GetCurrentName),
+          Yast::LanItems.GetBridgeableInterfaces(Yast::LanItems.GetCurrentName),
```
jreidinger: already fixed in master, I also notice it :)

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,71 @@
...
+    class Br < InterfaceConfigBuilder
+
+      include Yast::Logger
+
+      def initialize
+        super(type: "br")
+      end
+
+      # @return [Array<Interface>] list of interfaces usable in bridge
+      def select_bridgeable
```
jreidinger: the name is strange. You do not select anything. please use simple `bridgable` or `bridgable_interfaces`

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,71 @@
...
+  module InterfaceConfigBuilders
+    class Br < InterfaceConfigBuilder
+
+      include Yast::Logger
+
+      def initialize
+        super(type: "br")
+      end
+
+      # @return [Array<Interface>] list of interfaces usable in bridge
```
jreidinger: why array and not InterfaceCollection?

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,71 @@
...
+        return false if iface.name == @name
+        return true if !iface.configured
+
+        if interfaces.bond_index[iface.name]
+          log.debug("Excluding (#{iface.name}) - is bonded")
+          return false
+        end
+
+        # the iface is already in another bridge
+        if interfaces.bridge_index[iface.name] && interfaces.bridge_index[iface.name] != @name
```
jreidinger: btw those methods just show how strange is to know in collection attribute of interface. Ideally api should look like
```
if iface.bridge && iface.bridge.name != @name
```

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        # exclude interfaces of type unusable for bridge
+        case iface.type
+        when "br"
+          log.debug("Excluding (#{iface.name}) - is bridge")
+          return false
+        when "tun", "usb", "wlan"
+          log.debug("Excluding (#{iface.name}) - is #{iface.type}")
+          return false
+        end
```
jreidinger: btw I already notice this. at first it can be single branch and also it looks like constant for me. So ideally code like:
```
NON_BRIDGEABLE_TYPES = ["br", "tun", "usb", "wlan"]
...
if NON_BRIDGEABLE_TYPES.include?(iface.type)
  log.debug("Excluding (#{iface.name}) - is #{iface.type}")
  return false
end
```

src/lib/y2network/interface_config_builders/br.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        case iface.startmode
+        when "nfsroot"
+          log.debug("Excluding (#{iface.name}) - is nfsroot")
+          return false
+
+        when "ifplugd"
+          log.debug("Excluding (#{iface.name}) - ifplugd")
+          return false
+        end
```
jreidinger: and same here, can be single branch and also constant.

src/include/network/lan/complex.rb
```diff
@@ -401,7 +401,7 @@ def handleOverview(_key, event)
           if LanItems.IsCurrentConfigured
             builder.load_sysconfig(LanItems.GetCurrentMap())
             builder.load_s390_config(LanItems.s390_ReadQethConfig(builder.name))
-            LanItems.SetItem
+            LanItems.SetItem(builder: builder)
```
jreidinger: also already fixed in network-ng ;)

src/modules/LanItems.rb
```diff
@@ -1983,8 +1985,14 @@ def SetItem
 
       @description = BuildDescription(@type, @device, devmap, @Hardware)
 
-      SetDeviceVars(devmap, @SysconfigDefaults)
-      SetS390Vars(s390_devmap, @s390_defaults)
+      devmap = SetDeviceVars(devmap, @SysconfigDefaults)
+      s390_devmap = SetS390Vars(s390_devmap, @s390_defaults)
+
+      # FIXME: should raise an exception
```
jreidinger: in fact not needed. Simply remove default param in API. So `def SetItem(builder:)` and it will raise it for you that param is missing.

src/lib/y2network/widgets/slave_items.rb
```diff
@@ -13,13 +13,15 @@ module SlaveItems
       # Builds content for slave configuration dialog (used e.g. when configuring
       # bond slaves) according the given list of item_ids (see LanItems::Items)
       #
-      # @param [Array<Fixnum>] item_ids           list of indexes into LanItems::Items
+      # @param [Array<String>] slaves             list of device names
       # @param [Array<String>] enslaved_ifaces    list of device names of already enslaved devices
-      def slave_items_from(item_ids, enslaved_ifaces)
-        raise ArgumentError, "no slave device defined" if item_ids.nil?
+      def slave_items_from(slaves, enslaved_ifaces)
+        raise ArgumentError, "no slave device defined" if slaves.nil?
```
jreidinger: please use more meaningful error message like `slaves cannot be nil` and also maybe we need to check if slaves is not empty as it also does not make sense.

src/lib/y2network/interface_config_builders/bond.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      # Checks whether an interface can be enslaved in particular bond interface
+      #
+      # @param iface [Interface] an interface to be validated as bond_iface slave
+      # TODO: Check for valid configurations. E.g. bond device over vlan
+      # is nonsense and is not supported by netconfig.
+      # Also devices enslaved in a bridge should be excluded too.
+      def bondable?(iface)
+        Yast.import "Arch"
+        Yast.include self, "network/lan/s390.rb"
```
jreidinger: as already written before including whole network include hell to new API is bad idea as it horrible mess object. Cleaner solution is to have private class which include it and call it from there.

src/lib/y2network/interface_config_builders/br.rb
mchf: after all it still is only conversion of old code into new environment. I expect a lot of space for improvements.

src/include/network/lan/complex.rb
mchf: even better - double check ;-)

src/lib/y2network/interface_config_builders/bond.rb
mchf: yes it is already on my todo list - just wanted to move it to final place before doing so ;-)

src/lib/y2network/interface_config_builders/br.rb
mchf: done for both br, bond

src/lib/y2network/interface_config_builders/br.rb
mchf: done

src/lib/y2network/interface_config_builders/br.rb
mchf: done

src/modules/LanItems.rb
mchf: done

src/lib/y2network/widgets/slave_items.rb
mchf: (hopefully) improved

src/lib/y2network/interfaces_collection.rb
mchf: moved to the builder

src/lib/y2network/interfaces_collection.rb
mchf: renamed

src/lib/y2network/interface.rb
mchf: done

src/lib/y2network/interface.rb
mchf: done

src/lib/y2network/interface.rb
```diff
@@ -28,14 +28,46 @@ class Interface
     attr_accessor :description
     # @return [Symbol] Interface type
     attr_accessor :type
+    attr_reader :configured
+    attr_reader :hardware
+
+    # Shortcuts for accessing interfaces' ifcfg options
+    ["STARTMODE", "BOOTPROTO"].each do |ifcfg_option|
```
imobachgs: The idea is that `Y2Network::Interface` remains backend-agnostic, and these lines are defeating that purpose. Moreover, this information should be part of `Y2Network::ConnectionConfig` classes. However, I guess that you need those classes to be finished before using them :) so, for the time being, please add a note here that this logic should be moved in the near future.

src/lib/y2network/interface.rb
```diff
@@ -28,14 +28,46 @@ class Interface
...
     # @param name [String] Interface name (e.g., "eth0")
     def initialize(name, type: InterfaceType::ETHERNET)
       @name = name
       @description = ""
       @type = type
+      # @hardware and @name should not change during life of the object
+      @hardware = Hwinfo.new(name: name)
+
+      if !(name.nil? || name.empty?)
+        @name = name
```
imobachgs: This assignment is already done a few lines above.

src/lib/y2network/interface.rb
```diff
@@ -50,5 +82,19 @@ def ==(other)
     # eql? (hash key equality) should alias ==, see also
     # https://ruby-doc.org/core-2.3.3/Object.html#method-i-eql-3F
     alias_method :eql?, :==
+
+    def config
```
imobachgs: Please, document this method.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -70,6 +71,14 @@ def by_name(name)
       end
     end
 
+    # Returns list of interfaces of given type
+    #
+    # @param type [String] device type name ("eth", "br", ...)
+    # @return [Array<Interfaces>] list of found interfaces
+    def by_type(type)
```
imobachgs: What about returning an `InterfacesCollection`? The point is that we could concatenate several calls (`by_type(...).by_whatever(...)`). If you need an array, you could just use a `to_a` method (or implement `first` or whatever you need). It is different from the `by_name` case, because in that case you only expect one interface with a given name.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -70,6 +71,14 @@ def by_name(name)
       end
     end
 
+    # Returns list of interfaces of given type
+    #
+    # @param type [String] device type name ("eth", "br", ...)
```
imobachgs: We should use an `InterfaceType` instance, although we could be able to handle strings/symbols too.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -87,5 +96,70 @@ def ==(other)
     end
 
     alias_method :eql?, :==
+
+    # Returns the interfaces that are enslaved in the given bridge
+    #
+    # @param master [String] bridge name
+    # @return [Array<String>] a list of interface names
```
imobachgs: `Array<String>` ? I would have expected an `Array<Interface>` (or an `InterfacesCollection`).

src/lib/y2network/interfaces_collection.rb
```diff
@@ -87,5 +96,70 @@ def ==(other)
...
+      bridge_index.select { |_k, v| v == master }.keys
+    end
+
+    # Creates list of devices enslaved in the bond device.
+    #
+    # @param bond_iface [String] a name of an interface of bond type
+    # @return list of names of interfaces enslaved in the bond_iface
+    # TODO: move to class of interface type bond
+    def bond_slaves(bond_iface)
+      bond_map = Yast::NetworkInterfaces::FilterDevices("netcard").fetch("bond", {}).fetch(bond_iface, {})
```
imobachgs: `Y2Network::InterfacesCollection` is a class which holds a set of interfaces and allows you to query the collection (filtering by type, name, whatever). And that is its only responsibility.

So IMHO, coupling it with `NetworkInterfaces` is wrong. If you need to get the slaves for a given interface, you should be able to do so by querying the interfaces included in the collection. Otherwise, I fear that the logic does not belong to this class.

src/lib/y2network/interfaces_collection.rb
```diff
@@ -87,5 +96,70 @@ def ==(other)
...
+          index[slave] = bond_master
+        end
+      end
+
+      log.debug("bond slaves index: #{index}")
+
+      index
+    end
+
+    def all
```
imobachgs: I know there is a `fixme` :) but I feel it is wrong. If you need so, you could have created a wrapper class adding those bits there (and keeping the `InterfacesCollection` as clean as possible).

src/modules/Lan.rb
```diff
@@ -1087,11 +1086,15 @@ def configure_as_bridge!(ifcfg, bridge_name)
...
 
     # Convenience method that returns true if the current item has link and can
-    # be enslabed in a bridge.
+    # be enslaved in a bridge.
     #
     # @return [Boolean] true if it is bridgeable
     def connected_and_bridgeable?(bridge_name, item, config)
-      if !LanItems.IsBridgeable(bridge_name, item)
+      # FIXME: a workaround until we fully use builders in proposals
+      bridge_builder = Y2Network::InterfaceConfigBuilder.for("br")
```
imobachgs: Using `Y2Network::InterfaceType` would be desirable :)

src/modules/LanItems.rb
```diff
@@ -786,156 +786,6 @@ def GetItemModules(default_module)
...
 
-    # Searches map of known devices and decides if referenced lan item
-    # can be enslaved in a bond device
-    #
-    # @param [String] bondMaster    name of master device
-    # @param [Fixnum] itemId        index into LanItems::Items
-    # TODO: Check for valid configurations. E.g. bond device over vlan
-    # is nonsense and is not supported by netconfig.
-    # Also devices enslaved in a bridge should be excluded too.
-    def IsBondable(bondMaster, itemId)
```
imobachgs: Yeah, dropping code is nice :)

src/lib/y2network/interfaces_collection.rb
mchf: i'm moving old functionality to new place ... if i change everything I would like to ... and other things you would like ... I'll get another huge PR which won't get accepted.

I planned to create Interface class descendants for bond and br as a next step and clean this methods which should be moved into these new classes when moving.

src/modules/Lan.rb
mchf: well InterfaceConfigBuilder hasn't use InterfaceType up to now ... changing it deserve own PR in my POV as I haven't implement the ```.for``` functionality here

src/modules/LanItems.rb
mchf: it was purpose of this PR. Keep what can be kept from LanItems and move it to new API to relax dependency on LanItems

src/lib/y2network/interfaces_collection.rb
mchf: well I feel that it is wrong if interfaces collection works only with configured interfaces. On the other hand it makes sense to me to have possibility to list ```all``` or ```configured``` interfaces in the future .... this functionality is used already in old LanItems but under not so pretty names like ```find_configured```

src/lib/y2network/interface.rb
mchf: ah yeah  .... mistake from rebasing :-/

src/lib/y2network/interface.rb
mchf: I don't have functionality from underlying layers ... we currently use NetworkInterfaces as access to backend everywhere. This will have to be cleaned once we have generic layer for accessing backend.

src/lib/network/lan_items_summary.rb
```diff
@@ -68,11 +68,12 @@ def default
     # @return [String] summary in RichText
     def proposal
       items = []
+      config = Y2Network::Config.find(:yast)
 
       items << "<li>#{dhcp_summary}</li>" unless LanItems.find_dhcp_ifaces.empty?
       items << "<li>#{static_summary}</li>" unless LanItems.find_static_ifaces.empty?
-      items << "<li>#{bridge_summary}</li>" unless bridges.empty?
-      items << "<li>#{bonding_summary}</li>" unless bonds.empty?
+      items << "<li>#{bridge_summary}</li>" unless config.interfaces.by_type("br").empty?
```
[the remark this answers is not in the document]
mchf: well ... I've been implementing it when InterfaceType was not merged ... address everything now would be complicated, so I did it when it was a must. So I'd prefer to postpone it.

src/lib/y2network/interfaces_collection.rb
mchf: hm, makes sense I'll adapt.

src/lib/y2network/interfaces_collection.rb
mchf: I'd propose to add it as another TODO bcs I planed to rework this method when moving to proper class as written in the todo.

src/lib/network/lan_items_summary.rb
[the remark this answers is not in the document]
mchf: in the end I've implemented it when touching ```by_type``` anyway

src/lib/y2network/interface.rb
mchf: cleaned

src/lib/y2network/interfaces_collection.rb
imobachgs: But it does not belong to InterfacesCollection, but to readers/writers.

PR #846, opened 2019-06-19
Y2Network config readers/writers reorganization

imobachgs:
At the beginning of the Y2Network development, we created two different modules to keep configuration readers and writers. The idea was to keep one class for each possible backend in those modules. However, as we add more features, we need more classes, and the code layout is pretty cumbersome. For instance, there are some common bits like `Y2Network::SysconfigRoutesFile` or `Y2Network::SysconfigInterfaceFile` which are common to readers and writers and we need them to live directly in `Y2Network`.

This PR tries to organize readers/writers differently, creating a module per backend. We [already considered](https://gist.github.com/imobachgs/1bc825a3fa9eb4cd5070fa01ed076251#proposal-1) a layout like this. We could create more modules or change things a bit, but that's the approach I think we should follow.

```
src/lib/y2network/sysconfig
├── config_reader.rb <- configuration reader
├── config_writer.rb <- configuration writer
├── connection_config_reader_handlers
│   ├── eth.rb
│   └── wlan.rb
├── dns_reader.rb
├── dns_writer.rb
├── interface_file.rb <- interface file
├── interfaces_reader.rb
├── interfaces_writer.rb
└── routes_file.rb <- routes file
```

So if tomorrow we want to support NetworkManager, we should create a similar layout (omitting the ancillary classes and modules):

```
src/lib/y2network/nm
├── config_reader.rb
└── config_writer.rb
```
This PR relies on #843, but I will rebase it before merging.


teclator:
For me the reorganization looks reasonable.

teclator:
LGTM

PR #848, opened 2019-06-20
Add backend documentation

imobachgs:
Add some documentation about the backend. I've added the file to the list of extra ones which are included in RubyDoc because I wanted the links to the classes that are mentioned to be updated automatically.

### DIFF
--- a/doc/network-ng.md
+++ b/doc/network-ng.md
@@ -1,0 +1,100 @@
+# Network NG
+
+This document tries to describe the details of the network-ng data model and how all the pieces are
+combined.
+
+## Overall Approach
+
+{Y2Network::Config} and other associated classes represent the network configuration in a backend
+agnostic way. It should not matter whether you are using `sysconfig` files, NetworkManager or
+networkd.
+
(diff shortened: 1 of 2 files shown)


doc/network-ng.md
```diff
@@ -0,0 +1,100 @@
...
+| Interface type  | read | write |
+|-----------------|------|-------|
+| Ethernet        |  ✓  |   ⌛   |
+| Wireless        |  ✓  |   ⌛   |
+| InfiniBand      |  ⌛   |       |
+| Bridge          |  ⌛   |       |
+| Bonding         |      |       |
+| VLAN            |      |       |
+| TUN             |      |       |
+| Dummy           |      |       |
```
jreidinger: Should we add here also S390 specific types? - https://github.com/yast/yast-yast2/blob/master/library/network/src/modules/NetworkInterfaces.rb#L1020 and also what others like USB or TAP?

imobachgs: You are right. However, I would like to clarify if we plan to drop support for any of them. For the time being, I am adding USB, TAP and s390 types.

PR #849, opened 2019-06-20
backport cpu mitigations

jreidinger:


src/clients/firewall_stage1_proposal.rb
```diff
@@ -130,17 +135,17 @@ def main
           raise "INTERNAL ERROR: unknown action '#{@chosen_link}' for proposal client"
         end
 
-        SuSEFirewallProposal.SetChangedByUser(true)
+        SuSEFirewallProposal.SetChangedByUser(true)
```
teclator:
```suggestion
        SuSEFirewallProposal.SetChangedByUser(true)
```

test/firewall_stage1_proposal_test.rb
```diff
@@ -1,61 +0,0 @@
...
-require "yast"
-
-module Yast
-  extend Yast::I18n
-
-  import "SuSEFirewall4Network"
-  import "SuSEFirewallProposal"
-  import "WFM"
-
-  describe "FirewallStage1ProposalClient" do
```
teclator: Deleting tests is not so nice :(

jreidinger: yep, not so nice, but for pure clients it is hard to mock them and I need to mock that bootloader loading part as bootloader is not build requirement.

teclator:
JFR: It basically backports https://github.com/yast/yast-firewall/pull/123, but in SLE-12-SP5 the first stage firewall proposal was in this repository instead. It has been unified in SLE-15 when adding firewalld based proposal.

teclator:
LGTM once provided version bump & changelog.

PR #850, opened 2019-06-21
[WIP] Added more connection configs

teclator:


src/lib/y2network/connection_config/bonding.rb
```diff
@@ -34,6 +35,31 @@ def initialize
...
+      #
+      # @param interfaces [Y2Network::InterfacesCollection]
+      def update_interfaces!(interfaces)
+        return if slaves.empty?
+        new_slaves = []
+
+        slaves.each do |index, slave|
+          alternative = interfaces.by_name(slave)
+          unless alternative
+            alternative = Y2Network::FakeInterface.new(slave)
```
imobachgs: So we are modifying the interfaces collection here? I would expect all the interfaces to be created at `InterfacesReader`, but not here. We should check and document this behavior

doc/network-ng.md
```diff
@@ -61,7 +61,12 @@ The sysconfig backend support is composed by these files:
     ├── config_writer.rb <- WRITER
     ├── connection_config_reader.rb
     ├── connection_config_readers
+    │   ├── bonding.rb
+    │   ├── bridge.rb
+    │   ├── dummy.rb
     │   ├── ethernet.rb
+    │   ├── infiniband.rb
+    │   ├── vlan.rb
```
jreidinger: I do not think we need to name all of them here. That `...` is enough to get idea that there is more types.

src/lib/y2network/connection_config/base.rb
```diff
@@ -31,10 +31,12 @@ class Base
       # no specific interface then it could be activated by the first available
       # device.
       #
-      # #FIXME: Maybe it could be a matcher instead of an Interface, or just a
+      # @return [String] Connection name
+      attr_accessor :name
```
jreidinger: so connection name can be different to interface name and in sysconfig case it is same, right?

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -336,7 +405,7 @@ def fetch_collection(key, type)
       end
 
       def collection_keys(key)
-        collection_keys = defined_variables.select do |k|
+        collection_keys = (defined_variables || []).select do |k|
```
imobachgs: `#defined_variables` should return always an array.

src/lib/y2network/connection_config/base.rb
```diff
@@ -31,10 +31,12 @@ class Base
...
+      # #FIXME: Maybe it could be a matcher instead of an Interface or just
       # the interface name by now.
       #
-      # @return [Interface, nil]
+      # @return [Interface, String, nil]
```
jreidinger: document what each type mean. interface is expected. String should be only interface name? and nil is connection without interface?

src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -47,6 +47,7 @@ def config
         return @config if @config
         find_physical_interfaces
         find_connections
+        link_related_interfaces!
```
imobachgs: Why a `!`? There is no `#link_related_interface` method, right?

src/lib/y2network/connection_config/base.rb
```diff
@@ -62,6 +64,15 @@ def type
         const_name = self.class.name.split("::").last.upcase
         InterfaceType.const_get(const_name)
       end
+
+      # @param interfaces [Y2Network::InterfacesCollection]
+      def update_interfaces!(interfaces)
+        nil
```
jreidinger: ugh, what? intentions?

src/lib/y2network/connection_config/base.rb
```diff
@@ -62,6 +64,15 @@ def type
...
+      end
+
+      def virtual?
```
jreidinger: documentation

src/lib/y2network/connection_config/usb.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for Usb connections
+    class Usb < Base
```
imobachgs: I would say that we agreed to use USB (and VLAN, TUN, TAP, etc.).

src/lib/y2network/connection_config/bonding.rb
```diff
@@ -34,6 +35,31 @@ def initialize
...
+
+      # @see Y2Network::InterfacesCollection
+      #
+      # @param interfaces [Y2Network::InterfacesCollection]
+      def update_interfaces!(interfaces)
+        return if slaves.empty?
+        new_slaves = []
+
+        slaves.each do |index, slave|
+          alternative = interfaces.by_name(slave)
```
jreidinger: why alternative? do I get it correctly that it creates slaves if needed as fake interfaces if it is not found, right?

src/lib/y2network/connection_config/bridge.rb
```diff
@@ -0,0 +1,46 @@
...
+      # @return [Boolean] whether spanning tree protocol is enabled or not
+      attr_accessor :stp
+      # @return [Integer]
+      attr_accessor :forward_delay
+
+      def initialize
+        @ports = []
+        @stp = false
+        @forward_delay = 0
+      end
```
jreidinger: so here ports cannot be fake ones?

src/lib/y2network/connection_config/infiniband.rb
```diff
@@ -0,0 +1,37 @@
...
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for infiniband connections
+    #
+    # @see https://www.kernel.org/doc/Documentation/infiniband/ipoib.txt
+    class Infiniband < Base
+      # @return [String] transport mode ("datagram" or "connected")
+      attr_accessor :ipoib_mode
```
jreidinger: not sure, maybe enum for it?

src/lib/y2network/connection_config/base.rb
teclator: In network manager them can be different. In sysconfig it is basically the ifcfg name.

src/lib/y2network/connection_config/tun.rb
```diff
@@ -0,0 +1,45 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for TUN/TAP connections
```
jreidinger: ugh, what? Tun and Tap are different interface type. So e.g. type detection won't work correctly.

src/lib/y2network/sysconfig/connection_config_reader.rb
```diff
@@ -60,6 +60,13 @@ def find_handler_class(type)
                  "Connection handler could not be loaded: #{e.message}"
         nil
       end
+
+      # @param connection [Y2Network::ConnectionConfig]
+      # @param interfaces [Array<Y2Network::Interface>] list of configured interfaces
+      # @return [Boolean]
+      def update_interfaces!(connection, interfaces)
```
jreidinger: what?

src/lib/y2network/connection_config/base.rb
teclator: It is just for opening the discussion about what to use by now. In virtual connections, it is mainly a 1-1 relationship between the connection and the virtual interface, for other type of connection it could be a matcher. 

The doubt also arised when working with the readers, who should be responsible of add the virtual interface and what is preferable to use, the name or the interface as the association because at the time of reading the interface is still not present and the reader know nothing about current config..

src/lib/y2network/sysconfig/connection_config_writers/bonding.rb
```diff
@@ -0,0 +1,41 @@
...
+      # This class is responsible for writing the information from a ConnectionConfig::Bonding
+      # object to the underlying system.
+      class Bonding < Base
+
+      private
+
+        # @see Y2Network::ConnectionConfigWriters::Base#update_file
+        # @param conn [Y2Network::ConnectionConfig::Bonding] Configuration to write
+        def update_file(conn)
+          # FIXME:
```
jreidinger: fixme what?

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -231,6 +237,62 @@ def variable_name(param_name)
...
+      # @!attribute [r] bonding_module_opts
+      #   @return [String] options for the bonding module ('mode=active-backup
+      #                     miimon=100')
+      define_variable(:bonding_module_opts)
+
+      ## BRIDGE
+
+      # @!attribute [r] bridge_ports
+      #   @return [String] interfaces members of the bridge
+      define_variable(:bridge_ports)
```
jreidinger: why not array instead of string?

src/lib/y2network/connection_config/base.rb
teclator: Basically trying to look the best place and naming for replacing interface names by objects in master / slaves connections or vlan for example.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -278,7 +340,14 @@ def save
       #
       # @return [String] Interface's type depending on the file values
       def type
-        "eth"
+        return InterfaceType::DUMMY if @values["INTERFACETYPE"] == "dummy"
+        return InterfaceType::BONDING if defined_variables.any? { |k| k.start_with?("BOND") }
+        return InterfaceType::BRIDGE if defined_variables.any? { |k| k.start_with?("BRIDGE") }
+        return InterfaceType::WIRELESS if defined_variables.any? { |k| k.start_with?("WIRELESS") }
+        return InterfaceType::VLAN if defined_variables.include? ("ETHERDEVICE")
+        return InterfaceType::INFINIBAND if defined_variables.include? ("IPOIB_MODE")
```
jreidinger: tun/tap according to mode?

src/lib/y2network/connection_config/base.rb
teclator: As said before, virtual connection mainly means a 1:1 relation with virtual interfaces and we need a way to differentiate from fake interfaces.

src/lib/y2network/virtual_interface.rb
```diff
@@ -22,5 +22,14 @@
 module Y2Network
   # Virtual Interface Class (veth, bond, bridge, vlan, dummy...)
   class VirtualInterface < Interface
+    # Build connection
+    #
+    # @todo Would be possible to get the name from the connection?
```
jreidinger: you already passing it

src/lib/y2network/connection_config/bonding.rb
teclator: Yep, probably moving it to the interfaces reader will be better, so having to know or being responsible of replacing the name of interfaces related by Y2Network::Interface objects is something it could deal with

src/lib/y2network/connection_config/bonding.rb
teclator: It links the slaves with real or fake interfaces, yes. So definitely a bad naming.

src/lib/y2network/sysconfig/interface_file.rb
teclator: There was some error when checking readers but it has been already addressed. So will report in case of find again.

src/lib/y2network/connection_config/base.rb
mchf: to be honest I'm still lost in current sense of Interface classes. 

Existing device (i mean real NIC e.g. ethernet one) -> physical_interface - I'm fine with that. We currently don't "need" it but i can see it's purpose

bond / bridge / vlan and other beasts -> virtual_interface - why? It is configuration stuff and in theory it can easily change when a config is switched to another profile

non existing device (a configuration for a physical device which is not present) - currently can be also seen as a configuration stuff as we don't handle hotplug events and do not re-read interfaces during the module's life so it never can turn to physical interface.

So as I see it now we need just one class for physical interfaces (which is currently not much useful at this time anyway) - and keeping it bcs of faked renaming example (deleting existing virtual config and creating new one) is not a reason for keeping bunch of code living.

If i summarize it. If you want to detect connection type you can check whether it has an interface assigned. If yes then it is a physical one. If not then it is either virtual or fake. If you are able to detect virtual type (bond, br, ...) from the config you have a virtual connection. If not you have a fake one.

src/lib/y2network/connection_config/base.rb
mchf: i currently know of only one usage of this so putting it to generic ancestor makes no sense to me.

src/lib/y2network/connection_config/base.rb
teclator: > to be honest I'm still lost in current sense of Interface classes.
> 
> Existing device (i mean real NIC e.g. ethernet one) -> physical_interface - I'm fine with that. We currently don't "need" it but i can see it's purpose
> 
> bond / bridge / vlan and other beasts -> virtual_interface - why? It is configuration stuff and in theory it can easily change when a config is switched to another profile

So, we can avoid adding them to the list of interfaces, just to not declare all used by existent configurations but only define the virtual interface name.

> 
> non existing device (a configuration for a physical device which is not present) - currently can be also seen as a configuration stuff as we don't handle hotplug events and do not re-read interfaces during the module's life so it never can turn to physical interface.

Agree
> 
> So as I see it now we need just one class for physical interfaces (which is currently not much useful at this time anyway) - and keeping it bcs of faked renaming example (deleting existing virtual config and creating new one) is not a reason for keeping bunch of code living.

Well, them are abstractions that are currently valid and the minimal we thought that could make sense to add. For example you could read information about udev rules and has always associated an interface. A physical one in case of present or a fake in case of not.

When read you could also represent virtual interfaces that are currently listed but not managed but that is other history. 

The main point for me is how and what to associate with a connection.

> 
> If i summarize it. If you want to detect connection type you can check whether it has an interface assigned. If yes then it is a physical one. If not then it is either virtual or fake. If you are able to detect virtual type (bond, br, ...) from the config you have a virtual connection. If not you have a fake one.

src/lib/y2network/connection_config/bridge.rb
teclator: Why not?, Them are interfaces and a Fake interface is an Interface.

src/lib/y2network/connection_config/base.rb
imobachgs: I will try to summarize the rationale behind interfaces and connection configurations concepts. The idea of the interfaces class is to represent real system interfaces (for real, I mean interfaces under `/sys/class/net`, not only physical ones). The exception is the `FakeInterface` class that are interfaces which we *know* that will exist at some point (USB, hotplug, etc.).

Connection configurations are responsible to describe interfaces configurations. We usually consider a 1:1 relationship with interfaces, but it is not limited to that. For instance, in NM you can specify different profiles for a given interface (it may depend if you are connecting at home or at work, using different wifi networks, etc.). So, basically, these are different from the interfaces. In sysconfig, they usually correspond to `ifcfg-*` files.

Having said that, we can of course adapt or modify those concepts, but I would say that the idea will allow us to represent regular ifcfg-* files, NM profiles or, in the future, Wicked roles.

src/lib/y2network/sysconfig/interface_file.rb
teclator: Inferring types from values is still missing, and probably part of another card. I added these mainly for testing purpose.

src/lib/y2network/connection_config/base.rb
mchf: I understood the concept (at least I thought so) ... what i wanted to point was that we don't use the model (Interface hierarchy). Only case I currently see in the near future is udev for phy devices.

But then I saw a PRs for Conection config for bond / br where fake interfaces are created according to br / bond configuration and I've got lost as it breaks the concept above - bcs fake interfaces are created for ifcfgs not assigned to physical device. But in this case it was created bcs it could be a non-existent BRIDGE_PORT (e.g.)

So not to enlarge it ... I currently don't see a point in having interfaces hierarchy and insist on mapping of each connection config to an interface. I'm slightly afraid that we got hijacked and over engineering a bit

teclator: > I understood the concept (at least I thought so) ... what i wanted to point was that we don't use the model (Interface hierarchy). Only case I currently see in the near future is udev for phy devices.
> 
> But then I saw a PRs for Conection config for bond / br where fake interfaces are created according to br / bond configuration and I've got lost as it breaks the concept above - bcs fake interfaces are created for ifcfgs not assigned to physical device. But in this case it was created bcs it could be a non-existent BRIDGE_PORT (e.g.)

I see your point and reasons behind it. Basically we wanted to have in the `Y2Network::Config` all the interfaces that were somehow managed by YaST. That is, physical, virtual or fake interface, but of course, we could avoid doing that just based on connection relations.

> 
> So not to enlarge it ... I currently don't see a point in having interfaces hierarchy and insist on mapping of each connection config to an interface. I'm slightly afraid that we got hijacked and over engineering a bit

teclator:
This PR has been replaced by smaller one adding each specific types:

- https://github.com/yast/yast-network/pull/896 (dummy type)
- https://github.com/yast/yast-network/pull/899 (infiniband type)
- https://github.com/yast/yast-network/pull/901 (vlan type)
- https://github.com/yast/yast-network/pull/903 (tun / tap type)
- https://github.com/yast/yast-network/pull/904 (bridge type)
- https://github.com/yast/yast-network/pull/908 (bonding type)
- https://github.com/yast/yast-network/pull/909 (qeth type)
- https://github.com/yast/yast-network/pull/915 (hsi type)

So, will close it definitely.

PR #853, opened 2019-07-01
add initial test cases document

jreidinger:


doc/test-cases.md
```diff
@@ -0,0 +1,36 @@
...
+5. create dummy
+6. create vlan on top of dummy
+
+#### CLI
+
+For testing this scenario create script that take as params name of two physical devices and using yast2 lan CLI create testing setup.
+
+#### AutoYaST
+
+For testing this scenario create autoyast profile that can be passed to autoinstallation that creates required setup.
```
imobachgs: We should be able to reproduce the rest of scenarios but using AutoYaST profiles. Moreover, we should be able that the base case (no network configuration at all) still works.

doc/test-cases.md
```diff
@@ -0,0 +1,36 @@
+# Test Cases
+## Purpose
+Goal of this document is to document non trivial testing scenarious we can use to ensure network module is not seriously broken.
+
+## Test Cases
+
+### Complex setup
+
+Test case require VM with two network cards. Setup is:
+
```
teclator: I would start even with the more simplest ones. 

1. DHCP configuration (no configuration at all but different backends)
2. Static configuration ( 1 network card is enough, that will need some routing and DNS configuration by sure)
...
And then the complex ones.

A Goal for the future is to have the same network configuration done during the installation in the final machine independently of the selected backend (**not for now**)

jreidinger:
replaced by https://github.com/yast/yast-network/pull/947

PR #855, opened 2019-07-02
Merge master

imobachgs:
Merge master changes into master.


src/include/network/lan/complex.rb
```diff
@@ -328,9 +328,10 @@ def UpdateSlaves
         builder["BOOTPROTO"] = "none"
         case builder.type
         when "bond"
-          builder["STARTMODE"] = "hotplug"
-
-          LanItems.update_item_udev_rule!(:bus_id)
+          LanItems.startmode = "hotplug"
+          if LanItems.current_udev_rule.empty? || LanItems.GetItemUdev("KERNELS").empty?
```
jreidinger: hmm, can we have here comment why only in this case? @teclator probably remember why this if.

teclator: If there is already a udev rule based on the bus_id, then it does not make sense to update it. Update item_udev_rule ... or GetItemUdev(Item) was buggy. That is, when there was a udev rule based on :bus_id, if we updated it, then the attribute was duplicated.

PR #856, opened 2019-07-02
Fix installation

imobachgs:
* Fix DNS initialization during installation
* Fix reading IPADDR value during installation

### DIFF
--- a/src/lib/y2network/sysconfig/interface_file.rb
+++ b/src/lib/y2network/sysconfig/interface_file.rb
@@ -130,7 +130,7 @@ module Y2Network
       # @return [IPAddr,nil] IP address or nil if it is not defined
       def ip_address
         str = fetch("IPADDR")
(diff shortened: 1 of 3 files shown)


src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -130,7 +130,7 @@ def initialize(name)
       # @return [IPAddr,nil] IP address or nil if it is not defined
       def ip_address
         str = fetch("IPADDR")
-        str.empty? ? nil : IPAddr.new(str)
+        str.nil? || str.empty? ? nil : IPAddr.new(str)
```
jreidinger: we really have to adapt new rubocop to simply call just `str&.empty?`

imobachgs: Good point.

PR #857, opened 2019-07-02
Use the new builder when modifying the changes.

teclator:


src/include/network/lan/cmdline.rb
```diff
@@ -244,56 +243,37 @@ def EditHandler(options)
...
       else
-        LanItems.ipaddr = ""
-        LanItems.netmask = ""
+        builder["IPADDR"] = ""
+        builder["NETMASK"] = ""
       end
 
-      LanItems.startmode = Ops.get(options, "startmode", "auto")
-      if !Builtins.contains(["auto", "ifplugd", "nfsroot"], LanItems.startmode)
+      builder["STARTMODE"] = options.fetch("startmode", "auto")
```
teclator: startmode is even a not supported option ...

src/include/network/lan/cmdline.rb
```diff
@@ -244,56 +243,37 @@ def EditHandler(options)
       end
       LanItems.SetItem(builder: builder)
 
-      if Builtins.size(Ops.get_string(LanItems.getCurrentItem, "ifcfg", "")) == 0
-        NetworkInterfaces.Add
-        LanItems.operation = :edit
```
jreidinger: sadly removal of this does not work, as builder expect that Add and operation settings. ( I planned to add it there, but in the end do not do it )

teclator: I was checking if was needed at all. I was able to edit a non configured one. So, was not sure.

teclator:
```bash
$ yast2 lan edit id=0 ip=192.168.0.4 netmask=255.255.255.0
## THE OVERVIEW IS WRONG
RTL-8100/8101L/8139 PCI Fast Ethernet Adapter (Not connected)
MAC : 52:54:00:6e:7c:0e
BusID : 0000:00:0b.0
Device Name: ens11
The device is not configured. Press Edit

$ yast2 lan show id=0
RTL-8100/8101L/8139 PCI Fast Ethernet Adapter (Not connected)
MAC : 52:54:00:6e:7c:0e
BusID : 0000:00:0b.0
Device Name: ens11
Started automatically at boot
IP address: 192.168.0.4/24
to configure.

$ cat /etc/sysconfig/network/ifcfg-ens11
BOND_SLAVES='[]' ## This is incorrect.
BOOTPROTO='static'
BROADCAST=''
ETHTOOL_OPTIONS=''
IPADDR='192.168.0.4/24'
MTU=''
NAME=''
NETWORK=''
PREFIX=''
PREFIXLEN=''
REMOTE_IPADDR=''
STARTMODE='auto'

Comment (jreidinger) [reply src/include/network/lan/cmdline.rb]: try to add completelly new device like new bridge. This is where add is needed.

Comment (teclator) [reply src/include/network/lan/cmdline.rb]: But then it is a completely different handler, isn't it? This is the edit handler, where the interface should be listed as configured or unconfigured.

File: src/include/network/lan/cmdline.rb [hunk 3]
```diff
@@ -191,31 +192,22 @@ def AddHandler(options)
         return false
       end
 
+      builder = Y2Network::InterfaceConfigBuilder.for(LanItems.type)
+      builder.name = LanItems.GetCurrentName()
+
       case LanItems.type
```
jreidinger: NP: here you can query builder.type

src/include/network/lan/cmdline.rb
```diff
@@ -191,31 +192,22 @@ def AddHandler(options)
...
       when "bond"
-        LanItems.bond_slaves = options.fetch("slaves", "").split(" ")
+        builder["BOND_SLAVES"] = options.fetch("slaves", "").split(" ")
       when "vlan"
-        LanItems.vlan_etherdevice = options.fetch("ethdevice", "")
+        builder["ETHERDEVICE"] = options.fetch("ethdevice", "")
       when "br"
-        LanItems.bridge_ports = options.fetch("bridge_ports", "")
+        builder["BRIDGE_PORTS"] = options.fetch("bridge_ports", "")
       end
```
jreidinger: why this case is not in `update_builder_from_options!` ?

teclator: Well, basically because options where not allowed edition but just when adding. Although from my POV should be valid in both cases and the behavior could be also moved to the convenience method.

PR #858, opened 2019-07-02
Fixes

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -184,7 +184,8 @@ def device_sysconfig
       config = config.delete_if { |k, _| k == "IFPLUGD_PRIORITY" } if config["STARTMODE"] != "ifplugd"
 
       # all keys / values has to be strings, agent won't write it otherwise
-      config.map { |k, v| [k, v.to_s] }.to_h
+      # JR this does not work for aliases, BON_OPTIONS and SLAVES which is later processed
```
imobachgs: * "JR"?
* "which are processed later"
* On the other hand, should we drop this code?

jreidinger: I want to discuss it this code with @mchf  that is why I keep it commented out instead of complete removal. JR is @jreidinger, so michal can ask me when he see it. Will fix typo.

imobachgs:
Now even better.

PR #859, opened 2019-07-02
Fix DNS writing during installation

imobachgs:


test/dns_test.rb
```diff
@@ -156,7 +156,7 @@ module Yast
       end
 
       it "writes DNS settings" do
-        expect(dns_writer).to receive(:write).with(yast_config, system_config)
+        expect(dns_writer).to receive(:write).with(yast_config.dns, system_config.dns)
```
jreidinger: note for future, this for me looks like overmocking which cause that test does not catch potential issues. I more likely expect here check if something is really written down on fake env.

PR #860, opened 2019-07-02
fix creating vlan

jreidinger:


src/lib/y2network/widgets/vlan_id.rb
```diff
@@ -23,7 +23,7 @@ def init
       end
 
       def store
-        @config["VLAN_ID"] = value
+        @config["VLAN_ID"] = value.to_s
```
imobachgs: Could `value` be nil?

jreidinger: no, it is input selection. Only possibility is if widget is disabled, but in that case store method is not called.

PR #863, opened 2019-07-03
Do not raise an exception when no interface type is provided

teclator:
## Problem

Since https://github.com/yast/yast-network/pull/661 the device type is mandatory, but the help has not been updated and a exception is raised instead of an error reported.
```
$ yast lan add name=vlan50 ethdevice=eth0

# Reports
Error: Internal error. Please report a bug report with logs.
Run save_y2logs to get complete logs.

Caller:  /usr/share/YaST2/include/network/lan/cmdline.rb:194:in `AddHandler'
Details: Device type is mandatory.
```
https://bugzilla.suse.com/show_bug.cgi?id=1137346

## Solution

- Try to infer the device type when able to.
- Update the examples adding the new mandatory device 'type'.
- Report an error instead of raise an exception when no type is provided and there is no way to infer it from the given options.

## Tests

- New unit test created.
- Tested manually

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Wed Jul  3 13:18:05 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
 
 - bsc#1137346
-  - CLI: Report an error isntead of raise an exception when the
+  - CLI: Report an error instead of raising an exception when the
     device type is not provided and there is no way to infer it
     from the given options.
 - 4.1.50
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,8 +2,9 @@
 Wed Jul  3 13:18:05 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
 
 - bsc#1137346
-  - CLI: Do not raise an exception when the device type is not
-    provided but report an error instead.
+  - CLI: Report an error isntead of raise an exception when the
+    device type is not provided and there is no way to infer it
+    from the given options.
 - 4.1.50
 
 -------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Jul  3 13:18:05 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- bsc#1137346
+  - CLI: Report an error isntead of raise an exception when the
```
imobachgs: typo: instead

imobachgs: I would say "instead of raising", but I may be wrong.

test/cmdline_test.rb
```diff
@@ -21,4 +21,60 @@ def initialize
...
+    before do
+      allow(Yast::Report).to receive(:Error)
+      allow(Yast::LanItems).to receive(:Commit)
+    end
+
+    context "when called without type" do
+      context "and it cannot be infered from the given options" do
+        it "reports an error" do
+          expect(Yast::Report).to receive(:Error)
+          subject.AddHandler(options.reject { |k| k == "ethdevice" })
```
imobachgs: np: In order to reduce duplicity, you could move the `#reject` logic to a *let* block.

test/new_device_startmode_test.rb
```diff
@@ -53,9 +58,11 @@
         expect(result).to be_eql expected_startmode
       end
 
-      it "results to ifplugd when running on laptop" do
+      it "results to ifplugd when running on laptop with wicked and not virtual device" do
```
imobachgs: What exactly "not virtual device" means?

teclator: Well, basically this results in ifplugd if none of the other conditions are true, not just because running on a laptop

https://github.com/yast/yast-network/blob/3d5ca9554e975d78ecadcda3a428361029530e52/test/new_device_startmode_test.rb#L79

teclator: currently it means not a "bridge", "bond" or "vlan"

PR #868, opened 2019-07-04
Merges SLE-15-SP1 into master

imobachgs:
Brings some fixes (like #866) to master.

### DIFF
--- a/src/include/network/lan/cmdline.rb
+++ b/src/include/network/lan/cmdline.rb
@@ -182,41 +182,35 @@ module Yast
     # Handler for action "add"
     # @param [Hash{String => String}] options action options
     def AddHandler(options)
-      options = deep_copy(options)
       LanItems.AddNew
       Lan.Add
-      Ops.set(
-        LanItems.Items,
-        [LanItems.current, "ifcfg"],
-        Ops.get(options, "name", "")
-      )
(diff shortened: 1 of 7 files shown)


src/include/network/lan/cmdline.rb
```diff
@@ -182,41 +182,35 @@ def ListHandler(options)
...
-      if LanItems.type == "br"
-        LanItems.bridge_ports = Ops.get(options, "bridge_ports", "")
+
+      case LanItems.type
+      when "bond"
+        LanItems.bond_slaves = options.fetch("slaves", "").split(" ")
+      when "vlan"
+        LanItems.vlan_etherdevice = options.fetch("ethdevice", "")
+      when "br"
+        LanItems.bridge_ports = options.fetch("bridge_ports", "")
```
jreidinger: hmm, CLI is really useless, you basically cannot configure infiband datagram mode, also vlan id and other options like tun/tap device owner and I am pretty sure that more other options missing.

teclator: @jreidinger yep. It seems to be used to very specific use cases, because some options are not really supported, some others are limited, the output is not very user friendly... so, need some love, but just fix by now and collect use cases to see if it worth the effort to re-implement it

PR #869, opened 2019-07-04
Bring master changes to network-ng

imobachgs:
Bring master changes to the networ-ng branch.


test/new_device_startmode_test.rb
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env rspec
+
+require_relative "test_helper"
+
+require "yast"
+
+Yast.import "ProductFeatures"
+Yast.import "LanItems"
+
+describe "LanItemsClass#new_device_startmode" do
```
jreidinger: this is wrong as this method is no longer in network-ng, or am I wrong?

test/new_device_startmode_test.rb
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env rspec
```
teclator: This test was completely removed in network-ng

test/new_device_startmode_test.rb
imobachgs: Yes, you are right. Dropped.

test/new_device_startmode_test.rb
imobachgs: Yes, you are right. Dropped.

teclator:
@imobachgs @jreidinger there is a failing test because Commit expect now the builder. I can fix it later if you want as part of the CLI adaptions I'm doing for network-ng

teclator:
LGTM

PR #870, opened 2019-07-04
Several fixes across builders and removed some unneeded code

mchf:


src/lib/y2network/interface_config_builders/ib.rb
```diff
@@ -13,13 +13,17 @@ def initialize
...
+        else
+          @config["IPOIB_MODE"]
+        end
       end
 
       def save
         super
 
-        Yast::LanItems.ipoib_mode = ipoib_mode == "default" ? nil : ipoib_mode
+        @config["IPOIB_MODE"] = ipoib_mode == "default" ? nil : ipoib_mode
```
jreidinger: do you test this with real write? I sometimes found that transition from LanItems to NetworkInterfaces has many cave eats.

mchf: yes ... but testing even more just now

mchf: manually tested and seems to be ok

src/lib/y2network/interface_config_builders/dummy.rb
```diff
@@ -0,0 +1,18 @@
...
+require "y2network/interface_config_builder"
+
+module Y2Network
+  module InterfaceConfigBuilders
+    class Dummy < InterfaceConfigBuilder
+      def initialize
+        super(type: "dummy")
+      end
+
+      def save
```
teclator: In general miss documentation in `InterfaceConfigBuilder`methods, in particular do not expected to return sometimes the set of an attribute and sometimes `nil`.. specially when more than one attribute is responsibility of a specialized class.

src/lib/y2network/interface_config_builders/ib.rb
teclator: Same than above, please try to document as much as possible the new classes / methods.

test/y2network/interface_config_builders/ib_test.rb
```diff
@@ -27,14 +27,12 @@
 
     it "stores ipoib configuration" do
       subject.ipoib_mode = "datagram"
-      subject.save
-      expect(Yast::LanItems.ipoib_mode).to eq "datagram"
+      expect(subject.save).to eq "datagram"
```
teclator: I would prefert to check the resultant config than the return of the save method

test/y2network/interface_config_builders/ib_test.rb
```diff
@@ -27,14 +27,12 @@
...
     end
 
     it "stores nil to ipoib configuration if mode is 'default'" do
       subject.ipoib_mode = "default"
-      subject.save
-      expect(Yast::LanItems.ipoib_mode).to eq nil
+      expect(subject.save).to eq nil
```
teclator: I would expect to return something like true or false.

teclator:
LGTM

PR #872, opened 2019-07-04
fix s390 initialize

jreidinger:


src/include/network/lan/complex.rb
```diff
@@ -437,6 +438,12 @@ def handleOverview(_key, event)
               return nil
             end
 
+            type = LanItems.GetCurrentType()
+            # s390 does not have type yet set, but it can be read from hwinfo
+            type = LanItems.Items.dig(LanItems.current, "hwinfo", "type") if !type || type.empty?
```
teclator: np: I have to refresh all the ways to obtain the type, and how is the preferred way now but seems at least to me that if it is read from the hwinfo then it should be already returned by GetCurrentType

jreidinger: I hope we can kill it in near future :)

PR #874, opened 2019-07-09
Improve tests2

jreidinger:


test/y2network/sequences/interface_test.rb
```diff
@@ -0,0 +1,83 @@
...
+
+require "y2network/sequences/interface"
+
+Yast.import "Sequencer"
+
+describe Y2Network::Sequences::Interface do
+  let(:builder) do
+    res = Y2Network::InterfaceConfigBuilder.new
+    res.type = "eth"
+    res
```
imobachgs: np: You could use `#tap` but, in this case, it won't save you any keystroke :)

PR #875, opened 2019-07-09
Network ng write connection config

imobachgs:
This PR adds support to write ifcfg-* files for network interfaces (the so-called _ConfigurationConfig_ objects in the new data model). At this point in time, only Ethernet and Wireless connections are supported.

Note: there is a problem when writing the IP address. I am working on a fix.


src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -35,22 +35,78 @@ class << self
...
+          params << param_name
+
+          define_method param_name do
+            return @values[param_name] if @values.key?(param_name)
+            value = fetch(param_name.to_s.upcase)
             send("value_as_#{type}", value)
           end
+
+          define_method "#{param_name}=" do |value|
+            @values[param_name] = value
```
jreidinger: i would mention here in comment that write does not need convert, that write itself use always to_s for type. So all types have to have this method and also it have to return expected value. It is something I have to check and is not obvious from code.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -35,22 +35,78 @@ class << self
...
+          define_method "#{param_name}s" do
+            return @values[param_name] if @values.key?(param_name)
+            key = param_name.to_s.upcase
+            values = [fetch(key)]
+            values += Array.new(limit) do |idx|
+              value = fetch("#{key}_#{idx}")
+              next if value.nil?
+              send("value_as_#{type}", value)
+            end
+            values.compact
```
jreidinger: I am not sure if it is problem, but this compact basically change file. 
if there is KEY_1, KEY_3, then after write there will be KEY_1, KEY_2.

test/y2network/sysconfig/connection_config_writer_test.rb
```diff
@@ -0,0 +1,71 @@
...
+        interface: "eth0",
+        type:      Y2Network::InterfaceType::ETHERNET
+      )
+    end
+
+    let(:file) do
+      instance_double(
+        Y2Network::Sysconfig::InterfaceFile, save: nil, clean: nil
+      )
+    end
```
jreidinger: I would really feel more confident if we can simply switch SCR here and operate on real files. So we see what is real changes.

imobachgs: We rely on `Y2Network::Sysconfig::InterfaceFile` in which unit tests we switch SCR and operate on real files. See https://github.com/yast/yast-network/blob/fc128f537beb6f84a72783c08f4a121a13616ee8/test/y2network/sysconfig/interface_file_test.rb#L37.

src/lib/y2network/sysconfig/interface_file.rb
imobachgs: Good point. I do not think it is a problem as we are not going to write the array unless we have explicitly modified it. But it is true that it does not reflect the real content. I will fix it.

imobachgs: Thinking about that, we could solve this as part of [the upcoming support for multiple IPs](https://trello.com/c/0lBPS32M/997-network-ng-start-with-a-data-model-which-permits-to-use-different-network-connection-profiles).

doc/network-ng.md
```diff
@@ -40,29 +40,73 @@ other classes.
...
     ├── config_reader.rb <- READER
     ├── config_writer.rb <- WRITER
-    ├── connection_config_reader_handlers
-    │   ├── eth.rb
-    │   ├── wlan.rb
-    │   └── ...
     ├── connection_config_reader.rb
+    ├── connection_config_readers
+    │   ├── eth.rb
+    │   └── wlan.rb
```
jreidinger: maybe add there also that ... as it was before, so it is clear it is not all of them.

doc/network-ng.md
```diff
@@ -40,29 +40,73 @@ other classes.
...
 
-As you can see, there are many classes, but the relevant ones are just `ConfigReader` and `ConfigWriter`.
+{Y2Network::Sysconfig::ConfigReader} and {Y2Network::Sysconfig::ConfigWriter} are the reader and
+writer classes. Each of them cooperates with a set of ancillary classes in order to get the job
+done.
+
+{Y2Network::Sysconfig::DNSReader}, {Y2Network::Sysconfig::InterfacesReader} and
+{Y2Network::Sysconfig::ConnectionConfigReader} are involved in reading the configuration. The logic
+to read the configuration for a connection (e.g., `ifcfg-eth0`, `ifcfg-wlan0`, etc.) is implemented
+in a set of smaller classes (one for each time of connection) under
```
jreidinger: maybe mention here InterfaceType class that is used for that smaller class as delimeter.

doc/network-ng.md
```diff
@@ -40,29 +40,73 @@ other classes.
...
+{Y2Network::Sysconfig::ConnectionConfigReaders}.
+
+{Y2Network::Sysconfig::DNSWriter} and {Y2Network::Sysconfig::ConnectionConfigWriter}, including
+smaller classes under {Y2Network::Sysconfig::ConnectionConfigWriters}, are involved in writing the
+configuration. In this case, it does not exist a `InterfacesWriter` class, because when it comes to
+interfaces the configuration is handled through connection configuration files.
```
jreidinger: so if we have any attributes in interface that can be changed, it is done by ConnectionConfigWriter or is not allowed at all?

doc/network-ng.md
```diff
@@ -40,29 +40,73 @@ other classes.
...
+
+Last but not least, there are additional classes like {Y2Network::Sysconfig::RoutesFile} and
+{Y2Network::Sysconfig::InterfaceFile} which abstract the details of reading/writing `ifcfg` files.
+
+### AutoYaST
+
+AutoYaST is a special case in the sense that it reads the information from a profile, instead of
+using the running system as reference. Additionally, it does not implement a writer because the
+configuration will be written using a different backend (like sysconfig).
```
jreidinger: What is responsible for autoyast export?

src/lib/y2network/connection_config/base.rb
```diff
@@ -36,7 +36,10 @@ class Base
       #
       # @return [Interface, nil]
       attr_accessor :interface
-      # @return [String] Bootproto (static, dhcp, ,dhcp4, dhcp6, autoip,
+      # @return [Symbol] Choose when the interface should be set up (manual, auto, nfsroot, hotplug,
+      #   off)
+      attr_accessor :startmode
```
jreidinger: well, collision with my pr, but something that can be solved.

src/lib/y2network/ip_address.rb
```diff
@@ -0,0 +1,97 @@
...
+    IPV4_DEFAULT_PREFIX = 32
+    # @return [Integer] IPv6 address default prefix
+    IPV6_DEFAULT_PREFIX = 128
+
+    # Constructor
+    #
+    # @param address [String] IP address without the prefix
+    # @param prefix [Integer] IP prefix (number of bits). If not specified, 32 will be used for IPv4
+    #   and 128 for IPv6.
+    def initialize(address, prefix = nil)
```
jreidinger: so we drop netmask completelly?

src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -77,7 +78,8 @@ def write_interface_changes(config, old_config)
           file.remove
         end
 
-        nil
+        write_dns_settings(config, old_config)
```
jreidinger: so we do not write dns previously?

jreidinger: or better, why it have to be here and not in def write? can you document reason? as I see it written twice

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -0,0 +1,55 @@
...
+  module Sysconfig
+    class ConnectionConfigWriter
+      include Yast::Logger
+
+      # Writes connection config to the underlying system
+      #
+      # @param conn [Y2Network::ConnectionConfig::Base] Connection configuration to write
+      def write(conn)
+        file = Y2Network::Sysconfig::InterfaceFile.new(conn.interface)
+        handler_class = find_handler_class(conn.type.short_name)
```
jreidinger: ideally pass type here and not short_name

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    private
+
+      # Returns the class to handle a given interface type
+      #
+      # @param type [Symbol]
+      # @return [Class] A class which belongs to the ConnectionConfigWriters module
+      def find_handler_class(type)
+        require "y2network/sysconfig/connection_config_writers/#{type}"
+        ConnectionConfigWriters.const_get(type.to_s.capitalize)
```
jreidinger: and here use new methods of type like `type.path_name` and `type.class_name`

src/lib/y2network/sysconfig/connection_config_writers/eth.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Ethernet
+      # object to the underlying system.
+      class Eth
```
jreidinger: consistency. It should be Ethernet.

src/lib/y2network/sysconfig/connection_config_writers/wlan.rb
```diff
@@ -0,0 +1,93 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Wireless
+      # object to the underlying system.
+      class Wlan
```
jreidinger: and Wireless here

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -65,6 +125,11 @@ def define_parameter(name, type = :string)
       #   @return [Integer] Length in bits for all keys used
       define_parameter(:wireless_key_length, :integer)
 
+      # @return [Integer] Number of supported keys
```
jreidinger: how looks @return in constant case? in yardoc?

src/lib/y2network/sysconfig/connection_config_writers/eth.rb
imobachgs: Those are the names we used in sysconfig, hence the name of the class. I will move it.

src/lib/y2network/ip_address.rb
imobachgs: Yes, basically we keep the address and the prefix. The netmask, if needed, can be inferred from there. Take into account that this class exists only to workaround the original IPAddr limitations.

src/lib/y2network/sysconfig/config_writer.rb
imobachgs: Good catch. It was a problem with merging. I will fix it.

doc/network-ng.md
imobachgs: For the time being, everything is handled through connection configs (writing ifcfg-* files), but of course that could change in the future.

src/lib/y2network/sysconfig/interface_file.rb
imobachgs: it just displays the "return" type but, as it also shows the value, it is useless. Removed.

doc/network-ng.md
jreidinger: so maybe mention it here.

PR #876, opened 2019-07-10
Fixed deleting ifroute when an interface is deleted

mchf:


src/lib/y2network/config.rb
```diff
@@ -101,9 +101,14 @@ def initialize(interfaces: InterfacesCollection.new, connections: [], routing: R
...
+    # Writes only changes agains original configuration if the original configuration
+    # is provided
+    #
+    # @param original [Y2Network::Config] configuration used for detecting changes
+    #
     # @see Y2Network::ConfigWriter
-    def write
-      Y2Network::ConfigWriter.for(source).write(self)
+    def write(original: nil)
+      Y2Network::ConfigWriter.for(source).write(self, old_config = original)
```
jreidinger: is assignment to old_config really needed?

src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -37,31 +37,47 @@ def write(config, old_config = nil)
...
           file.routes = routes
           file.save
         end
 
-        # update /etc/sysconfig/network/routes file
-        file = routes_file_for(nil)
-        file.routes = find_routes_for(nil, config.routing.routes)
-        file.save
+        # Actions needed for removed interfaces
+        removed_ifaces = old_config ? old_config.interfaces.to_a - config.interfaces.to_a : []
```
jreidinger: so old config can be nil? If so, please document in method documentation.

test/y2network/sysconfig/config_writer_test.rb
```diff
@@ -36,7 +36,7 @@
         source:     :sysconfig
       )
     end
-    let(:old_config) { instance_double(Y2Network::Config, dns: double("dns")) }
+    let(:old_config) { instance_double(Y2Network::Config, dns: double("dns"), interfaces: nil) }
```
jreidinger: I expect here more code that tests also that removal of files and other parts that is added.

BTW is only network-ng affected or also master?

mchf: should be network-ng specific, but will check in master too.

PR #877, opened 2019-07-10
AY: Use current sysconfig readers and infer type from config

teclator:
## Problem

1. In case of an Autoinstallation, the configuration is imported and converted by `Lan.FromAY`. The main problem here is that, in case of virtual interfaces, as them are still not present in the system the interface type cannot be inferred from the sysfs. Once the detection from the name was removed recently, at least the interface configuration should be given in order to try to determine the type from them. Otherwise `NetworkInterfaces` imported will be indexed by 'eth' instead of the virtual interface type ('bond', 'br', 'vlan', 'dummy'...)

2. When xen, qemu or kvm packages are installed (and not in a s390 machine), save_network propose a network configuration for virtualization. That is, it creates a bridge when not exist. The problem is that normally the network configuration is done during the second stage and in the moment of proposed is even not importer yet. (Currently a crash when checking the current yast_config.interfaces which does not exist)

## Solution

1. Use the interface type detection passing the ifcfg data.
2. Initialize the list of Config interfaces from sysconfig at least to have a proper InterfacesCollection and not an array.
3. Propose the virtualization configuration after configurin the lan in case of autoinst.

### DIFF
--- a/src/lib/y2network/autoinst/config_reader.rb
+++ b/src/lib/y2network/autoinst/config_reader.rb
@@ -44,7 +44,7 @@ module Y2Network
       # @return [Y2Network::Config] Network configuration
       def config
         attrs = { source: :sysconfig }
-        # FIXME: implement prper readers for AutoYaST
+        # FIXME: implement proper readers for AutoYaST
         attrs[:interfaces] =  interfaces_reader.interfaces
         attrs[:connections] = interfaces_reader.connections
         attrs[:routing] = RoutingReader.new(section.routing).config if section.routing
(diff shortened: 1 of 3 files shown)


src/lib/y2network/autoinst/config_reader.rb
```diff
@@ -43,26 +44,21 @@ def initialize(section)
       # @return [Y2Network::Config] Network configuration
       def config
         attrs = { source: :sysconfig }
-        attrs[:interfaces] = find_interfaces
+        # FIXME: implement prper readers for AutoYaST
```
jreidinger: typo prper

PR #878, opened 2019-07-10
Add a Y2Network::IPAddress class

imobachgs:
This class should solve a problem we have with Ruby stdlib IPAddr class. Details are documented in the class description.

### DIFF
--- a/src/lib/y2network/ip_address.rb
+++ b/src/lib/y2network/ip_address.rb
@@ -1,0 +1,73 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


src/lib/y2network/ip_address.rb
```diff
@@ -0,0 +1,86 @@
...
+  # This class represents an IP address
+  #
+  # The problem with the IPAddr from the Ruby standard library, is that it drops
+  # the host part according to the netmask.
+  #
+  # @example IPAddr from standard library behavior
+  #  ip = IPAddr.new("192.168.122.1/24")
+  #  ip.to_s #=> "192.168.122.0/24"
+  #
+  # However, what we need is to be able to keep the host part
```
jreidinger: please document why we need it to keep it.

teclator:
It looks OK to me.

PR #879, opened 2019-07-10
fix default value for bond options

jreidinger:


src/lib/y2network/interface_config_builders/bond.rb
```diff
@@ -19,10 +19,22 @@ def bondable_interfaces
...
+          @bond_options = @config["BONDING_MODULE_OPTS"]
+        end
+        @bond_options
+      end
+
       def save
         super
 
         Yast::LanItems.bond_slaves = @config["BOND_SLAVES"]
+        Yast::LanItems.bond_option = bond_options
```
teclator: It looks tricky to me that one reads from @config and the other depends on if it is in LanItems or not..

jreidinger: in the end they are same, both are read value or modified one. I just do not create shortcut for bond_slaves for now.

teclator:
LGTM, there is not BONDOPTION now in the generated ifcfg file

PR #880, opened 2019-07-10
Network ng

jreidinger:
trello: https://trello.com/c/IwYpz97V/1097-test-and-merge-into-master

The entry point for the documentation should be [this one](https://github.com/yast/yast-network/tree/network-ng#documentation). However, until merged, links won't work. However, the diagram is a nice starting point.

## The data model

We are developing a data model which enables us to read/write the network configuration using different backends (at this time, only a `sysconfig` backend is available). You can read more about it in [network-ng.md](https://github.com/yast/yast-network/blob/network-ng/doc/network-ng.md).

## Interface configuration builders

In the old code, there was no clear separation between UI and business logic. In order to improve the situation, we introduced the concept of [interface configuration builders](https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/interface_config_builder.rb).

We already have implemented support for several interface types. You can find them under the [Y2Network::InterfaceConfigBuilders namespace](https://github.com/yast/yast-network/tree/network-ng/src/lib/y2network/interface_config_builders).

## Widgets

General workflow is that Interface sequences drives sequences of dialogs how they should follow up. Dialogs then contain widgets usually grouped in tabs. Content of dialog is usually heavily dependent on interface type.

* Sequences:
  *  [Sequences::Interface](https://github.com/yast/yast-network/blob/358bcd13b4e92e7c4e9c0e477c83196ca67b578e/src/lib/y2network/sequences/interface.rb)
* Dialogs:
  * [Dialogs::AddInterface](https://github.com/yast/yast-network/blob/358bcd13b4e92e7c4e9c0e477c83196ca67b578e/src/lib/y2network/dialogs/add_interface.rb)
  * [Dialogs::EditInterface](https://github.com/yast/yast-network/blob/358bcd13b4e92e7c4e9c0e477c83196ca67b578e/src/lib/y2network/dialogs/edit_interface.rb)
* [Y2Network::Widgets](https://github.com/yast/yast-network/tree/358bcd13b4e92e7c4e9c0e477c83196ca67b578e/src/lib/y2network/widgets)

## What is this PR about

This PR merges quite a lot of logic related to interface configuration builders and the UI. During this sprint we have been working in the data model too ([reading/writing interfaces configuration](https://github.com/yast/yast-network/pull/875)), but we do not want to merge it into master yet.


src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -0,0 +1,141 @@
...
+Yast.import "NetworkInterfaces"
+
+module Y2Network
+  module Sysconfig
+    # This class reads interfaces configuration from sysconfig files
+    #
+    # * Physical interfaces are read from the hardware.
+    # * Virtual interfaces + Connections are read from sysconfig.
+    #
+    # @see Y2Network::Interface
```
mvidner:
```suggestion
    # @see Y2Network::InterfacesCollection
```

src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -0,0 +1,141 @@
...
+    # @see Y2Network::Connection::Connection
```
mvidner:
```suggestion
    # @see Y2Network::ConnectionConfig
```

src/lib/y2network/fake_interface.rb
```diff
@@ -0,0 +1,38 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/interface"
+
+module Y2Network
+  # This class represents an interface which does not exists in the system
```
mvidner: Then why does it exist here? For example:
> It's because each {Foo} must have an {Interface} so when {Gandalf} does not provide one we use a {FakeInterface}

src/lib/y2network/interface_config_builder.rb
```diff
@@ -0,0 +1,313 @@
...
+require "y2firewall/firewalld"
+require "y2firewall/firewalld/interface"
+
+Yast.import "LanItems"
+Yast.import "NetworkInterfaces"
+
+module Y2Network
+  # Collects data from the UI until we have enough of it to create a {Y2Network::Interface}.
+  # {Yast::LanItemsClass#Commit Yast::LanItems.Commit(builder)} uses it.
+  class InterfaceConfigBuilder
```
mvidner: Maybe nitpicking, but shouldn't it be called `InterfaceBuilder` then? Because a `ConnectionConfigBuilder` builds a `ConnectionConfig`, not a `Connection`.

src/lib/y2network/sysconfig/config_reader.rb
```diff
@@ -22,44 +22,47 @@
...
+      include Yast::Logger
+
       def initialize(_opts = {})
       end
 
       # @return [Y2Network::Config] Network configuration
       def config
-        interfaces = find_interfaces
-        routing_tables = find_routing_tables(interfaces)
+        routing_tables = find_routing_tables(interfaces_reader.interfaces)
```
mvidner: Following the naming patterns, I was surprised not to find a Sysconfig::RoutingReader class. Should it exist? 

Same for Sysconfig::RoutingWriter.

src/lib/y2network/sysconfig/connection_config_readers/eth.rb
```diff
@@ -0,0 +1,49 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/ethernet"
+
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigReaders
+      # This class is able to build a ConnectionConfig::Ethernet object given a
+      # Sysconfig::InterfaceFile object.
+      class Eth
```
mvidner: `class Ethernet` for consistency with the ConnectionConfig namespace?

src/lib/y2network/sysconfig/connection_config_readers/wlan.rb
```diff
@@ -0,0 +1,61 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/wireless"
+
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigReaders
+      # This class is able to build a ConnectionConfig::Wireless object given a
+      # Sysconfig::InterfaceFile object.
+      class Wlan
```
mvidner: `class Wireless` for consistency with the ConnectionConfig namespace?

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -0,0 +1,193 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "pathname"
+require "ipaddr"
+
+module Y2Network
+  module Sysconfig
+    # This class represents a sysconfig file containing an interface configuration
```
mvidner: An InterfaceFile does not know its file path, unlike a RoutesFile. Maybe it should be made consistent?

src/lib/y2network/config_writer.rb
```diff
@@ -23,9 +23,9 @@ module ConfigWriter
     # @param source [Symbol] Source name (e.g., :sysconfig)
     # @return [#config] Configuration writer from {Y2Network::ConfigReader}
```
mvidner:
```suggestion
     # @return [Y2Network::Sysconfig::ConfigWriter]
```

also note that Y2Network::Autoinst::ConfigWriter does not exist

src/lib/y2network/config_reader.rb
```diff
@@ -28,9 +28,9 @@ module ConfigReader
     # @param opts   [Hash] Reader options
     # @return [#config] Configuration reader from {Y2Network::ConfigReader}
```
mvidner:
```suggestion
     # @return [Y2Network::Autoinst::ConfigReader,Y2Network::Sysconfig::ConfigReader]
```
also the class comment above is outdated.

src/lib/y2network/interface.rb
```diff
@@ -16,17 +16,50 @@
...
   class Interface
-    # @return [String] Device name (eth0, wlan0, etc.)
+    # @return [String] Device name ('eth0', 'wlan0', etc.)
     attr_accessor :name
+    # @return [String] Interface description
+    attr_accessor :description
+    # @return [Symbol] Interface type
+    attr_accessor :type
+    attr_reader :configured
+    attr_reader :hardware
```
mvidner:
```suggestion
    # @return [HwInfo]
    attr_reader :hardware
```

src/lib/y2network/interface.rb
```diff
@@ -16,17 +16,50 @@
...
   # Network interface.
   class Interface
-    # @return [String] Device name (eth0, wlan0, etc.)
+    # @return [String] Device name ('eth0', 'wlan0', etc.)
     attr_accessor :name
+    # @return [String] Interface description
+    attr_accessor :description
+    # @return [Symbol] Interface type
+    attr_accessor :type
+    attr_reader :configured
```
mvidner:
```suggestion
    # @return [Boolean,nil]
    attr_reader :configured
```

`init` allows a nil value, which I think is really a minor type bug

src/lib/y2network/fake_interface.rb
jreidinger: Main reason is hot-plug devices. So it is physical devices that is not physically attached. I am not sure about virtual devices built from fake one, if it is also fake or not. ( I expect yes, as it is not on system till underlaying device is attached )

src/lib/y2network/sysconfig/connection_config_readers/eth.rb
jreidinger: we plan to address it in this sprint with unified InterfaceType everywhere which will have method for consistent names.

src/lib/y2network/sysconfig/connection_config_readers/wlan.rb
jreidinger: Same as above.

src/lib/y2network/fake_interface.rb
mvidner:
```suggestion
  # A physical interface that is not plugged in.
```

src/autoyast-rnc/networking.rnc
```diff
@@ -280,15 +280,13 @@ gateway = element gateway { text }
 dns =
   element dns {
     dhcp_hostname ? &
-    write_hostname ? &
```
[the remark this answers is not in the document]
teclator: So, we should mark it as deprecated and just ignore by now

src/lib/y2network/sysconfig/interface_file.rb
imobachgs: Yes, you are right. Actually, I am doing a heavy refactoring on InterfaceFile to act more like the RoutesFile, including `#load` and `#path` methods as part of [adding support for multiple IPs configurations](https://trello.com/c/sNisjEg2/1117-2-support-multiple-ips-configuration).

src/lib/y2network/sysconfig/connection_config_readers/eth.rb
jreidinger: addressed in https://github.com/yast/yast-network/pull/888 I think we should postpone this change as it is quite big and can invalidate testing we did.

PR #885, opened 2019-07-12
Improve and update network-ng.md document

imobachgs:
Merge after #28 

It updates and improves the `network-ng.md` document.

* Update current status.
* Add a plan.
* Add information about the UI layer and interface builders.

### DIFF
--- a/doc/network-ng.md
+++ b/doc/network-ng.md
@@ -3,7 +3,9 @@
 This document tries to describe the details of the network-ng data model and how all the pieces are
 combined.
 
-## Overall Approach
+## Data Model
+
+### Overall Approach
 
 {Y2Network::Config} and other associated classes represent the network configuration in a backend
 agnostic way. It should not matter whether you are using `sysconfig` files, NetworkManager or
@@ -25,7 +27,7 @@ filesystem by using a *configuration writer*.
 Obviously, we should implement configuration readers and writers for whathever backend we would like
 to support. At this point of time, only `Sysconfig` and `Autoinst` are supported.
 
-## The Configuration Classes
+### The Configuration Classes
 
 {Y2Network::Config} offers and API to deal with network configuration, but it collaborates with
 other classes.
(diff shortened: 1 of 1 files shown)


doc/network-ng.md
```diff
@@ -38,31 +40,77 @@ other classes.
...
     src/lib/y2network/sysconfig
     ├── config_reader.rb <- READER
     ├── config_writer.rb <- WRITER
-    ├── connection_config_reader_handlers
-    │   ├── eth.rb
-    │   ├── wlan.rb
-    │   └── ...
     ├── connection_config_reader.rb
+    ├── connection_config_readers
+    │   ├── eth.rb
```
jreidinger: just curios is it eth or according to new types ethernet? just that we should be consistent.

doc/network-ng.md
```diff
@@ -38,31 +40,77 @@ other classes.
...
-## Accessing the Configuration
+Last but not least, there are additional classes like {Y2Network::Sysconfig::RoutesFile} and
+{Y2Network::Sysconfig::InterfaceFile} which abstract the details of reading/writing `ifcfg` files.
 
-The `Yast::Lan` module is still the entry point to read and write the network configuration. Basically, it keeps two configuration objects, one for the running system and another want for the wanted configuration.
+#### AutoYaST
+
+AutoYaST is a special case in the sense that it reads the information from a profile, instead of
+using the running system as reference. Additionally, it does not implement a writer because the
+configuration will be written using a different backend (like sysconfig).
```
jreidinger: cannot we have writter for clonning and so writting profile?

doc/network-ng.md
```diff
@@ -101,3 +170,9 @@ reader and will write it to the final system using the one corresponding to the
 | USB             |      |       |
 | Dummy           |      |       |
 | s390 types      |      |       |
+
+## (Short Term) Plan
+
+- [ ] Finish reading/writing interfaces
+- [ ] Move missing UI logic to interface configuration builders
+- [ ] Replace LanItems with the new data model
```
jreidinger: maybe also use that new type class everywhere. So everything use same class to represent types. I think now we need in short term focus on unifying to existing new solutions like that new type class.

imobachgs:
Included in #875.

PR #886, opened 2019-07-12
Add aliases to the builder sysconfig devmap

teclator:
## Problem

When modified the list of aliases through the builder, the new devmap is overwritten by the device_sysconfig which is not exporting the aliases.

- https://github.com/yast/yast-network/blob/network-ng/src/modules/LanItems.rb#L1723
- https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/interface_config_builder.rb#L170

Previously save has updated the NetworkInterfaces.Current and LanItems with the set of aliases but that is not enough while we still uses the newdev to modify the changes

## Solution

Export also the aliases by device_sysconfig.

### DIFF
--- a/src/lib/y2network/widgets/additional_addresses.rb
+++ b/src/lib/y2network/widgets/additional_addresses.rb
@@ -86,6 +86,16 @@ module Y2Network
         end
 
         Yast::UI.ChangeWidget(Id(:address_table), :Items, table_items)
+        Yast::UI.ChangeWidget(
+          Id(:edit_address),
+          :Enabled,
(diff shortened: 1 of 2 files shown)


src/lib/y2network/widgets/additional_addresses.rb
```diff
@@ -86,6 +86,16 @@ def refresh_table
         end
 
         Yast::UI.ChangeWidget(Id(:address_table), :Items, table_items)
```
teclator: Otherwise when initialized with aliases it cannot be edited or deleted.

PR #887, opened 2019-07-15
Added missing s390 interfaces

teclator:


src/lib/y2network/interface_type.rb
```diff
@@ -84,5 +84,9 @@ def to_human_string
     define_type "TUN", N_("TUN"), "tun"
     define_type "TAP", N_("TAP"), "tap"
     define_type "USB", N_("USB"), "usb"
+    # s390
+    define_type "QETH", N_("QETH"), "qeth"
+    define_type "LCS", N_("LCS"), "lcs"
+    define_type "HIPERSOCKETS", N_("HiperSockets"), "hsi"
```
jreidinger: I worry this is not all what we currently support. See https://github.com/yast/yast-yast2/blob/master/library/network/src/modules/NetworkInterfaces.rb#L1019

teclator: @jreidinger I added by now the ones I now are supported as some of them the support has been already removed.

https://www.suse.com/documentation/sles-15/book_sle_admin/data/sec_network_yast.html#sec_network_yast_s390_ctc

teclator: And `ctc_configure` is also deprecated which is still used for lcs interfaces.

teclator: Added FICON, but ESCON, CTC and IUCV are not officially supported anymore:

https://www.suse.com/documentation/sles-15/singlehtml/book_sle_deployment/book_sle_deployment.html#sec.zseries.requirements.hw.network

PR #888, opened 2019-07-15
Use InterfaceType everywhere

jreidinger:


src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -108,7 +108,7 @@ def build_physical_interface(data)
         Y2Network::PhysicalInterface.new(data["dev_name"]).tap do |iface|
           iface.description = data["name"]
           type = data["type"] || Yast::NetworkInterfaces.GetTypeFromSysfs(iface.name)
-          iface.type = type.nil? ? :eth : type.to_sym
+          iface.type = type.nil? ? "eth" : type
```
teclator: Shouldn't it be an `InterfaceType`?

teclator: At least according to https://github.com/yast/yast-network/pull/888/files#diff-c8c2a4d640ee3258722656a5a52190f7R30

jreidinger: you are right.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -158,9 +158,9 @@ def fetch(key)
       #
       # @todo Borrow logic from https://github.com/yast/yast-yast2/blob/6f7a789d00cd03adf62e00da34720f326f0e0633/library/network/src/modules/NetworkInterfaces.rb#L291
       #
-      # @return [Symbol] Interface's type depending on the file values
+      # @return [String] Interface's type depending on the file values
       def type
-        :eth
+        "eth"
```
mchf: why is type a string here?

src/lib/y2network/sysconfig/interfaces_reader.rb
mchf: ```Interface::type``` should be an ```InterfaceType``` not string, shouldn't it?

test/default_route_test.rb
```diff
@@ -101,9 +102,8 @@
     subject.Read
     subject.current = 0
 
-    builder = Y2Network::InterfaceConfigBuilder.new
+    builder = Y2Network::InterfaceConfigBuilder.for("eth")
```
mchf: well I know that ```InterfaceConfigBuilder::for``` handles even strings, I'm just curious if using string here was an intention.

src/lib/y2network/sysconfig/interface_file.rb
jreidinger: well, reason is that in this case it can be also read from file, so I use value that can be specified in file as INTERFACETYPE . Or do you think it is bad approach?

src/lib/y2network/sysconfig/interfaces_reader.rb
jreidinger: right as knut pointed above.

test/default_route_test.rb
jreidinger: well, I start on same places using `builder = Y2Network::InterfaceConfigBuilder.for(Y2Network::InterfaceType::ETHERNET)` but it is incredible long with all that scoping.

src/lib/y2network/sysconfig/interface_file.rb
mchf: slightly misleading in the context of this pr, but makes sense in this case ;-)

mchf:
fine with me

src/lib/y2network/sysconfig/interface_file.rb
jreidinger: well, I need to change it as there is now basically only two representation short_name as string and InterfaceType as class. So no symbols anymore.

PR #889, opened 2019-07-16
Classes for bonding / bridge interfaces

mchf:


src/lib/y2network/interface.rb
```diff
@@ -49,6 +52,19 @@ class Interface
...
     end
 
+    # Creates an instance of interface class based on the given interface type
+    #
+    # @param name [String] interface name (eth0, ...)
+    # @param type [Y2Network::InterfaceType] interface type (InterfaceType::ETHERNET, ...)
+    def self.for(name, type)
+      type_name = type.short_name
+      require "y2network/interfaces/#{type_name}_interface"
+      Interfaces.const_get(type_name.to_s.capitalize + "Interface").new(name, type: type)
```
jreidinger: I suggest to wait a bit or base it on my interface_type pr, which adds methods for file_name and class_name. Also it is inconsistent to others which does not have suffix as namespace is enough. So ideally it should be Ethernet class and not EthernetInterface class.

jreidinger: also I think it would be nice to support also type as short name string.

src/lib/y2network/interface.rb
```diff
@@ -49,6 +52,19 @@ class Interface
...
+    rescue LoadError => e
+      log.info "Specialed interface class for #{type.short_name} not found. Fallbacking to default. #{e.inspect}"
```
jreidinger: typo specialized

src/lib/y2network/interface_config_builders/bond.rb
```diff
@@ -16,7 +16,7 @@ def initialize
 
       # @return [Array<Interface>] list of interfaces usable for the bond device
       def bondable_interfaces
-        interfaces.all.select { |i| bondable?(i) }
+        interfaces.all.to_a.select { |i| bondable?(i) }
```
jreidinger: why to_a is needed?

src/lib/y2network/interfaces/bond_interface.rb
```diff
@@ -0,0 +1,51 @@
...
+
+require "y2network/interfaces/virtual_interface"
+
+module Y2Network
+  module Interfaces
+    class BondInterface < VirtualInterface
+      # Returns list of interfaces enslaved in the bond interface
+      #
+      # @return [Y2Network::InterfacesCollection] interfaces enslaved in the bond
+      # TODO: reading from backend should be handled elsewhere
```
jreidinger: yep, it should be done by reader or created by builder.

src/lib/y2network/interfaces/bond_interface.rb
```diff
@@ -0,0 +1,51 @@
...
+    private
+
+      # Creates list of devices enslaved in the bond device.
+      #
+      # @param bond_iface [String] a name of an interface of bond type
+      # @return list of names of interfaces enslaved in the bond_iface
+      def bond_slaves(bond_iface)
+        bond_map = Yast::NetworkInterfaces::FilterDevices("netcard").fetch("bond", {}).fetch(bond_iface, {})
+
+        bond_map.select { |k, _| k.start_with?("BONDING_SLAVE") }.values
```
jreidinger: what happens if there are two bonds? It is real case in our OPS, when there are two cards to internal and two cards to external network for VPN server. or is it already selected above by that FilterDevices? But then why is needed also this select?

src/lib/y2network/interfaces/br_interface.rb
```diff
@@ -0,0 +1,50 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/interfaces/virtual_interface"
+
+module Y2Network
+  module Interfaces
+    class BrInterface < VirtualInterface
```
jreidinger: why Br? It should be Bridge, not?

src/lib/y2network/interfaces/bond_interface.rb
```diff
@@ -0,0 +1,51 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/interfaces/virtual_interface"
+
+module Y2Network
+  module Interfaces
+    class BondInterface < VirtualInterface
```
jreidinger: Bonding?

src/lib/y2network/interfaces/fake_interface.rb
```diff
@@ -20,18 +20,20 @@
...
+    class FakeInterface < Interface
+      class << self
+        # Build connection
+        #
+        # @todo Would be possible to get the name from the connection?
+        #
+        # @param name [String] Interface name
+        # @param conn [ConnectionConfig] Connection configuration related to the
+        #   network interface
+        def from_connection(name, conn)
```
jreidinger: one important thing, is virtual interface build on top of fake interface also fake or not?

src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -105,7 +105,7 @@ def find_connections
       # @option data [String] "name"     Device description
       # @option data [String] "type"     Device type ("eth", "wlan", etc.)
       def build_physical_interface(data)
-        Y2Network::PhysicalInterface.new(data["dev_name"]).tap do |iface|
+        Y2Network::Interfaces::PhysicalInterface.new(data["dev_name"]).tap do |iface|
```
jreidinger: hmm, how this will wotk with that specialization? e.g. if we have specialized wlan one?

src/lib/y2network/interfaces/bond_interface.rb
mchf: selection above selects particular bond device (see last fetch)

mchf: select is needed bcs bonding slaves are defined by options "BONDING_SLAVE0, BONDING_SLAVE1, ..." It differs from bridge where you have "BRIDGE_PORTS" with space separated list of ports / slaves.

src/lib/y2network/interfaces/bond_interface.rb
mchf: same reason as with bridge

src/lib/y2network/sysconfig/interfaces_reader.rb
mchf: I don't know and I didn't check this code in depth yet. I think it is in a prototype state now. It also is not creating virtual devices yet.

src/lib/y2network/interfaces/fake_interface.rb
mchf: well ... e.g. bond device will be created by wicked as long as it has at least one "non fake" slave (e.g. eht1 - unconfigured physical device, eth3 - fake interface). But will fail when slave(s) are only fake ones. So, It is somehow in between but I tend to see it as non fake device with kind of invalid configuration when only fake slaves are used.

src/lib/y2network/interfaces/bonding.rb
```diff
@@ -0,0 +1,51 @@
...
+
+require "y2network/interfaces/virtual_interface"
+
+module Y2Network
+  module Interfaces
+    class Bonding < VirtualInterface
+      # Returns list of interfaces enslaved in the bond interface
+      #
+      # @return [Y2Network::InterfacesCollection] interfaces enslaved in the bond
+      # TODO: reading from backend should be handled elsewhere
```
jreidinger: yeah, big topic. Should connection config knows it? As you can change slaves in UI, then I think using connection config makes sense for holding information about slaves. @teclator what do you think? Also what is order of dependency? connection_config depends on interface or vice versa?

mchf: well, this was intended as temporary solution for isolating code from lanitems and using new API in the rest of the code. So, it is not perfect. I'd say that reading slaves should be in connections. In fact I'm slowly working on it as a part of https://trello.com/c/6CBQadMq/1146-2-infer-interfaces-type-correctly-when-reading-them however I don't have working prototype yet.

PR #891, opened 2019-07-18
add startmode and boot protocol object

jreidinger:


mchf:
hm seems to be a bit overkill and netconfig centric ... but i have to read it in more depth

jreidinger:
@mchf well, I discuss it also with knut how it looks like on NM and there is also something like startmode, but they support different values ( only autoconnect or not ). So in future I expect that we need also to have a flag for each class what backend it supports.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -129,6 +133,34 @@ def firewall_zone=(value)
...
+    end
+
+    def startmode=(name)
+      mode = Startmode.create(name)
+      # assign only if it is not already this value. This helps with ordering of ifplugd_priority
+      @connection_config.startmode = mode if @connection_config.startmode.name != mode.name
+      @config["STARTMODE"] = mode.name
+    end
+
+    def ifplugd_priority=(value)
```
imobachgs: Please, document (at least the @param). Same for `startmode=` and `ifplugd_priority`

src/lib/y2network/startmodes/auto.rb
```diff
@@ -0,0 +1,38 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/startmode"
+
+module Y2Network
+  module Startmodes
+    class Auto < Startmode
```
imobachgs: Please, document. I know that for us the meaning is clear, but I am thinking about newcomers. And the same for all startmode classes.

imobachgs:
Oh, wait, you need to solve the conflicts first.

PR #892, opened 2019-07-19
Avoid cycle

jreidinger:


package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Fri Jul 19 08:10:14 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- avoid dependency on yast2-installation
```
teclator: A bug number will probably be needed

mchf: also it is not dependency on yast2-installation

jreidinger: well, it is not accepted to target project, it should work without it

teclator: OK, so then fine with me

teclator:
```suggestion
- avoid dependency on autoyast2-installation
```

PR #893, opened 2019-07-19
Network ng multiple ips

imobachgs:
* IP related settings are stored in a new class `IPConfig`.
* A connection config can have several IPConfig objects.
* InterfaceFile has a CFA-like interface now (although it is still using SCR). Moreover, now it is able to handle suffixed variables. See (5) ifcfg for further information.

~As soon as #875 is merged, this PR's target should be changed to `network-ng`~

## TODO

- [x] Support to read NETMASK from ifcfg-* files
- [x] Write IPADDR_* and related stuff


src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,52 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+module Y2Network
+  module ConnectionConfig
+    class IPConfig
+      # @return [IPAddress] IP address
+      attr_accessor :address
+      # @return [Symbol] Address scope (:global, :site, :link or :host)
+      attr_accessor :scope
```
jreidinger: what is this value in old implementation?

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,52 @@
...
+      # @return [IPAddress,nil] Remote IP address of a point to point connection
+      attr_accessor :remote_address
+      # @return [IPAddress,nil] Broadcast address
+      attr_accessor :broadcast
+
+      # Constructor
+      #
+      # @param address [IPAddress]
+      # @param id      [Symbol,String] ID (needed for sysconfig backend in order to write suffixes in
+      #   ifcfg-* files)
```
jreidinger: cannot we simply generate ids ourself? is only requirement to be unique or something else?

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,52 @@
...
+      # @param scope   [Symbol]
+      # @param label   [String,nil]
+      # @param remote_address [IPaddress,nil]
+      # @param broadcast [IPaddress,nil]
+      def initialize(address:, id: nil, scope: nil, label: nil, remote_address: nil, broadcast: nil)
```
jreidinger: is needed to be first param also named? maybe forcing it to be the first one is not bad.

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
```diff
@@ -17,31 +17,22 @@
...
-
+      class Ethernet < Base
         # @return [Y2Network::ConnectionConfig::Ethernet]
         def connection_config
           Y2Network::ConnectionConfig::Ethernet.new.tap do |conn|
             conn.bootproto = file.bootproto
             conn.description = file.name
             conn.interface = file.interface
-            conn.ip_address = file.ip_address
+            conn.ip_configs = ip_configs
```
jreidinger: hmm, considering  that in future maybe more shared stuff will be there. what about protected method that take conn and file and assign it?

I expect it can already read bootproto, description, interface and those ip_configs. as it is probably for all interfaces.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -44,144 +63,167 @@ def find(interface)
...
-          @params ||= []
+        def variables
+          @variables ||= {}
+        end
+
+        # Parameter name to internal variable name
+        #
+        # @param param_name [Symbol]
+        # @return [String] Convert a parameter name to the expected internal name
+        def variable_name(param_name)
```
jreidinger: private method?

src/lib/y2network/connection_config/ip_config.rb
imobachgs: It is slightly tricky (actually, I would rather prefer to drop this `id`). Think about this case (these entries are written by the user):

```ini
IPADDR_SOMEID='192.168.122.1/24'
LABEL_SOMEID='my-label'
```

In the current version, this "alias" is shown in the interface, but it is not handled properly. This is the result when the settings are written:

```ini
IPADDR_SOMEID='192.168.122.1/24'
LABEL_SOMEID='my-label'
IPADDR_1='192.168.122.1/24'
LABEL_1='my-label'
```

Nice, uh? So the idea of the `id` is to keep the `_SOMEID` part, so we do not create duplicated entries. However, I was thinking about dropping it and just rewriting the whole collection like:

```ini
IPADDR_1='192.168.122.1/24'
LABEL_1='my-label'
```
We would drop the "user defined" suffix, but I think it is acceptable. What's your opinion?

jreidinger: not sure as I think some customers will complain if they use some meaningful suffix. We will need that generator if we want to hide this suffix stuff from user, but we can generate it elsewhere and not in this code.

src/lib/y2network/connection_config/ip_config.rb
teclator: I think it is out of scope and nothing we should deal with..

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -200,35 +246,24 @@ def path
...
-      # Fetches a key
+      # Loads values from the configuration file
       #
-      # @param key [String] Interface key
-      # @return [Object] Value for the given key
-      def fetch(key)
-        path = Yast::Path.new(".network.value.\"#{interface}\".#{key}")
-        Yast::SCR.Read(path)
+      # @return [Hash<String, Object>] All values from the file
+      def load
```
teclator: So the load method loads all the attributes even if it does not belongs to an specific type isn't it?. Thus, inferring the type from the current values is harder as there is no distinction between present or not.

teclator: Can be inferred from `@defined_variables so forget what I said :)

src/lib/y2network/connection_config/ip_config.rb
imobachgs: In such a case, I fear that we should keep the `id` as it is by now (read from the ifcfg-* files).

src/lib/y2network/connection_config/ip_config.rb
imobachgs: OK, dropped.

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
imobachgs: ip_configs is already a method (from `Base` class) and the `file` is available as an instance variable.

src/lib/y2network/connection_config/base.rb
```diff
@@ -38,17 +38,20 @@ class Base
       attr_accessor :interface
       # @return [BootProtocol] Bootproto
       attr_accessor :bootproto
-      # @return [IPAddr,nil]
-      attr_accessor :ip_address
-      # @return [Array<IPAddr>]
-      attr_accessor :secondary_ip_addresses
+      # @return [Array<IPConfig>]
+      attr_accessor :ip_configs
```
jreidinger: maybe some helper for default would be useful? as current UI use only default, not?

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,50 @@
...
+    class IPConfig
+      # @return [IPAddress] IP address
+      attr_accessor :address
+      # @return [String,nil] Address label
+      attr_accessor :label
+      # @return [IPAddress,nil] Remote IP address of a point to point connection
+      attr_accessor :remote_address
+      # @return [IPAddress,nil] Broadcast address
+      attr_accessor :broadcast
+      # @return [Symbol,String] ID (needed for sysconfig backend in order to write suffixes in
```
jreidinger: having option to pass in constructor symbol or string is fine, but I think we should convert it and use it always same here. Symbol only?

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,50 @@
...
+      #
+      # @param address [IPAddress]
+      # @param id      [Symbol,String] ID (needed for sysconfig backend in order to write suffixes in
+      #   ifcfg-* files)
+      # @param label   [String,nil]
+      # @param remote_address [IPaddress,nil]
+      # @param broadcast [IPaddress,nil]
+      def initialize(address, id: nil, label: nil, remote_address: nil, broadcast: nil)
+        @address = address
+        @id = id || :default
```
jreidinger: as default is a bit specific, I would like to see constant for it and maybe also ask method like `default?`. Also why not having in params `id: DEFAULT` directly? as even documentation does not mention nil as allowed value neither document that it will be default if nothing is specified.

src/lib/y2network/ip_address.rb
```diff
@@ -67,12 +62,18 @@ def from_string(str)
     def initialize(address, prefix = nil)
       @address = IPAddr.new(address)
       @prefix = prefix
-      @prefix ||= @address.ipv4? ? IPV4_DEFAULT_PREFIX : IPV6_DEFAULT_PREFIX
     end
 
     # Returns a string representation of the address
     def to_s
-      host? ? @address.to_s : "#{@address}/#{@prefix}"
+      prefix? ? "#{@address}/#{@prefix}" : @address.to_s
```
jreidinger: maybe also document that output can omit prefix if prefix is not defined in constructor. Ideally in example in top level class documentation.

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
jreidinger: this is actually not what I mean. I mean that you repeat in all specialized class stuff like
```
conn.bootproto = BootProtocol.from_name(file.bootproto || "static")
￼conn.description = file.name
conn.interface = file.interface
conn.ip_configs = ip_configs
conn.startmode = Startmode.create(file.startmode || "manual")
conn.startmode.priority = file.ifplugd_priority if conn.startmode.name == "ifplugd"
```

so I expect it should be in that base class and onlly class specific stuff in this config constructor.

src/lib/y2network/sysconfig/connection_config_writers/ethernet.rb
```diff
@@ -17,28 +17,23 @@
...
         # Writes connection information to the interface configuration file
         #
         # @param conn [Y2Network::ConnectionConfig::Base] Configuration to write
         def write(conn)
           file.bootproto = conn.bootproto.name
-          file.ipaddr = conn.ip_address
           file.name = conn.description
           file.startmode = conn.startmode.to_s
           file.ifplugd_priority = conn.startmode.priority if conn.startmode.name == "ifplugd"
+          write_ip_configs(conn.ip_configs)
```
jreidinger: why not simple super that do all shared writtings like that ip_configs and probably also startmore or name?

src/lib/y2network/connection_config/base.rb
imobachgs: Yes, that's true. I will add it the subsequent PR so we can merge this one as soon as possible (@teclator is working on top of this moving branch).

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -35,8 +38,41 @@ def initialize(file)
...
+
       private
 
+        # Returns the class of the connection configuration
+        #
+        # @note This method should be redefined by derived classes.
+        #
+        # @return [Class]
+        def connection_class
+          raise NotImplementedError
```
jreidinger: btw you can easily compute it from self.class. So you can have something like 
```
class_name = self.class.to_s[/::([^:])*$/, 1]
Y2Network::ConnectionConfig.const_get(class_name)
```

this avoids to redefine it everywhere.

src/lib/y2network/sysconfig/connection_config_writers/base.rb
```diff
@@ -33,8 +33,38 @@ def initialize(file)
...
+
       private
 
+        # Returns the class of the connection configuration
+        #
+        # @note This method should be redefined by derived classes.
+        #
+        # @return [Class]
+        def connection_class
+          raise NotImplementedError
```
jreidinger: same here ;)

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -89,6 +89,7 @@ def ip_configs
               broadcast:      file.broadcasts[id]
             )
           end
+          configs.sort_by { |c| c.id.nil? ? -1 : 0 }
```
jreidinger: please document in methot that it returns default one first.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -330,7 +330,7 @@ def fetch_scalar(key, type)
       # @return [Hash<String, Object>]
       def fetch_collection(key, type)
         collection_keys(key).each_with_object({}) do |k, h|
-          index = key == k ? :default : k.sub(key, "")
+          index = k.sub(key, "")
```
jreidinger: so here suffix with be "" or nil? as nil will cause crash

test/y2network/sysconfig/connection_config_writers/ethernet_test.rb
```diff
@@ -48,12 +48,12 @@ def file_content(scr_root, file)
   let(:ip_configs) do
     [
       Y2Network::ConnectionConfig::IPConfig.new(
-        Y2Network::IPAddress.from_string("192.168.122.1/24"), id:        :default,
-                                                              broadcast: Y2Network::IPAddress.from_string("192.168.122.255")
+        Y2Network::IPAddress.from_string("192.168.122.1/24"),
+        id: "", broadcast: Y2Network::IPAddress.from_string("192.168.122.255")
```
jreidinger: now I am a bit lost, default is "" or nil?

imobachgs: I've decided to go for an empty string.

src/lib/y2network/sysconfig/connection_config_writers/base.rb
imobachgs: Actually, this one is not even needed :)

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -0,0 +1,112 @@
...
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/ip_config"
+require "y2network/ip_address"
```
teclator:
```suggestion
require "y2network/ip_address"
require "y2network/boot_protocol"
require "y2network/startmode"
```

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -0,0 +1,112 @@
...
+          end
+        end
+
+      private
+
+        # Returns the class of the connection configuration
+        #
+        # @return [Class]
+        def connection_class
+          self.class.to_s[/::([^:]+)$/, 1]
```
teclator:
```suggestion
          class_name = self.class.to_s[/::([^:]+)$/, 1]
```

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
```diff
@@ -17,6 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "y2network/sysconfig/connection_config_readers/base"
 require "y2network/connection_config/ethernet"
 require "y2network/boot_protocol"
```
teclator: Added to base reader

src/lib/y2network/sysconfig/connection_config_readers/wireless.rb
```diff
@@ -17,6 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "y2network/sysconfig/connection_config_readers/base"
 require "y2network/connection_config/wireless"
 require "y2network/boot_protocol"
```
teclator: Same than above

src/lib/y2network/sysconfig/connection_config_readers/wireless.rb
```diff
@@ -26,39 +27,38 @@ module Sysconfig
...
       # Sysconfig::InterfaceFile object.
-      class Wireless
-        # @return [Y2Network::Sysconfig::InterfaceFile]
-        attr_reader :file
+      class Wireless < Base
+      private
 
-        def initialize(file)
-          @file = file
+        # @return [Y2Network::ConnectionConfig::Ethernet]
```
teclator:
```suggestion
        # @return [Y2Network::ConnectionConfig::Wireless
```

src/lib/y2network/sysconfig/connection_config_readers/base.rb
jreidinger: hmm and test passed, that is strange. It should be covered.

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
imobachgs: It may be that I am confused, but I see this line in my code. And same below.

src/lib/y2network/sysconfig/connection_config_readers/base.rb
imobachgs: I forgot to drop the old methods :man_facepalming:

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,50 @@
...
+    class IPConfig
+      # @return [IPAddress] IP address
+      attr_accessor :address
+      # @return [String,nil] Address label
+      attr_accessor :label
+      # @return [IPAddress,nil] Remote IP address of a point to point connection
+      attr_accessor :remote_address
+      # @return [IPAddress,nil] Broadcast address
+      attr_accessor :broadcast
+      # @return [String,nil] ID (needed for sysconfig backend in order to write suffixes in
```
jreidinger: if default is "" then I would avoid nil here.

src/lib/y2network/connection_config/ip_config.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      # Constructor
+      #
+      # @param address [IPAddress]
+      # @param id      [String] ID (needed for sysconfig backend in order to write suffixes in
+      #   ifcfg-* files)
+      # @param label   [String,nil]
+      # @param remote_address [IPaddress,nil]
+      # @param broadcast [IPaddress,nil]
+      def initialize(address, id: nil, label: nil, remote_address: nil, broadcast: nil)
```
jreidinger: and same here `id: ""` instead of nil.

src/lib/y2network/sysconfig/connection_config_readers/ethernet.rb
teclator: It is in the code, I suggest to remove boot_protocol and startmode requirement and add just to the base class.

imobachgs: Already done.

src/lib/y2network/sysconfig/connection_config_readers/wireless.rb
```diff
@@ -17,48 +17,40 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "y2network/sysconfig/connection_config_readers/base"
 require "y2network/connection_config/wireless"
```
teclator: And now with last changes this require is even not needed, isn't it?

imobachgs: Yes, it is not needed. I removed that line from `ethernet.rb` but not from here.

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -62,6 +62,8 @@ def connection_config
         # @return [Class]
         def connection_class
           class_name = self.class.to_s.split("::").last
+          file_name = class_name.gsub(/(\w)([A-Z])/, "\\1_\\2").downcase
```
jreidinger: NP: instead of this magic you can load a type from class_name and use its methods for file_name and class_name

PR #894, opened 2019-07-22
add helpers to type and fix its usage

jreidinger:


src/lib/y2network/interface_type.rb
```diff
@@ -79,6 +79,27 @@ def file_name
       name.downcase
     end
 
+    def respond_to_missing?(method_name, _include_private = false)
```
teclator: For me the PR looks good, just maybe would add some documentation with an example.

jreidinger: ok, btw this is standard method - https://thoughtbot.com/blog/always-define-respond-to-missing-when-overriding

newer rubocop require it

teclator:
LGTM

PR #895, opened 2019-07-23
Replace deprecated operator

jreidinger:
goal of this pull request is to remove direct access to `@config` in InterfaceConfigBuilder. This will allow to modify InterfaceConfigBuilder to try e.g. store to both ConnectionConfig and LanItems in easier way.

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -1866,10 +1866,10 @@ module Yast
         Builtins.y2debug("%1", NetworkInterfaces.ConcealSecrets1(newdev))
 
         # configure bridge ports
-        bridge_ports = builder["BRIDGE_PORTS"]
-        if bridge_ports
+        if builder.type.bridge?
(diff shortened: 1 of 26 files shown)


src/modules/LanItems.rb
```diff
@@ -1978,15 +1978,15 @@ def ProposeItem(item_id)
       type = Items().fetch(item_id, {}).fetch("hwinfo", {})[type]
       builder = Y2Network::InterfaceConfigBuilder.for(type)
 
-      builder["MTU"] = "1492" if Arch.s390 && Builtins.contains(["lcs", "eth"], type)
-      builder["IPADDR"] = ""
-      builder["NETMASK"] = ""
-      builder["BOOTPROTO"] = "dhcp"
+      builder.mtu = "1492" if Arch.s390 && Builtins.contains(["lcs", "eth"], type)
```
imobachgs: What about using a constant here instead of "1492"?

PR #896, opened 2019-07-24
Added dummy connection config.

teclator:


src/lib/y2network/sysconfig/connection_config_writers/dummy.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Sysconfig
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Dummy
+      # object to the underlying system.
+      class Dummy < Base
+      private
+
+        # @see Y2Network::ConnectionConfigWriters::Base#update_file
+        def update_file(_conn)
+          file.interfacetype = "dummy"
```
jreidinger: I though that this is discouraged by wicked people

mchf: no, dummy is the "only" (ok there is one more exception, but ... ;-) type when you have to specify it.

imobachgs: Perhaps we could add a comment there to clarify that it is an exception (so we do not remove it in the future during a clean-up :) ).

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -57,6 +58,7 @@ class << self
         # @param interface [String] Interface name
         # @return [Sysconfig::InterfaceFile,nil] Sysconfig
         def find(interface)
+          Yast.import "FileUtils"
```
imobachgs: Please, put this line as it was a regular `require`.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -278,7 +284,14 @@ def save
       #
       # @return [String] Interface's type depending on the file values
```
imobachgs: Are we returning a string?

src/lib/y2network/sysconfig/connection_config_writers/dummy.rb
teclator: Not for dummy interfaces. It is also part of the man (man ifcfg-dummy), basically in case of dummy interfaces, them can only be inferred by the interface name "dummyX" or by the 'interfacetype' attribute.

src/lib/y2network/sysconfig/interface_file.rb
teclator: According to my last changes should be an `InterfaceType`, so.. updating it.

src/lib/y2network/sysconfig/connection_config_writers/dummy.rb
teclator: see https://bugzilla.suse.com/show_bug.cgi?id=1129552

PR #897, opened 2019-07-24
network-ng: list kernel modules

imobachgs:


src/lib/y2network/hwinfo.rb
```diff
@@ -96,6 +96,19 @@ def description
       @hwinfo ? @hwinfo.fetch("name", "") : ""
     end
 
+    # Returns the list of kernel modules
+    #
+    # The list of modules is internally represented as:
+    #
+    #   [[mod1name, mod1args], [mod2name, mod2args]]
```
jreidinger: we probably will also need driver options as we also allow to set them. So maybe showing current args can be useful?

PR #898, opened 2019-07-24
set more connection config attributes

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -58,12 +60,17 @@ def self.for(type)
     #
     # Load with reasonable defaults
     # @param type [Y2Network::InterfaceType] type of device
-    def initialize(type:)
+    # @param config [Y2Network::ConnectionConfig::Base] existing configuration of device or nil for
```
imobachgs: np: Just add the `, nil` there.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -232,6 +240,23 @@ def aliases
     # @param value [Array<Hash>] see #aliases for hash values
     def aliases=(value)
       @aliases = value
+
+      # connection config
+      # keep only default as aliases does not handle default ip config
+      @connection_config.ip_configs.delete_if { |c| c.id != "" }
```
imobachgs: We could add `ConnectionConfig::Base#main_ip_config` and `ConnectionConfig::Base#aliases` (or something like this) so you do not have to worry about this.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -259,6 +284,14 @@ def ip_address
     # @param [String] value
     def ip_address=(value)
       @config["IPADDR"] = value
+
+      # connection_config
+      if value.nil? || value.empty?
+        # in such case remove default config
+        @connection_config.ip_configs.delete_if { |c| c.id == "" }
+      else
+        ip_config_default.address.address = value
```
imobachgs: Same than above.

src/lib/y2network/interface_config_builder.rb
jreidinger: yep, when added, feel free to change this code ;)

PR #899, opened 2019-07-24
network-ng: Add infiniband connection

teclator:


src/lib/y2network/connection_config/infiniband.rb
```diff
@@ -0,0 +1,37 @@
...
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for infiniband connections
+    #
+    # @see https://www.kernel.org/doc/Documentation/infiniband/ipoib.txt
+    class Infiniband < Base
+      # @return [String] transport mode ("datagram" or "connected")
```
imobachgs: Why not an enum? (`BootProtocol`-like).

jreidinger: what means empty value? Use default?

src/lib/y2network/ipoib_mode.rb
```diff
@@ -0,0 +1,57 @@
...
+
+module Y2Network
+  # This class represents the supported IPoIB transport modes.
+  class IpoibMode
+    class << self
+      # Returns all the existing modes
+      #
+      # @return [Array<IpoibMode>]
+      def all
+        @all ||= IpoibMode.constants
```
imobachgs: Not for this PR, but we might think about putting this code in a common place (a mixin or something like that).

test/y2network/sysconfig/connection_config_writers/infiniband_test.rb
```diff
@@ -0,0 +1,78 @@
...
+    ]
+  end
+
+  let(:conn) do
+    instance_double(
+      Y2Network::ConnectionConfig::Infiniband,
+      name:        "ib0",
+      interface:   "ib0",
+      description: "",
+      ipoib_mode:  "datagram",
```
imobachgs: This should be a IpoibMode instance, right?

teclator: Right

PR #900, opened 2019-07-25
initial prototype of comparing backends

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -528,5 +538,13 @@ def ip_config_default
       end
       default
     end
+
+    # method that allows easy change of backend for providing data
+    # it also logs error if result differs
+    def select_backend(old, new)
```
mchf: do we need such "simple" backend access? I mean when we switch backends we have to reload whole config anyway and definitely cannot be done during life of the builder (adding, editing iface) in a simple way.

jreidinger: this method is intended mainly for testing. You can easily change what it returns and it also logs if values in backend differs. As you can you in my previous PR, currently I write values to both backends. So it hope it should be possible to read it from both backends and compare result and also test it in UI by simply changing this method.
It is not something for production.

mchf: then it deserve a "todo" or "fixme"

imobachgs: We might even use an (undocumented) environment variable, so you do not need to modify this method when testing.

src/include/network/lan/complex.rb
```diff
@@ -305,13 +305,20 @@ def enableDisableButtons
 
     # Automatically configures slaves when user enslaves them into a bond or bridge device
     def UpdateSlaves
+      # TODO: sometimes it is called when no device is selected and then it crashes
+      return if LanItems.GetCurrentName.nil? || LanItems.GetCurrentName.empty?
+
       current = LanItems.current
-      master_builder = Y2Network::InterfaceConfigBuilder.for(LanItems.GetCurrentType())
+      config = Lan.yast_config.copy
+      connection_config = config.connections.find { |c| c.name == LanItems.GetCurrentName }
```
imobachgs: We might think about adding a find_connection method to `Y2Network::Config`. Or to promote the array of connection configs to a "collection" (similar to InterfacesCollection).

src/lib/y2network/interface_config_builder.rb
imobachgs: We might even use an (undocumented) environment variable, so you do not need to modify this method when testing.

jreidinger: yes, it is also option

src/include/network/lan/complex.rb
jreidinger: yes, same as with ip_configs default. Some helper would be nice. and BTW also some easy way to assign back connection_config to yast_config after modifications would be nice. As currently changes are lost as I use copy of yast_config

PR #901, opened 2019-07-25
network-ng: Added vlan connection config.

teclator:


src/lib/y2network/connection_config/vlan.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      # Replace the name of the etherdevice by a real Y2Network::Interface
+      #
+      # @see Y2Network::InterfacesCollection
+      #
+      # @param interfaces [Y2Network::InterfacesCollection]
+      def link_related_interfaces(interfaces)
+        return unless etherdevice
+        interface = interfaces.by_name(etherdevice)
+        self.etherdevice = interface if interface
```
jreidinger: this is strange construct, as it basically means that etherdevice can be String or Interface dependending when you read it. I do not like much such non-clear contracts of classes.

teclator: Yep, it is strange. We could have two attributes, one for the name and other for the interface, but also tried to avoid bad use of it although from other point of view give more flexibility.

@jreidinger @imobachgs what do think or propose?

teclator: We could continue using just interface names and implement proper callbacks when renaming or things like that.

mchf: etherdevice is defined to be an Interface a few lines above ... so using it in by_name will always fail anyway.

src/lib/y2network/connection_config/vlan.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      def related_interfaces
+        [etherdevice].compact
```
mchf: why do we need to return an array when vlan can be assigned to one device?

teclator: It will be a common method for connections that could have related interfaces. Basically, only needed for reading and linked interfaces.

src/lib/y2network/connection_config/vlan.rb
teclator: It is a problem when reading the configuration. It should be an interface but when reading we have just the name. So, the main discussion was about who and where to put the responsibility of reading.

See https://github.com/yast/yast-network/pull/901/files/7df8eb82498d9206f3df134f8f587447d1919d14#diff-2ac28654c480f56fd97a76755a0dfb15R32

At that point the interface could already exist or not. So we discussed about a two times reading, first use the name in the ifcfg file secondly link to the proper object (at that point all the interfaces should be already declared)

jreidinger: @teclator well, my opinion is not have that name here and handle it in reader where it will assign name and in second round assign Interface. All done inside of reader and clearly document that objects are complete only after reader finish.

teclator: @jreidinger @imobachgs @mchf so, do we prefer to maintain just the names by now for the related interfaces and just the interfaces reader to take care of adding it to the list?

mchf: i agree with @jreidinger

teclator: Ok, so will commit some changes and ask for opinions. Thnx both!

teclator:
@imobachgs @jreidinger @mchf removed the addition of virtual and fake interfaces to the list. I think it maybe could be useful for the transition from LanItems and that the FakeInterfaces where introduced mainly because of that and missed from the list, so not so sure about removing the addition...

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -244,6 +244,16 @@ def variable_name(param_name)
       #   @return [String] IPOIB mode ("connected" or "datagram")
       define_variable(:ipoib_mode)
 
+      ## VLAN
+
+      # !@attribute [r] etherdevice
+      #   @return [String] Real device for the virtual LAN
+      define_variable(:etherdevice)
```
jreidinger: Hmm, I am not sure here if String is good. Still how renaming should work?

Consider here. We have physical device A that is renamed to B. And we have old vlan, which has in etherdevice A and new vlan, which has there B. And now consider that we have device that is B and is renamed to A.

How this will work in writer? Always use original name of interface?

jreidinger: ah, this should not go here, but to Vlan ConnectionConfig

teclator: Exactly, I was confused by the comment xDD This is the interface file.

teclator: > Hmm, I am not sure here if String is good. Still how renaming should work?
> 
> Consider here. We have physical device A that is renamed to B. And we have old vlan, which has in etherdevice A and new vlan, which has there B. And now consider that we have device that is B and is renamed to A.

That is what we were discussing about, if we do not reference with a Interface directly, then we will need a `service object` or `callback` or whatever that takes the responsibility of updating interface name references when an interface is renamed. 

Probably we should modify the parent device by a `ParentVlan` object that could be an interface, a connection uuid or a mac like NM permits. But for the time being I just used a simple String there.
> 
> How this will work in writer? Always use original name of interface?

jreidinger: I actually prefer to have there Interface. Only part that I do not like previously was that class is responsible for that name to interface translation. It should be done by Config reader, that reads interfaces, connection configs, create necessary interfaces if needed and result should be connection config with proper interfaces. So it should be reader responsibility to do linking to interface as during read we are sure names are correct, later it can change.

imobachgs: The problem is that there is a chance that, when the reader for a connection configuration runs, the underlying interface does not exist yet (think of virtual interfaces or even related connection config objects).

In such a case, we need to do a second pass. I would say that for that second pass, there are two options:

* Extending ConnectionConfig classes to know how to link its interfaces (a `#link` method could get the list of known interfaces and sets its own links).
* Extending readers with this logic.

Is this OK?

jreidinger: @imobachgs yes, for me the preferred solution is extending reader with logic as linking is from my POV responsibility of reader and not connection config, which will have other business.

PR #902, opened 2019-07-25
Add some clarification about Interface class role

imobachgs:


src/lib/y2network/interface.rb
```diff
@@ -22,6 +22,17 @@
 
 module Y2Network
   # Network interface.
+  #
+  # It represents network interfaces, no matter whether they are physical or virtual ones. This
+  # class (including its subclasses) are basically responsible for holding the hardware
+  # configuration (see {Hwinfo}), including naming and driver information.
+  #
+  # Logical configuration, like TCP/IP or WIFI settings, are handled through
+  # Y2Network::ConnectionConfig::Base classes.
```
jreidinger: maybe mention here also that for virtual devices its slaves is part of config and not interface. So it will be clear.

PR #903, opened 2019-07-29
network-ng: Added TUN / TAP connections

teclator:

### DIFF
--- a/src/lib/y2network/sysconfig/connection_config_writers/tap.rb
+++ b/src/lib/y2network/sysconfig/connection_config_writers/tap.rb
@@ -17,20 +17,21 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "y2network/sysconfig/connection_config_writers/tun"
+require "y2network/sysconfig/connection_config_writers/base"
 
 module Y2Network
   module Sysconfig
     module ConnectionConfigWriters
       # This class is responsible for writing the information from a ConnectionConfig::Tap
       # object to the underlying system.
-      class Tap < Tun
+      class Tap < Base
       private
 
         # @see Y2Network::ConnectionConfigWriters::Base#update_file
         def update_file(conn)
-          super
           file.tunnel = "tap"
+          file.tunnel_set_owner = conn.owner
+          file.tunnel_set_group = conn.group
         end
       end
     end
--- a/src/lib/y2network/sysconfig/connection_config_writers/tap.rb
+++ b/src/lib/y2network/sysconfig/connection_config_writers/tap.rb
@@ -17,21 +17,20 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "y2network/sysconfig/connection_config_readers/base"
-require "y2network/ipoib_mode"
+require "y2network/sysconfig/connection_config_writers/tun"
 
 module Y2Network
   module Sysconfig
-    module ConnectionConfigReaders
-      # This class is able to build a ConnectionConfig::Infiniband object given a
-      # SysconfigInterfaceFile object.
-      class Infiniband < Base
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Tap
+      # object to the underlying system.
+      class Tap < Tun
       private
 
-        # @param conn [Y2Network::ConnectionConfig::Infiniband]
-        # @see Y2Network::Sysconfig::ConnectionConfigReaders::Base#update_connection_config
-        def update_connection_config(conn)
-          conn.ipoib_mode = IpoibMode.from_name(file.ipoib_mode.to_s)
+        # @see Y2Network::ConnectionConfigWriters::Base#update_file
+        def update_file(conn)
+          super
+          file.tunnel = "tap"
         end
       end
     end
--- a/src/lib/y2network/sysconfig/interface_file.rb
+++ b/src/lib/y2network/sysconfig/interface_file.rb
@@ -349,17 +349,18 @@ module Y2Network
 
     private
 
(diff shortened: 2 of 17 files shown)


src/lib/y2network/sysconfig/connection_config_writers/tap.rb
```diff
@@ -0,0 +1,38 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/sysconfig/connection_config_writers/tun"
+
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Tap
+      # object to the underlying system.
+      class Tap < Tun
```
mchf: i think that inheritance should somehow reflect real world. Since Tap is not any extension to tun and both technologies work even on different layers of Iso / osi model, I don't think that this direction of inheritance is correct.

teclator: But that is some underlying thing. I mean, for us it only differs in the 'mode' that is the 'tunnel' attribute.

Even iproute2 uses `ip tuntap ...`

It was basically for DRY. Originally was even thinking in just add it as tuntap and the mode, but preferred to separate the types as we had for InterfaceType.

teclator: SO, if you prefer will not inherit the connection config.

teclator: @imobachgs @mchf 

For the reader & writer, I would keep at it is, but as you prefer

teclator: Also removed the inheritance.

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -301,14 +315,7 @@ def save
...
       def type
-        return InterfaceType::DUMMY if @values["INTERFACETYPE"] == "dummy"
-        return InterfaceType::BONDING if defined_variables.any? { |k| k.start_with?("BOND") }
-        return InterfaceType::BRIDGE if defined_variables.any? { |k| k.start_with?("BRIDGE") }
-        return InterfaceType::WIRELESS if defined_variables.any? { |k| k.start_with?("WIRELESS") }
-        return InterfaceType::VLAN if defined_variables.include? "ETHERDEVICE"
-        return InterfaceType::INFINIBAND if defined_variables.include? "IPOIB_MODE"
-
-        InterfaceType::ETHERNET
+        type_from_keys || type_from_values || InterfaceType::ETHERNET
```
imobachgs: Not it looks more readable. Thanks! :+1:

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -324,6 +331,26 @@ def clean
 
     private
 
+      # Determine the Interface type based on specific values
+      #
+      # @return [Y2Network::InterfaceType, nil]
+      def type_from_values
+        return InterfaceType::DUMMY if @values["INTERFACETYPE"] == "dummy"
```
imobachgs: Why not using `#interfacetype`? And what about `#tunnel?` in the lines below?

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -324,6 +331,26 @@ def clean
...
+        return InterfaceType::TUN if @values["TUNNEL"] == "tun"
+        return InterfaceType::TAP if @values["TUNNEL"] == "tap"
+      end
+
+      # Determine the Interface type based on defined variables
```
imobachgs: np: _Determines_

test/y2network/sysconfig/connection_config_readers/tun_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:file) do
+    Y2Network::Sysconfig::InterfaceFile.find(interface_name).tap(&:load)
+  end
+
+  describe "#connection_config" do
+    it "returns a tun connection config object" do
+      tun_conn = handler.connection_config
+      expect(tun_conn.interface).to eq("tun0")
+      expect(tun_conn.type).to eq(Y2Network::InterfaceType::TUN)
+      expect(tun_conn.bootproto).to eq(Y2Network::BootProtocol::STATIC)
```
imobachgs: np: I guess you do not need to check this. And, if you do, please do it in the tap test aswell.

test/y2network/sysconfig/connection_config_writers/tun_test.rb
```diff
@@ -0,0 +1,87 @@
...
+require "y2network/sysconfig/connection_config_writers/tun"
+require "y2network/startmode"
+require "y2network/boot_protocol"
+require "y2network/connection_config/tun"
+require "y2network/connection_config/ip_config"
+
+describe Y2Network::Sysconfig::ConnectionConfigWriters::Tun do
+  subject(:handler) { described_class.new(file) }
+
+  def file_content(scr_root, file)
```
imobachgs: Maybe we should start thinking about a helper (we have the same method in many places).

test/y2network/sysconfig/connection_config_readers/tun_test.rb
teclator: I try to add some of the common options defined in ifcfg-file not only the type specific.

test/y2network/sysconfig/connection_config_writers/tun_test.rb
teclator: Definitely

teclator: Will try to do it later when implementing a new connection type, probably with the bonding one.

mchf:
no more comments from my side

PR #904, opened 2019-07-29
network-ng: Added bridge connection config

teclator:


src/lib/y2network/sysconfig/connection_config_readers/bridge.rb
```diff
@@ -0,0 +1,39 @@
...
+  module Sysconfig
+    module ConnectionConfigReaders
+      # This class is able to build a ConnectionConfig::Bridge object given a
+      # SysconfigInterfaceFile object.
+      class Bridge < Base
+      private
+
+        # @see Y2Network::Sysconfig::ConnectionConfigReaders::Base#update_connection_config
+        def update_connection_config(conn)
+          conn.ports = file.bridge_ports.split(" ")
```
imobachgs: np: What if I have "1  2 3"? I would replace multiple spaces with just one before splitting.

PR #905, opened 2019-07-29
network-ng: use connection config writers

imobachgs:
This PR replaces `NetworkInterfaces.Write` with `Y2Network::Config#write`. `Yast::NetworkAutoconfiguration` is out of scope.

### DIFF
--- a/src/modules/LanItems.rb
+++ b/src/modules/LanItems.rb
@@ -736,10 +736,6 @@ module Yast
       end
 
       LanItems.WriteUdevRules if !Stage.cont && InstallInfConvertor.instance.AllowUdevModify
-
-      # FIXME: remove this code
-      # FIXME: hack: no "netcard" filter as biosdevname names it diferently (bnc#712232)
-      # NetworkInterfaces.Write("")
     end
 
     # Exports configuration for use in AY profile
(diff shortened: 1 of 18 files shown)


teclator:
For me it looks good

test/network_autoconfiguration_test.rb
```diff
@@ -47,6 +48,14 @@ def probe_netcard_factory(num)
 end
 
 describe Yast::NetworkAutoconfiguration do
+  let(:yast_config) { Y2Network::Config.new(source: :sysconfig) }
+  let(:system_config) { yast_config.copy }
+
+  before do
+    Yast::Lan.add_config(:yast, yast_config)
+    Yast::Lan.add_config(:system, system_config)
```
mchf: I'd prefer ```Y2Network::Config.add``` here. You don't need dependency on ```Lan``` module then. And ```add_config``` is only wrapper for it anyway.

src/modules/LanItems.rb
```diff
@@ -737,8 +737,9 @@ def write
 
       LanItems.WriteUdevRules if !Stage.cont && InstallInfConvertor.instance.AllowUdevModify
 
+      # FIXME: remove this code
       # FIXME: hack: no "netcard" filter as biosdevname names it diferently (bnc#712232)
-      NetworkInterfaces.Write("")
+      # NetworkInterfaces.Write("")
```
mchf: ok we usually don't keep old code as comments, so if there is a reason for it it would deserve a comment.

PR #906, opened 2019-07-30
packaging: Require hostname

DimStar77:
Since the code calls /bin/hostname in multiple places, we really ought
to ensure /bin/hostname is present.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Jul 30 11:14:04 UTC 2019 - Dominique Leuenberger <dimstar@opensuse.org>
+
+- Requires hostname: there are many places where the module calls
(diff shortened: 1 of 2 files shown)


DimStar77:
to be clear, this should fix issues like
  https://openqa.opensuse.org/tests/994890#step/installation/31

imobachgs:
Thanks for the fix. Actually, we already had [a PR](https://github.com/yast/yast-yast2/pull/951) for the yast2.rpm package which should fix this issue too. But as we call hostname directly in yast2-network, I would say that your fix still applies.

Please, could you add the bug number [boo#1142595](https://bugzilla.opensuse.org/show_bug.cgi?id=1142595) to the changes file?

Thanks!

DimStar77:
boo# added, force pushed

imobachgs:
Thanks!

PR #908, opened 2019-07-31
network-ng: Add bonding connection config

teclator:


src/lib/y2network/sysconfig/connection_config_readers/bonding.rb
```diff
@@ -0,0 +1,49 @@
...
+          conn.options = file.bonding_module_opts
+        end
+
+        # Convenience method to obtain the bonding slaves defined in the file
+        #
+        # @return [Array<String>] bonding slaves defined in the file
+        def slaves
+          return [] unless file.bonding_slaves
+
+          file.bonding_slaves.map do |_id, name|
```
imobachgs: Why not call `file.bonding_slaves.values`?

src/lib/y2network/sysconfig/connection_config_writers/bonding.rb
```diff
@@ -0,0 +1,49 @@
...
+          file.bonding_slaves = file_slaves(conn)
+          file.bonding_module_opts = conn.options
+        end
+
+        # Convenience method to obtain the map of bonding slaves in the file
+        # format
+        #
+        # @return [Hash<Integer, String>] indexed bonding slaves
+        def file_slaves(conn)
+          {}.tap do |h|
```
imobachgs: To be honest, this usage of `tap` looks weird to me. I would prefer using `#reduce` or `#each_with_object` or a simple `#each`.

teclator: Mainly used because of the index

PR #913, opened 2019-08-01
Fix Enum comparisons

imobachgs:
Implement `#==` to compare `BootProtocol` and `IpoibMode` objects.

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -28,6 +28,7 @@
 #
 require "y2firewall/helpers/interfaces"
 require "y2network/dialogs/edit_interface"
+require "y2network/boot_protocol"
 
 module Yast
   module NetworkLanAddressInclude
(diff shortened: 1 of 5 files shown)


src/include/network/lan/address.rb
```diff
@@ -73,9 +74,9 @@ def AddressDialog(builder:)
 
         # IP is mandatory for static configuration. Makes no sense to write static
         # configuration without that.
-        return ret if bootproto == "static" && ipaddr.empty?
+        return ret if bootproto == Y2Network::BootProtocol::STATIC && ipaddr.empty?
```
teclator: np: would be nice to have a `bootproto.static?` method like we defined for interface_type. We already have `dhcp?` defined and probably both are the most commonly used.

imobachgs: I see your point. But in the `dhcp` case it was done mainly because it included several values (DHCP4, DHCP6, etc.). This case is slightly different and I am not sure if it really pays off. Unless we consider more protocols to be considered as "static"...

PR #916, opened 2019-08-05
network-ng: Added ctc connection config

teclator:

### DIFF
--- a/src/lib/y2network/connection_config/ctc.rb
+++ b/src/lib/y2network/connection_config/ctc.rb
@@ -21,13 +21,32 @@ require "y2network/connection_config/base"
 
 module Y2Network
   module ConnectionConfig
-    # Configuration for ctc connections
+    # Configuration for ctc connections.
+    #
+    # @note The use of this connection is deprecated or not recommended as it
+    #   will not be officially supported in future SLE versions.
     class Ctc < Base
+      # For CCW devices and CCW group devices, this device ID is the device
+      # bus-ID.
+      #
+      # The device bus-ID is of the format 0.<subchannel_set_ID>.<devno>,
+      # for example, 0.0.8000.
+      #
+      # The CTCM device driver requires two I/O subchannels for each interface,
+      # a read subchannel and a write subchannel
+      #
+      # @see https://www.ibm.com/developerworks/linux/linux390/index.html
       # @return [String] read bus id
       attr_accessor :read_channel
       # @return [String] write bus id
       attr_accessor :write_channel
-      # @return [Integer] connection protocol
+      # @return [Integer] connection protocol (0, 1, 3, or 4)
+      #   0 Compatibility with peers other than OS/390®.
+      #   1 Enhanced package checking for Linux peers.
+      #   3 For compatibility with OS/390 or z/OS peers.
+      #   4 For MPC connections to VTAM on traditional mainframe operating systems.
+      # @see https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.ljdd/ljdd_t_ctcm_wrk_protocol.html
+      # @see https://github.com/SUSE/s390-tools/blob/master/ctc_configure#L16
       attr_accessor :protocol
     end
   end
--- a/src/lib/y2network/connection_config/ctc.rb
+++ b/src/lib/y2network/connection_config/ctc.rb
@@ -21,25 +21,14 @@ require "y2network/connection_config/base"
 
 module Y2Network
   module ConnectionConfig
-    # Configuration for qeth connections
-    class Qeth < Base
+    # Configuration for ctc connections
+    class Ctc < Base
       # @return [String] read bus id
       attr_accessor :read_channel
       # @return [String] write bus id
       attr_accessor :write_channel
-      # @return [String] data bus id
-      attr_accessor :data_channel
-      # @return [Boolean] whether layer2 is enabled or not
-      attr_accessor :layer2
-      # @return [Integer] port number
-      attr_accessor :port_number
-
-      # Constructor
-      def initialize
-        super()
-        @layer2 = false
-        @port_number = 0
-      end
+      # @return [Integer] connection protocol
+      attr_accessor :protocol
     end
   end
 end
--- a/src/lib/y2network/connection_config/qeth.rb
+++ b/src/lib/y2network/connection_config/qeth.rb
@@ -23,6 +23,22 @@ module Y2Network
   module ConnectionConfig
     # Configuration for qeth connections
     class Qeth < Base
+      # For CCW devices and CCW group devices, this device ID is the device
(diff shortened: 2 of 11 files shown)


src/lib/y2network/connection_config/qeth.rb
```diff
@@ -23,6 +23,23 @@ module Y2Network
...
     # Configuration for qeth connections
     class Qeth < Base
+      # @return [String] read bus id
+      attr_accessor :read_channel
+      # @return [String] write bus id
+      attr_accessor :write_channel
+      # @return [String] data bus id
+      attr_accessor :data_channel
+      # @return [Boolean] whether layer2 is enabled or not
+      attr_accessor :layer2
```
mchf: at first view i miss logical link address (LLADDR) ... is it somewhere else?

src/lib/y2network/connection_config/ctc.rb
```diff
@@ -0,0 +1,34 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for ctc connections
+    class Ctc < Base
+      # @return [String] read bus id
+      attr_accessor :read_channel
```
jreidinger: can you add example how it looks like for others not so familiar with s390?

src/lib/y2network/connection_config/ctc.rb
```diff
@@ -0,0 +1,34 @@
...
+      # @return [String] write bus id
+      attr_accessor :write_channel
+      # @return [Integer] connection protocol
+      attr_accessor :protocol
```
jreidinger: link to more details as it looks strange to have protocol identified just by number.

src/lib/y2network/interface_type.rb
```diff
@@ -132,6 +132,8 @@ def method_missing(method_name, *arguments, &block)
     LCS = new(N_("LCS"), "lcs")
     # HiperSockets s390 network device
     HSI = new(N_("HSI"), "hsi")
+    # Channel To Channel
```
jreidinger: it is s390 specific. If you have link with more details, it will be welcomed.

src/lib/y2network/sysconfig/connection_config_readers/ctc.rb
```diff
@@ -0,0 +1,36 @@
...
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigReaders
+      # This class is able to build a ConnectionConfig::Ctc object given a
+      # Sysconfig::InterfaceFile object.
+      class Ctc < Base
+      private
+
+        # @see Y2Network::Sysconfig::ConnectionConfigReaders::Base#update_connection_config
+        def update_connection_config(_conn)
```
jreidinger: this looks strange. Why not having empty method in Base and redefine only when needed?

src/lib/y2network/sysconfig/connection_config_writers/ctc.rb
```diff
@@ -0,0 +1,36 @@
...
+module Y2Network
+  module Sysconfig
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Ctc
+      # object to the underlying system.
+      class Ctc < Base
+      private
+
+        # @see Y2Network::ConnectionConfigWriters::Base#update_file
+        def update_file(_conn)
```
jreidinger: same here. Defining empty method looks strange. Simply have empty method in Base and redefine only when needed.

src/lib/y2network/connection_config/qeth.rb
teclator: Yes, it was added to base:

https://github.com/yast/yast-network/pull/909/files#diff-828c91bf00f25c07b26cb97ff7952eddR56

src/lib/y2network/connection_config/ctc.rb
teclator: @imobachgs , maybe something you could comment.

src/lib/y2network/sysconfig/connection_config_readers/ctc.rb
teclator: @imobachgs , maybe something you could comment.

imobachgs: Yes, fine with having the base method.

src/lib/y2network/connection_config/ctc.rb
```diff
@@ -0,0 +1,53 @@
...
+      # a read subchannel and a write subchannel
+      #
+      # @see https://www.ibm.com/developerworks/linux/linux390/index.html
+      # @return [String] read bus id
+      attr_accessor :read_channel
+      # @return [String] write bus id
+      attr_accessor :write_channel
+      # @return [Integer] connection protocol (0, 1, 3, or 4)
+      #   0 Compatibility with peers other than OS/390®.
+      #   1 Enhanced package checking for Linux peers.
```
imobachgs: Just curiosity: where is the number "2"?

src/lib/y2network/connection_config/qeth.rb
```diff
@@ -23,6 +23,39 @@ module Y2Network
   module ConnectionConfig
     # Configuration for qeth connections
     class Qeth < Base
+      # For CCW devices and CCW group devices, this device ID is the device
```
imobachgs: To be clear, the "read channel" is the "device ID" you are talking about, right?

src/lib/y2network/connection_config/ctc.rb
teclator: Probably does not make sense in Linux. To be honest, No idea

teclator: see https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.ljdd/ljdd_t_ctcm_wrk_protocol.html

src/lib/y2network/connection_config/qeth.rb
teclator: Well, the device ID is for both, the read channel and write channel. That is for example "0.0.8000" is the read channel device id and "0.0.8001" is the write channel device id.

https://www.ibm.com/support/knowledgecenter/linuxonibm/com.ibm.linux.z.lgdd/lxctcrot.jpg
https://github.com/SUSE/s390-tools/blob/master/ctc_configure#L10

teclator:
> In general, it looks good to me. My only concern is whether it makes sense to have a reader/writer which do not implement `update_file` or `update_connection_config`.

The thing is that each connection looks for a reader or writer of its class, when all of them does not have a particular reader or writer. That is, will use mainly the same base options.

teclator:
I will merge it as it is. But will create a macro with ccw group device information in next PR to add to s390 connnection config types.

PR #919, opened 2019-08-12
network-ng: Added usb connection config

teclator:

### DIFF
--- a/src/lib/y2network/connection_config/usb.rb
+++ b/src/lib/y2network/connection_config/usb.rb
@@ -1,0 +1,28 @@
+# Copyright (c) [2019] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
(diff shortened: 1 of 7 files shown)


src/lib/y2network/connection_config/usb.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/connection_config/base"
+
+module Y2Network
+  module ConnectionConfig
+    # Configuration for Usb connections
+    class Usb < Base
```
jreidinger: it is a bit strange that in reader/writter it is ethernet and here only base. So what if read want to write some eth specific attribute?

teclator: It was mainly because of the `update_file` and `update_connection_config` methods. As there was not particular attributes in the USB connection, but once these methods have been just implemented in the Base class there is no reason to inherit from Ethernet anymore.

PR #920, opened 2019-08-15
Infer interface type

mchf:
Tested manually and seems to be working as expected.

### DIFF
--- a/src/lib/y2network/connection_configs_collection.rb
+++ b/src/lib/y2network/connection_configs_collection.rb
@@ -53,6 +53,14 @@ module Y2Network
       connection_configs.find { |c| c.name == name }
     end
 
+    # Returns a connection configuration tied to the given interface
+    #
+    # @param iface [Y2Network::Interface]
+    # @return [ConnectionConfig, nil] Connection config tied to the interface or nil if not found
+    def by_interface(iface)
+      connection_configs.find { |c| c.interface == iface }
+    end
+
     # Adds or updates a connection configuration
     #
     # @note It uses the name to do the matching.
--- a/src/lib/y2network/connection_configs_collection.rb
+++ b/src/lib/y2network/connection_configs_collection.rb
@@ -53,14 +53,6 @@ module Y2Network
       connection_configs.find { |c| c.name == name }
     end
 
-    # Returns a connection configuration tied to the given interface
-    #
-    # @param iface [Y2Network::Interface]
-    # @return [ConnectionConfig, nil] Connection config tied to the interface or nil if not found
-    def by_interface(iface)
-      connection_configs.find { |c| c.interface == iface }
-    end
-
     # Adds or updates a connection configuration
     #
     # @note It uses the name to do the matching.
--- a/src/lib/y2network/interface_type.rb
+++ b/src/lib/y2network/interface_type.rb
@@ -112,7 +112,7 @@ module Y2Network
     # Infiniband card
     INFINIBAND = new(N_("Infiniband"), "ib")
     # Infiniband card child device. Used in IPoIB (IP-over-InfiniBand)
-    INFINIBANDCHILD = new(N_("Infiniband Child"), "ibchild")
+    INFINIBAND_CHILD = new(N_("Infiniband Child"), "ibchild")
     # Bonding device
     BONDING = new(N_("Bonding"), "bond")
     # bridge device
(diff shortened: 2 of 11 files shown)


src/lib/y2network/connection_configs_collection.rb
```diff
@@ -53,6 +53,14 @@ def by_name(name)
       connection_configs.find { |c| c.name == name }
     end
 
+    # Returns a connection configuration tied to the given interface
+    #
+    # @param iface [Y2Network::Interface]
+    # @return [ConnectionConfig, nil] Connection config tied to the interface or nil if not found
+    def by_interface(iface)
+      connection_configs.find { |c| c.interface == iface }
```
jreidinger: with profiles it can be more then one, not?

mchf: eh ... this function is not used at all - i tried to create something and in the end went by another way ... I will drop it.

src/lib/y2network/interface_type.rb
```diff
@@ -111,6 +111,8 @@ def method_missing(method_name, *arguments, &block)
     WIRELESS = new(N_("Wireless"), "wlan")
     # Infiniband card
     INFINIBAND = new(N_("Infiniband"), "ib")
+    # Infiniband card child device. Used in IPoIB (IP-over-InfiniBand)
+    INFINIBANDCHILD = new(N_("Infiniband Child"), "ibchild")
```
jreidinger: why not INFINIBAND_CHILD?

src/lib/y2network/interface_type.rb
```diff
@@ -134,5 +136,19 @@ def method_missing(method_name, *arguments, &block)
...
+    # IPv6 Tunnel interface
+    IPV6TNL = new(N_("IPv6 Tunnel"), "ip6tnl")
+    # IPv6 over IPv4
+    SIT = new(N_("IPv6 over IPv4 Tunnel"), "sit")
+    # IP over IPv4
+    GRE = new(N_("IP over IPv4 Tunnel"), "gre")
+    # Infrared
+    IRDA = new(N_("Infrared"), "irda")
+    # Loopback
+    LO = new(N_("Loopback"), "lo")
```
jreidinger: wow, that is a lot of new types, can we somehow handle it? we need to add at least popup that we cannot configure it.

src/lib/y2network/interface_type.rb
mchf: can be ... I was not aware of any policy in naming

src/lib/y2network/interface_type.rb
mchf: none of these types was never created by y2-network. But we were able to detect them (and e.g. filter such ifcfg files out of parsing). However it would be nice to have ability to e.g. create these IP* over IP* tunnel devices in the future.

src/lib/y2network/sysconfig/type_detector.rb
```diff
@@ -0,0 +1,119 @@
...
+
+        # Provides interface's sysconfig configuration
+        #
+        # @return [Hash{String => Object}] parsed configuration
+        def devmap(iface)
+          scr_path = ".network.value.\"#{iface}\""
+          values = Yast::SCR.Dir(Yast::Path.new(scr_path))
+
+          # provide configuration in canonicalized format
+          devmap = Yast::NetworkInterfaces.generate_config(scr_path, values)
```
jreidinger: cannot we use what we currently have in connection config? using for generating config network interfaces looks strange for me.

src/lib/y2network/type_detector.rb
```diff
@@ -0,0 +1,141 @@
...
+      # @return [Y2Network::InterfaceType, nil] type of given interface or nil if cannot be recognized
+      def type_of(iface)
+        type_by_sys(iface) || type_by_config(iface) || nil
+      end
+
+    private
+
+      include Yast::Logger
+
+      SYS_TYPE_NUMBERS = {
```
jreidinger: any link or documentation for this magic numbers?

mchf: i'm not aware of any ... taken from wicked sources

src/lib/y2network/sysconfig/type_detector.rb
mchf: ok I'll try to polish it a bit.

jreidinger:
replaced by https://github.com/yast/yast-network/pull/927

PR #921, opened 2019-08-20
network-ng add interfaces renaming support

imobachgs:
This PR implements support to rename an interface using a udev rule.

* [x] Handle udev rules
* [x] Add an API to rename interfaces
* [x] Adapt UI
* [ ] Rename routes
* [ ] Support for hotplug devices
* [ ] Do not overwrite custom rules

### DIFF
--- a/src/lib/network/edit_nic_name.rb
+++ b/src/lib/network/edit_nic_name.rb
@@ -19,33 +19,17 @@ module Yast
     # @return [String] current udev match criteria
     attr_reader :old_key
 
-    # udev rule attribute for MAC address
-    MAC_UDEV_ATTR   = "ATTR{address}".freeze
-
-    # udev rule attribute for BUS id
-    BUSID_UDEV_ATTR = "KERNELS".freeze
-
-    def initialize
+    # Constructor
+    #
+    # @param [Y2Network::InterfaceConfigBuilder] Interface configuration
+    def initialize(settings)
       textdomain "network"
-
-      Yast.include self, "network/routines.rb"
-
-      current_item = LanItems.getCurrentItem
-      current_rule = LanItems.current_udev_rule
-
-      @old_name = LanItems.current_udev_name
-      unless current_rule.empty?
-        @old_key = :mac unless LanItems.GetItemUdev(MAC_UDEV_ATTR).empty?
-        @old_key = :bus_id unless LanItems.GetItemUdev(BUSID_UDEV_ATTR).empty?
-      end
-
-      if current_item["hwinfo"]
-        @mac = current_item["hwinfo"]["permanent_mac"]
-        @bus_id = current_item["hwinfo"]["busid"]
-      else
-        @mac = ""
-        @bus_id = ""
-      end
+      @settings = settings
+      interface = settings.interface
+      @old_name = interface.name
+      @old_key = interface.renaming_mechanism
+      @mac = interface.hardware.mac
+      @bus_id = interface.hardware.busid
     end
 
     # Opens dialog for editing NIC name and runs event loop.
@@ -61,22 +45,16 @@ module Yast
         next if ret != :ok
 
         new_name = UI.QueryWidget(:dev_name, :Value)
+        udev_type = UI.QueryWidget(:udev_type, :CurrentButton)
 
         if CheckUdevNicName(new_name)
-          LanItems.rename(new_name)
+          @settings.rename_interface(new_name, udev_type)
         else
           UI.SetFocus(:dev_name)
           ret = nil
 
           next
         end
-
-        udev_type = UI.QueryWidget(:udev_type, :CurrentButton)
-
-        # FIXME: it changes udev key used for device identification
-        #  and / or its value only, name is changed elsewhere
-        LanItems.update_item_udev_rule!(udev_type) if udev_type && (old_key != udev_type)
-        LanItems.rename_current_device_in_routing(old_name) if new_name != old_name
       end
 
       close
@@ -146,11 +124,12 @@ module Yast
     # @return [boolean] false if name is invalid
     def CheckUdevNicName(name)
       # check if the name is assigned to another device already
-      if LanItems.GetNetcardNames.include?(name) && name != LanItems.GetCurrentName
+      if @settings.name_exists?(name)
         Popup.Error(_("Configuration name already exists."))
         return false
       end
-      if !ValidNicName(name)
+
+      if !@settings.valid_name?(name)
         Popup.Error(_("Invalid configuration name."))
         return false
       end
--- a/src/lib/network/edit_nic_name.rb
+++ b/src/lib/network/edit_nic_name.rb
@@ -21,7 +21,7 @@ module Yast
 
     # Constructor
     #
-    # @param [Y2Network::InterfaceConfigBuilder] Interface configuration
+    # @param settings [Y2Network::InterfaceConfigBuilder] Interface configuration
     def initialize(settings)
       textdomain "network"
       @settings = settings
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -129,6 +129,17 @@ module Y2Network
         routing == other.routing && dns == other.dns
     end
 
+    # Renames a given interface and the associated connections
+    #
+    # @param old_name  [String] Old interface's name
+    # @param new_name  [String] New interface's name
+    # @param mechanism [Symbol] Property to base the rename on (:mac or :bus_id)
+    def rename_interface(old_name, new_name, mechanism)
+      interface = interfaces.by_name(old_name)
+      interface.rename(new_name, mechanism)
+      connections.by_interface(old_name).each { |c| c.interface = new_name }
+    end
+
     alias_method :eql?, :==
   end
 end
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -45,6 +45,8 @@ module Y2Network
     attr_reader :configured
     # @return [HwInfo]
     attr_reader :hardware
+    # @return [Symbol] Mechanism to rename the interface (nil -no rename-, :bus_id or :mac)
+    attr_accessor :renaming_mechanism
 
     # Shortcuts for accessing interfaces' ifcfg options
     #
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -17,6 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"
 require "y2network/interface_type"
 require "y2network/hwinfo"
 
@@ -35,6 +36,8 @@ module Y2Network
   # @see Y2Network::VirtualInterface
   # @see Y2Network::FakeInterface
   class Interface
+    include Yast::Logger
+
     # @return [String] Device name ('eth0', 'wlan0', etc.)
     attr_accessor :name
     # @return [String] Interface description
@@ -112,6 +115,7 @@ module Y2Network
     # @param new_name  [String] New interface's name
     # @param mechanism [Symbol] Property to base the rename on (:mac or :bus_id)
     def rename(new_name, mechanism)
+      log.info "Rename interface '#{name}' to '#{new_name}' using the '#{mechanism}'"
       @name = new_name
       @renaming_mechanism = mechanism
     end
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -107,6 +107,15 @@ module Y2Network
       hardware.drivers
     end
 
+    # Renames the interface
+    #
+    # @param new_name  [String] New interface's name
+    # @param mechanism [Symbol] Property to base the rename on (:mac or :bus_id)
+    def rename(new_name, mechanism)
+      @name = new_name
+      @renaming_mechanism = mechanism
+    end
+
   private
 
     def system_config(name)
--- a/src/lib/y2network/sysconfig/interfaces_reader.rb
+++ b/src/lib/y2network/sysconfig/interfaces_reader.rb
@@ -75,10 +75,23 @@ module Y2Network
       # Physical interfaces are read from the old LanItems module
       def find_physical_interfaces
         return if @interfaces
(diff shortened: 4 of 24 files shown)


src/lib/y2network/interface.rb
```diff
@@ -45,6 +45,8 @@ class Interface
     attr_reader :configured
     # @return [HwInfo]
     attr_reader :hardware
+    # @return [Symbol] Mechanism to rename the interface (nil -no rename-, :bus_id or :mac)
+    attr_accessor :renaming_mechanism
```
jreidinger: this looks quite limiting for me as I would hope that in future we can allow to rename interface more flexible, e.g. by user provided bus_id or by different udev value, which is currently not supported ( and means that for e.g. usb attached wifi you cannot set udev name, as udev rules are empty in dialog ).

src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -144,6 +146,20 @@ def add_interface(name, conn)
...
+      # @param name [String] Interface's name
+      # @return [Symbol,nil] :mac (MAC address), :bus_id (BUS ID) or nil (no renaming)
+      def renaming_mechanism_for(name)
+        rule = UdevRule.find_for(name)
+        return nil unless rule
+        if rule.parts.any? { |p| p.key == "ATTR{address}" }
+          :mac
+        elsif rule.parts.any? { |p| p.key == "KERNELS" }
+          :bus_id
+        end
```
jreidinger: don't we kill some complex udev rules there? E.g. if customer has his nice handcrafted udev rule for network card, if it survive run of yast2-network?

src/lib/y2network/sysconfig/interfaces_writer.rb
```diff
@@ -0,0 +1,64 @@
...
+          Y2Network::UdevRule.new_mac_based_rename(iface.name, iface.hardware.mac)
+        when :bus_id
+          Y2Network::UdevRule.new_bus_id_based_rename(iface.name, iface.hardware.busid, iface.hardware.dev_port)
+        end
+      end
+
+      # Refreshes udev service
+      def update_udevd
+        Yast::Execute.on_target!("/usr/bin/udevadm", "control", "--reload")
+        Yast::Execute.on_target!("/usr/bin/udevadm", "trigger", "--subsystem-match=net", "--action=add")
```
jreidinger: hmm, internal error if it failed. Maybe we should catch it later and show some nicer popup?

src/lib/y2network/udev_rule.rb
```diff
@@ -0,0 +1,125 @@
...
+  #   rule = UdevRule.find_for("eth0")
+  #   rule.to_s #=> "ACTION==\"add\", SUBSYSTEM==\"net\", ATTR{address}==\"?*31:78:f2\", NAME=\"eth0\""
+  class UdevRule
+    class << self
+      # Returns the udev rule for a given device
+      #
+      # @param device [String] Network device name
+      # @return [UdevRule] udev rule
+      def find_for(device)
+        rules_map = Yast::SCR.Read(Yast::Path.new(".udev_persistent.net")) || {}
```
jreidinger: how expensive is this call?

src/lib/y2network/udev_rule.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      # Helper method to create a rename rule based on the BUS ID
+      #
+      # @param name     [String] Interface's name
+      # @param mac      [String] BUS ID (e.g., "0000:08:00.0")
+      # @param dev_port [String] Device port
+      def new_bus_id_based_rename(name, bus_id, dev_port = nil)
+        parts = [UdevRulePart.new("KERNELS", "=", bus_id)]
+        parts.push(UdevRulePart.new("ATTR{dev_port}", "=", dev_port)) if dev_port
```
jreidinger: why not simple `parts << UdevRulePart.new("ATTR{dev_port}", "=", dev_port) if dev_port`

src/lib/y2network/udev_rule.rb
```diff
@@ -0,0 +1,125 @@
...
+      # The network rule includes some parts by default.
+      #
+      # @param parts [Array<UdevRulePart] Additional rule parts
+      # @return [UdevRule] udev rule
+      def new_network_rule(parts = [])
+        base_parts = [
+          UdevRulePart.new("SUBSYSTEM", "==", "net"),
+          UdevRulePart.new("ACTION", "==", "add"),
+          UdevRulePart.new("DRIVERS", "==", "?*"),
+          UdevRulePart.new("ATTR{type}", "==", "1")
```
jreidinger: some explanation here would be nice.

src/lib/y2network/udev_rule.rb
```diff
@@ -0,0 +1,125 @@
...
+        ]
+        new(base_parts.concat(parts))
+      end
+
+      # Writes udev rules to the filesystem
+      #
+      # @param udev_rules [Array<UdevRule>] List of udev rules
+      def write(udev_rules)
+        Yast::SCR.Write(Yast::Path.new(".udev_persistent.rules"), udev_rules.map(&:to_s))
+        Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
```
jreidinger: explanation for this tricky call?

src/lib/y2network/udev_rule_part.rb
```diff
@@ -0,0 +1,81 @@
...
+  # that logic later if required.
+  class UdevRulePart
+    # Regular expression to match a udev rule part
+    PART_REGEXP = Regexp.new("\\A(?<key>[A-Za-z\{\}]+)(?<operator>[^\"]+)\"(?<value>.+)\"\\Z")
+
+    class << self
+      # Returns a rule part from a string
+      #
+      # @example Simple case
+      #   part = UdevRulePart.from_string('ACTION=="add"')
```
jreidinger: is it mandatory to have key without quotes and value with quotes? according to `man udev` Or is it defined by our agent for reading it?

src/lib/y2network/udev_rule_part.rb
```diff
@@ -0,0 +1,81 @@
...
+      # @return [UdevRule] udev rule object
+      def from_string(str)
+        match = PART_REGEXP.match(str)
+        return if match.nil?
+        new(match[:key], match[:operator], match[:value])
+      end
+    end
+    # @return [String] Key name
+    attr_accessor :key
+    # @return [String] Operator ("==", "!=", "=", "+=", "-=", ":=")
```
jreidinger: important part is that first two are matching operators and rest are assign operator.

src/lib/y2network/udev_rule_part.rb
```diff
@@ -0,0 +1,81 @@
...
+    attr_accessor :operator
+    # @return [String] Value to match or assign
```
jreidinger: maybe mention that value support some kind of globs. Ideally mention in top of class `man udev` which contain interesting information.

jreidinger:
what I am missing here is actual renaming in internal structures. So what I mean? That if interface A is renamed to interface B, that it will correctly use name interface B everywhere and also mark properly all objects that uses modified interface as modified, so it is written down.

src/lib/y2network/sysconfig/interfaces_reader.rb
imobachgs: The current implementations kills them anyway.

src/lib/y2network/udev_rule.rb
imobachgs: Write changes to the file. We do something similar for sysconfig. I will add a note.

src/lib/y2network/udev_rule_part.rb
imobachgs: Yes, that's it.

src/lib/y2network/sysconfig/interfaces_writer.rb
imobachgs: > what I am missing here is actual renaming in internal structures. So what I mean? That if interface A is renamed to interface B, that it will correctly use name interface B everywhere and also mark properly all objects that uses modified interface as modified, so it is written down.

The PR is not finished yet :) I am writing a method in `Y2Network::Config` to handle such renames and keep consistency between connections and interfaces.

src/lib/y2network/interface.rb
imobachgs: You are right, I did not though about hotplug devices. Maybe we could add a `renaming_value` (provided by the user). But I would keep the `renaming_mechanism` so we know against which attribute we should match.

src/lib/y2network/udev_rule.rb
imobachgs: > what I am missing here is actual renaming in internal structures. So what I mean? That if interface A is renamed to interface B, that it will correctly use name interface B everywhere and also mark properly all objects that uses modified interface as modified, so it is written down.


The PR is not finished yet :) I am writing a method in `Y2Network::Config` to handle such renames and keep consistency between connections and interfaces.

src/lib/y2network/udev_rule_part.rb
imobachgs: OK, I will clarify it. Actually, I am using a [glob in the tests](https://github.com/yast/yast-network/pull/921/commits/6658f26a2efde12c7b4e93aa4127a6619a8cb5cb#diff-c4b8fec620000db22c30682f986fe97eR33).

src/lib/y2network/interface.rb
imobachgs: Or we could even take those `renaming_mechanism/renaming_value` to a separate class.

src/lib/y2network/udev_rule.rb
imobachgs: It does not look expensive, but maybe it can sum up if we have dozens of cards. I will memoize the result.

imobachgs:
> what I am missing here is actual renaming in internal structures. So what I mean? That if interface A is renamed to interface B, that it will correctly use name interface B everywhere and also mark properly all objects that uses modified interface as modified, so it is written down.

The PR is not finished yet :) I am writing a method in Y2Network::Config to handle such renames and keep consistency between connections and interfaces.

src/lib/y2network/udev_rule_part.rb
imobachgs: I am following our agent convention but I also checked some examples here and there and it looks like the right syntax. Even rules under `/etc/udev/rules.d` follow this syntax.

src/lib/y2network/interface.rb
jreidinger: maybe it will be more flexible to have it in own class.

src/lib/y2network/sysconfig/interfaces_reader.rb
jreidinger: yep, but we should aim to behave better as I see many customers use fine tuned udev rules.

src/lib/y2network/udev_rule.rb
jreidinger: I thought it looks differently with value nil and ".udev_persistent" as path, but I can be wrong.

imobachgs: I expected that too, but I learn from the old code that, in this case, it is different. I tried it manually and the "expected" version does not work.

src/lib/network/edit_nic_name.rb
```diff
@@ -19,33 +19,17 @@ class EditNicName
...
-        @bus_id = current_item["hwinfo"]["busid"]
-      else
-        @mac = ""
-        @bus_id = ""
-      end
+      @settings = settings
+      interface = settings.interface
+      @old_name = interface.name
+      @old_key = interface.renaming_mechanism
+      @mac = interface.hardware.mac
```
jreidinger: what happens when hardware is nil? ( sadly we now call it even for fake devices )

src/lib/y2network/config.rb
```diff
@@ -129,6 +129,17 @@ def ==(other)
...
 
+    # Renames a given interface and the associated connections
+    #
+    # @param old_name  [String] Old interface's name
+    # @param new_name  [String] New interface's name
+    # @param mechanism [Symbol] Property to base the rename on (:mac or :bus_id)
+    def rename_interface(old_name, new_name, mechanism)
+      interface = interfaces.by_name(old_name)
+      interface.rename(new_name, mechanism)
+      connections.by_interface(old_name).each { |c| c.interface = new_name }
```
jreidinger: btw we probably need to update also hostname  tab where it can link to interface. Or does we use object there?

src/lib/y2network/interface.rb
```diff
@@ -105,6 +107,15 @@ def drivers
       hardware.drivers
     end
 
+    # Renames the interface
+    #
+    # @param new_name  [String] New interface's name
+    # @param mechanism [Symbol] Property to base the rename on (:mac or :bus_id)
+    def rename(new_name, mechanism)
```
jreidinger: please add logging as I expect some troubles there.

test/y2network/config_test.rb
```diff
@@ -146,4 +159,18 @@
...
+      config.rename_interface("eth0", "eth1", :mac)
+      eth1 = config.interfaces.by_name("eth1")
+      expect(eth1.renaming_mechanism).to eq(:mac)
+    end
+
+    it "adjusts the connection configurations for that interface" do
+      config.rename_interface("eth0", "eth1", :mac)
+      eth1_conns = config.connections.by_interface("eth1")
+      expect(eth1_conns).to_not be_empty
+    end
```
jreidinger: can you just for sure add test for tricky double rename? that was eth0, then eth1 and then eth2? In past we also have issues with renames of multiple interfaces and with editing renamed interface, but it is something we should test manually.

src/lib/network/edit_nic_name.rb
imobachgs: I will fix this in a separate PR. Actually, the widget must be updated to display a textfield where the user can specify the value

src/lib/y2network/config.rb
imobachgs: I will check.

imobachgs: I've created [a PBI](https://trello.com/c/2ZfEE3yB/3456-fix-dhcpsethostname-handling) a card because dhcp_hostname is not handled correctly.

PR #926, opened 2019-08-26
Drop old backend

jreidinger:
note: not all is dropped to keep at least basic functionality.

Also already some tricky behavior dropped, like setting BOND_MASTER to yes there, but maybe it is needed.


src/lib/y2network/widgets/interfaces_table.rb
```diff
@@ -0,0 +1,75 @@
...
+          _("Device"),
+          _("Note")
+        ]
+      end
+
+      def items
+        # TODO: unconfigured devices
+        config = Yast::Lan.yast_config.copy
+        # TODO: handle unconfigured
+        config.connections.map do |conn|
```
teclator: As it is an interfaces table and not a connection table it should use config.interfaces isn't it? When adding a new connection with a Fake interface yast should already contain it.

jreidinger: yes, I need to check how it is done. It hopefully solve also unconfigured ones.

src/include/network/lan/cmdline.rb
```diff
@@ -184,17 +184,14 @@ def ListHandler(options)
...
+      type = options.fetch("type", infered_type(options))
+      if type.empty?
         Report.Error(_("The device type is mandatory."))
         return false
       end
 
-      builder = Y2Network::InterfaceConfigBuilder.for(LanItems.type)
-      builder.name = LanItems.GetCurrentName()
+      builder = Y2Network::InterfaceConfigBuilder.for(Y2Network::InterfaceType.from_short_name(type))
+      builder.name = options.fetch("name")
```
imobachgs: If `options["name"]` it will crash. But that's better than continuing with a `nil` value for `builder.name`.

src/lib/y2network/config.rb
```diff
@@ -140,6 +140,15 @@ def rename_interface(old_name, new_name, mechanism)
       connections.by_interface(old_name).each { |c| c.interface = new_name }
     end
 
+    # deletes interface and all its config. If interface is physical, it is kept.
```
imobachgs: You mean that only the configuration is removed, right?

src/lib/y2network/interface_config_builders/bonding.rb
```diff
@@ -25,46 +25,28 @@ module Y2Network
...
-        @slaves = value
-      end
-
-      # (see Y2Network::InterfaceConfigBuilder#save)
-      def save
-        super
-
-        Yast::LanItems.bond_slaves = @slaves
-        Yast::LanItems.bond_option = bond_options
+      # current options for bonding
```
imobachgs: np: Please, annotate the type. Given that it is called `options`, somebody could expect an array.

src/lib/y2network/interface_config_builders/infiniband.rb
```diff
@@ -29,31 +28,22 @@ def initialize(config: nil)
...
+        connection_config.ipoib_mode = IpoibMode.from_name(value)
+      end
 
       # Returns current value of infiniband mode
       #
       # @return [String] particular mode or "default" when not set
       def ipoib_mode
-        @ipoib_mode ||= if [nil, ""].include?(@config["IPOIB_MODE"])
-          "default"
+        case connection_config.ipoib_mode.name
```
imobachgs: If you are expecting only two branches, what about using an `if`?

src/lib/y2network/interface_config_builders/tap.rb
```diff
@@ -0,0 +1,41 @@
...
+      def initialize(config: nil)
+        super(type: InterfaceType::TAP, config: config)
+      end
+
+      # @return [Array(2)<String,String>] user and group of tunnel
+      def tunnel_user_group
+        [connection_config.owner, connection_config.group]
+      end
+
+      def assign_tunnel_user_group(user, group)
```
imobachgs: Are user/group names or numerical IDs?

src/lib/y2network/interface_config_builders/tun.rb
```diff
@@ -0,0 +1,41 @@
...
+      def initialize(config: nil)
+        super(type: InterfaceType::TUN, config: config)
+      end
+
+      # @return [Array(2)<String,String>] user and group of tunnel
+      def tunnel_user_group
+        [connection_config.owner, connection_config.group]
+      end
+
+      def assign_tunnel_user_group(user, group)
```
imobachgs: Same than the `TAP` case.

src/lib/y2network/config.rb
jreidinger: yes, we cannot remove physical interface :)

src/lib/y2network/interface_config_builders/infiniband.rb
jreidinger: makes sense, it is relict of old way where all types are managed manually.

src/lib/y2network/interface_config_builders/tap.rb
jreidinger: both are possible.

src/lib/y2network/config.rb
imobachgs: hehe, I know :) But I was thinking about some newcomer reading the code (making the distinction between the config and the interface itself).

src/lib/y2network/interface_config_builders/tap.rb
imobachgs: Please, document :)

PR #927, opened 2019-08-27
Infer interfaces

jreidinger:
replaces https://github.com/yast/yast-network/pull/920
just merge conflicts

### DIFF
--- a/src/lib/y2network/sysconfig/interfaces_reader.rb
+++ b/src/lib/y2network/sysconfig/interfaces_reader.rb
@@ -26,9 +26,9 @@ require "y2network/fake_interface"
 require "y2network/sysconfig/connection_config_reader"
 require "y2network/interfaces_collection"
 require "y2network/connection_configs_collection"
+require "y2network/sysconfig/type_detector"
 
 Yast.import "LanItems"
-Yast.import "NetworkInterfaces"
 
 module Y2Network
   module Sysconfig
@@ -97,11 +97,6 @@ module Y2Network
 
       # Instantiates an interface given a hash containing hardware details
       #
-      # If there is not information about the type, it will rely on NetworkInterfaces#GetTypeFromSysfs.
-      # This responsability could be moved to the PhysicalInterface class.
-      #
-      # @todo Improve detection logic according to NetworkInterfaces#GetTypeFromIfcfgOrName.
-      #
       # @param data [Hash] hardware information
       # @option data [String] "dev_name" Device name ("eth0")
       # @option data [String] "name"     Device description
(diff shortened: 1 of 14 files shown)


src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -111,27 +111,15 @@ def find_connections
...
-          type = data["type"] || Yast::NetworkInterfaces.GetTypeFromSysfs(iface.name)
           iface.renaming_mechanism = renaming_mechanism_for(iface.name)
-          iface.type = case type
-          when nil then InterfaceType::ETHERNET
-          when ::String then InterfaceType.from_short_name(type)
-          when InterfaceType then type
-          else
-            raise "Unexpected value in interface type #{type.class.inspect}:#{type.inspect}"
-          end
+          iface.type = TypeDetector.type_of(iface.name)
```
teclator: What about s390 devices. There does not have at all until enabled the group device.

jreidinger: good question, but I have no clue.

jreidinger: I suggest to postpone it till michal get back.

teclator: Something like:

```
iface.type = InterfaceType.from_short_name(data["type"]) || TypeDetector.type_of(iface.name)
```

Originally it was ```type = data["type"] || Yast::NetworkInterfaces.GetTypeFromSysfs(iface.name)``` so, in most of the cases the type was obtained from hardware. The inference of types is mainly needed for connection_configs where the interface could not be obtained from the device information but from the configuration.

src/lib/y2network/type_detector.rb
```diff
@@ -0,0 +1,141 @@
...
+  # @see Yast::NetworkInterfaces.GetType
+  class TypeDetector
+    class << self
+      # Finds type of given interface
+      #
+      # @param iface [String] interface name
+      #
+      # @return [Y2Network::InterfaceType, nil] type of given interface or nil if cannot be recognized
+      def type_of(iface)
+        type_by_sys(iface) || type_by_config(iface) || nil
```
teclator: NP: would still try to infer the type from the ifcfg-name (ifcfg-dummy0 or ifcfg-ib0 are good examples of interface where could be useful)

jreidinger: @mchf it is something for you. Please consider when you get back. I will merge it for now to prevent more collisions

mchf: there is a hook for infering type from name here: https://github.com/yast/yast-network/pull/927/files#diff-0901740064f05799d30009695e74334dR449
 It currently considers only loopback's ifcfg file but can be extended easily when needed.

PR #928, opened 2019-08-27
network-ng: rename any interface

imobachgs:
This PR is the follow-up of #921. It includes:

* Support for hotplug devices
* EditNicName dialog is replaced by a set of smaller widgets
* Improved udev rules according to feedback from the wicked team

Before merging, I would like to clean-up old ifcfg files and set down the link when renaming an interface.

Still pending:

- [ ] Rename routes
- [ ] Do not overwrite custom rules


src/lib/y2network/dialogs/rename_interface.rb
```diff
@@ -0,0 +1,96 @@
...
+    # This dialog allows the user to rename a network interface
+    #
+    # Is works in a slightly different way depending on the interface.
+    #
+    # * For physical interfaces, it allows the user to select between using
+    #   the MAC adddress or the Bus ID, which are present in the Hwinfo object
+    #   associated to the interface.
+    # * For not connected interfaces ({FakeInterface}), as the hardware is not present,
+    #   the user must specify the MAC or the Bus ID by hand (NOT IMPLEMENTED YET).
+    # * For virtual interfaces, like bridges, only the name can be chaned (no hardware
```
jreidinger: do we need for this third case this dialog at all? I though that simply allow to specify name in general tab is enough, not?

src/lib/y2network/dialogs/rename_interface.rb
```diff
@@ -0,0 +1,96 @@
...
+        @old_name = interface.name
+      end
+
+      # Runs the dialog
+      def run
+        ret = super
+        return unless ret == :ok
+        renaming_mechanism = rename_hwinfo_widget.value unless virtual_interface?
+        @builder.rename_interface(name_widget.value, renaming_mechanism)
+        name_widget.value
```
jreidinger: if it is intention to return new name which is different to common run method, please document it. Also document that it returns nil if not confirmed. And last but not least, please also document that modify builder.

src/lib/y2network/hwinfo.rb
```diff
@@ -25,11 +25,68 @@ module Y2Network
...
+
+      # Returns hardware information for the given device
+      #
+      # It relies on the {Yast::LanItems} module.
+      #
+      # @param name [String] Interface's name
+      # @return [Hwinfo,nil] Hardware info or nil if not found
+      def hwinfo_from_hardware(name)
+        hw = Yast::LanItems.Hardware.find { |h| h["dev_name"] == name }
+        return nil if hw.nil?
```
jreidinger: this collides with https://github.com/yast/yast-network/pull/927/files#diff-d0c6d9ea3e0b7a970cde5a5b107ea158R124 . Ideally we should have class for that reading of hardware.

src/lib/y2network/hwinfo.rb
```diff
@@ -25,11 +25,68 @@ module Y2Network
...
+        hw["dev_port"] = raw_dev_port unless raw_dev_port.empty?
+        new(hw)
+      end
+
+      # Returns hardware information for the given device
+      #
+      # It relies on udev rules.
+      #
+      # @param name [String] Interface's name
+      # @return [Hwinfo,nil] Hardware info or nil if not found
```
jreidinger: I do not see that nil.

src/lib/y2network/hwinfo.rb
```diff
@@ -25,11 +25,68 @@ module Y2Network
...
+      end
+    end
+
+    # Constructor
+    #
+    # @param hwinfo [Hash<String,Object>] Hardware information
+    def initialize(hwinfo = {})
       # FIXME: store only what's needed.
-      @hwinfo = load_hwinfo(name)
+      @hwinfo = Hash[hwinfo.map { |k, v| [k.to_s, v] }] if hwinfo
```
jreidinger: so if hwinfo is nil, that @hwinfo is also left nil?

src/lib/y2network/hwinfo.rb
```diff
@@ -88,7 +145,7 @@ def initialize(name:)
     alias_method :name, :dev_name
 
     def exists?
-      !@hwinfo.nil?
+      !@hwinfo.empty?
```
jreidinger: and here :boom:

src/lib/y2network/udev_rule.rb
```diff
@@ -53,7 +53,15 @@ def find_for(device)
       # @param mac  [String] MAC address
       def new_mac_based_rename(name, mac)
         new_network_rule(
-          [UdevRulePart.new("ATTR{address}", "=", mac), UdevRulePart.new("NAME", "=", name)]
+          [
+            # Guard to not try to rename everything with the same MAC address (e.g. vlan devices
+            # inherit the MAC address from the underlying device).
+            UdevRulePart.new("KERNEL", "==", "eth*"),
```
jreidinger: well, problem with that rule is that it works only for Ethernet. For WIFI this won't work. And I worry also for IB and other types.

jreidinger: So guard is good, but kernel name depends on interface type.

src/lib/y2network/virtual_interface.rb
```diff
@@ -31,5 +31,14 @@ class VirtualInterface < Interface
     def self.from_connection(name, conn)
       new(name, type: conn.type)
     end
+
+    # Determines whether the interface can be renamed
+    #
+    # A virtual interface can always be renamed.
+    #
+    # @return [Boolean]
+    def can_be_renamed?
```
jreidinger: is it used?

src/lib/y2network/widgets/custom_interface_name.rb
```diff
@@ -0,0 +1,88 @@
...
+      # Saves the current value so it can be queried after the widget is closed
+      # @see CWM::AbstractWidget#init
+      def store
+        @value = value
+      end
+
+      # Current value
+      #
+      # @return [String,nil]
+      def value
```
jreidinger: I found a bit confusing to overload that value with also stored one. Usually I use different method like `result` which is later used. But just matter of taste so NP

src/lib/y2network/widgets/custom_interface_name.rb
```diff
@@ -0,0 +1,88 @@
...
+        @value || super
+      end
+
+      # The value is valid when it does not contain unexpected characters
+      # and it is not taken already.
+      #
+      # @return [Boolean]
+      #
+      # @see CWM::AbstractWidget#opt
```
jreidinger: why see opt? I do not get it. Why not validate instead of opt.

src/lib/y2network/widgets/rename_hwinfo.rb
```diff
@@ -0,0 +1,88 @@
...
+    private
+
+      def current_value
+        Yast::UI.QueryWidget(Id(:udev_type), :Value)
+      end
+
+      def radio_buttons
+        buttons = []
+        buttons << Left(RadioButton(Id(:mac), _("MAC address: %s") % @mac)) if @mac
+        buttons << Left(RadioButton(Id(:bus_id), _("BusID: %s") % @bus_id)) if @bus_id
```
jreidinger: Is there way to say that I do not want udev name and keep kernel name?

src/lib/y2network/dialogs/rename_interface.rb
imobachgs: I prefer to use the same workflow for all renames. It should be easy to change, anyway.

src/lib/y2network/dialogs/rename_interface.rb
imobachgs: I just tried to keep the same behaviour than [EditNicName](https://github.com/yast/yast-network/blob/master/src/lib/network/edit_nic_name.rb#L84). Regarding modifying the builder, I could move the change to the calling widget if I return the renaming mechanism and the name. Is it ok for you?

src/lib/y2network/hwinfo.rb
imobachgs: Yes, I agree (something like `HardwareReader`), but it was of the scope of this PR. On the other hand, I will fix the collision.

src/lib/y2network/udev_rule.rb
imobachgs: And not only that: we have detected that it won't work when using predictable names (for instance, think about openSUSE). So after discussing with @teclator, we have decided to keep the old behaviour and add a new PBI to do some research about the best way to handle it.

src/lib/y2network/widgets/custom_interface_name.rb
imobachgs: Using `result` makes sense. I will change it.

src/lib/y2network/widgets/custom_interface_name.rb
imobachgs: Oh, it was a mistake :)

src/lib/y2network/widgets/rename_hwinfo.rb
imobachgs: No, there is no way to do such a thing even in the new implementation. The problem is that you will need to write the `ifcfg-*` files but, at this point, you do not know which name the kernel well assign if you remove the rule.

src/lib/y2network/dialogs/rename_interface.rb
imobachgs: Updated. Now the dialog is only shown when the required hardware information is available.

src/lib/y2network/dialogs/rename_interface.rb
imobachgs: Refactored and simplified.

src/lib/y2network/virtual_interface.rb
imobachgs: It was :) After last night refactoring, it is not needed anymore.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -309,7 +308,7 @@ def aliases=(value)
 
     # gets interface name that will be assigned by udev
     def udev_name
-      # cannot cache as EditNicName dialog can change it
+      # cannot cache as it can be changed
```
jreidinger: can we replace that value now? by new backend instead of old one?

src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -394,6 +394,12 @@ def clean
         @defined_variables = nil
       end
 
+      # Removes the file
+      def remove
+        return unless Yast::FileUtils.Exists(path.to_s)
+        Yast::SCR.Execute(Yast::Path.new(".target.remove"), path.to_s)
+      end
```
jreidinger: :+1:

src/lib/y2network/sysconfig/interfaces_writer.rb
```diff
@@ -54,10 +57,40 @@ def udev_rule_for(iface)
...
+        # and rcnetwork reload actually works (bnc#749365)
+        Yast::Execute.on_target("/usr/bin/udevadm", "settle")
+        sleep(1)
+      end
+
+      # Cleans and shutdowns renamed interfaces
+      #
+      # @param interfaces [InterfacesCollection] Interfaces
+      def clean_up_old_interfaces(interfaces)
+        interfaces.to_a.select(&:old_name).each do |iface|
```
jreidinger: what happens if I switch names. So I have eth0 and eth1 and I change it to eth1 and eth0? This looks like it it is removed  and shutted down, not?

src/lib/y2network/sysconfig/interfaces_writer.rb
```diff
@@ -54,10 +57,40 @@ def udev_rule_for(iface)
...
+          ifroute = Y2Network::Sysconfig::RoutesFile.find(iface.old_name)
+          ifroute && ifroute.remove
+        end
+      end
+
+      # Sets the interface down
+      #
+      # @param iface_name [String] Interface's name
+      def set_interface_down(iface_name)
+        Yast::Execute.on_target("/sbin/ifdown", iface_name) unless Yast::Mode.autoinst
```
jreidinger: this is really old code. It should use ip or wicked for shutting down interface.

src/lib/y2network/sysconfig/interfaces_writer.rb
imobachgs: Yes, that's it. And you need to do so in order to rename the interfaces later (if the interface is still up, it won't be renamed).

src/lib/y2network/interface_config_builder.rb
imobachgs: Yes, you are right. I will have a look and check whether we can drop this old line of code.

src/lib/y2network/sysconfig/interfaces_writer.rb
imobachgs: You are right. But I would prefer to address it later. I will make a note.

PR #930, opened 2019-08-29
network-ng: keep custom udev rules

imobachgs:
Do not overwrite rules for unknown devices (an unknown device is that interface which is not mentioned in the configuration -not even in renames, e.g. a hotplug device with no configuration file-).

Take into account that `.udev_persistent.net` only reads `NAME=` rules (ignoring the rest). The good thing is that, when writing, the "rest" is kept as it is (and only `NAME=` rules are replaced).

### DIFF
--- a/test/y2network/sysconfig/interfaces_writer_test.rb
+++ b/test/y2network/sysconfig/interfaces_writer_test.rb
@@ -41,6 +41,7 @@ describe Y2Network::Sysconfig::InterfacesWriter do
     before do
       allow(Yast::Execute).to receive(:on_target)
       allow(eth0).to receive(:hardware).and_return(hardware)
+      allow(writer).to receive(:sleep)
     end
 
     around do |example|
--- a/test/y2network/sysconfig/interfaces_writer_test.rb
+++ b/test/y2network/sysconfig/interfaces_writer_test.rb
@@ -111,6 +111,21 @@ describe Y2Network::Sysconfig::InterfacesWriter do
           subject.write(interfaces)
         end
       end
+
+      context "when there is some rule for an unknown interface" do
+        let(:unknown_rule) { Y2Network::UdevRule.new_mac_based_rename("unknown", "00:11:22:33:44:55:66") }
+
+        before do
+          allow(Y2Network::UdevRule).to receive(:all).and_return([unknown_rule])
+        end
+
+        it "keeps the rule" do
+          expect(Y2Network::UdevRule).to receive(:write) do |rules|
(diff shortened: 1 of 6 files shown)


test/y2network/sysconfig/interfaces_writer_test.rb
```diff
@@ -111,6 +112,21 @@
...
       end
+
+      context "when there is some rule for an unknown interface" do
+        let(:unknown_rule) { Y2Network::UdevRule.new_mac_based_rename("unknown", "00:11:22:33:44:55:66") }
+
+        before do
+          allow(Y2Network::UdevRule).to receive(:all).and_return([unknown_rule])
+        end
+
+        it "keeps the rule" do
```
jreidinger: can you try some specific rule that is not created by yast? like based on something that we do not use like KERNEL assigned name or whatever?

imobachgs: I will improve the test to really write to the filesystem so we can check what is the real result.

imobachgs: `.udev_persistent.net` is not writing to the scr_root. I might need to investigate why. Anyway, I've tried this example:

```
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{type}=="1", TAG="XX", ATTR{dev_id}=="0x0"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", KERNEL=="eth*", TAG=="tagged", NAME="unknown"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{type}=="1", ATTR{dev_id}=="0x0", ATTR{address}=="52:54:00:68:54:fb", NAME="eth5"
```

And in this case, only the last rule is rewritten.

imobachgs: > And in this case, only the last rule is rewritten.

I mean, the last one was rewritten, and the rest were kept.

jreidinger: agent probably does not use scr_root, because originally yast uses agent on target chroot, so it do chroot and call own agent that is perl based. But then there was feature that yast has to be able to install system without yast on target system. So we change way how agents are called and instead agent get parameter that specify changed root. But probably this agent is not adapted (when digging in history it looks like it happens many years ago :)

jreidinger: btw this commit should add term that set root https://github.com/yast/yast-core/commit/6cc105daf7f6bc3cca2692a661282fdf2dd67b46

PR #939, opened 2019-09-03
network-ng: drop fake interface

imobachgs:
This method improves `Y2Network::Hwinfo` API adding a `#netcards` method and implementing memoization. It does not have any performance impact because, at this time, the hardware info is stored in `LanItems` anyway.

Additionally, the hardware info is passed to `PhysicalInterface` through its constructor. It solves a problem where, although we had the hardware info for unnamed interfaces, the `PhysicalInterface` was unable to find it because of the missing name.


src/lib/y2network/hwinfo.rb
```diff
@@ -184,6 +237,17 @@ def drivers
...
 
+    # Determines whether the hardware is available (plugged)
+    #
+    # If the hardware layer was able to get its type, it consider the hardware to be connected. Bear
+    # in mind that it is not possible to just rely in #exists? because it could include some info
+    # from udev rules.
+    #
+    # @return [Boolean]
+    def present?
+      !!type
```
jreidinger: is type really so deterministic? I though that type could be part of udev rule, not? What about dev_name?

imobachgs: I would say that the `type` is always there (but perhaps there is a better candidate) and, from udev, we are only able to infer the bus ID and the MAC. On the other hand, we want to avoid the dev_name because it might be missing (s390) and that's the main problem we try to address.

PR #940, opened 2019-09-03
network-ng: add virtual devices

imobachgs:
This PR fixes these issues:

* Do not rely on LanItems to add an interface (hence do not crash on `LanItems.update_routing_devices?` method anymore).
* Correctly displays virtual devices in the interfaces table.
* Use the correct interface builder for bridges (and s390 types).


src/lib/y2network/config.rb
```diff
@@ -150,6 +151,17 @@ def delete_interface(name)
       interfaces.reject! { |i| i.name == name }
     end
 
+    # Adds or update a connection config
+    #
+    # If it is a virtual connection, it adds interface if it does not exist.
```
jreidinger: what happens with "fake" interface? should not be also added here?

imobachgs: Good point. I will fix it.

jreidinger: still it does not mention here that it can also create not attached physical interface. Basically it always create interface if it does not exist already.

src/lib/y2network/sysconfig/connection_config_readers/bridge.rb
```diff
@@ -29,7 +29,7 @@ class Bridge < Base
 
         # @see Y2Network::Sysconfig::ConnectionConfigReaders::Base#update_connection_config
         def update_connection_config(conn)
-          conn.ports = file.bridge_ports.split(" ")
+          conn.ports = file.bridge_ports.to_s.split(" ")
```
jreidinger: hmm, so this to_s is to avoid nil?

PR #941, opened 2019-09-03
implement changing hostname for interface

jreidinger:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -328,13 +332,25 @@ def subnet_prefix=(value)
...
     end
 
     # @param [String] value
     def hostname=(value)
-      # TODO: write it
+      @hostname = value
+    end
+
+    def save_hostname
+      # avoid unncessary modification
```
imobachgs: typo:  unnecessary

PR #942, opened 2019-09-03
network-ng: builder fixes

imobachgs:
Several unrelated fixes:

* Do not duplicate interfaces when an already existing one is renamed.
* Use the correct list of interfaces when determining which of them can be used in a bridge or a bonding.
* Take the boot protocol into account when writing the hostname.

### DIFF
--- a/src/lib/y2network/interface_config_builder.rb
+++ b/src/lib/y2network/interface_config_builder.rb
@@ -343,13 +343,16 @@ module Y2Network
       @hostname = value
     end
 
+    # Saves the hostname
+    #
+    # It needs to take into account whether the old configuration and the boot protocol.
     def save_hostname
-      # avoid unnecessary modification
-      return if @original_ip_config == @connection_config.ip && @original_hostname == hostname
+      if !required_ip_config? || hostname.empty?
+        Yast::Host.remove_ip(@original_ip_config.address.to_s)
+        return
+      end
 
-      # remove old ip
-      Yast::Host.remove_ip(@original_ip_config.address.to_s) if @original_ip_config != @connection_config.ip
-
+      return if @original_ip_config == connection_config.ip && @original_hostname == hostname
       Yast::Host.Update(@original_hostname, hostname, @connection_config.ip.address.to_s)
     end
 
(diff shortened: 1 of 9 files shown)


src/lib/y2network/interface_config_builder.rb
```diff
@@ -343,13 +343,16 @@ def hostname=(value)
       @hostname = value
     end
 
+    # Saves the hostname
+    #
+    # It needs to take into account whether the old configuration and the boot protocol.
```
jreidinger: this sentence is hard to understand for me. Maybe mention that it needs to clear old hostname entry and set a new one even when ip and hostname changes.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -343,13 +343,16 @@ def hostname=(value)
...
     def save_hostname
-      # avoid unnecessary modification
-      return if @original_ip_config == @connection_config.ip && @original_hostname == hostname
-
-      # remove old ip
-      Yast::Host.remove_ip(@original_ip_config.ip.address.to_s) if @original_ip_config != @connection_config.ip
+      if !required_ip_config? || hostname.empty?
```
jreidinger: are you sure that it covers case when ip is different and hostname is different? I worry this handle only case when static to dynamic or hostname delete. And in fact Update handles also empty hostname and properly remove it. I think we definitivelly need to document all possible cases.

src/lib/y2network/interface_config_builder.rb
```diff
@@ -436,11 +439,25 @@ def find_interface
       yast_config.interfaces.by_name(name)
     end
 
+    # Returns the interfaces collection
+    #
+    # @return [Y2Network::InterfacesCollection]
+    def interfaces
+      yast_config.interfaces
```
jreidinger: when you add this method ( good one ), please also try to replace all `yast_config.interfaces` in file.

test/y2network/interface_config_builder_test.rb
```diff
@@ -161,4 +161,45 @@
...
+        config_builder.boot_protocol = "none"
+        config_builder.ip_address = "192.168.122.1"
+      end
+
+      it "removes old hostname if it exists" do
+        expect(Yast::Host).to receive(:remove_ip)
+        config_builder.save_hostname
+      end
+    end
+  end
```
jreidinger: :+1:

src/lib/y2network/interface_config_builder.rb
imobachgs: I've added some test cases and fixed a few issues because even the comparisons where not working at all.

PR #943, opened 2019-09-04
autoyast interfaces

jreidinger:
It is not in perfect shape, but it can do initial export/import which is better then now.

### DIFF
--- a/src/clients/lan_auto.rb
+++ b/src/clients/lan_auto.rb
@@ -133,50 +133,10 @@ module Yast
     # @return [Hash] autoyast network settings
     def ToAY(settings)
       settings = deep_copy(settings)
-      interfaces = []
-      discard = ["UDI", "_nm_name"]
-      Builtins.y2milestone("Devices: #{settings["devices"].inspect})")
-      Builtins.foreach(settings.fetch("devices", {}).fetch("interfaces", [])) do |devsmap|
-        newmap = {}
-        Builtins.foreach(devsmap) do |key, val|
-          Builtins.y2milestone("Adding: %1=%2", key, val)
-          if key != "_aliases"
-            if Ops.greater_than(Builtins.size(Convert.to_string(val)), 0) &&
-                !Builtins.contains(discard, key) &&
-                !Builtins.contains(discard, Builtins.tolower(key))
-              Ops.set(newmap, Builtins.tolower(key), Convert.to_string(val))
-            end
-          else
-            # handle aliases
-            Builtins.y2debug("val: %1", val)
-            # if aliases are empty, then ommit it
-            if Ops.greater_than(Builtins.size(Convert.to_map(val)), 0)
-              # replace key "0" into "alias0" (bnc#372678)
-              Builtins.foreach(
-                Convert.convert(
-                  val,
-                  from: "any",
-                  to:   "map <string, map <string, any>>"
-                )
-              ) do |k, v|
-                Ops.set(
-                  newmap,
-                  Builtins.tolower("aliases"),
-                  Builtins.add(
-                    Ops.get_map(newmap, Builtins.tolower("aliases"), {}),
-                    Builtins.sformat("alias%1", k),
-                    v
-                  )
-                )
-              end
-            end
-          end
-        end
-        interfaces = Builtins.add(interfaces, newmap)
-      end
+      interfaces = settings["interfaces"] || []
+      Builtins.y2milestone("interfaces: #{interfaces.inspect})")
 
       # Modules
-
       s390_devices = []
       Builtins.foreach(Ops.get_map(settings, "s390-devices", {})) do |_device, mod|
         s390_devices = Builtins.add(s390_devices, mod)
--- a/src/clients/lan_auto.rb
+++ b/src/clients/lan_auto.rb
@@ -135,50 +135,44 @@ module Yast
       settings = deep_copy(settings)
       interfaces = []
       discard = ["UDI", "_nm_name"]
-      Builtins.foreach(Ops.get_map(settings, "devices", {})) do |_type, devsmap|
-        Builtins.foreach(
-          Convert.convert(devsmap, from: "map", to: "map <string, map>")
-        ) do |device, devmap|
-          newmap = {}
-          Builtins.foreach(
-            Convert.convert(devmap, from: "map", to: "map <string, any>")
-          ) do |key, val|
-            Builtins.y2milestone("Adding: %1=%2", key, val)
-            if key != "_aliases"
-              if Ops.greater_than(Builtins.size(Convert.to_string(val)), 0) &&
-                  !Builtins.contains(discard, key) &&
-                  !Builtins.contains(discard, Builtins.tolower(key))
-                Ops.set(newmap, Builtins.tolower(key), Convert.to_string(val))
-              end
-            else
-              # handle aliases
-              Builtins.y2debug("val: %1", val)
-              # if aliases are empty, then ommit it
-              if Ops.greater_than(Builtins.size(Convert.to_map(val)), 0)
-                # replace key "0" into "alias0" (bnc#372678)
-                Builtins.foreach(
-                  Convert.convert(
-                    val,
-                    from: "any",
-                    to:   "map <string, map <string, any>>"
+      Builtins.y2milestone("Devices: #{settings["devices"].inspect})")
+      Builtins.foreach(settings.fetch("devices", {}).fetch("interfaces", [])) do |devsmap|
+        newmap = {}
+        Builtins.foreach(devsmap) do |key, val|
+          Builtins.y2milestone("Adding: %1=%2", key, val)
+          if key != "_aliases"
+            if Ops.greater_than(Builtins.size(Convert.to_string(val)), 0) &&
+                !Builtins.contains(discard, key) &&
+                !Builtins.contains(discard, Builtins.tolower(key))
+              Ops.set(newmap, Builtins.tolower(key), Convert.to_string(val))
+            end
+          else
+            # handle aliases
+            Builtins.y2debug("val: %1", val)
+            # if aliases are empty, then ommit it
+            if Ops.greater_than(Builtins.size(Convert.to_map(val)), 0)
+              # replace key "0" into "alias0" (bnc#372678)
+              Builtins.foreach(
+                Convert.convert(
+                  val,
+                  from: "any",
+                  to:   "map <string, map <string, any>>"
+                )
+              ) do |k, v|
+                Ops.set(
+                  newmap,
+                  Builtins.tolower("aliases"),
+                  Builtins.add(
+                    Ops.get_map(newmap, Builtins.tolower("aliases"), {}),
+                    Builtins.sformat("alias%1", k),
+                    v
                   )
-                ) do |k, v|
-                  Ops.set(
-                    newmap,
-                    Builtins.tolower("aliases"),
-                    Builtins.add(
-                      Ops.get_map(newmap, Builtins.tolower("aliases"), {}),
-                      Builtins.sformat("alias%1", k),
-                      v
-                    )
-                  )
-                end
+                )
               end
             end
           end
-          newmap["device"] = device
-          interfaces = Builtins.add(interfaces, newmap)
         end
+        interfaces = Builtins.add(interfaces, newmap)
       end
 
       # Modules
--- a/src/lib/y2network/autoinst/config_reader.rb
+++ b/src/lib/y2network/autoinst/config_reader.rb
@@ -44,13 +44,9 @@ module Y2Network
 
       # @return [Y2Network::Config] Network configuration
       def config
-        attrs = { source: :sysconfig }
-        # How it works? For autoyast it reads current config and apply autoyast on top of it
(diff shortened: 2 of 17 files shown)


src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -0,0 +1,320 @@
...
+      def self.attributes
+        [
+          { name: :bootproto },
+          { name: :broadcast },
+          { name: :device },
+          { name: :name }, # has precedence over device
+          { name: :ipaddr },
+          { name: :remote_ipaddr },
+          { name: :netmask },
+          { name: :network }, # TODO: what it is? looks like ipaddr with applied prefix
```
mchf: it can be "network address" - exactly what is in the comment. However I've never seen it in any profile, so what about trying to remove it silently ?

src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -0,0 +1,320 @@
...
+          { name: :bonding_slave0 },
+          { name: :bonding_slave1 },
+          { name: :bonding_slave2 },
+          { name: :bonding_slave3 },
+          { name: :bonding_slave4 },
+          { name: :bonding_slave5 },
+          { name: :bonding_slave6 },
+          { name: :bonding_slave7 },
+          { name: :bonding_slave8 },
+          { name: :bonding_slave9 },
```
mchf: isn't it nice in comparison to boring ```bridge_ports``` ? :-/

jreidinger: both are horrible. It should be array. But I cannot change it due to backward compatibility.

src/lib/y2network/autoinst_profile/interface_section.rb
jreidinger: Well, my plan is to ignore it for now, but keep it in rnc for backward compatibility. That also reminds me that we should improve official SLE documentation for autoyast network as current one is really insuficient.

src/lib/y2network/autoinst/config_reader.rb
```diff
@@ -44,21 +45,28 @@ def initialize(section)
...
 
     private
 
       # Returns an interfaces reader instance
       #
       # @return [SysconfigInterfaces] Interfaces reader
-      def interfaces_reader
-        @interfaces_reader ||= Y2Network::Sysconfig::InterfacesReader.new
+      def system_interfaces
+        @system_interfaces ||= Y2Network::Sysconfig::InterfacesReader.new
```
imobachgs: We are coupling AutoYaST import process to sysconfig. Another approach could be to read the system configuration and pass that `Y2Network::Config` object, so AutoYaST can read the interfaces from it (I am not sure if, in `Mode.autoinst`, that configuration is even accessible via `Yast::Lan.system_config`).

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,66 @@
...
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2network/dns"
+require "ipaddr"
+Yast.import "IP"
+
+module Y2Network
+  module Autoinst
+    # This class is responsible of importing the AutoYast dns section
```
imobachgs: I guess the description is wrong ("dns section").

src/lib/y2network/autoinst/config_reader.rb
jreidinger: I am also not so sure about this approach. On other hand I found quite important to somehow match system settings and on top of it apply autoyast settings.

src/lib/y2network/autoinst/interfaces_reader.rb
jreidinger: yes, copy and paste

src/lib/y2network/autoinst/config_reader.rb
imobachgs: Yes, but my point is that this class could get current system settings from outside.

jreidinger: so maybe we should try to first get config from Lan.system_config and if failed load it?

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,150 @@
...
+
+        ConnectionConfigsCollection.new(configs)
+      end
+
+    private
+
+      def create_config(interface_section)
+        # TODO: autoyast backend for type detector?
+        # TODO: TUN/TAP interface missing for autoyast?
+        return ConnectionConfig::Bonding.new if interface_section.bonding_slave0 && !interface_section.bonding_slave0.empty?
```
imobachgs: In Rails, there are a pair of methods to deal with these situations ([#blank?](https://apidock.com/rails/Object/blank%3F) and [#present?](https://apidock.com/rails/Object/present%3F)). I am starting to think about using a similar approach but using refinements (to not pollute the String class).

imobachgs: Probably an `if-else` construction would be more readable (and you could get rid of all those `return` keywords).

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,150 @@
...
+
+        config.startmode = Startmode.create(interface_section.startmode) if interface_section.startmode
+        config.startmode.priority = interface_section.ifplugd_priority if config.startmode.name == "ifplugd" && interface_section.ifplugd_priority
+        config.mtu = interface_section.mtu.to_i if interface_section.mtu
+        config.ethtool_options = interface_section.ethtool_options if interface_section.ethtool_options
+        config.firewall_zone = interface_section.zone if interface_section.zone
+      end
+
+      def load_wireless(config, interface_section)
+        config.mode = interface_section.wireless_mode if interface_section.wireless_mode
```
imobachgs: np: What about having a hash to make the correspondence?

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,150 @@
...
+        config.auth_mode = interface_section.wireless_auth_mode.to_sym if interface_section.wireless_auth_mode
+        config.nick = interface_section.wireless_nick if interface_section.wireless_nick
+        config.nwid = interface_section.wireless_nwid if interface_section.wireless_nwid
+        config.wpa_anonymous_identity = interface_section.wireless_wpa_anonid if interface_section.wireless_wpa_anonid
+        config.wpa_identity = interface_section.wireless_wpa_identity if interface_section.wireless_wpa_identity
+        config.wpa_password = interface_section.wireless_wpa_password if interface_section.wireless_wpa_password
+        config.wpa_psk = interface_section.wireless_wpa_psk if interface_section.wireless_wpa_psk
+        config.keys = []
+        (0..3).each do |i|
+          key = interface_section.public_send(:"wireless_key#{i}")
```
imobachgs: The section object could have a `#wireless_keys` method that abstracts these logic details.

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,150 @@
...
+      def load_bridge(config, interface_section)
+        config.ports = interface_section.bridge_ports.split
+        config.stp = interface_section.bridge_stp == "on" if interface_section.bridge_stp
+        config.forward_delay = interface_section.bridge_forward_delay.to_i if interface_section.bridge_forward_delay
+      end
+
+      def load_bonding(config, interface_section)
+        config.options = interface_section.bonding_module_opts if interface_section.bonding_module_opts
+        config.slaves = []
+        (0..9).each do |i|
```
imobachgs: Same than above. Having some kind of `#bonding_slaves` method would be nice.

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -0,0 +1,150 @@
...
+
+      def create_config(interface_section)
+        # TODO: autoyast backend for type detector?
+        # TODO: TUN/TAP interface missing for autoyast?
+        return ConnectionConfig::Bonding.new if interface_section.bonding_slave0 && !interface_section.bonding_slave0.empty?
+        return ConnectionConfig::Bridge.new if interface_section.bridge_ports && !interface_section.bridge_ports.empty?
+        return ConnectionConfig::Vlan.new if interface_section.etherdevice && !interface_section.etherdevice.empty?
+        return ConnectionConfig::Wireless.new if interface_section.wireless_essid && !interface_section.wireless_essid.empty?
+
+        ConnectionConfig::Ethernet.new # TODO: use type detector to read it from sys
```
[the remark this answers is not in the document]
jreidinger: well, it makes sense only if interface already exists and have matching name. Current TypeDetector use as one source data from sysconfig see https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/type_detector.rb#L134 and https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/sysconfig/type_detector.rb#L33

So here I need to ideally create autoyast specific type detector that do stuff like I did above.

src/lib/y2network/autoinst/interfaces_reader.rb
jreidinger: Actually it is maybe not needed as I did initialize at the start of section. So everything should be set to "" - https://github.com/yast/yast-network/pull/943/files#diff-d2f2fbfa1adfc18366a8b1bbfc5b22e7R247

I have to test it without this nil check

src/lib/y2network/autoinst/interfaces_reader.rb
jreidinger: I am a bit lost. What you mean here?

src/lib/y2network/autoinst/interfaces_reader.rb
jreidinger: you mean as helper? makes sense.

src/lib/y2network/autoinst/config_reader.rb
mchf: I'd say we should use common access to system config as we do always - AY do not need to know that it is (e.g.) sysconfig source. AY reader should also deal with differences as read from the profile and process the merge of configurations as done in old code.

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: May be not in this case but in general the @imobachgs proposal makes sense. Checks like if !nil && !empty? are quite common.

src/lib/y2network/autoinst_profile/interface_section.rb
mchf: I know, I was just crying.

src/lib/y2network/autoinst/interfaces_reader.rb
jreidinger: @mchf well, still I do not like it. We should unify what is default value, if empty string or nil. Having both is very confusing and error prone.

src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -290,6 +290,31 @@ def init_from_config(config)
         true
       end
 
+      # Helper to get wireless keys as array
+      # @return [Array<String>]
+      def wireless_keys
+        keys = []
+        (0..3).each do |i|
+          key = public_send(:"wireless_key#{i}")
+          keys << key unless key.empty?
```
mchf: key can be even nil when one of wireless_keyX values is missing. Or not?

mchf: ok, seems I was wrong bcs of a to_s conversion.

src/lib/y2network/autoinst/type_detector.rb
```diff
@@ -0,0 +1,54 @@
...
+require "y2network/type_detector"
+require "y2network/sysconfig/interface_file"
+
+module Y2Network
+  module Autoinst
+    # Detects type of given interface based on autoyast profile
+    class TypeDetector < Y2Network::TypeDetector
+      class << self
+        def type_of(iface, section)
+          type_by_sys(iface) || type_by_config(section) || InterfaceType::ETHERNET
```
mchf: I'm thinking if we should try to do a collision detection here (e.g. device is ethernet according to the /sys but something else according to the profile) and change the precendence (prefer type from AY profile over the /sys one) in such case.

jreidinger: well, detection by sys is used mainly for physical interfaces and I am not sure how we can change it. How we can change ib to eth?

src/clients/lan_auto.rb
```diff
@@ -133,56 +133,10 @@ def main
...
-                end
-              end
-            end
-          end
-          newmap["device"] = device
-          interfaces = Builtins.add(interfaces, newmap)
-        end
-      end
+      interfaces = settings["interfaces"] || []
+      Builtins.y2milestone("interfaces: #{interfaces.inspect})")
```
imobachgs: np: why not `log.info`?

jreidinger: well, I just adapt existing code. As it is untested client I try to do only minimal modifications.

PR #944, opened 2019-09-04
network-ng: fix dhclient set hostname handling

imobachgs:
Fixes proper support for DHCLIENT_SET_HOSTNAME. There is one drawback: if more than one interface has the `DHCLIENT_SET_HOSTNAME` set to `yes`, YaST will not warn the user. It will simply write the first one when writing changes to disk. I meant [this message](https://github.com/yast/yast-network/blob/e1f0db2fecf41f773856767629c3c093e1b1a0c8/src/include/network/routines.rb#L872).

I decided to put this setting into `Y2Network::DNS#dhcp_hostname` instead of adding it to `ConnectionConfig::Base` because it looks more natural to me. Of course, it can be changed.

### DIFF
--- a/src/lib/y2network/sysconfig/connection_config_writers/base.rb
+++ b/src/lib/y2network/sysconfig/connection_config_writers/base.rb
@@ -45,7 +45,9 @@ module Y2Network
           file.lladdr = conn.lladdress
           file.startmode = conn.startmode.to_s
           file.ifplugd_priority = conn.startmode.priority if conn.startmode.name == "ifplugd"
-          file.ethtool_options = conn.ethtool_options unless conn.ethtool_options.empty?
+          if conn.ethtool_options && !conn.ethtool_options.empty?
+            file.ethtool_options = conn.ethtool_options
+          end
           file.zone = conn.firewall_zone
           add_ips(conn)
           update_file(conn)
--- a/src/lib/y2network/sysconfig/dns_reader.rb
+++ b/src/lib/y2network/sysconfig/dns_reader.rb
@@ -18,6 +18,7 @@
 # find current contact information at www.suse.com.
 require "yast"
 require "y2network/dns"
+require "y2network/sysconfig/interface_file"
 
 Yast.import "Hostname"
 Yast.import "Mode"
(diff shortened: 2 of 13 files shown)


src/lib/y2network/sysconfig/connection_config_writers/base.rb
```diff
@@ -45,7 +45,9 @@ def write(conn)
           file.lladdr = conn.lladdress
           file.startmode = conn.startmode.to_s
           file.ifplugd_priority = conn.startmode.priority if conn.startmode.name == "ifplugd"
-          file.ethtool_options = conn.ethtool_options unless conn.ethtool_options.empty?
+          if conn.ethtool_options && !conn.ethtool_options.empty?
```
jreidinger: hmm, I start worry that we start having it all over. Why simply not allowing nil and also assign at least `""` which is default?

src/lib/y2network/sysconfig/dns_reader.rb
```diff
@@ -126,10 +127,15 @@ def searchlist
...
-      # @return [Boolean]
+      # @return [String,:any,:none] Interface to set the hostname based on DHCP settings;
+      #   :any for any interface; :none for ignoring the hostname assigned via DHCP
       def dhcp_hostname
         value = Yast::SCR.Read(Yast::Path.new(".sysconfig.network.dhcp.DHCLIENT_SET_HOSTNAME"))
-        value == "yes"
+        return :any if value == "yes"
+        files = InterfaceFile.all
+        files.each(&:load)
+        file = files.find { |f| f.dhclient_set_hostname == "yes" }
```
jreidinger: NP: more effective will be to have load in find, so if there are 100 files and first contain this entry, it will be loaded faster.

src/lib/y2network/sysconfig/connection_config_writers/base.rb
imobachgs: For me, a `nil` means that the key is not in the file and an empty string that it is there but with no value. Do we need to make such a distinction? Probably no, so we could simplify as you have proposed and do not write anything with empty values.

PR #945, opened 2019-09-05
network-ng: fix autoyast dns handling

imobachgs:


src/lib/y2network/hostname_reader.rb
```diff
@@ -0,0 +1,97 @@
...
+
+      host, _domain = *Yast::Hostname.SplitFQ(fqdn)
+      host.empty? ? nil : host
+    end
+
+    # Reads the hostname from the underlying system
+    #
+    # @return [String] Hostname
+    def hostname_from_system
+      Yast::Execute.on_target!("/bin/hostname", "--fqdn", stdout: :capture)
```
jreidinger: I think now prefered is to run it as `/usr/bin/hostname`

src/lib/y2network/autoinst/dns_reader.rb
```diff
@@ -40,9 +41,9 @@ def initialize(section)
       def config
         Y2Network::DNS.new(
           dhcp_hostname:      section.dhcp_hostname,
-          hostname:           section.hostname,
+          hostname:           section.hostname || default_hostname,
           nameservers:        valid_ips(section.nameservers),
-          resolv_conf_policy: section.resolv_conf_policy,
+          resolv_conf_policy: section.resolv_conf_policy || "auto",
```
kobliha: Ideally not a string, but a constant (and used elsewhere too)

src/lib/y2network/hostname_reader.rb
```diff
@@ -0,0 +1,97 @@
...
+    def hostname_from_system
+      Yast::Execute.on_target!("/usr/bin/hostname", "--fqdn", stdout: :capture).strip
+    rescue Cheetah::ExecutionFailed
+      name = Yast::SCR.Read(Yast::Path.new(".target.string"), "/etc/hostname").to_s.strip
+      name.empty? ? nil : name
+    end
+
+    # @return [Array<String>] Valid chars to be used in the random part of a hostname
+    HOSTNAME_CHARS = (("a".."z").to_a + ("0".."9").to_a).freeze
+    private_constant :HOSTNAME_CHARS
```
kobliha: Looks like Punycode is not supported, see https://en.wikipedia.org/wiki/Punycode or https://es.wikipedia.org/wiki/Punycode

kobliha: Hmm, taking back, this seems to be used just for random hostname, not for validation...

imobachgs: These characters are only used to generate a random hostname (`linux-ab3d`, for instance).

src/lib/y2network/autoinst/dns_reader.rb
imobachgs: Yes, actually we might think about having an enum, but it is basically only used here. I will move it to a constant though.

PR #947, opened 2019-09-10
network-ng: test cases document (draft)

imobachgs:


teclator:
@mchf @imobachgs I guess this is still a WIP, so we shall continue improving it with new cases isn't it?, In that case would be nice to cover the cases that affected the hostname configuration during installation #1000 .

jreidinger:
lets merge it as it is and we can improve it later, as it makes no sense to just keep it open.

teclator (approved):
LGTM

PR #948, opened 2019-09-10
Small fixes

imobachgs:


src/lib/y2network/config.rb
```diff
@@ -140,10 +140,12 @@ def rename_interface(old_name, new_name, mechanism)
...
     # deletes interface and all its config. If interface is physical,
     # it is not removed as we cannot remove physical interface.
+    #
+    # @param name [String] Interface's name
     def delete_interface(name)
-      connections.reject! { |c| c.name == name }
+      connections.reject! { |c| c.interface == name }
       interface = interfaces.by_name(name)
-      return if interface.hardware.exists?
+      return unless interface.is_a?(VirtualInterface)
```
jreidinger: why only for VirtualInterfaces? We want to delete also hotplug interface, no? Otherwise we cannot delete it if we previously create it.

PR #949, opened 2019-09-11
Merged SLE-12-SP5

mchf:


package/yast2-network.changes
```diff
@@ -1,3 +1,17 @@
+-------------------------------------------------------------------
+Tue Sep 10 07:40:04 UTC 2019 - Michal Filka <mfilka@suse.com>
+
+- bnc#1149234
+  - apply udev rule from AY profile according to device's mac
+    value when permanent_mac is missing in list of the device's
+    options
+- backport: - change proposal naming and add option to set cpu
+  mitigations (bsc#1128707)
```
jreidinger: this should not go there as it modify installer and have other dependencies. It was exclusivelly for SP5 and SP1.

jreidinger: I see that change is not applied, but it also should not be mentioned in changes.

PR #955, opened 2019-09-12
Autoyast udev rules

jreidinger:


src/lib/y2network/config.rb
```diff
@@ -162,6 +162,7 @@ def add_or_update_connection_config(connection_config)
       connections.add_or_update(connection_config)
       interface = interfaces.by_name(connection_config.interface)
       return if interface
+      log.info "Creating new interface"
```
imobachgs: What about logging some useful info from the connection config object? Unless there is no useful info at this point...

jreidinger: only useful info is already logged few lines above[1], so I do not want to write it again to log. I use it mainly to debug that duplicate interfaces after importing configs.

jreidinger: an [1] https://github.com/yast/yast-network/pull/955/files#diff-28640d69554356babb600a69c10ad80bR161

imobachgs: Then it is perfectly fine.

PR #956, opened 2019-09-13
Network ng write drivers options

imobachgs:
This PR adds support for drivers handling. It is rather long, and it includes these changes:

* `Y2Network::Config` includes now a list of drivers (`#drivers`). The reason to keep a unified list if that drivers options are shared for all interfaces.
* A new widget to specify driver name and options have been added.
* When an interface has a custom driver (a driver specified by the user), it is written to `/etc/udev/rules.d/79-yast2-drivers.rules`.
* Drivers options are placed at `/etc/modprobe.d/50-yast.conf`.

# TODO

- [ ] Rename Driver#params to Driver#options (to keep consistency with the rest).
- [ ] Keep driver options when the selection changes. For instance, let's say that you modify "virtio_net" and set options to "csum=1". If you change your selection, these options are lost.
- [ ] Drop udev code from the old backend.

### DIFF
--- a/src/lib/y2network/driver.rb
+++ b/src/lib/y2network/driver.rb
@@ -33,6 +33,19 @@ module Y2Network
         params_string = params.map { |k, v| "#{k}=#{v}" }.join(" ")
         new(name, params_string)
       end
+
+      # Writes driver options to the underlying system
+      #
+      # @param drivers [Y2Network::Driver] Drivers to write options
+      def write_options(drivers)
+        drivers.each(&:write_options)
+        commit
+      end
+
+      # Commits drivers options to disk
+      def commit
+        Yast::SCR.Write(Yast::Path.new(".modules"), nil)
+      end
     end
 
     # @return [String] Kernel module name
@@ -57,8 +70,16 @@ module Y2Network
       name == other.name && params == other.params
     end
 
-    # Saves driver parameters to the underlying system
-    def save
+    # Adds driver parameters to be written to the underlying system
+    #
+    # Parameters are not written to disk until Y2Network::Driver.commit
+    # is called. The reason is that writing them is an expensive operation, so it is
+    # better to write parameters for all drivers at the same time.
+    #
+    # You might prefer to use Y2Network::Driver.write_options instead.
+    #
+    # @see Y2Network::Driver.commit
+    def write_options
       parts = params.split(/ +/)
       params_hash = parts.each_with_object({}) do |param, hash|
         key, value = param.split("=")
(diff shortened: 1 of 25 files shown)


src/lib/y2network/udev_rule.rb
```diff
@@ -35,18 +46,23 @@ module Y2Network
...
       #
       # @return [Array<UdevRule>] Persistent network rules
-      def all
-        return @all if @all
-        rules_map = Yast::SCR.Read(Yast::Path.new(".udev_persistent.net")) || {}
-        @all = rules_map.values.map do |parts|
-          udev_parts = parts.map { |p| UdevRulePart.from_string(p) }
-          new(udev_parts)
-        end
+      def all(group = :net)
```
jreidinger: well, I do not like much this all with param. What about helpers like "naming_rules" and "driver_rules"

src/lib/y2network/driver.rb
```diff
@@ -17,29 +17,77 @@
...
       name == other.name && params == other.params
     end
 
+    # Adds driver parameters to be written to the underlying system
+    #
+    # Parameters are not written to disk until Y2Network::Driver.commit
+    # is called. The reason is that writing them is an expensive operation, so it is
+    # better to write parameters for all drivers at the same time.
+    #
+    # You might prefer to use Y2Network::Driver.write_options instead.
```
jreidinger: link here with `{Y2Network::Driver.write_options}`?

src/lib/y2network/widgets/driver.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      def store
+        @builder.driver = Y2Network::Driver.new(kernel_module_widget.value, kernel_options_widget.value)
+      end
+
+    private
+
+      def kernel_module_widget
+        return @kernel_module_width if @kernel_module_width
```
jreidinger: type here? I expect you want to check widget and not width

imobachgs: :man_facepalming:

src/lib/y2network/udev_rule.rb
```diff
@@ -60,17 +60,30 @@ class << self
...
+        find_rules(:drivers)
       end
 
       # Returns the udev rule for a given device
       #
       # @param device [String] Network device name
       # @return [UdevRule] udev rule
       def find_for(device)
-        all.find { |r| r.device == device }
+        naming_rules.find { |r| r.device == device }
```
jreidinger: here should be documented that it is only naming rule. Not driver rule for given device

PR #957, opened 2019-09-13
autoconfiguration convert

jreidinger:


src/lib/network/network_autoconfiguration.rb
```diff
@@ -45,29 +45,31 @@ class NetworkAutoconfiguration
     #
     # returns [Boolean] true when at least one interface is active
     def any_iface_active?
-      network_cards.any? { |c| configured?(c) && active_config?(c) }
+      config.interfaces.any? { |c| config.connections.by_name(c.name) && active_config?(c.name) }
```
imobachgs: We could have a `Y2Network::Config#configured_interface?` or something like that. Or even a `Y2Network::Config#configured_interfaces`.

PR #958, opened 2019-09-13
Allow to select no driver (auto)

imobachgs:
In my tests I ended in this scenario: let's say that I select a `custom` driver for a given interface, which caused a udev rule to be written. The next time that I open the hardware dialog again, I can only switch between the `custom` and the `virtio_net` driver. But, in both cases, a udev rule will be written (or updated). But what I want to do is to not force any driver at all. The only option is to manually delete the file.

As a solution, I have extended the drivers dialog to support an `auto` option, in which case no udev rule is written (and the kernel will decide which module to use).

driver-auto

### DIFF
--- a/src/lib/y2network/driver.rb
+++ b/src/lib/y2network/driver.rb
@@ -67,8 +67,8 @@ module Y2Network
 
     # Determines whether two drivers are equal
     #
-    # @param other [Driver] Driver to compare with
-    # @return [Boolean]
+    # @param other [Object] Driver to compare with
+    # @return [Boolean] true if +other+ is a Driver instance with the same name and params; false otherwise.
     def ==(other)
       return false unless other.is_a?(Driver)
       name == other.name && params == other.params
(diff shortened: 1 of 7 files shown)


src/lib/y2network/driver.rb
```diff
@@ -70,6 +70,7 @@ def initialize(name, params = "")
     # @param other [Driver] Driver to compare with
```
jreidinger: please document that it can accept also non Driver

test/y2network/interface_config_builder_test.rb
```diff
@@ -70,26 +70,34 @@
...
 
-      it "sets the interface driver" do
-        expect(eth0).to_not receive(:custom_driver=)
-        config_builder.driver = driver
+      it "sets the interface driver to nil" do
+        expect(eth0).to receive(:custom_driver=).with(nil)
+        subject.save
+      end
+
+      it "does not update any driver" do
```
jreidinger: do we want not update or explicit udev rule delete?

imobachgs: This one is about setting the [Driver](https://github.com/yast/yast-network/blob/network-ng/src/lib/y2network/driver.rb#L27) options.

PR #959, opened 2019-09-16
Drop udev writing/export capabilities from LanItems

imobachgs:
Drop udev writing/export capabilities from LanItems in order to detect if anything is missing in the new implementation.


src/modules/Lan.rb
```diff
@@ -739,15 +739,8 @@ def Import(settings)
     # @return dumped settings
     def Export
       profile = Y2Network::AutoinstProfile::NetworkingSection.new_from_network(yast_config)
-      devices = NetworkInterfaces.Export("")
-      udev_rules = LanItems.export(devices)
       ay = {
         "dns"                  => profile.dns ? profile.dns.to_hashes : {},
-        "s390-devices"         => Ops.get_map(
```
jreidinger: this is still problem as we first need to implement s390 activation in autoyast.

imobachgs: Yes, I know. I just wanted to drop the udev handling code from the old backend and replace with the new implementation.

PR #962, opened 2019-09-17
Ng fixes

jreidinger:


src/lib/y2network/interface_config_builders/bonding.rb
```diff
@@ -50,10 +50,33 @@ def bond_options
         connection_config.options
       end
 
+      # Checks if any of given device is already configured and need adaptation for bridge
```
imobachgs: Please, annotate the `devices` param.

src/lib/y2network/interface_config_builders/bridge.rb
```diff
@@ -92,8 +101,8 @@ def bridgeable?(iface)
           return false
         end
 
-        if NONBRIDGEABLE_STARTMODE.include?(iface.startmode)
-          log.debug("Excluding (#{iface.name}) - is #{iface.startmode}")
+        if NONBRIDGEABLE_STARTMODE.include?(config.startmode.to_s)
```
imobachgs: np: We could add a "bridgeable?" method to the connection.

jreidinger: yes and same with bonding, but I am not sure about connection as part of detection is also non existing connection :)

PR #966, opened 2019-09-17
detect that connections changed and fix removal of it

jreidinger:


test/y2network/sysconfig/config_writer_test.rb
```diff
@@ -137,8 +137,9 @@
       let(:old_config) do
         instance_double(
           Y2Network::Config,
-          dns:        double("dns"),
-          interfaces: [eth0, eth1]
+          dns:         double("dns"),
+          interfaces:  [eth0, eth1],
```
imobachgs: Actually, interfaces should be an `InterfacesCollection` object and connections an `ConnectionConfigsCollection`

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -36,6 +36,14 @@ def write(conn)
         file.save
       end
 
+      # Removes connection config from the underlying system
+      #
+      # @param name [String] Connection name to remove
+      def remove(name)
```
imobachgs: np: I would call the method remove_ifcfg or something like that.

jreidinger: well, it looks consistent for me with write. It is also `write` and not `write_ifcfg`

PR #967, opened 2019-09-18
Fix IP aliases handling

imobachgs:


src/lib/y2network/interface_config_builder.rb
```diff
@@ -475,5 +452,32 @@ def original_hostname
...
+        result << ConnectionConfig::IPConfig.new(ipaddr, label: map[:label], id: id.to_s)
+      end
+    end
+
+    # Returns a free numeric ID for an IP aliases
+    #
+    # @param used_ids   [Array<Integer>] Already used IDs
+    # @param current_id [Integer] Current used ID
+    def find_free_alias_id(used_ids, last_id)
+      loop do
```
jreidinger: btw with new rubocop we can use `(last_id..)` but not with current one, as it is supported in new version.

PR #968, opened 2019-09-18
network-ng: merge master

imobachgs:
Merge the `master` branch into `network-ng`. It includes an additional commit (894c2b47) to use `permanent_mac` whenever is available instead of `mac`.

### DIFF
--- a/test/network_autoconfiguration_test.rb
+++ b/test/network_autoconfiguration_test.rb
@@ -221,11 +221,6 @@ describe Yast::NetworkAutoconfiguration do
       let(:interfaces) { Y2Network::InterfacesCollection.new([eth0]) }
       let(:proposal) { true }
 
-      after(:each) do
-        # some methods might have sideeffects - clen them :-/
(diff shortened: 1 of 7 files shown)


test/network_autoconfiguration_test.rb
```diff
@@ -221,6 +221,11 @@ def probe_netcard_factory(num)
       let(:interfaces) { Y2Network::InterfacesCollection.new([eth0]) }
       let(:proposal) { true }
 
+      after(:each) do
+        # some methods might have sideeffects - clen them :-/
```
jreidinger: typo clean

test/network_autoconfiguration_test.rb
```diff
@@ -221,6 +221,11 @@ def probe_netcard_factory(num)
...
+        Yast::NetworkInterfaces.Devices.reject! { |k, _| k == "br" }
```
jreidinger: and do we still need it? I  though that after my adaptation it is no longer needed as we do not read NetworkInterfaces.

PR #969, opened 2019-09-18
network-ng: LanItems module clean up

imobachgs:
An attempt to drop unused stuff from `LanItems` module.

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -100,6 +100,8 @@ module Yast
 
       @backend = nil
 
+      @modified = false
+
       # Y2Network::Config objects
       @configs = {}
     end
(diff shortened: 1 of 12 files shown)


src/modules/Lan.rb
```diff
@@ -1039,16 +1035,21 @@ def connected_and_bridgeable?(bridge_builder, interface)
       true
     end
 
+    # Refreshes YaST network configuration
+    #
+    # It does not modified the system configuration that was already read.
```
jreidinger: is this really correct? maybe we need reread system configuration for e.g. s390 where new devices can appear or maybe when we need to load new kernel module for new interface? not sure.

imobachgs: To be honest, I am not sure. We always work on the `yast_config` and use the `system_config` as a reference to decide whether we have something to write to the underlying system or not.

jreidinger: yep, but s390 stuff we do immediatelly. So looks like our system_config is obsolete. But it will probably work, maybe just writter have some challenges.

imobachgs: After discussing via IRC, we have found that what we want to do in this case is just refresh the list of interfaces, but keeping the rest of the configuration untouched.

PR #971, opened 2019-09-19
Ng fixes5

jreidinger:


src/lib/y2network/connection_config/base.rb
```diff
@@ -65,6 +65,7 @@ class Base
       def initialize
         @ip_aliases = []
         @bootproto = BootProtocol::STATIC # TODO: maybe do test query if physical interface is attached?
+        @ip = IPConfig.new(IPAddress.from_string("0.0.0.0/36"))
```
imobachgs: 36 or 32?

jreidinger: ah, really, it should be 32

PR #973, opened 2019-09-20
network-ng: fix connections hostnames

imobachgs:
Initially, the objective of this PR was to fix the handling of connection configs hostnames, which are written to `/etc/hosts`. However, I realized that cleaning the old value and updating the new one was kind of inconvenient. The writer (`ConnectionConfigWriter` class) does not have access to the old value -no old name, no old IP-, so it is needed to always clean it up in advance and let the writer to blindly set the final value.

When renaming devices, we had the same problem. In that case, we added an `old_name` attribute so we can perform the rename in the same place. However, I did not want to start add `old_*` attributes to be honest.

I decided that it might be a good idea to pass the new and the old config objects to the writer. But, it would not fly at all if the name was changed (we should start checking for old names too). What I did was to add an artificial `id` to connection configs (similar to what `storage-ng` does). We could extend this behaviour to other objects, like interfaces or routes if needed. It has some advantages:

* You have a unique identifier for each element.
* You can easily check which elements have been added without relying on the IDs.
* You have enough information to decide whether to skip and update if a connection has not been modified.

We could extend this idea to interfaces too but I did not want to go that far in this PR.


src/lib/y2network/connection_config/base.rb
```diff
@@ -166,6 +175,13 @@ def find_master(configs)
         end
       end
 
+      # Returns whether the connection is configured as static
+      #
+      # @return [Boolean] return true if the connection is static; false otherwise
+      def static?
+        bootproto == BootProtocol::STATIC
```
jreidinger: I am not sure if it is good place here as connection config quite start growing. What about doing it same as I did it for interface type and add helpers to BootProtocol? so it can be `bootproto.static?`

src/lib/y2network/connection_configs_collection.rb
```diff
@@ -61,6 +61,22 @@ def by_interface(interface_name)
...
+    #
+    # @param id [Integer] Internal ID
+    # @return [ConnectionConfig::Base] Connection config with the given ID
+    def by_id(id)
+      connection_configs.find { |c| c.id == id }
+    end
+
+    # Returns connections with any of the given internal IDs
+    #
+    # @param ids [Integer] Internal IDs
```
jreidinger: is it Integer or Array<Integer> ?

src/lib/y2network/connection_configs_collection.rb
```diff
@@ -61,6 +61,22 @@ def by_interface(interface_name)
...
+    # @return [ConnectionConfig::Base] Connection config with the given IDs
+    def by_ids(*ids)
+      connection_configs.select { |c| ids.include?(c.id) }
+    end
```
jreidinger: and btw it is inconsistent with rest of methods that do not have both variants. so if needed, I would use the second one and define that it always return array.

src/lib/y2network/connection_configs_collection.rb
```diff
@@ -61,6 +61,22 @@ def by_interface(interface_name)
...
+    def by_id(id)
+      connection_configs.find { |c| c.id == id }
+    end
+
+    # Returns connections with any of the given internal IDs
+    #
+    # @param ids [Integer] Internal IDs
+    # @return [ConnectionConfig::Base] Connection config with the given IDs
+    def by_ids(*ids)
+      connection_configs.select { |c| ids.include?(c.id) }
```
jreidinger: why not use the new method select?

src/lib/y2network/connection_configs_collection.rb
```diff
@@ -61,6 +61,22 @@ def by_interface(interface_name)
...
+    # @param id [Integer] Internal ID
+    # @return [ConnectionConfig::Base] Connection config with the given ID
+    def by_id(id)
+      connection_configs.find { |c| c.id == id }
+    end
+
+    # Returns connections with any of the given internal IDs
+    #
+    # @param ids [Integer] Internal IDs
+    # @return [ConnectionConfig::Base] Connection config with the given IDs
```
jreidinger: It should be in fact Array now or ConnectionsCollection if you use later suggestion.

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -18,30 +18,40 @@
...
       # Removes connection config from the underlying system
       #
-      # @param name [String] Connection name to remove
-      def remove(name)
-        ifcfg = Y2Network::Sysconfig::InterfaceFile.find(name)
+      # @param conn [Y2Network::Conn] Connection name to remove
+      def remove(conn)
+        ifcfg = Y2Network::Sysconfig::InterfaceFile.find(conn.interface)
         ifcfg && ifcfg.remove
+        Yast::Host.remove_ip(conn.ip.address.address.to_s) if conn.ip
```
jreidinger: what about removal of aliases here also?

src/lib/y2network/sysconfig/interfaces_writer.rb
```diff
@@ -105,8 +105,6 @@ def clean_up_old_interfaces(interfaces)
           set_interface_down(iface.old_name)
           ifcfg = Y2Network::Sysconfig::InterfaceFile.find(iface.old_name)
           ifcfg && ifcfg.remove
```
jreidinger: is it needed? it should be done in connection config writer, not?

imobachgs: Yes, now it should be covered.

src/lib/y2network/sysconfig/connection_config_writer.rb
imobachgs: Do you mean IP aliases? It is done automatically when writing the file.

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -35,7 +35,7 @@ class ConnectionConfigWriter
       # @param conn [Y2Network::ConnectionConfig::Base] Connection configuration to write
       # @param old_conn [Y2Network::ConnectionConfig::Base,nil] Connection configuration to write
       def write(conn, old_conn = nil)
-        return if conn == old_conn
+        return if old_conn && conn == old_conn
```
jreidinger: this looks like mistake of `operator==` which should not crash when comparison to nil, so please fix  it there.

src/lib/y2network/connection_config/base.rb
```diff
@@ -90,6 +90,7 @@ def initialize
       # @return [Boolean] true when both connections are same
       #                   false otherwise
       def ==(other)
+        return false if other.nil?
```
jreidinger: what about more generic `return false if self.class != other.class`? this way it also handle comparison of different types.

PR #985, opened 2019-10-01
fix CLI

jreidinger:
bsc: https://bugzilla.suse.com/show_bug.cgi?id=1151242
trello: https://trello.com/c/OautSDuO/1324-p1-get-yast2-network-cmdline-to-an-usable-state

- [x] show
- [x] list
- [x] add
- [x] edit
- [x] delete


src/include/network/lan/cmdline.rb
```diff
@@ -125,58 +126,37 @@ def getItem(options, config)
...
-          descr.gsub!(/(<br>)|(<\/li>)/, "\n")
-          descr.gsub!(/<[^>]+>/, "")
+      config = Yast::Lan.yast_config
+      return false unless validateId(options, config.interfaces)
 
-          # to keep it backward compatible. Previously echo was called which adds additional new line"
-          descr << "\n"
+      presenter = Y2Network::Presenters::InterfaceSummary.new(config.interfaces.to_a[options["id"].to_i].name, config)
+      text = presenter.text
+      # create plain text from formated HTML
```
teclator: NP: maybe something for the presenter? like plain_text method

jreidinger: well, also list handler deserve own presenter, but due to its current state, I prefer easy and straighforward fix. Also looks like CLI is not much used, so goal is clear. Make it working, but do not invest much.

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Oct  2 09:05:14 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Fix again network CLI with improved error reporting (bsc#1151242)
```
teclator: "again" in the changelog looks like .. not ugly but something that happen with some .. let say frequency :P

jreidinger: agree, I want to mention that network-ng. So I will reformulate.

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Oct  2 09:05:14 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
+
+- Adapt network CLI to new data model with improved error reporting
+  (bsc#1151242)
+- 4.1.16
```
imobachgs: Version number is wrong.

PR #986, opened 2019-10-01
network-ng: s390 autoyast configuration

teclator:
- https://trello.com/c/D3n3PbVH/1278-5-network-ng-autoyast-support

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Mon Oct  7 06:05:22 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
 
 - AutoYaST: Added back the activation of s390 devices during the
-  first stage.
+  first stage as it was removed when switched to the new data model
 
 -------------------------------------------------------------------
 Fri Oct  4 13:51:42 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Oct  7 06:05:22 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- AutoYaST: Added back the activation of s390 devices during the
+  first stage.
+
+-------------------------------------------------------------------
 Fri Oct  4 13:51:42 UTC 2019 - Josef Reidinger <jreidinger@suse.com>
 
 - Fix CLI tests for network (bsc#957537)
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -18,6 +18,10 @@
 # find current contact information at www.suse.com.
 
 require "yast"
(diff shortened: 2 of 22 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -113,6 +117,28 @@ def configure_lan(write: false)
...
+    def activate_s390_devices(section = nil)
+      profile_devices = section || ay_networking_section["s390-devices"] || {}
+      devices_section = Y2Network::AutoinstProfile::S390DevicesSection.new_from_hashes(profile_devices)
+      connections = Y2Network::Autoinst::S390DevicesReader.new(devices_section).config
+      connections.each do |conn|
+        begin
+          builder = Y2Network::InterfaceConfigBuilder.for(conn.type, config: conn)
+          activator = Y2Network::S390DeviceActivator.for(builder)
+          if activator.configure
+            builder.name = activator.configured_interface
```
jreidinger: what is the goal is assigning name to builder when you do not save it?

src/lib/y2network/autoinst_profile/s390_device_section.rb
```diff
@@ -0,0 +1,93 @@
...
+      # @param config [Y2Network::ConnectionConfig]
+      # @return [Boolean]
+      def init_from_config(config)
+        @type = config.type.short_name
+        case config
+        when ConnectionConfig::Qeth
+          @chanids = config.device_id
+          @layer2 = config.layer2
+        when ConnectionConfig::Ctc
+          @protocol = config.protocol
```
jreidinger: what about LCS?

src/modules/LanItems.rb
```diff
@@ -401,7 +402,7 @@ def Import(settings)
       # - creates s390 device eth emulation
       # So, it belongs partly into Import and partly into Write. Note, that
       # the code is currently unable to revert already created emulated device.
-      settings.fetch("s390-devices", {}).each { |rule| createS390Device(rule) } if Arch.s390
+      NetworkAutoYast.instance.activate_s390_devices(settings.fetch("s390-devices", {}))
```
jreidinger: I would keep here that s390 condition, so it do not try fancy things if profile contain it accidentally.

jreidinger:
just few notes. overall looks good.

src/lib/network/network_autoyast.rb
teclator: It is not needed at all. It makes sense when activating the device for editing, but in AY it is irrelevant.

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Oct  7 06:05:22 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- AutoYaST: Added back the activation of s390 devices during the
```
jreidinger: maybe add added back after what. Like `after switch to new data model`. So it is clear it is not something broken in SP1.

PR #989, opened 2019-10-03
fix calling Yast.import "LanItems"

jreidinger:


src/lib/y2network/widgets/slave_items.rb
```diff
@@ -29,6 +27,10 @@ module SlaveItems
       include Yast::Logger
       include Yast::I18n
 
+      def initialize
+        Yast.import "NetworkInterfaces"
```
teclator: @jreidinger maybe you prefer to omit this change, if not, then we could move the `textomain "network"` here like we usually do.

jreidinger: fine for now. For future I am not sure if we really need it.

teclator: I just removed the LanItems import and left it as it was..

PR #990, opened 2019-10-03
Fix blink position

jreidinger:


kobliha:
Hint: It's usually good to attach a screenshot when fixing UI issues.

PR #991, opened 2019-10-03
 Write sysctl settings to /etc/sysctl.d

imobachgs:
Adapts sysctl handling to use the new [CFA::Sysctl](https://github.com/yast/yast-yast2/pull/966) class so the configuration is written to a YaST specific file under `/etc/sysctl.d`.

### DIFF
--- a/src/lib/y2network/sysconfig/config_reader.rb
+++ b/src/lib/y2network/sysconfig/config_reader.rb
@@ -17,7 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 require "yast"
-require "yast2/cfa/sysctl"
+require "cfa/sysctl"
 require "y2network/config"
 require "y2network/interface"
 require "y2network/routing"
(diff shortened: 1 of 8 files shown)


src/lib/y2network/sysconfig/config_reader.rb
```diff
@@ -106,14 +107,14 @@ def load_routes_from(path = nil)
...
-        Yast::SCR.Read(Yast::Path.new(SYSCTL_IPV4_PATH)) == "1"
+        sysctl_file.forward_ipv4 == "1"
       end
 
       # Reads IPv6 forwarding status
       #
       # return [Boolean] true when IPv6 forwarding is allowed
       def forward_ipv6?
-        Yast::SCR.Read(Yast::Path.new(SYSCTL_IPV6_PATH)) == "1"
+        sysctl_file.forward_ipv6 == "1"
```
jreidinger: when I see this, I think that logic for understanding values should be in CFA model. So maybe add helpers in sysctl model to have `forward_ipv6?` which returns boolean and ideally also `forward_ipv6=` that understand boolean and convert it to 1/0.

imobachgs: To some extent, we are representing the file content, so I wanted to avoid conversions. Adding a `#forward_ipv6?` method is OK.

PR #993, opened 2019-10-09
Consistent delete/rename

jreidinger:
https://trello.com/c/9q8UEvBD/1326-5-network-handle-properly-removal-and-rename-of-interfaces
https://bugzilla.suse.com/show_bug.cgi?id=1153460

### DIFF
--- a/src/lib/y2network/config.rb
+++ b/src/lib/y2network/config.rb
@@ -152,6 +152,8 @@ module Y2Network
     #
     # @param name [String] Interface's name
     def delete_interface(name)
+      delete_dependents(name)
+
       connections.reject! { |c| c.interface == name }
       interface = interfaces.by_name(name)
       return if interface.is_a?(PhysicalInterface) && interface.present?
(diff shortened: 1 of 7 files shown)


src/lib/y2network/config.rb
```diff
@@ -205,6 +210,54 @@ def configured_interface?(iface_name)
       !connections.by_interface(iface_name).empty?
     end
 
+    # @note does not work recursivelly. So for delete it needs to be called for all modified vlans.
```
imobachgs: np: recursively

src/lib/y2network/config.rb
```diff
@@ -143,7 +143,10 @@ def rename_interface(old_name, new_name, mechanism)
       interface = interfaces.by_name(old_name || new_name)
       interface.rename(new_name, mechanism)
       return unless old_name # do not modify configurations if it is just renaming mechanism
-      connections.by_interface(old_name).each { |c| c.interface = new_name }
+      connections.by_interface(old_name).each do |connection|
+        connection.interface = new_name
+        rename_dependencies(old_name, new_name, connection)
```
imobachgs: np: Perhaps you could omit the `old_name` or the `new_name` as it is contained in the `connection` instance.

jreidinger: where it is contained? I only found it in interface class, but not in connection one.

imobachgs: Is not `connection.interface`?

jreidinger: connection.interface is only the new name ( it is string ). So this way it is more explicit for me.

PR #994, opened 2019-10-11
New rubocop

jreidinger:


src/clients/lan_auto.rb
```diff
@@ -189,28 +185,18 @@ def ToAY(settings)
...
-      end
-      if Ops.greater_than(Builtins.size(s390_devices), 0)
-        Ops.set(ret, "s390-devices", s390_devices)
-      end
-      if Ops.greater_than(Builtins.size(net_udev), 0)
-        Ops.set(ret, "net-udev", net_udev)
-      end
+      Ops.set(ret, "interfaces", interfaces) if Ops.greater_than(Builtins.size(interfaces), 0)
+      Ops.set(ret, "s390-devices", s390_devices) if Ops.greater_than(Builtins.size(s390_devices), 0)
+      Ops.set(ret, "net-udev", net_udev) if Ops.greater_than(Builtins.size(net_udev), 0)
```
mvidner: You may want to use the eager mode of the Zombie Killer for this, https://github.com/yast/zombie-killer/pull/9 , to translate `Ops.greater_than(Builtins.size(net_udev), 0)` into `!net_udev.empty?`

src/include/network/lan/address.rb
```diff
@@ -65,10 +63,10 @@ def AddressDialog(builder:)
         if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
           # fixed bug #73739 - if dhcp is used, dont set default gw statically
           # but also: reset default gw only if DHCP* is used, this branch covers
-          #		 "No IP address" case, then default gw must stay (#460262)
+          #     "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
           yast_config = Y2Network::Config.find(:yast)
-          if yast_config && yast_config.routing && yast_config.routing.default_route
+          if yast_config&.routing && yast_config.routing.default_route
```
mvidner: Another `&.` case missed?

src/lib/y2network/sysconfig/interfaces_reader.rb
```diff
@@ -87,14 +89,15 @@ def find_physical_interfaces
 
       # Finds the connections configurations
       def find_connections
-        @connections ||=
+        @connections =
```
mvidner: This is not an equivalent change, is it?

src/modules/Lan.rb
```diff
@@ -713,13 +725,13 @@ def Import(settings)
     def Export
       profile = Y2Network::AutoinstProfile::NetworkingSection.new_from_network(yast_config)
       ay = {
-        "dns"                  => profile.dns ? profile.dns.to_hashes : {},
-        "net-udev"             => profile.udev_rules ? profile.udev_rules.udev_rules.map(&:to_hashes) : [],
-        "s390-devices"         => profile.s390_devices ? profile.s390_devices.to_hashes["devices"] : [],
+        "dns"                  => profile.dns&.to_hashes || {},
+        "net-udev"             => profile.udev_rules.udev_rules&.map(&:to_hashes) || [],
```
mvidner:
```suggestion
        "net-udev"             => profile.udev_rules&.udev_rules.map(&:to_hashes) || [],
```

src/modules/Lan.rb
```diff
@@ -713,13 +725,13 @@ def Import(settings)
...
+        "s390-devices"         => profile.s390_devices&.to_hashes&.fetch("devices", []) || [],
         "config"               => NetworkConfig.Export,
-        "interfaces"           => profile.interfaces ? profile.interfaces.interfaces.map(&:to_hashes) : [],
+        "interfaces"           => profile.interfaces.interfaces&.map(&:to_hashes) || [],
```
mvidner:
```suggestion
        "interfaces"           => profile.interfaces&.interfaces.map(&:to_hashes) || [],
```

src/include/network/lan/address.rb
jreidinger: you are right

src/clients/lan_auto.rb
jreidinger: well, goal if this pr is to adopt new rubocop. Feel free to open another one with zombie killer improvements

src/lib/y2network/sysconfig/interfaces_reader.rb
jreidinger: no, it is not, but I analyze code and there is no change as result of this call is cached, so it was in fact double caching.

src/modules/Lan.rb
jreidinger: no, this change will crash. as `&.` can return nil and in such case `nil.map` crash.

src/modules/Lan.rb
jreidinger: and same here.

src/modules/Lan.rb
mvidner: I think RuboCop has a bug in not distinguishing one `udev_rules` from the other. I cannot really blame it, such API is crazy.
Still, your change is not equivalent. You may want to use TWO `&.` then.

jreidinger: ah, right, two `&.` looks correct for me.

src/modules/Lan.rb
```diff
@@ -713,13 +725,13 @@ def Import(settings)
     def Export
       profile = Y2Network::AutoinstProfile::NetworkingSection.new_from_network(yast_config)
       ay = {
-        "dns"                  => profile.dns ? profile.dns.to_hashes : {},
-        "net-udev"             => profile.udev_rules ? profile.udev_rules.udev_rules.map(&:to_hashes) : [],
-        "s390-devices"         => profile.s390_devices ? profile.s390_devices.to_hashes["devices"] : [],
+        "dns"                  => profile&.dns&.to_hashes || {},
```
mvidner: But `profile` cannot be `nil`, we have just constructed it:
https://github.com/yast/yast-network/pull/994/files#diff-f0657dacb6a78e1b8c4dbfc977fc0e54R72-R82

This is the bad thing about RuboCop: it encourages you to make subtle logic changes, hidden among its monstrous near-whitespace diffs.

jreidinger: ah, right, I do too much. Lets revert this change.

PR #997, opened 2019-10-21
Live instalation: Make network more useful

teclator:
## Problem

- In a live installation we assumes that wicked is the network service by default which is completely wrong. The point is that we can check the network backend in use as systemd is running.
- The network client "lan" only shows the "general tab" in a running system which means that it always tries to configure the network assuming it uses wicked which is also wrong.
- At the end of the installation "save_network" configures the dns, which calls `netconfig update`. Netconfig checks internally the link of the network.service for determining which resolv.conf configuration should use (NM or wicked) overriding the one from the instsys and in many cases breaking the names resolution.

- https://trello.com/c/qvY7Oygp/1375-3-osdistribution-p1-1151291-build-1330-openqa-test-fails-in-awaitinstall-yast-breaks-network-during-install

## Solution

- When Systemd is running determine the backend in use (https://github.com/yast/yast-yast2/pull/972) and show the general tab in the "network" configuration dialog.
- Skip the network configuration when NetworkManager is in use.
- Do not call netconfig update if the NetworkService is not wicked.
- It depends on yast/yast-yast2#972 (That is, yast2-4.2.31) 
  - Will update the spec with the requirement once approved and merged that PR just for not breaking dependencies when travis is run

### DIFF
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -55,6 +55,7 @@ module Yast
       Yast.import "CWMTab"
       Yast.import "Stage"
       Yast.import "LanItems"
+      Yast.import "Systemd"
 
       Yast.include include_target, "network/routines.rb"
       Yast.include include_target, "network/lan/help.rb"
@@ -258,7 +259,7 @@ module Yast
       caption = _("Network Settings")
       widget_descr = {
         "tab" => CWMTab.CreateWidget(
-          "tab_order"    => if Stage.normal
+          "tab_order"    => if Stage.normal || Systemd.Running
                               ["global", "overview", "resolv", "route"]
                             else
                               ["overview", "resolv", "route"]
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -259,7 +259,7 @@ module Yast
       caption = _("Network Settings")
       widget_descr = {
         "tab" => CWMTab.CreateWidget(
-          "tab_order"    => if Stage.normal || Systemd.Running
+          "tab_order"    => if Systemd.Running
                               ["global", "overview", "resolv", "route"]
                             else
                               ["overview", "resolv", "route"]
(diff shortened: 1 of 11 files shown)


src/include/network/lan/complex.rb
```diff
@@ -258,7 +259,7 @@ def MainDialog(init_tab)
       caption = _("Network Settings")
       widget_descr = {
         "tab" => CWMTab.CreateWidget(
-          "tab_order"    => if Stage.normal
+          "tab_order"    => if Stage.normal || Systemd.Running
```
jreidinger: why not just Systemd.Running always?

src/lib/network/clients/save_network.rb
```diff
@@ -99,6 +100,10 @@ def CopyConfiguredNetworkFiles
         { dir: ETC, file: DNSClass::HOSTNAME_FILE }
       ]
 
+      if Y2Network::ProposalSettings.instance.network_service == :network_manager
```
jreidinger: please document "why" we are doing it

src/lib/y2network/sysconfig/dns_writer.rb
```diff
@@ -30,13 +30,13 @@ class DNSWriter
       #
       # @param dns [Y2Network::DNS] DNS configuration
       # @param old_dns [Y2Network::DNS] Old DNS configuration
-      def write(dns, old_dns)
+      def write(dns, old_dns, netconfig_update: true)
```
jreidinger: please document new parameter

src/lib/y2network/sysconfig/dns_writer.rb
```diff
@@ -92,7 +92,7 @@ def update_mta_config
       # Updates /etc/sysconfig/network/config
       #
       # @param dns [Y2Network::DNS]
-      def update_sysconfig_config(dns)
+      def update_sysconfig_config(dns, netconfig: true)
```
jreidinger: and same here

src/modules/DNS.rb
```diff
@@ -189,9 +189,11 @@ def Read
     # Includes Host,NetworkConfig::Write
     # @todo Update GUI
     # @return true if success
-    def Write(_gui: true)
+    def Write(_gui: true, netconfig_update: true)
```
jreidinger: if gui is not used, maybe remove it completelly? and document new param

PR #564, opened 2017-09-25
Sets DHCLIENT_SET_HOSTNAME based on linuxrc or from the control file (bsc#1054933)

teclator:
- [Trello Card](https://trello.com/c/Vj69gtYP/1168-5-caasp-20-hostname-via-dhcp-work-according-to-control-file)
- This PR depends on openSUSE/linuxrc#146 and openSUSE/linuxrc#147

Basically, it sets the DHCLIENT_SET_HOSTNAME in /etc/sysconfig/network/dhcp based on linuxrc cmdline sethostname option if provided  or obtains the default from the control file if not.

https://github.com/yast/skelcd-control-SLES/blob/master/control/control.SLES.xml#L21

## Screenshot after inst_setup_dhcp client (just after wellcome screen)
screenshot_testing_2017-09-25_11 08 43

## Screenshot after manual installation without special linuxrc cmdline options
screenshot_testing_2017-09-25_11 39 50

## CaaSP 2.0 Manual Network config

manualconfig

screenshot

## Notes and tricks

The libvirt network config used for the host:
```
      <host mac='dc:e4:cc:27:94:c2' name='teclator.suse.com' ip='192.168.122.60'/>
```

The tests have been done with a modified CaaSP iso with last linuxrc modification and also modified with a DUD containing this code.

```
mksusecd --initrd initrd --create /home/knut/isos/SUSE-CaaS-Platform-2.0-DVD-x86_64-Linuxrc.iso /home/knut/isos/SUSE-CaaS-Platform-2.0-DVD-x86_64-Build0071-Media1.iso
mksusecd --initrd network.dud --create /home/knut/isos/SUSE-CaaS-Platform-2.0-DVD-x86_64-dud.iso /home/knut/isos/SUSE-CaaS-Platform-2.0-DVD-x86_64-Linuxrc.iso
```

## Replace widget instead of just the Items to show the complete interface name
fit_selector

PR #1000, opened 2019-11-04
Hostname in installer

mchf:
Fate319639

Hotfix for network-ng to set hostname according to the fate above in the installer. It was only partly implemented in the code before network-ng. It basically means that:
- if no other way is used the hostname is set to a default (currently install)
- if a hostname is provided via dhcp then it is used (completely new functionality )
- if a hostname is provided via linuxrc's hostname option then it is used (was not working in network-ng)

The hostname is set in order described above. Also, the hostname is written to the target only in third case (when set using hostname option).

Yast2-network cooperates with linuxrc here bcs linuxrc is e.g. the one who sets the default hostname and so on.

What should be done next:
- in the old code the hostname and dns handling was mixed together. As this features are not fully related it would be nice  to have it in separate classes in network-ng (will follow)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Nov 21 07:13:28 UTC 2019 - Michal Filka <mfilka@suse.com>
 
-- fate#319639
+- bnc#1153198, fate#319639
   - implemented proper hostname setup in installation for the new
     backend
 - 4.2.29
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Thu Nov 21 07:13:28 UTC 2019 - Michal Filka <mfilka@suse.com>
+
+- fate#319639
+  - implemented proper hostname setup in installation for the new
+    backend
+- 4.2.29
+
+-------------------------------------------------------------------
 Tue Nov 19 12:05:21 UTC 2019 - David Diaz <dgonzalez@suse.com>
 
 - Do not crash when wrapping text (bsc#1157161)
--- a/src/lib/network/wicked.rb
+++ b/src/lib/network/wicked.rb
@@ -41,15 +41,40 @@ module Yast
     # @param iface [String] network device
     # @return [Array<String>] list of NTP servers
     def parse_ntp_servers(iface)
+      query_wicked(iface, "%{//ntp/server}")
+    end
+
+    # Parses wicked runtime configuration and returns hostname if set
+    #
+    # @param iface [String] network device
+    # @return [String] hostname
+    def parse_hostname(iface)
+      result = query_wicked(iface, "%{//hostname}")
+
+      raise "Malformed wicked runtime configuration" if result.count > 1
+
+      result.first
+    end
+
+    # Parses wicked runtime dhcp lease file for the given query
+    #
+    # It parses both ipv4 and ipv6 lease files at once.
+    #
+    # @param iface [String] queried interface
+    # @param query [String] xpath query
+    # @return [String] result of the query
+    def query_wicked(iface, query)
       raise ArgumentError, "A network device has to be specified" if iface.nil? || iface.empty?
       if !NetworkService.is_wicked
-        raise "Parsing NTP Servers not supported for network service in use"
+        raise "Parsing not supported for network service in use"
       end
 
       lease_files = ["ipv4", "ipv6"].map { |ip| "/var/lib/wicked/lease-#{iface}-dhcp-#{ip}.xml" }
       lease_files.find_all { |f| File.file?(f) }.reduce([]) do |stack, file|
-        result = SCR.Execute(BASH_OUTPUT_PATH,
-          "/usr/sbin/wicked xpath --file #{file.shellescape} \"%{//ntp/server}\"")
+        result = SCR.Execute(
+          BASH_OUTPUT_PATH,
+          "/usr/sbin/wicked xpath --file #{file.shellescape} \"#{query}\""
+        )
 
         stack + result.fetch("stdout", "").split("\n")
       end
--- a/src/lib/network/wicked.rb
+++ b/src/lib/network/wicked.rb
@@ -65,9 +65,7 @@ module Yast
     # @return [String] result of the query
     def query_wicked(iface, query)
       raise ArgumentError, "A network device has to be specified" if iface.nil? || iface.empty?
-      if !NetworkService.is_wicked
-        raise "Parsing not supported for network service in use"
-      end
+      raise "Parsing not supported for network service in use" if !NetworkService.is_wicked
 
       lease_files = ["ipv4", "ipv6"].map { |ip| "/var/lib/wicked/lease-#{iface}-dhcp-#{ip}.xml" }
       lease_files.find_all { |f| File.file?(f) }.reduce([]) do |stack, file|
--- a/src/lib/y2network/hostname_reader.rb
+++ b/src/lib/y2network/hostname_reader.rb
@@ -37,6 +37,8 @@ module Y2Network
   # @example Read hostname
   #   Y2Network::HostnameReader.new.hostname #=> "foo"
   class HostnameReader
+    attr_reader :install_inf_hostname
+
     include Yast::Logger
     include Yast::Wicked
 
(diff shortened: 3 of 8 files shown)


src/lib/network/wicked.rb
```diff
@@ -41,15 +41,38 @@ def reload_config(devs)
...
+
+      result.first
+    end
+
+    # Parses wicked runtime dhcp lease file for the given query
+    #
+    # It parses both ipv4 and ipv6 lease files at once.
+    #
+    # @param iface [String] queried interface
+    # @param query [String] xpath query
```
jreidinger: maybe mention that it is just subset of xpath ( as specified in man wicked section xpath )

src/lib/y2network/hostname_reader.rb
```diff
@@ -34,20 +37,23 @@ module Y2Network
...
-    # @note If the hostname cannot be determined, generate a random one.
+    # @note If the hostname cannot be determined, generate a random one
+    # in installed system (do not generate one in the installer).
     #
     # @return [String]
     def hostname
-      if (Yast::Mode.installation || Yast::Mode.autoinst) &&
-          Yast::FileUtils.Exists("/etc/install.inf")
-        fqdn = hostname_from_install_inf
+      if Yast::Mode.installation || Yast::Mode.autoinst
```
jreidinger: will it work during update?

src/lib/y2network/hostname_reader.rb
```diff
@@ -94,5 +122,26 @@ def random_hostname
...
       "linux-#{suffix}"
     end
+
+  private
+
+    # Runs workflow for querying hostname in the installer
+    #
+    # @return [String] Hostname
+    def hostname_for_installer
+      @install_inf_hostname = hostname_from_install_inf
```
jreidinger: why only this part is catched and others not?

src/lib/y2network/sysconfig/dns_reader.rb
```diff
@@ -31,9 +33,11 @@ class DNSReader
       #
       # @return [Y2Network::DnsConfig] DNS configuration
       def config
+        installer = Yast::Mode.installation || Yast::Mode.autoinst
```
jreidinger: what about autoinst mode in second stage?

src/lib/y2network/sysconfig/dns_reader.rb
```diff
@@ -75,7 +79,15 @@ def resolv_conf_policy
...
-        HostnameReader.new.hostname
+        @hostname_reader = HostnameReader.new
+        @hostname_reader.hostname
+      end
+
+      # Checks whether the hostname was read from install.inf
+      #
+      # @return [Boolean]
+      def hostname_from_install_inf?
+        !@hostname_reader.install_inf_hostname.nil?
```
jreidinger: this is really nasty side effect that really depends on fact that hostname entry is constructed before save_hostname entry. What I really prefer is method `hostname_reader` that returns cached object and those two methods just call it.

mchf: I've follow up of this PR. I separate Hostname and DNS classes there exactly for this (and one above) reason. So, would you prefer to merge those two patches together?

src/lib/y2network/sysconfig/dns_reader.rb
mchf: network configuration runs completely in first stage. So in my POV when something invokes it in second stage then it should behave as on installed system.

src/lib/y2network/sysconfig/dns_reader.rb
jreidinger: I am fine with merging it together, so I can see final result.

src/lib/y2network/hostname_reader.rb
mchf: see bellow comment about separating hostname class, this is another reason. In fact Hostname and DNS has nothing in common. It stayed together just because it was together in the old code.

src/lib/y2network/hostname_reader.rb
mchf: no, but network is not modified during update at all. I mean ... hostname will not be updated in the updated system, but it will be displayed correctly e.g. in bash prompt (bcs it is preset by linuxrc)

src/lib/network/wicked.rb
mchf: ok, I'll point to the wicked doc.

teclator:
I would like to see more documentation and description in general. Behavior, testing corner cases and so on. 

The PR description only mentions a fate entry, which was for SLE-12-SP2, a link to the trello card would be nice to have or at least to the bug that reported it for SLE-15-SP2.

The fate is from 2016, there were some changes related to DHCLIENT_SET_HOSTNAME after it:

See https://github.com/yast/yast-network/pull/564/files

So, documenting different scenarios and the workflow with the different options knowing exactly how the static or transient hostname it is set in each case would be very nice to have.

Even linuxrc has it:

https://github.com/openSUSE/linuxrc/blob/master/linuxrc_hostname.md

teclator:
@mchf I was asked about reviewing the PR but it still contains conflicts and the description basically says nothing. What the problem it, what it fixes, what is still missing..., of course there is a old fate number referenced but linking to it would be better for others to check it quickly and also as reference for the future as well as the trello card URL and any other useful link.

As we were discussing, we do not enforce a particular PR template but the [one](https://github.com/yast/yast-storage-ng/blob/master/.github/pull_request_template.md) provided by @lslezak is a good one for using at least as an example.

I would like to also see some documentation about use cases, and different scenarios as it is not the first time we face a problem and different expectations affecting the hostname setup, but for course taking in account that is not part of the acceptance criteria probably we could omit it.

I already mentioned that in previous comment, but have not seen an answer or movement from there in advance, so not sure what more I could door what is expected from my side at all :(

jsrain:
@teclator @mchf I fully understand that this PR's description says almost nothing (and this concern has not been addressed since more than two weeks now). Please, adjust the PR and get it merged ASAP. Currently, yast2-network fails to build in Staging (one test case failing) and effectively blocks merging any further changes of YaST. Thanks!

teclator:
> Fate319639
> 
> Hotfix for network-ng to set hostname according to the fate above in the installer. It was only partly implemented in the code before network-ng. It basically means that:
> 
>     * if no other way is used the hostname is set to a default (currently install)

At the end of the installation, if not hostname was set and no DHCP hostname what will be set? linux-(random), install or nothing?

> 
>     * if a hostname is provided via dhcp then it is used (completely new functionality )

Well, it was already the case since: https://github.com/yast/yast-network/pull/564 but probably lost in network-ng.

> 
>     * if a hostname is provided via linuxrc's hostname option then it is used (was not working in network-ng)
> 
> 
> The hostname is set in order described above. Also, the hostname is written to the target only in third case (when set using hostname option).
> 
> Yast2-network cooperates with linuxrc here bcs linuxrc is e.g. the one who sets the default hostname and so on.
> 
> What should be done next:
> 
>     * in the old code the hostname and dns handling was mixed together. As this features are not fully related it would be nice  to have it in separate classes in network-ng (will follow)

In the past there was even a separate dialog for setting the hostname but then it was removed and mixed with dns. Agree about moving the responsibility elsewhere.

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Nov 21 07:13:28 UTC 2019 - Michal Filka <mfilka@suse.com>
+
+- fate#319639
```
teclator: I would prefer to use the bug number instead of the fate, so, if you wanted to mention it then add both.

mchf:
> > Fate319639
> > Hotfix for network-ng to set hostname according to the fate above in the installer. It was only partly implemented in the code before network-ng. It basically means that:
> > ```
> > * if no other way is used the hostname is set to a default (currently install)
> > ```
> 
> At the end of the installation, if not hostname was set and no DHCP hostname what will be set? linux-(random), install or nothing?

not specified in the feature - so a default. As the feature says that hostname should be updated only if set via the hostname option.
> 
> > ```
> > * if a hostname is provided via dhcp then it is used (completely new functionality )
> > ```
> 
> Well, it was already the case since: #564 but probably lost in network-ng.

Yes, partly ... however stopped to work after transition to the network-ng
> 
> > ```
> > * if a hostname is provided via linuxrc's hostname option then it is used (was not working in network-ng)
> > ```
> > 
> > 
> > The hostname is set in order described above. Also, the hostname is written to the target only in third case (when set using hostname option).
> > Yast2-network cooperates with linuxrc here bcs linuxrc is e.g. the one who sets the default hostname and so on.
> > What should be done next:
> > ```
> > * in the old code the hostname and dns handling was mixed together. As this features are not fully related it would be nice  to have it in separate classes in network-ng (will follow)
> > ```
> 
> In the past there was even a separate dialog for setting the hostname but then it was removed and mixed with dns. Agree about moving the responsibility elsewhere.

sure, by the old code I mean state before switching to network-ng.

PR #1001, opened 2019-11-10
Fix wpa psk config

teclator:
- https://bugzilla.suse.com/show_bug.cgi?id=1155639

## Problem

When choosing the wireless WPA PSK auth mode, the config is expected to have a method `write_wpa-psk_auth_settings(conn)' but it is defined omitting WPA which is already implicit known.

See https://github.com/yast/yast-network/blob/master/src/lib/y2network/sysconfig/connection_config_writers/wireless.rb#L74

## Solution

- We could downcase the auth mode selection and rename the`wpa-psk` method but as we also omit `WPA` prefix in the file attribute I preferred to just omit the prefix completely from the Combo Box option values.

- see https://github.com/yast/yast-network/blob/master/src/lib/y2network/connection_config/wireless.rb#L35

With this changes the WPA PSK config is written correctly

### DIFF
--- a/package/yast2-network.spec
+++ b/package/yast2-network.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.2.24
+Version:        4.2.26
 Release:        0
 Summary:        YaST2 - Network Configuration
 License:        GPL-2.0-only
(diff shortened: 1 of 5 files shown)


package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.2.24
```
teclator: Apparently I didn't increase the version in #997 when rebasing. So, it was supposed to be yast2-network-4.2.25 already and we bump it now to 4.2.26

PR #1002, opened 2019-11-11
implement reading dhclient_set_hostname for autoyast profiles (bsc#11…

jreidinger:
…52021)

trello: https://trello.com/c/vpFJV0jV/1428-3-sles15-sp2-p2-1152021-autoyast-does-not-allow-to-specify-an-interface-to-set-the-hostname-via-dhcp

what is not yet implemented is sanitization of input like multiple interfaces with dhclient set to yes.

Manual testing needs to be done ( initial draft for first impressions and comments ).


mchf:
It seems fine ... as i believe it reads values from AY profile ;-) However, there is no kind of validation. At least it could log an error when an insecure configuration is going to be used - like global DHCLIENT_SET_HOSTNAME set to yes and no per interface setup done, or two or more interfaces set to yes etc. In such case you can obtain hostname from any of networks and result is unpredictable.

src/lib/y2network/connection_config/base.rb
```diff
@@ -70,6 +70,8 @@ class Base
       attr_accessor :firewall_zone
       # @return [String] interface's hostname
       attr_accessor :hostname
+      # @return [Boolean] set to true if dhcp from this interfae sets machine hostname
```
teclator:
```suggestion
      # @return [Boolean] set to true if dhcp from this interface sets machine hostname
```

teclator: And not sure. I think we should permit also nil. File configuration overrides global. 

If set globally to set the hostname by dhcp but not per interface we should not write anything at all, or in other case we will override the global configuration.

If I get it correctly we will always set it so, it will be always be written in the ifcfg-file.

src/lib/y2network/sysconfig/connection_config_readers/base.rb
```diff
@@ -58,6 +58,7 @@ def connection_config
             conn.startmode.priority = file.ifplugd_priority if conn.startmode.name == "ifplugd"
             conn.ethtool_options = file.ethtool_options
             conn.firewall_zone = file.zone
+            conn.dhclient_set_hostname = file.dhclient_set_hostname == "yes"
```
teclator: see below

src/lib/y2network/sysconfig/connection_config_writers/base.rb
```diff
@@ -44,6 +44,7 @@ def write(conn)
           file.name = conn.description
           file.lladdr = conn.lladdress
           file.startmode = conn.startmode.to_s
+          file.dhclient_set_hostname = conn.dhclient_set_hostname ? "yes" : "no"
```
teclator: see above

teclator:
> It seems fine ... as i believe it reads values from AY profile ;-) However, there is no kind of validation. At least it could log an error when an insecure configuration is going to be used - like global DHCLIENT_SET_HOSTNAME set to yes and no per interface setup done, or two or more interfaces set to yes etc. In such case you can obtain hostname from any of networks and result is unpredictable.

Well, in an AY configuration I would not expect that kind of validation that we do in a manual config, basically because it usually it is an exported configuration, that is, the user already declared how it should like.

mchf:
That's why I don't insist on any kind of hard stop. But I think that warning for user, saying "your configuration is weird, but if you know what you're doing feel free to continue" together with a log entry would be in place here.

And regarding tristate value ... we currently propose a value when creating a configuration via UI. So, we can behave the same even here for explicitly unconfigured ifaces.

teclator:
> That's why I don't insist on any kind of hard stop. But I think that warning for user, saying "your configuration is weird, but if you know what you're doing feel free to continue" together with a log entry would be in place here.
> 
> And regarding tristate value ... we currently propose a value when creating a configuration via UI. So, we can behave the same even here for explicitly unconfigured ifaces.

But in AY the profile already defines the content of ifcfg files I mean:

```xml
<networking>
  <dns>
      ## Global option (/etc/sysconfig/network/dhcp)
      <dhcp_hostname config:type="boolean">yes</dhcp_hostname>
  </dns>
    <interface>
      <bootproto>dhcp</bootproto>
      <device>eth0</device>
      ## Do we also support this option in network-ng?
      <dhclient_set_default_route>yes</dhclient_set_default_route>
      ## Per interface ifcfg-eth0
      <dhclient_set_hostname>no</dhclient_set_hostname>
      <startmode>auto</startmode>
    </interface>
    <interface>
      <bootproto>dhcp</bootproto>
      <device>eth1</device>  
      # Not defined, then no reason for setting it.
      <startmode>auto</startmode>
    </interface>
  </interfaces>
</networking>

```

jreidinger:
btw `dhclient_set_default_route` is properly written down when doing testing. So looks like supported.

PR #1003, opened 2019-11-19
Add unknown interface type

teclator:
## Problem

With the re-factorization of yast2-network, we have not implemented support for some network interfaces that were marked as deprecated (i.e s390 IUCV).

When the physical interfaces present in the system are read, we determine the interface type, which, in this case, is provided by hwinfo. But if there is the InterfaceType is not declared we try to infer it from the interface name. 

In the case of an IUCV interface, as we are not supporting it and the `device_name` exported by ReadHardware is nil, the lan client basically crashes.

- https://trello.com/c/wGKdDJAo/1437-1-ostumbleweed-p5-1156285-s390x-yast-lan-crashes-when-there-are-iucv-interfaces-deprecated
- https://bugzilla.suse.com/show_bug.cgi?id=1156285

## Solution

Added a special type for the "Unknown" ones, which basically are omited from the interfaces list.

## To discuss

- Shall we ignore the **unknowns** from the list of the interfaces or shall we add it and then just do not list in the interfaces table?

- Shall we try to detect all the cards available and infer the type also for not supporting ones being able to show them but maybe as disabled? 

  For example, now we do not support **IUVC** intercaces but a `ifcfg-iucv0` file could exist, it probably will add a physical interface as not present because we didn't add it to the list because of unsupported.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Nov 25 09:36:43 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- Added a special type for the "Unknown" interfaces omitting them
+  from the interfaces list (bsc#1156285)
+- 4.2.32
+
+-------------------------------------------------------------------
 Mon Nov 25 08:34:55 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
 
 - Ignores invalid udev rules parts (bsc#1157361)
--- a/src/lib/y2network/interface_type.rb
+++ b/src/lib/y2network/interface_type.rb
@@ -149,7 +149,7 @@ module Y2Network
     IRDA = new(N_("Infrared"), "irda")
     # Loopback
     LO = new(N_("Loopback"), "lo")
-    # Not supported interfaces
-    UNSUPPORTED = new(N_("Unsupported"), "unsupported")
+    # Unknown interfaces
+    UNKNOWN = new(N_("Unknown"), "unknown")
   end
 end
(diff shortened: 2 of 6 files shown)


mchf:
missing changes, otherwise lgtm

src/lib/y2network/interface_type.rb
```diff
@@ -149,5 +149,7 @@ def method_missing(method_name, *arguments, &block)
     IRDA = new(N_("Infrared"), "irda")
     # Loopback
     LO = new(N_("Loopback"), "lo")
+    # Not supported interfaces
+    UNSUPPORTED = new(N_("Unsupported"), "unsupported")
```
imobachgs: Why not `UNKNOWN`? After all, that's the word you have used in the PR's description.

teclator: I was not sure, because the type is recognized by hwinfo, so, it is more a known one but unsupported ... so, not have a strong opinion about which one should be used.

teclator: And you have already realized I used **unknown** instead of unsupported in the description as I was already doubting the term to use

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Nov 25 09:36:43 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- Added a special type for the "Unknown" interfaces omitting them
+  from the the interfaces list (bsc#1156285)
```
imobachgs: "the the"

src/lib/y2network/interface_type.rb
mchf: well, unknown and unsupported have slightly different meaning. We can e.g. know the type but do not support its configuration ;-)

package/yast2-network.changes
teclator: Fixed, and added with --amend to not add an extra commit just for fixing the typo.

PR #1006, opened 2019-11-25
Skip not valid udev rule parts when reading drivers

teclator:
## Problem

When reading the existent udev rules, the parts assigned to the udev rule could have 'nil' entries corresponding to invalid udev rule parts.

- https://bugzilla.suse.com/show_bug.cgi?id=1157361

## Solution

Skip not valid udev rule parts when reading drivers

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Nov 25 08:34:55 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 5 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Nov 25 08:34:55 UTC 2019 - Knut Anderssen <kanderssen@suse.com>
+
+- When reading udev rules, skip from the assigned udev rule parts
```
imobachgs: np: I would simplify this sentence: "Ignores invalid udev rules parts." or something like that.

PR #1007, opened 2019-11-25
Wireless: fix auth_mode initialization

teclator:
## Problem

There is some type mismatch when reading existing wireless interface files and when initializing the wireless mode widget because of different types used. In some cases we use strings and in others we use symbols.

- https://bugzilla.opensuse.org/show_bug.cgi?id=1157394

## Solution

Use strings for `wireless mode` by now as the 'ad-hoc' mode needs some special handling in the interface file in case of symbol. By now we just ensure the mode is downcased when assigned to the connection config.

## Test

- Tested manually configuring a wireless interface with different modes verifying attributes are written correctly and also not lost when the client is called again with an existing wireless configuration.

### DIFF
--- a/src/lib/y2network/connection_config/wireless.rb
+++ b/src/lib/y2network/connection_config/wireless.rb
@@ -26,8 +26,8 @@ module Y2Network
       # wireless options
       #
       # FIXME: Consider an enum
-      # @return [Symbol] (:ad-hoc, :managed, :master)
-      attr_accessor :mode
+      # @return [String] (ad-hoc, managed, master)
+      attr_reader :mode
       # @return [String]
       attr_accessor :essid
       # @return [String] Network ID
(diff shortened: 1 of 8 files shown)


src/lib/y2network/connection_config/wireless.rb
```diff
@@ -26,7 +26,7 @@ class Wireless < Base
       # wireless options
       #
       # FIXME: Consider an enum
-      # @return [String] (Ad-hoc, Managed, Master)
+      # @return [Symbol] (:ad-hoc, :managed, :master)
```
dgdavid: NP: `:ad_hoc`

teclator: It is not just a NP, the interface_file , does a only a downcase.to_sym, which is fine when reading but as you commented we need an underscore substitution and do the inverse operation when writing the file. 

So, by now will maintin the original string and just ensure we are handling it correctly

dgdavid: Ah! right!

I simply thought that you make a typo in the comment, but it didn't come to my mind the described problem while I was doing the review.

Good point!

PR #1008, opened 2019-11-26
Fix mocking in hostname_reader_test

teclator:
## Problem

When running the unit tests with NM, the test raises an exception because obtaining the hostname by dhcp is not mocking.

Related to #1000 

## Solution

Just mock the read of the hostname by dhcp.

## Test

- Unit tests adapted and tested manually.

### DIFF
--- a/test/y2network/hostname_reader_test.rb
+++ b/test/y2network/hostname_reader_test.rb
@@ -25,11 +25,13 @@ describe Y2Network::HostnameReader do
 
   describe "#hostname" do
     let(:install_inf_hostname) { "linuxrc" }
+    let(:dhcp_hostname) { "dhcp" }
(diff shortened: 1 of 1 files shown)


mchf:
I assume that changes was left intentionally ...

teclator:
> I assume that changes was left intentionally ...

If you mean, changelog and bump version, yes, it was. It is not really needed and I prefer to have it in a separate PR, so, will be submitted by next one.

dgdavid:
> If you mean, changelog and bump version, yes, it was. It is not really needed and I prefer to have it in a separate PR, so, will be submitted by next one.

@teclator I use to put something like "No version bump/changelog entry required" somewhere in the PR description. Just an idea, in case it could help :)

PR #1009, opened 2019-11-26
Added supported interface types.

teclator:
##  Problem

There are many network interfaces that are rarely used or the support in SUSE has been deprecated. As we do not plant to support them in network-ng we should remove them in order to not be offered when adding a new connection.

The current list is obtained by the 'Yast::NetworkInterfaces.GetDeviceTypes' method.. see https://github.com/yast/yast-yast2/pull/981

In that PR, it was [recommended](https://github.com/yast/yast-yast2/pull/981/files#r348982058) to remove the dependency moving it to ther InterfaceType class.

Trello Card: https://trello.com/c/qg8k5n8y/1379-3-feature-jscsle-7753-drop-deprecated-network-technologies-in-yast2

## Solution

Define in the InterfaceType class the supported interfaces and remove the dependency on Yast::NetworkInterfaces for adding new connections.

## Test

- Added unit test for the new `Y2Network::InterfaceType.supported` class method.
- Tested manually

## Screenshots

Add Interface

Add Interface Ncurses

### DIFF
--- a/src/lib/y2network/interface_type.rb
+++ b/src/lib/y2network/interface_type.rb
@@ -19,6 +19,8 @@
 
 require "yast"
 
+Yast.import "Arch"
+
 module Y2Network
   # This class represents the interface types which are supported.
   # Class have helpers to check if given type is what needed. It check name and also short name:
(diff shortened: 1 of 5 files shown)


src/lib/y2network/interface_type.rb
```diff
@@ -40,6 +42,13 @@ def all
           .select { |c| c.is_a?(InterfaceType) }
       end
 
+      # Returns all the supported interfaces for the current arquitecture
```
teclator:
```suggestion
      # Returns all the supported interfaces for the current architecture
```

teclator:
@mchf @imobachgs thnx for reviewing it!

PR #1010, opened 2019-11-29
Separated Hostname class from DNS one

mchf:
## Problem

Hostname class is responsible for handling /etc/hostname which has almost nothing to do with DNS. It was part of DNS class for a historical reasons. As we implement a completely new backend it makes no sense to take this old sins with us.

- https://trello.com/c/uhlRip1d/1368-3-hostname-not-properly-set-during-installation
- https://fate.suse.com/319639

## Solution

I've also moved some logic from hostname_reader to newly created hostname class where it fits better from my POV. Especially logic for proposing hostname under different circumstances does not belong to the reader. It also makes easier and cleaner to implement [fate#31939](https://fate.suse.com/319639) and replaces hotfix from https://github.com/yast/yast-network/pull/1000

## TODO:

- ~~tests on new Hostname class~~
- ~~re-test manually for latest fixes~~


src/lib/y2network/autoinst/hostname_reader.rb
```diff
@@ -0,0 +1,57 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2network/hostname"
+require "y2network/sysconfig/hostname_reader"
+
+module Y2Network
+  module Autoinst
+    # This class is responsible of importing the AutoYast dns section
```
teclator: Probably copied from the dns_reader, needs some rewording

src/lib/y2network/autoinst/hostname_reader.rb
```diff
@@ -0,0 +1,57 @@
...
+      def config
+        Y2Network::Hostname.new(
+          dhcp_hostname: section.dhcp_hostname,
+          hostname:      section.hostname || default_hostname
+        )
+      end
+
+    private
+
+      # Returns a default hostname proposal for installer
```
teclator: If it is a proposal maybe would do it in a separate method even class..

src/lib/y2network/hostname.rb
```diff
@@ -0,0 +1,129 @@
+# Copyright (c) [2019] SUSE LLC
```
teclator: Shall we defined it here or maybe better in yast2?

I already tried to open a discussion with this PR: https://github.com/yast/yast-yast2/pull/921

What in cases where we have a minimal image without yast2-network but still being able to handle hostname configuration, could be that the case?, as you commented, it has nothing to be with dns, so.. maybe we should have it in yast2.. although stil under y2network namespace...

src/lib/y2network/autoinst/hostname_reader.rb
mchf: fixed

src/lib/y2network/hostname.rb
mchf: problem is that only UI we currently have is part of "yast2 lan". Moving it elsewhere was rejected. I'm currently not aware of any other use of hostname API

src/lib/y2network/autoinst/hostname_reader.rb
mchf: well the proposal most probably is not the right name here. Mostly it is about which hostname is picked as the one which is presented to the user under different circumstances.

src/lib/y2network/sysconfig/hostname_reader.rb
```diff
@@ -0,0 +1,131 @@
...
+Yast.import "Hostname"
+Yast.import "IP"
+Yast.import "Stage"
+Yast.import "NetHwDetection"
+
+module Y2Network
+  module Sysconfig
+    # This class is responsible for reading the hostname
+    #
+    # Depending on different circunstamces, the hostname can be read from different places (from a
```
teclator:
```suggestion
    # Depending on different circumstances, the hostname can be read from different places (from a
```

test/y2network/hostname_test.rb
```diff
@@ -0,0 +1,137 @@
...
+          expect(hostname.proposal).to eql "net_hostname"
+        end
+      end
+
+      context "and no hostname is known" do
+        let(:static_hostname) { nil }
+        let(:transient_hostname) { nil }
+
+        it "creates a random hostname" do
+          expect(hostname.proposal).to match(/linux-[a-z0-9]{4}$/)
```
teclator: So, during the installation, the proposal does not propose nothing, but it is proposed in case of a running system?

test/y2network/hostname_test.rb
```diff
@@ -0,0 +1,137 @@
...
+  describe "#save_hostname?" do
+    context "When used in installer" do
+      before(:each) do
+        allow(Yast::Stage).to receive(:initial).and_return(true)
+      end
+
+      context "without explicitly set hostname" do
+        let(:installer_hostname) { nil }
+
+        it "fails" do
```
teclator: np: why not explain what it returns? fails or succeeds looks like performing an action and this is more a question

teclator:
In general looks good to me, what is the state of the TODO?, is there still something missing, have it been tested manually already in a normal installation, autoyast etc?

test/y2network/hostname_test.rb
mchf: yes. I know that it is weird, but installation currently behaves according to the fate request in the description and running system behaves as before. In my POV this random stuff is obsolete thing, but I'm not sure whether it is safe to drop it.

test/y2network/hostname_test.rb
mchf: ok, will try to improve.

mchf:
> In general looks good to me, what is the state of the TODO?, is there still something missing, have it been tested manually already in a normal installation, autoyast etc?

I've updated the desc. Complex manual tests missing.

teclator:
@mchf it missed the changelog entry and version bump

PR #1012, opened 2019-12-05
Use a popup instead of a dialog (add interface config)

teclator:
## Problem

With the refactorization of network-ng, some workflows and dialogs have changed a little bit. That is the case of the dialog for adding a new interface configuration. 

During the review of #1009 , it was proposed to enhance the current dialog which looks ugly specially in big screens.

| pre_network-ng Qt|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
|   :---:      | :---:   |     :---:     |    :---:       |
| PreviousRefactorizationQt | CurrentQtBigScreen | CurrentNcursesBigScreen | CurrentNcurses80x24 |

Bear in mind that it is mainly for not present or virtual interfaces as interfaces which are already present but not configured will be recognized by the type read from the hardware information. Thus, the workflow is the same as editing a configured interface.

- https://trello.com/c/3myTtYoM/1478-3-enhance-the-selection-of-the-interface-type-in-network

## Solution

This PR implements one **(option 2)** of the proposed alternatives and also expose others with screenshots and descriptions in order to select the preferable one. There are probably other ways and maybe better alternatives to improve the current dialog, so suggestions are welcomed.

## Alternatives

Adding a new interface configuration is somehow an expert action. In that context the interface type name is already self descriptive. Therefore, it seems that improving the current description could be nice as part of the help dialog but not so relevant for clarifying the option to be chosen. Alternatives **(1, 2 & 5)**  are based on that opinion, while alternatives **(3 & 4)** shows screenshots using the new SingleItemSelector.

### 1. Current dialog but adding a title and centering the content horizontally and veritically

|   Qt    | Ncurses 80x24 | Ncurses 132x43 | Installation |
| :---:   |     :---:     |    :---:       | :---: |
| CenterDialogQt  | CenterDialogNcurses80x24 | CenterDialogNcurses134x43 | Installation |

#### Adding ad MarginBox and a title (this PR changes)

|   Qt    | Ncurses 80x24 | Ncurses 132x43 | Installation |
| :---:   |     :---:     |    :---:       |   :---:      |
| OnlyMargin_Qt | OnlyMargin_Ncurses80x24 | OnlyMargin_Ncurses132x43 | OnlyMarginInstallation_Qt |

#### With MarginBox & HVSquash

|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
| :---:   |     :---:     |    :---:       |
|         | OnlyMargin_Ncurses80x24 | OnlyMargin_Ncurses132x43 |

#### The current dialog but using the new SingleItemSelector

|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
| :---:   |     :---:     |    :---:       |
| SingleItemSelectorQt | | SingleItemSelectorNcurses |

#### The current dialog but using the new SingleItemSelector and Squashing the content horizontally

|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
| :---:   |     :---:     |    :---:       |
| SingleItemSelector_Qt_HSquash | HSquash_SingleItemSelector_ncurses80x24 | HSquash_SingleItemSelector_ncurses_full_screen

#### The current dialog but using the new SingleItemSelector and Squashing the content horizontally and verically

|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
| :---:   |     :---:     |    :---:       |
| SingleItemSelector_Qt_HVSquash | | |

### 4. Use SingleItemSelector in a popup dialog
    
|   Qt    | Ncurses 80x24 | Ncurses 132x43 |
| :---:   |     :---:     |    :---:       |
| PopupSingleItemSelectorQt | | PopupSingleItemSelectorNcurses132x43 |
    
#### example from storage-ng

MenuButton

## NOTES

- In the ncurses popup, the buttons do not respect the order and are not aligned as in the wizard, something that is ok in the Qt.
- When running the client in Gnome with xdg-su, the style is basically lost, and frames do not use a background or border in QT.

### DIFF
--- a/src/lib/y2network/dialogs/add_interface.rb
+++ b/src/lib/y2network/dialogs/add_interface.rb
@@ -17,7 +17,7 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
-require "cwm/dialog"
+require "cwm/popup"
 require "y2network/widgets/interface_type"
 require "y2network/interface_config_builder"
 
@@ -29,13 +29,17 @@ Yast.import "NetworkInterfaces"
 module Y2Network
   module Dialogs
     # Dialog which starts new device creation
-    class AddInterface < CWM::Dialog
+    class AddInterface < CWM::Popup
       def initialize(default: nil)
         @type_widget = Widgets::InterfaceType.new(default: default ? default.short_name : nil)
       end
 
+      def title
+        _("Add interface configuration")
+      end
+
       def contents
-        HBox(
+        MarginBox(1, 0.4,
           @type_widget
         )
       end


shundhammer:
Whichever we choose, it should have

- a headline saying what the sub-workflow is all about ("Add Interface Configuration"; notice the headline capitalization!)

- a hint indicating what the selection is ("Device Type")

- if there are multiple steps, a "Back" and a "Next" button.

If we only have descriptions that are just unabbreviated versions of the item itself, they don't make sense, so using an ItemSelector does not add any benefit:

- **VLAN**
  Virtual LAN
- **Bridge**
  Network bridge

This is pointless (it's equally pointless as help texts BTW). If we can come up with descriptions that really explain what that option is, it might be useful; but not like this.

shundhammer:
A frame that almost touches an outer frame is also quite pointless. A frame is useful if it is grouping something. If there is nothing or almost nothing else in the window, make sure to have some margins to the window border (at least to the left and to the right border).

teclator:
> Whichever we choose, it should have
> 
>     * a headline saying what the sub-workflow is all about ("Add Interface Configuration"; notice the headline capitalization!)
> 
>     * a hint indicating what the selection is ("Device Type")
> 
>     * if there are multiple steps, a "Back" and a "Next" button.

That is one of the side effects of the Popup dialog. Somehow the sequence will start once selected the type to add, and then it will work as editing an existent one. So, in case of choosing the popup I would propose to not permitting going back from the sequence.

> 
> 
> If we only have descriptions that are just unabbreviated versions of the item itself, they don't make sense, so using an ItemSelector does not add any benefit:
> 
>     * **VLAN**
>       Virtual LAN
> 
>     * **Bridge**
>       Network bridge
> 
> 
> This is pointless (it's equally pointless as help texts BTW). If we can come up with descriptions that really explain what that option is, it might be useful; but not like this.

In all the versions at least we should sort them having the most common ones first. Also take in account that 'Adding a connection configuration in YaST' is somehow already an expert option. That is, when the interface is a physical one and already plugged in the system it should be already listed in the table list and in that case the dialog will not shown, the dialog is mainly for virtual interfaces or not present ones.

Bearing that in mind, I do not find so useful to have long descriptions. We could try to improve the description of each interface type for being part of the help instead of the useless text we currently have, but in most cases, the type short name is self descriptive for most of the users.

teclator:
@kwwii : Would be nice if you can take a look and give us your opinion about the different alternatives.

teclator:
This PR was created mainly for discussing. It has been replaced by #1014 which was already merged, closing it by now.

PR #1015, opened 2019-12-16
Fixed the proposal of the virtualization configuration

teclator:
## Problem

With network-ng, a regression was introduced in the network virtualization proposal.

That is, if for example there is only one interface configured by dhcp , it is skipped as bridge port candidate because it is already configured (of course same happen with any configured interface):

https://github.com/yast/yast-network/blob/master/src/modules/Lan.rb#L752

Introduced by:

https://github.com/yast/yast-network/commit/4e0a75e588a480bcbe3d5b366e6a20eb65cc65bb#diff-a44342e8d97412ca7486bd2d61e8e429L864

The configuration of the interface should be copied to the bridge and it should be added as a bridge port as it was in the past.

The issue was detected as consequence of these bugs, which has been closed as the crash should not happen with last changes in network-ng but will face the described regression:

- https://bugzilla.suse.com/show_bug.cgi?id=1153373
- https://bugzilla.suse.com/show_bug.cgi?id=1155852

https://trello.com/c/Dh9D0SIi/1443-sles15-sp2-p2-1156986-yast-does-not-propose-configured-interfaces-for-virtualization-bridges

## Solution

The configuration is created again and the logic has been moved to its own class.

## Tests

- Created unit tests and tested manually in a running system with the virtualization client

### DIFF
--- a/src/lib/y2network/interface_config_builder.rb
+++ b/src/lib/y2network/interface_config_builder.rb
@@ -347,7 +347,13 @@ module Y2Network
 
     # @param [String] value
     def remote_ip=(value)
-      ip_config_default.remote_address = IPAddress.from_string(value)
+      return unless ip_config_default
+
+      ip_config_default.remote_address = if value.nil? || value.empty?
+        nil
+      else
+        IPAddress.from_string(value)
+      end
     end
 
     # Gets Maximum Transition Unit
@@ -365,6 +371,9 @@ module Y2Network
     def configure_as_slave
       self.boot_protocol = "none"
       self.aliases = []
+      self.ip_address = nil
+      self.subnet_prefix = ""
+      self.remote_ip = ""
     end
 
     # @param info [Hash<String,Object>] Hardware information
--- a/src/lib/y2network/interface_config_builder.rb
+++ b/src/lib/y2network/interface_config_builder.rb
@@ -334,7 +334,6 @@ module Y2Network
       @connection_config.hostname = value
     end
 
-    # sets remote ip for ptp connections
     # @return [String]
     def remote_ip
       default = @connection_config.ip
@@ -345,7 +344,10 @@ module Y2Network
       end
     end
 
-    # @param [String] value
+    # Sets remote ip for ptp connections
+    #
+    # @param [String, nil] value
+    # @return [IPAddress, nil]
     def remote_ip=(value)
       return unless ip_config_default
 
--- a/src/lib/y2network/virtualization_config.rb
+++ b/src/lib/y2network/virtualization_config.rb
@@ -79,19 +79,20 @@ module Y2Network
 
   private
 
-    # Convenience method that returns true if the current item has link and can
-    # be enslaved in a bridge.
+    # Convenience method that returns true if the interface given is connected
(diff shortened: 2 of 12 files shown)


src/lib/y2network/interface_config_builder.rb
```diff
@@ -347,7 +347,13 @@ def remote_ip
 
     # @param [String] value
```
mchf: doc for nil case missing

src/modules/Lan.rb
```diff
@@ -741,27 +741,35 @@ def UseNetworkManager
...
+
+      interface_names.each do |interface_name|
+        interface = yast_config.interfaces.by_name(interface_name)
         bridge_builder = Y2Network::InterfaceConfigBuilder.for("br")
         bridge_builder.name = yast_config.interfaces.free_name("br")
 
         next unless connected_and_bridgeable?(bridge_builder, interface)
 
-        next if yast_config.connections.by_name(interface.name)
+        connection = yast_config.connections.by_name(interface.name)
```
mchf: well ```interface.name``` vs ```interface_name``` is a bit confusing, so I'd expect that just one of them is in use

src/lib/y2network/virtualization_config.rb
```diff
@@ -0,0 +1,120 @@
...
+      true
+    end
+
+  private
+
+    # Convenience method that returns true if the current item has link and can
+    # be enslaved in a bridge.
+    #
+    # @return [Boolean] true if it is bridgeable
+    def connected_and_bridgeable?(bridge_builder, interface)
```
mchf: needs some love in doc ... refers to "item" or "lan item" in doc or log messages

src/lib/y2network/virtualization_config.rb
```diff
@@ -0,0 +1,120 @@
...
+      if !bridge_builder.bridgeable_interfaces.map(&:name).include?(interface.name)
+        log.info "The interface #{interface.name} cannot be proposed as bridge."
+        return false
+      end
+
+      hwinfo = interface.hardware
+      if !hwinfo.present? || !hwinfo.link
```
mchf: NP: an idea ... this could be masked / hidden in something like ```interface.link?``` ... looks more pretty to me than searching for hardware config.

src/lib/y2network/virtualization_config.rb
```diff
@@ -79,18 +79,19 @@ def create
...
     def connected_and_bridgeable?(bridge_builder, interface)
       if !bridge_builder.bridgeable_interfaces.map(&:name).include?(interface.name)
         log.info "The interface #{interface.name} cannot be proposed as bridge."
         return false
       end
 
-      hwinfo = interface.hardware
-      if !hwinfo.present? || !hwinfo.link
+      unless interface.connected?
         log.warn("Lan item #{interface.inspect} has link:false detected")
```
mchf: lan item is no longer true

teclator: Ups leftover :P

mchf: yeah it does not surrender that easily ;-)

PR #1018, opened 2019-12-23
Improved exception handling when querying transient hostname

mchf:
Catched by QA.

Follow up of https://github.com/yast/yast-network/pull/1010

### DIFF
--- a/src/lib/y2network/sysconfig/hostname_reader.rb
+++ b/src/lib/y2network/sysconfig/hostname_reader.rb
@@ -95,11 +95,13 @@ module Y2Network
         host.empty? ? nil : host
       end
 
-      # Reads the hostname known to the resolver
+      # Reads the (transient) hostname known to the resolver
       #
-      # @return [String] Hostname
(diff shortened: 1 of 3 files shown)


src/lib/y2network/sysconfig/hostname_reader.rb
```diff
@@ -95,11 +95,13 @@ def hostname_from_install_inf
...
       end
 
-      # Reads the hostname known to the resolver
+      # Reads the (transient) hostname known to the resolver
       #
-      # @return [String] Hostname
+      # @return [String, nil] Hostname or nil if transient hostname is not known
       def hostname_from_resolver
         Yast::Execute.on_target!("/usr/bin/hostname", "--fqdn", stdout: :capture).strip
+      rescue Cheetah::ExecutionFailed
```
jreidinger: just not sure, is it also logged in case that command failed? I think yes, but not sure.

mchf: yes it is ... by cheetah

PR #1020, opened 2020-01-02
Fixed compatibility between old and new DNS / Hostname code

mchf:
## Problem ##

Detected by QA.
When user tried to store hostname and internal error was raised (no method).

Related to https://github.com/yast/yast-network/pull/1010

## Solution ##

Introduced reasonable aliases

### DIFF
--- a/test/dns_test.rb
+++ b/test/dns_test.rb
@@ -39,7 +39,7 @@ describe Yast::DNS do
     Y2Network::DNS.new
   end
   let(:hostname_config) do
-    Y2Network::Hostname.new(hostname: "install", dhcp_hostname: true)
+    Y2Network::Hostname.new(static: "install", dhcp_hostname: true)
   end
 
   subject { Yast::DNS }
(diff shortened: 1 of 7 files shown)


mchf:
@dgdavid 
yes, removing obsolete code is intentional as I found it by chance when working on testsuite for this PR. However, I have some technical issue when trying to build run testsuite locally so will update the PR asap.

test/dns_test.rb
```diff
@@ -218,4 +218,21 @@
...
+
+  describe "#hostname" do
+    it "provides static hostname" do
+      expect(subject.hostname).to eq "install"
+    end
+  end
+
+  describe "#hostname=" do
+    let(:hostname_config) do
+      Y2Network::Hostname.new
```
dgdavid: NP: I wonder why is necessary to do that to test the  `#hostname=` alias method. I'd expect to be able to set/change the hostname/static value even with the initial configuration given.

mchf: just didn't want to corrupt it ;-)

PR #1021, opened 2020-01-03
Support for reading / writing MTU

mchf:
## Problem ##

https://bugzilla.suse.com/show_bug.cgi?id=1159970

As described in title - it was completely missing.

## Solution ##

Updated interface file class together with connection reader / writer.


jreidinger:
code wise LGTM, I just would like to see it covered by tests.

PR #1022, opened 2020-01-06
Test that ruby-bindings and CommandLine.rb propagate errors

mvidner:
Related to https://github.com/yast/yast-ruby-bindings/pull/237

I put it here because it would get run by openQA out of the box: https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/2a5be2f7b668279c5db275de190e7d5cd92c24f7/tests/console/yast2_cmdline.pm

### DIFF
--- a/t/exit-codes.t
+++ b/t/exit-codes.t
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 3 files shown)


t/exit-codes.t
```diff
@@ -0,0 +1,43 @@
...
+tapfail() {
+  echo -n "not "
+}
+
+YAST=/usr/sbin/yast
+
+TEST="1 An unknown command produces an error exit code"
+echo "# $TEST"
+! $YAST lan nosuchcommand || tapfail
+echo "ok $TEST"
```
jreidinger: well, as small smoke test it looks fine, but it does not test whole range of various scenarios. So if openqa willing to add it and we are willing to fix it ( not as network tap test :)

mvidner: Yes, this is a quick and easy fix for the common case. It fixes the problem that I had, I thought you may have been testing some other things.

mvidner:
@jreidinger this one depends on https://github.com/yast/yast-yast2/pull/1004, please review that one too

PR #1023, opened 2020-01-13
Update only needed info when an interface already have an udev rule (bsc#1158025)

teclator:
## Problem

When there is some `change` in the network `configuration` state, **YaST** writes the `udev` rules in its **own form**, that is, we get rid of some of the parts.

For example, the rules created by `write_net_rules` are:

```
# PCI device 0x1af4:0x1041 (virtio-pci)
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="virtio-pci", ATTR{dev_id}=="0x0", KERNELS=="0000:01:00.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth0"

# PCI device 0x8086:0x10d3 (e1000e)
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="e1000e", ATTR{dev_id}=="0x0", KERNELS=="0000:07:00.0", ATTR{type}=="1", KERNEL=="eth*", NAME="eth1"
```

And after changing some general configuration in network the rules are **rewritten** as:

```
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{type}=="1", KERNELS=="0000:07:00.0", NAME="eth1"
SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{type}=="1", KERNELS=="virtio0", NAME="eth0"
```

Especially when a `virtio` netcard, YaST overwrites existing udev rules **based on** `busid` with the hwinfo `busid` which is `virtio`, but in that case it should use the `parent_bus_id` "0000:01:00.0". That could be handle by `hwinfo`.

So, in general the bug is about not removing default udev parts created by the udev rule generator if it is not strictly necessary.

- Trello Card: https://trello.com/c/gf2QyfLL
- https://bugzilla.suse.com/show_bug.cgi?id=1158025

## Solution

- Only the udev rule parts that has changed or need to be replaced are now touched.
- When the hardware is read, we use the parent_bus_id as the busid for virtio netcards. The busid exported by hwinfo is stored as a `sub_device_busid`.

## Test

- Added unit test
- Tested manually **(In Build 122.1 with a dud)**

### DIFF
--- a/src/lib/y2network/sysconfig/interfaces_writer.rb
+++ b/src/lib/y2network/sysconfig/interfaces_writer.rb
@@ -34,7 +34,6 @@ module Y2Network
     #
     # @see Y2Network::InterfacesCollection
     class InterfacesWriter
-      include Yast::Logger
       # Writes interfaces hardware configuration and refreshes udev
       #
       # @param interfaces [Y2Network::InterfacesCollection] Interfaces collection
@@ -44,22 +43,6 @@ module Y2Network
       end
 
     private
-
-      # Creates an udev rule to rename the given interface
-      #
-      # @param iface [Interface] Interface to generate the udev rule for
-      # @return [UdevRule,nil] udev rule or nil if it is not needed
-      def renaming_udev_rule_for(iface)
-        log.info("Generating a new udev rule for #{iface.name} based on: " \
-                 "#{iface.renaming_mechanism.inspect}")
-        case iface.renaming_mechanism
-        when :mac
-          Y2Network::UdevRule.new_mac_based_rename(iface.name, iface.hardware.mac)
-        when :bus_id
-          Y2Network::UdevRule.new_bus_id_based_rename(iface.name, iface.hardware.busid,
-            iface.hardware.dev_port)
-        end
-      end
 
       # Creates an udev rule to set the driver for the given interface
       #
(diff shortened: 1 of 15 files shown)


src/lib/y2network/sysconfig/interfaces_writer.rb
```diff
@@ -77,15 +80,41 @@ def update_udevd(interfaces)
...
+
+      # Writes down the current interfaces udev rules and the custom rules that
+      # were present when read and that are still valid
+      #
+      # @see Y2Network::UdevRule#write_net_rules
       def update_renaming_udev_rules(interfaces)
-        udev_rules = interfaces.map { |i| renaming_udev_rule_for(i) }.compact
+        udev_rules = interfaces.map do |iface|
+          update_renaming_udev_rule(iface) if iface.udev_rule
+          iface.udev_rule ||= renaming_udev_rule_for(iface)
```
jreidinger: this is a bit strange. If there is already udev_rule it will be updated. And if there is not yet such udev_rule it is created. Why not having it in one method?

jreidinger: also I a bit expect to be a method in interface and not only in writer.

teclator: Originally I put the logic in one method, but was not sure about merging the logic for new rules and for existing ones.

PR #1025, opened 2020-01-21
Fix error while creating bridge under virtualization

goldwynr:
connected_and_bridgeable is using log functions. So, class
VirtualizationConfig needs to include Yast::Logger in order to use
log.* functions, else it bugs with:

Caller: 
/usr/share/YaST2/lib/y2network/virtualization_config.rb:100:in `connected_and_bridgeable?'
undefined local variable or method `log' for #Did you mean? load


teclator:
@goldwynr thnx for the patch!

PR #1026, opened 2020-01-21
Bump version & changelog for (#1025)

teclator:
Changes needed for #1025

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,8 @@
 Tue Jan 21 07:24:03 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
 - bsc#1156986
-  - Fix Yast::Logger dependency in Y2Network::VirtualizationConfig
+  - Fix crash when logging that an interface is not connected or
+    is not bridgeable when propising the virtualization config.
 - 4.2.46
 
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 21 07:24:03 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- bsc#1156986
+  - Fix Yast::Logger dependency in Y2Network::VirtualizationConfig
```
jreidinger: well, better to have here real user use case. This is more developer oriented message. I expect something like "fix crash when ..."

teclator: Ok, updated according to your suggestions.

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue Jan 21 07:24:03 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- bsc#1156986
+  - Fix crash when logging that an interface is not connected or
+    is not bridgeable when propising the virtualization config.
```
jreidinger: typo propising

PR #1027, opened 2020-01-22
[bsc#1161234] Do not read missing ifcfg files

imobachgs:
Do not try to read an `ifcfg-*` file that we know in advance that it is missing. See [bsc#1161234](https://bugzilla.suse.com/show_bug.cgi?id=1161234) for further information.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Jan 23 09:16:03 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not try to read an ifcfg-* file that does not exist
+  (bsc#1161234).
+
+-------------------------------------------------------------------
 Wed Jan 22 16:43:21 UTC 2020 - Ancor Gonzalez Sosa <ancor@suse.com>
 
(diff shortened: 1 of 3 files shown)


test/y2network/sysconfig/interface_file_test.rb
```diff
@@ -65,6 +65,26 @@ def file_content(scr_root, file)
...
+
+    context "when the file for the given interface exists" do
+      let(:interface) { "eth0" }
+
+      it "returns true" do
+        expect(file.exist?).to eq(true)
+      end
+    end
+
+    context "when the file for the given interface exists" do
```
teclator:
```suggestion
    context "when the file for the given interface does not exist" do
```

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Thu Jan 23 09:16:03 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not try to read an ifcfg-* file that does not exist
+  (bsc#1161234).
```
teclator: Any reason to not bump the version?

imobachgs: Yes, it is a change that has a minimal impact, so I would prefer it to be released with the next version.

PR #1029, opened 2020-01-28
S390 activation dialog

teclator:
## Problem

When the new interfaces table widget was introduced, the [call](https://github.com/yast/yast-network/pull/926/files#diff-da30c24a016201e350c6d2d9e78a978bL471) for activating the group devices was accidentally removed and it goes directly to the connection config configuration.

Apart of that, the interfaces table shows one line per group device channel. The interfaces are basically read by a `hwinfo --netcard` call although no of them have a name yet (it was the interfaces table id). Which means that in case of editing it will find the first interface without a name instead of the one selected.

InterfacesPerChannel

Last but not least, the internal config was not refreshed when one of the interfaces was selected and enabled with the activation dialog. That is, the new interface should be shown and the channel devices needed to be removed.

- https://trello.com/c/dWOFuHGr/1584-sles15-sp2-p2-1160997-sles-15-sp2-beta1-cannot-configure-additional-network-interfaces-during-installation-yast
- https://bugzilla.suse.com/show_bug.cgi?id=1160997

## Solution

- S390 offline group devices are shown in the overview dialog and can be enabled pressing 'Edit'. Only the s390 group device is listed (in the past it listed a line per device channel)

OverviewOfflineS390Devices
S390ActivationDialog

- When a device is activated the overview (Interfaces Table) is refreshed even if canceled in the connection config sequence.

OverviewAfterActivation

- Activation errors details will be shown in the popup.

WrongChannelGiven
WrongOptionsGiven
ActivationErrorDetails

## Tests

- Added unit tests
- Tested manually

## TODO

- In case of a connection config exist for the activated interface (for example ifcfg-eth1). It should be merged with the one created for S390 group device. It is already a problem in previous SLE versions so no new because of network-ng. See the image below from SLE-12-SP4 configuration:
SLE-12-SP4_s390_activation_with_existent_ifcfg-file
 - Having a button at the right or somewhere for showing the s390 group devices summary would be nice to have as we have for DASD (separate installation step). That maybe could permit to activate or deactivate s390 group devices (currently only activate is permitted)

DASDActivation

### DIFF
--- a/src/lib/y2network/dialogs/s390_device_activation.rb
+++ b/src/lib/y2network/dialogs/s390_device_activation.rb
@@ -80,26 +80,21 @@ module Y2Network
       end
 
       def run
-        ret = nil
-        loop do
-          ret = super
-          if ret == :next
-            _stdout, stderr, status = activator.configure
-            configured = status.zero?
+        ret = super
+        if ret == :next
+          _stdout, stderr, status = activator.configure
+          configured = status.zero?
 
-            if configured
-              interface_name = activator.configured_interface
-              builder.name = interface_name
-              add_interface(interface_name)
-            end
-
-            if !configured || builder.name.empty?
-              show_activation_error(stderr)
-              ret = :redraw
-            end
+          if configured
+            interface_name = activator.configured_interface
+            builder.name = interface_name
+            add_interface(interface_name)
           end
 
-          break if ret != :redraw
+          if !configured || builder.name.empty?
+            show_activation_error(stderr)
+            return run
+          end
         end
 
         ret
@@ -130,8 +125,7 @@ module Y2Network
       end
 
       def add_interface(name)
-        interface = reader.interfaces.by_name(name)
-        config.interfaces << interface if interface
+        config.interfaces << reader.interfaces.by_name(name)
       end
     end
   end
--- a/src/lib/y2network/dialogs/s390_device_activation.rb
+++ b/src/lib/y2network/dialogs/s390_device_activation.rb
@@ -21,6 +21,9 @@ require "cwm/dialog"
 require "y2network/s390_device_activator"
 require "y2network/widgets/s390_common"
 require "y2network/widgets/s390_channels"
+require "y2network/sysconfig/interfaces_reader"
+
+Yast.import "Lan"
 
 module Y2Network
   module Dialogs
@@ -64,6 +67,7 @@ module Y2Network
         @activator = activator
         @activator.propose!
         @builder = activator.builder
+        @builder.newly_added = false
       end
 
       def title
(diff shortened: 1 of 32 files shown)


src/lib/y2network/dialogs/s390_device_activation.rb
```diff
@@ -75,23 +80,26 @@ def contents
...
+              add_interface(interface_name)
+            end
+
+            if !configured || builder.name.empty?
+              show_activation_error(stderr)
+              ret = :redraw
+            end
           end
+
+          break if ret != :redraw
```
jreidinger: I am not sure I like this whole loop just for redraw. What about instead something like `return run if ret == :redraw` and kill that loop? Of course if you trying it for long it can be result in stack overflow, but it is really quite big.

src/lib/y2network/dialogs/s390_device_activation.rb
```diff
@@ -100,6 +108,31 @@ def run
...
+        @reader ||= Y2Network::Sysconfig::InterfacesReader.new
+      end
+
+      def config
+        Yast::Lan.yast_config
+      end
+
+      def add_interface(name)
+        interface = reader.interfaces.by_name(name)
+        config.interfaces << interface if interface
```
jreidinger: this looks a bit strange for me. In which case it can fail to find interface?

src/lib/y2network/interface_config_builder.rb
```diff
@@ -64,6 +64,8 @@ def self.for(type, config: nil)
     attr_writer :renaming_mechanism
     # @return [Y2Network::Interface,nil] Underlying interface if it exists
     attr_reader :interface
+    # @return [Boolean] True when it is a new connection
+    attr_accessor :newly_added
```
jreidinger: well, why not only writer? there is already `newly_added?` reader below.

src/lib/y2network/dialogs/s390_device_activation.rb
teclator: It should not fail, I could assign directly the reader.interfaces.by_name(name) interface.

src/lib/y2network/dialogs/s390_device_activation.rb
teclator: well, it is really about not closing the dialog in case of an error, I initially used nil instead of :redraw so, we could do a return run in case of error as you propose instead of the loop

src/lib/y2network/dialogs/s390_device_activation.rb
```diff
@@ -75,23 +80,26 @@ def contents
...
+
+            if configured
+              interface_name = activator.configured_interface
+              builder.name = interface_name
+              add_interface(interface_name)
+            end
+
+            if !configured || builder.name.empty?
+              show_activation_error(stderr)
+              ret = :redraw
```
teclator: @jreidinger that is what you suggest, isn't it? apart of removing the loop of course

```suggestion
              return run
```

PR #1031, opened 2020-02-03
Fix call to remote_address method

teclator:
## Problem

The configuration summary breaks when a remote IP is configured

RemoteAddressError

## Solution

Fix method call.

## Tests

- Tested manually
- Added unit test


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Feb  3 10:44:31 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix connection configuration summary when using a remote
+  ip address (bsc#1162483)
```
imobachgs: np: IP

PR #1033, opened 2020-02-05
Hostname(s) documentation

mchf:
## Problem

The hostname / DNS setup has changed recently in YaST and the user miss some of the options that YaST allowed in the past. Thus, the documentation should be updated reflecting the new changes and clarifying why some of the options do not make sense anymore.

- https://trello.com/c/9zNXXS8E/1611-hostname-hostname-host-we-need-documentation


doc/hostname.md
```diff
@@ -0,0 +1,33 @@
...
+
+YaST is able to take care of this setup too. If you want to configure ```DHCLIENT_SET_HOSTNAME``` option via YaST then go to ```Hostname / DNS``` tab and set desired configuration in "Set Hostname via DHCP" combobox. Only safe configurations are allowed there. If you pick one of predefined setups there, YaST will set global even local options for you.
+
+However, we still talk about naming local system here. This name is also called "Transient hostname" in systemd world. You can display it e.g. by ```hostnamectl --transient```.
+
+## Static network hostname(s) (/etc/hosts) ##
+
+We're getting off the local system here.
+
+In networking world we are used to access remote computers by human readable names instead of networking addresses (IPs). E.g. we want to use google.com instead of 216.239.36.117 when browsing internet. There are severak ways how to achieve this. If you don't want to bother with DNS, you can use local database in /etc/hosts. This file is used as local database translating IP addresses to hostnames - one IP per line.
```
teclator:
```suggestion
In networking world we are used to access remote computers by human readable names instead of networking addresses (IPs). E.g. we want to use google.com instead of 216.239.36.117 when browsing internet. There are several ways how to achieve this. If you don't want to bother with DNS, you can use local database in /etc/hosts. This file is used as local database translating IP addresses to hostnames - one IP per line.
```

doc/hostname.md
```diff
@@ -0,0 +1,33 @@
...
+
+We have several hostnames in linux. Every of them has different usage and different place where to set it up. Some of them can be configured in the YaST's lan module.
+
+## Local system hostname / static hostname (/etc/hostname) ##
+
+You can put a string into /etc/hostname. This string is than used as so called "local system hostname" which is set in the kernel during the boot. This name is used for identification of your local system. This name is typically used e.g. in logs where individual log entries can be prepended by this name. More importantly this name is not related to networking, so putting e.g. FQDN (Fully Qualified Domain Name) has no sense.
+
+Content of this file can be displayed by ```hostname``` command or in systemd world by ```hostnamectl --static```
+
+In YaST you can edit this hostname on Hostname / DNS tab, hostname field.
```
teclator: Why about adding a screenshot? And taking in account that neither the kernel nor hostnamectl have validation or restriction about using a FQDN how YaST handle that situation?

doc/hostname.md
```diff
@@ -0,0 +1,33 @@
...
+
+YaST is able to take care of this setup too. If you want to configure ```DHCLIENT_SET_HOSTNAME``` option via YaST then go to ```Hostname / DNS``` tab and set desired configuration in "Set Hostname via DHCP" combobox. Only safe configurations are allowed there. If you pick one of predefined setups there, YaST will set global even local options for you.
+
+However, we still talk about naming local system here. This name is also called "Transient hostname" in systemd world. You can display it e.g. by ```hostnamectl --transient```.
+
```
teclator: What about adding some explanation what is the default during an installation?

mchf: added a few sentences

teclator:
NP: I would split the PR moving the fixes to a different PR

mchf:
> NP: I would split the PR moving the fixes to a different PR

as customer wishes ;-)

doc/hostname.md
```diff
@@ -0,0 +1,49 @@
+## Hostname in Linux ##
+
+We have several hostnames in linux. Every of them has different usage and different place where to set it up. Some of them can be configured in the YaST's lan module.
```
teclator:
```suggestion
We have several hostnames in linux. Each of them has different usage and different place where to set it up. Some of them can be configured by the YaST's lan module.
```

doc/hostname.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Local system hostname / static hostname (/etc/hostname) ##
+
+You can put a string into /etc/hostname. This string is than used as so called "local system hostname" which is set in the kernel during the boot. This name is used for identification of your local system. This name is typically used e.g. in logs where individual log entries can be prepended by this name. More importantly this name is not related to networking, so putting e.g. FQDN (Fully Qualified Domain Name) has no sense.
+
+Content of this file can be displayed by ```hostname``` command or in systemd world by ```hostnamectl --static```
+
+In YaST you can edit this hostname on Hostname / DNS tab, hostname field. Unlike other ways mentioned above, YaST validates your input.
```
teclator: would complete it with ... allowing or not allowing the use of .. FQDN

doc/hostname.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Dynamic system hostname(s) (via DHCP) ##
+
+In modern networks is quite common to use DHCP for configuring workstations. If DHCP server supports that you can even configure hostname for its clients. If you want to use hostname from DHCP server on the client, you have to allow it. You have to configure ```DHCLIENT_SET_HOSTNAME``` option on the client. This option can be configured on global level for all interfaces in ```/etc/sysconfig/network/dhcp``` file - this setup is then used as global default for all interfaces without that option explicitly configured. You can also configure this option per interface in the interface's ifcfg file. It is very easy to put the system into wrong state here and you will get unpredictable hostname(s) for your system.
```
teclator:
```suggestion
In modern networks is quite common to use DHCP for configuring workstations. If the DHCP server supports that it can even configure hostname for its clients. If you want to use hostname from DHCP server on the client, you have to allow it. You have to configure ```DHCLIENT_SET_HOSTNAME``` option on the client. This option can be configured on global level for all interfaces in ```/etc/sysconfig/network/dhcp``` file - this setup is then used as global default for all interfaces without that option explicitly configured. You can also configure this option per interface in the interface's ifcfg file. It is very easy to put the system into wrong state here and you will get unpredictable hostname(s) for your system.
```

doc/hostname.md
```diff
@@ -0,0 +1,49 @@
+## Hostname in Linux ##
+
+We have several hostnames in linux. Every of them has different usage and different place where to set it up. Some of them can be configured in the YaST's lan module.
+
+## Local system hostname / static hostname (/etc/hostname) ##
+
+You can put a string into /etc/hostname. This string is than used as so called "local system hostname" which is set in the kernel during the boot. This name is used for identification of your local system. This name is typically used e.g. in logs where individual log entries can be prepended by this name. More importantly this name is not related to networking, so putting e.g. FQDN (Fully Qualified Domain Name) has no sense.
+
+Content of this file can be displayed by ```hostname``` command or in systemd world by ```hostnamectl --static```
```
mvidner: Markdown: Inline, a pair of single backquotes is enough. Triple backquotes are for multiline code
```md
    ```sh
    cat /etc/hostname
    cat /etc/HOSTNAME
    ```
```

vs
```md
    Blah `cat /etc/hostname` blah
```

mvidner: To be pedantic, which is the point of this file: No.

`hostname` displays the kernel hostname, like `uname -n` does. That kernel hostname is initialized from `/etc/hostname` (probably) by `hostnamectl`. Just running `hostnamectl` reads the file and sets it. (even if using the `status` subcomand!)

doc/hostname.md
mvidner: In such a long suggestion, I can't see the difference :(

doc/hostname.md
mchf: Right. I've updated the description to be (hopefully) more precise.

PR #1034, opened 2020-02-05
Reflect network configuration changes in the proposal (bsc#1162796)

teclator:
## Problem

When changing the network configuration during the installation it is not properly reflected in the proposal summary

## Solution

- Adapt the code using the new connection config and summary classes instead of the old `LanItems` summary
- The configuration is now shown in nested lists.

## Tests

- Tested manually
- Added unit tests for the new presenters

### Before 

Before

### After

Screenshot_testing_2020-02-07_11:20:42
Screenshot_testing_2020-02-07_11:22:41


mchf:
So in short the issue was, that we used old cache which is not updated anymore bcs of switching to connections & co.

lgtm

teclator:
> So in short the issue was, that we used old cache which is not updated anymore bcs of switching to connections & co.

Yep, mainly that is the issue, and the same with other summaries, so I'm cleaning it up.
> 
> lgtm

src/include/network/services/dns.rb
```diff
@@ -635,7 +635,9 @@ def DNSMainDialog(_standalone)
     #
     # @return [Boolean] true when an interface uses DHCP config
     def dhcp?
-      !LanItems.find_dhcp_ifaces.empty?
+      config = Yast::Lan.yast_config
+
+      !!config&.connections&.any? { |c| c.bootproto&.dhcp? }
```
imobachgs: np: I would prefer something more readable like:

```ruby
return false unless config&.connections
config.connections.any? { |c| c.bootproto&.dhcp? }
```

src/lib/y2network/presenters/dns_summary.rb
```diff
@@ -54,6 +52,12 @@ def text
 
     private
 
+      def dhcp_hostname?
+        return false unless config.connections.any? { |c| c.bootproto&.dhcp? }
```
imobachgs: Given that we are using `c.booproto&.dhcp?` more than once, what about adding a `dhcp?` method to the `ConnectionConfig` class?

src/lib/y2network/presenters/dns_summary.rb
```diff
@@ -54,6 +52,12 @@ def text
...
+
+        hostname.dhcp_hostname && hostname.dhcp_hostname != :none
```
imobachgs: I would say that the first part of the conditions is not needed (`nil != :none`). Actually, if `hostname.dhcp_hostname` is nil, this method will return `nil` instead of a boolean.

test/y2network/presenters/dns_summary_test.rb
```diff
@@ -22,13 +22,44 @@
...
+  subject(:presenter) { described_class.new(config) }
 
   let(:dns) do
     Y2Network::DNS.new(
       nameservers: nameservers, searchlist: searchlist
     )
   end
+
+  let(:config) do
+    Y2Network::Config.new(interfaces: interfaces, connections: connections,
```
imobachgs: This indentantion looks weird. What about: 

```ruby
Y2Network::Config.new(
  interfaces: interfaces, connections: connections, source: :testing # etc.
)

**Comment by teclator** [comment]
@imobachgs @dgdavid @mchf I merged #1037 already So, if this PR still looks good to you I will merge it ;)

File: src/lib/y2network/presenters/config_summary.rb [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+      # Network config RichText summary
+      #
+      # @return [String]
+      def text
+        "#{interfaces_summary.text}#{dns_summary.text}#{routing_summary.text}"
+      end
+
+    private
+
+      # Convenicente method to obtain the current config interfaces summary
```
dgdavid: typo: convenience

src/lib/y2network/presenters/config_summary.rb
```diff
@@ -0,0 +1,62 @@
...
+      def interfaces_summary
+        @interfaces_summary ||= Summary.for(config, "interfaces")
+      end
+
+      # Convenicente method to obtain the current config routing summary
```
dgdavid: typo: convenience

src/lib/y2network/presenters/config_summary.rb
```diff
@@ -0,0 +1,62 @@
...
+      def routing_summary
+        @routing_summary ||= Summary.for(config, "routing")
+      end
+
+      # Convenicente method to obtain the current config dns summary
```
dgdavid: typo: convenience

test/y2network/presenters/dns_summary_test.rb
```diff
@@ -42,30 +75,38 @@
...
+        eth0.bootproto = Y2Network::BootProtocol::DHCP
+        hostname.dhcp_hostname = "eth0"
+      end
+
+      it "is shown as the hostname" do
+        expect(presenter.text).to include("Hostname: Set by DHCP")
+      end
+    end
+
+    context "when the config does not contains a system hostname" do
```
dgdavid: NP: does not contain

test/y2network/presenters/dns_summary_test.rb
```diff
@@ -42,30 +75,38 @@
...
 
       it "does not show the hostname" do
-        text = presenter.text
-        expect(text).to_not include("Hostname")
+        expect(presenter.text).to_not include("Hostname")
       end
     end
 
-    context "when name servers are given" do
+    context "when the config does not contains name servers" do
```
dgdavid: NP: does not contain

test/y2network/presenters/dns_summary_test.rb
```diff
@@ -42,30 +75,38 @@
...
 
       it "does not show the name servers" do
-        text = presenter.text
-        expect(text).to_not include("Name Servers")
+        expect(presenter.text).to_not include("Name Servers")
       end
     end
 
-    context "when search domains are given" do
+    context "when the config does not contains search domains" do
```
dgdavid: NP: does not contain

test/y2network/presenters/proposal_summary_test.rb
```diff
@@ -0,0 +1,67 @@
...
+  before do
+    allow(presenter).to receive(:interfaces_summary).and_return(interfaces_summary)
+    allow(presenter).to receive(:dns_summary).and_return(dns_summary)
+    allow(presenter).to receive(:routing_summary).and_return(routing_summary)
+  end
+
+  describe "#text" do
+    it "returns a summary in text form" do
+      text = presenter.text
+      expect(text).to be_a(::String)
```
dgdavid: NP: As below, you can put the expectation in a single line. `expect(presenter.text).to be...`

src/lib/y2network/presenters/config_summary.rb
teclator: Thnx, you also can add them as suggestions making easier to merge as you are already proposing the fix.

teclator:
> Again, changes LGTM, although I left some comments for typos.

Updated :)

> 
> Feel free to merge it whey you considered it finished.

:stuck_out_tongue_winking_eye: 

> 
> Thanks

PR #1035, opened 2020-02-06
Do not assign interface section empty values (bsc#1162671)

teclator:
## Problem

When AutoYaST interfaces_section reader tries to initialize some connection config attributes it checks whether the value if defined in the section. The problem is that all the attributes are initialized to an empty string and the check becomes useless.

https://github.com/yast/yast-network/blob/master/src/lib/y2network/autoinst_profile/interface_section.rb#L264

- https://trello.com/c/YHW3GY5E/1627-sles15-sp2-p1-1162671-sles-15-sp2-beta2-internal-error-during-autoyast-install-leads-to-unfinished-installation

## Solution

For the time being, we just check whether the value to be assigned from the interfaces_section is empty or not only assigning it in case of not empty.

It probably would be better to not initialize all the interfaces_section attributes to an empty string but marked as `TODO` to be done later.

### DIFF
--- a/src/lib/y2network/autoinst_profile/interface_section.rb
+++ b/src/lib/y2network/autoinst_profile/interface_section.rb
@@ -259,6 +259,9 @@ module Y2Network
       def initialize(*_args)
         super
 
+        # TODO: Initializing all the attributes to an empty string makes
+        # hard to know whether the value was defined or not at all. We probably
+        # should ommit this initialization
         self.class.attributes.each do |attr|
           # init everything to empty string
           public_send(:"#{attr[:name]}=", "")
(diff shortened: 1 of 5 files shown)


src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -259,6 +259,9 @@ def self.new_from_network(connection_config)
       def initialize(*_args)
         super
 
+        # TODO: Initializing all the attributes to an empty string makes
+        # hard to know whether the value was defined or not at all. We probably
+        # should ommit this initialization
```
jreidinger: well, on other hand we should not have nil at all. So it is better then state with either nil or "" everywhere. At least that was idea.

PR #1036, opened 2020-02-07
Fixed validation of hostname

mchf:
https://trello.com/c/VrZGjzNP

## Problem ##

Empty hostname was not allowed and erasing /etc/hostname was impossible using yast.

## Solution ##

Fixed the way how yast validates and stores hostname

## Before ##

hostname-old

## After ##

hostname-new

### DIFF
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -427,7 +427,7 @@ module Yast
     # Validator for hostname, no_popup
     # @param key [String] the widget being validated
     # @param _event [Hash] the event being handled
-    # @return whether valid
+    # @return [Boolean] whether the current static hostname is valid or not
     def ValidateHostname(key, _event)
       value = UI.QueryWidget(Id(key), :Value).to_s
 
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -431,8 +431,9 @@ module Yast
     def ValidateHostname(key, _event)
       value = UI.QueryWidget(Id(key), :Value).to_s
 
-      # empty hostname is allowed - /etc/hostname gets cleared in such case
-      value.empty? || Hostname.Check(value)
+      # 1) empty hostname is allowed - /etc/hostname gets cleared in such case
+      # 2) FQDN is allowed
+      value.empty? || Hostname.Check(value.tr('.',''))
     end
 
     # Validator for the search list
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -57,8 +57,7 @@ module Yast
       # are lists and their widgets are suffixed.
       @hn_settings = {}
 
-      # TODO:
-      # It would be nice to display also transient hostname here - e.g. that
+      # TODO: It would be nice to display also transient hostname here - e.g. that
       # one received from dhcp
       @widget_descr_dns = {
         "HOSTNAME"        => {
@@ -436,7 +435,7 @@ module Yast
 
       # 1) empty hostname is allowed - /etc/hostname gets cleared in such case
       # 2) FQDN is allowed
-      value.empty? || Hostname.Check(value.tr('.',''))
+      value.empty? || Hostname.Check(value.tr(".", ""))
     end
 
     # Validator for the search list
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -64,7 +64,7 @@ module Yast
       @widget_descr_dns = {
         "HOSTNAME"        => {
           "widget"            => :textentry,
-          "label"             => Label.HostName,
+          "label"             => "Static H&ostname",
           "opt"               => [],
           "help"              => Ops.get_string(@help, "hostname_global", ""),
           "valid_chars"       => Hostname.ValidChars,
@@ -433,15 +433,10 @@ module Yast
     # @param _event [Hash] the event being handled
     # @return whether valid
     def ValidateHostname(key, _event)
-      dhn = dhcp? && use_dhcp_hostname?
-      # If the names are set by dhcp, the user may enter backup values
-      # here - N#28427. That is, host and domain name are optional then.
-      # For static config, they are mandatory.
-      value = Convert.to_string(UI.QueryWidget(Id(key), :Value))
+      value = UI.QueryWidget(Id(key), :Value).to_s
 
-      return Hostname.Check(value) if !dhn || value != ""
-
-      true
+      # empty hostname is allowed - /etc/hostname gets cleared in such case
+      value.empty? || Hostname.Check(value)
     end
 
     # Validator for the search list
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -19,11 +19,7 @@
 # you may find current contact information at www.novell.com
 #
 # **************************************************************************
-# File:  include/network/lan/dialogs.ycp
-# Package:  Network configuration
-# Summary:  Summary, overview and IO dialogs for network cards config
-# Authors:  Michal Svec <msvec@suse.cz>
-#
+
 module Yast
   module NetworkServicesDnsInclude
     # CWM wants id-value pairs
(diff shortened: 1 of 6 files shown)


src/lib/y2network/sysconfig/hostname_writer.rb
```diff
@@ -73,8 +73,15 @@ def update_sysconfig_dhcp(hostname, old_hostname)
       #
       # @param hostname [Y2Network::Hostname] Hostname configuration
       def update_hostname(hostname)
-        Yast::Execute.on_target!("/usr/bin/hostname", hostname.hostname.split(".")[0])
-        Yast::SCR.Write(Yast::Path.new(".target.string"), HOSTNAME_PATH, "#{hostname.hostname}\n")
+        hostname = hostname.static
+        # 1) when user asked for ereasing hostname from /etc/hostname, we keep runtime as it is
```
teclator:
```suggestion
        # 1) when user asked for erasing hostname from /etc/hostname, we keep runtime as it is
```

teclator: In case of erasing and no dhcp, shouldn't it be `localhost`?

teclator: Note: We probably should use hostnamectl in most of the cases, and systemd-firstboot during installation, but it is something we can discuss about.

mchf: well, empty hostname is treated as invalid by hostname utility and empty /etc/hostname is automatically changed to ```localhost``` during the boot.

However, ```hostname ''``` even ```hostname -b --file /etc/hostname``` do not work (expected in first case, but not so expected in the second one according to the man pages). So, runtime remains untouched (hostname cmd still returns previous hostname).

```hostnamectl set-hostname ''``` works in a sense that it deletes /etc/hostname, erases static hostname and sets transient hostname to localhost, so ```hostname``` cmd then correctly shows ```localhost``` as result. However, things like bash promt remains untouched (which is also true when modifying hostname to a nonempty string).

We can put localhost directly into /etc/hostname to mimic the behavior but in my POV it is something different than user wanted to achieve.

So, we can do exactly what hostnamectl does - delete ```/etc/hostname```. However, until we switch to ```hostnamectl ``` we most probably should warn user by a popup that runtime hostname stays untouched until reboot..

**Comment by mchf** [comment]
yep tests are on the way ... once we confirm the solution ;-)

File: src/include/network/services/dns.rb [hunk 2]
```diff
@@ -64,7 +64,7 @@ def initialize_network_services_dns(include_target)
       @widget_descr_dns = {
         "HOSTNAME"        => {
           "widget"            => :textentry,
-          "label"             => Label.HostName,
+          "label"             => "Static H&ostname",
```
kobliha: _("...")?

mchf: true but without question mark ;-)

test/dns_service_test.rb
```diff
@@ -0,0 +1,54 @@
...
+  subject { DummyDnsService.new }
+
+  describe "#ValidateHostname" do
+    it "allows empty hostname" do
+      allow(Yast::UI).to receive(:QueryWidget).and_return("")
+
+      expect(subject.ValidateHostname("", {})).to be true
+    end
+
+    it "allows valid characters in hostname" do
```
teclator: But a FQDN is allowed or not?, it is not described by the test, so would also add it. The test should be ok for those who knows which are the valid characters.

src/include/network/services/dns.rb
```diff
@@ -433,15 +429,10 @@ def HandleResolverData(key, _event)
     # @param _event [Hash] the event being handled
     # @return whether valid
```
teclator:
```suggestion
    # @return [Boolean] whether the current static hostname is valid or not
```

test/dns_service_test.rb
teclator: BTW, the valid_chars check is done by the widget, and looks a test more for it than for the validate method

src/lib/y2network/sysconfig/hostname_writer.rb
```diff
@@ -74,7 +74,7 @@ def update_sysconfig_dhcp(hostname, old_hostname)
       # @param hostname [Y2Network::Hostname] Hostname configuration
       def update_hostname(hostname)
```
teclator: What about adding a unit test for this writer?

test/dns_service_test.rb
mchf: true ... updated with test for FQDN

src/lib/y2network/sysconfig/hostname_writer.rb
teclator: It seems that the `SENDMAIL` ... constant is not used, so probably copied from the `dns_writer`

https://github.com/yast/yast-network/blob/e7d6c742f4617ba4dd0cbd5cdbeba0ece4053195/src/lib/y2network/sysconfig/hostname_writer.rb#L45

teclator:
np: a screenshot would be nice to have and maybe a TODO section with action items to be done in the future, like for example show the current dhcp hostname

test/y2network/hostname_writer_test.rb
```diff
@@ -0,0 +1,74 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require_relative "../test_helper"
+require "y2network/sysconfig/hostname_writer"
+require "y2network/hostname"
+
+describe Y2Network::Sysconfig::HostnameWriter do
+  subject { Y2Network::Sysconfig::HostnameWriter.new }
+
+  describe ".update_hostname" do
```
teclator: Update hostname is private, and you are really testing the write call that is ok.

```suggestion
  describe "#write" do
```

test/y2network/hostname_writer_test.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    let(:old_hostname_container) do
+      instance_double(
+        Y2Network::Hostname,
+        static:         "old#{hostname}",
+        dhcp_hostname:  false,
+        save_hostname?: true
+      )
+    end
+
```
teclator: Miss when no change context

test/y2network/hostname_writer_test.rb
teclator: please mock the call to the syconfig dhcp update.

test/y2network/hostname_writer_test.rb
```diff
@@ -0,0 +1,92 @@
...
+          .with(anything, anything, /#{hostname}/)
+
+        subject.write(hostname_container, old_hostname_container)
+      end
+    end
+
+    context "when no change in hostname" do
+      let(:hostname) { "hostname" }
+
+      it "do not try to update anything" do
```
teclator:
```suggestion
      it "does not try to update anything" do
```

test/y2network/hostname_writer_test.rb
```diff
@@ -0,0 +1,92 @@
...
+        expect(Yast::Execute)
+          .not_to receive(:on_target!)
+        expect(Yast::SCR)
+          .not_to receive(:Write)
+          .with(anything, anything, /#{hostname}/)
```
teclator: you can omit the .with

teclator:
```suggestion
```

test/y2network/hostname_writer_test.rb
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) [2019] SUSE LLC
```
teclator:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

test/dns_service_test.rb
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env rspec
+
+# Copyright (c) [2019] SUSE LLC
```
teclator:
```suggestion
# Copyright (c) [2020] SUSE LLC
```

PR #1037, opened 2020-02-07
Added new presenters for the config and proposal summaries (bsc#1162796)

teclator:
## Problem

When changing the network configuration during the installation it is not properly reflected in the proposal summary

## Solution

- Adapt the code using the new connection config and summary classes instead of the old `LanItems` summary
- The configuration is now shown in nested lists.

## Tests

- Tested manually
- Added unit tests for the new presenters

### Before 

Before

### After

Screenshot_testing_2020-02-07_11:20:42
Screenshot_testing_2020-02-07_11:22:41

## TODO

- Remove LanItems.summary completely


src/lib/network/clients/network_proposal.rb
```diff
@@ -74,11 +75,19 @@ def ask_user(args)
 
   private
 
+    def config
+      Yast::Lan.yast_config
```
dgdavid: Maybe you can use `Yast::Lan.yast_config` directly in the proposal summary.

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+        protocols << "DHCP" if config.connections.any?(&:dhcp?)
+        protocols << "STATIC" if config.connections.any?(&:static?)
+
+        output << protocols.first if protocols.uniq.size == 1
+
+        case config.connections.size
+        when 0
+          return Yast::Summary.NotConfigured
+        when 1
+          output << config.connections.first.interface
```
dgdavid: Why not asking here for the protocol?

```ruby
connection = config.connections.first
protocol = "DHCP" if connection.dhcp?
protocol ||= "STATIC" if connection.static?

[protocol, connection.interface].compact.join(" / ")
```

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+        else
+          output << _("Multiple Interfaces")
+        end
+
+        output.join(" / ")
+      end
+
+      # Generates a summary in RichText format for the configured interfaces
```
dgdavid: NP: I'd say in HTML format

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+
+      # Return a summary of the interfaces configurew with DHCP
+      #
+      # @return [String] interfaces configured with DHCP summary
+      def dhcp_summary
+        # TRANSLATORS: %s is the list of interfaces configured by DHCP
+        _("Configured with DHCP: %s") % dhcp_ifaces.sort.join(", ")
+      end
+
+      def dhcp_ifaces
```
dgdavid: NP: although the method name is descriptive enough... maybe it worth to add a short documentation

# Returns DHCP interfaces

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+        items << list_item_for(dhcp_summary) unless dhcp_ifaces.empty?
+        items << list_item_for(static_summary) unless static_ifaces.empty?
+        items << list_item_for(bridge_summary) unless bridge_connections.empty?
+        items << list_item_for(bonding_summary) unless bonding_connections.empty?
+        items << list_item_for(Yast::Summary.NotConfigured) if items.empty?
+
+        Yast::Summary.DevicesList(items)
+      end
+
+      # Return a summary of the interfaces configurew with DHCP
```
dgdavid: typo, `configured`

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+
+      def dhcp_ifaces
+        config.connections.select(&:dhcp?).map(&:interface)
+      end
+
+      # Return a summary of the interfaces configured statically
+      #
+      # @return [String] statically configured interfaces summary
+      def static_summary
+        # TRANSLATORS: %s is the list of interfaces configured by DHCP
```
dgdavid: ...configured statically

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+        _("Statically configured: %s") % static_ifaces.sort.join(", ")
+      end
+
+      # Return a summary of the configured bridge interfaces
+      #
+      # @return [String] bridge configured interfaces summary
+      def bridge_summary
+        _("Bridges: %s") % bridge_connections.map do |connection|
```
dgdavid: Don't forget the comment for translators

src/lib/y2network/presenters/interfaces_summary.rb
```diff
@@ -50,6 +53,115 @@ def text
...
+          "#{connection.name} (#{connection.ports.sort.join(", ")})"
+        end
+      end
+
+      # Return a summary of the configured bonding interfaces
+      #
+      # @return [String] bonding configured interfaces summary
+      def bonding_summary
+        _("Bonds: %s") % bonding_connections.map do |connection|
```
dgdavid: As above, I miss the translators comment

src/lib/y2network/presenters/summary.rb
```diff
@@ -0,0 +1,65 @@
...
+          summary.public_send(type)
+        end
+
+        # Config summary for a given section
+        #
+        # @param config [Y2Network::Config]
+        # @param section [String]
+        def for(config, section)
+          require "y2network/presenters/#{section}_summary"
+          summary_class =
```
dgdavid: What about 

```ruby
Module.const_get("#{section.capitalize}Summary")
```

test/y2network/presenters/interfaces_summary_test.rb
```diff
@@ -69,5 +73,62 @@
...
+      let(:connections) { Y2Network::ConnectionConfigsCollection.new([eth0]) }
+
+      it "returns the interface bootproto and interface name" do
+        expect(subject.one_line_text).to eql "DHCP / eth0"
+      end
+    end
+
+    context "when there are multiple interfaces" do
+      context "sharing the same bootproto" do
+        it "returns the bootproto and 'Multiple Interfaces'" do
```
dgdavid: Actually returns "Protocol / Multiple_interfaces", isn't it?

src/lib/y2network/presenters/interfaces_summary.rb
ancorgs: The `%s` syntax is common to all programming languages and the word "bridges" should be obvious in the network context. Why do you miss a comment?

src/lib/y2network/presenters/interfaces_summary.rb
ancorgs: Same than above.

src/lib/y2network/presenters/interfaces_summary.rb
ancorgs: Is not HTML. Is our own RichText format, the same that our RichText widget supports.

dgdavid: Then... let's use always the same term. But wait, it just a NP :)

src/lib/y2network/presenters/interfaces_summary.rb
dgdavid: Because we usually put that `# TRANLATOR` message. But, if it is not necessary, I'm fine skipping it.

src/lib/y2network/presenters/interfaces_summary.rb
teclator: Probably can be written in a better way but the procol is also used in case there are multiple connections using the same one. That is, it can be like:
- `DHCP / eth0`: Only one with DHCP
- `STATIC / eth1`: Only one configured statically
- `DHCP / Multiple Interfaces`: More than one but all sharing the dhcp protocol
- `STATIC / Multiple Interfaces`: More tan one but all configured statically
- `Multiple Interfaces`: More than one with different protocols

src/lib/network/clients/network_proposal.rb
teclator: Well, it currently uses the same config that is modified during the installation but could be the autoYaST one or maybe one from the proposal and other for the running system so I prefer to pass it as a parameter to the presenter as we do with others.

src/lib/y2network/presenters/summary.rb
teclator: 'dns' does not follow that pattern, then we should use `upcase` for it. 

We could do something like

```
           summary_class_for(section).new(config)
           ...
        end

      private

        def summary_class_for(section)
          Y2Network::Presenters.const_get(class_name_for(section))
        end

        def class_name_for(section)
          upcase = ["dns"]
          section_class_name = upcase.include?(section) ? section.upcase : section.capitalize
          "#{section_class_name}Summary"
        end

```

What do you prefer?

teclator: @dgdavid I will merge the PR by now as it will be merged in the `hostname_summary` branch.

dgdavid: It was just a suggestion. I didn't realize about the `dns` section. I agree keeping it as it originally was.

Thank you @teclator

PR #1039, opened 2020-02-19
[SLE-7188] Add an NtpServer class

imobachgs:
This PRs adds a `NtpServer` class so it can be used by other modules. The main problem it tries to solve is moving the logic to calculate the default servers to a single point.

**Related PRs:**
* https://github.com/yast/yast-installation/pull/838
* https://github.com/yast/yast-country/pull/243
* https://github.com/yast/yast-ntp-client/pull/155
* https://github.com/yast/yast-caasp/pull/44

**Trello card:** https://trello.com/c/lbOwsD5J

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Feb 19 11:52:25 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Add a class to represent NTP servers (jsc#SLE-7188).
+- 4.2.55
+
+-------------------------------------------------------------------
 Fri Feb  7 17:31:31 UTC 2020 - Michal Filka <mfilka@suse.com>
 
 - bsc#1162271
--- a/src/lib/y2network/ntp_server.rb
+++ b/src/lib/y2network/ntp_server.rb
@@ -1,0 +1,85 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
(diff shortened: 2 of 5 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Feb 19 11:52:25 UTC 2020 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Add a class to represent NTP servers (jsc#SLE-7188).
```
teclator: I'm not sure if like the idea about this class in network module. Initially it sounds more for ntp-client class. So, need to check the needs about obtaining the dhcp ntp servers from CaaSP and  from the installer before doing a proper review, so maybe @lslezak and @mchf could have a better opinion about it.

imobachgs: Yes, I thought pretty much the same. But, to be honest: 1) yast2-network already contains NTP specific code and 2) I did not want to put another dependency (on yast2-ntp-client) in the related packages. Anyway, I am fine with moving to yast2-ntp-client if required.

src/lib/y2network/ntp_server.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        host =
+          if base_products.any? { |p| p["name"] =~ /openSUSE/i }
+            "opensuse"
+          else
+            "suse"
+          end
+
+        (0..DEFAULT_SERVERS - 1).map { |i| new("#{i}.#{host}.#{DEFAULT_SUBDOMAIN}") }
+      end
```
teclator: maybe it could expose also the ntp_servers method...

```ruby
    def ntp_servers(use_fallback: true)
      # TODO: use Yast::NtpClient.ntp_conf if configured
      # to better handle going back
      servers = dhcp_ntp_servers
      servers = ntp_fallback if servers.empty? && use_fallback

      servers
    end
```

package/yast2-network.changes
teclator: I think that the benefits far outweighs the risks, so, it is fine with me.

src/lib/y2network/ntp_server.rb
teclator: The logic is currently duplicated and once this module already provide a class for NtpServer why not also unify it here?

imobachgs: Yes, it would be nice. But I prefer to keep it for the next iteration :)

teclator: Ok then, the rest look good to me

PR #1040, opened 2020-02-21
Add missing import

teclator:
## Problem

yast2-nisserver uses the `Yast::DNS` module for checking if some host is local or not. Unfortunately it crashes because of a missing `Yast::Lan` import.

- https://trello.com/c/9r2m9E4r/1645-1-littlebugostumbleweed-p2-1163305-yast2-nisserver-crashes-during-reading-of-the-configfiles
- https://bugzilla.suse.com/show_bug.cgi?id=1163305

## Solution

- Added the missing import
- Get rid of the NetworkInterfaces dependencies and adapted the code for using the new network-ng classes for checking if a host is local or not.

## Test

- Adapted `Yast::DNS.IsHostLocal` unit test

### DIFF
--- a/src/modules/DNS.rb
+++ b/src/modules/DNS.rb
@@ -239,7 +239,13 @@ module Yast
       # loopback interface or localhost hostname
       return true if ["127.0.0.1", "::1", "localhost", "localhost.localdomain"].include?(check_host)
 
-      ip_addresses = connections.map { |c| c.all_ips.map { |i| i&.address&.address.to_s } }.flatten
+      ip_addresses = []
+      connections.each do |conn|
+        conn.all_ips.each do |ip|
+          address = ip&.address&.address.to_s
+          ip_addresses << address if !address.empty? && !ip_addresses.include?(address)
+        end
+      end
 
       # IPv4 address
       if IP.Check4(check_host)
--- a/src/modules/DNS.rb
+++ b/src/modules/DNS.rb
@@ -80,6 +80,7 @@ module Yast
       Yast.import "UI"
       textdomain "network"
 
+      Yast.import "Lan"
       Yast.import "Arch"
       Yast.import "Hostname"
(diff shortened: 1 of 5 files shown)


src/modules/DNS.rb
```diff
@@ -241,29 +239,19 @@ def IsHostLocal(check_host)
       # loopback interface or localhost hostname
       return true if ["127.0.0.1", "::1", "localhost", "localhost.localdomain"].include?(check_host)
 
+      ip_addresses = connections.map { |c| c.all_ips.map { |i| i&.address&.address.to_s } }.flatten
```
jreidinger: this can potentially create a lot of empty strings. What about cleaning it?

src/modules/DNS.rb
```diff
@@ -239,7 +239,13 @@ def IsHostLocal(check_host)
       # loopback interface or localhost hostname
       return true if ["127.0.0.1", "::1", "localhost", "localhost.localdomain"].include?(check_host)
 
-      ip_addresses = connections.map { |c| c.all_ips.map { |i| i&.address&.address.to_s } }.flatten
+      ip_addresses = []
```
jreidinger: you can use here each_with_object with benefits like
```
ip_addresses = connections.each_with_object([]) do |conn, res|
  conn.all_ips.each_with_object(res) do |ip, ips|
     address = ip&.address&.address.to_s
     ips << address if !address.empty? && !ip_addresses.include?(address)
   end
end
```

teclator: Was not sure about using it, the benefits could depends on the eyes reading it :P do not know if prefer it over the original. But if is fine with me

teclator: Apart that the connections does not implement the each_with_object method, so we could add to it, or omit by now

teclator: Adapted anyway

PR #1041, opened 2020-02-25
Using SysctlConfig class instead of Sysctl class

schubi2:
**Problem**
Sysctl entries are stored in different files in different directories:
/run/sysctl.d,
/etc/sysctl.d
/usr/local/lib/sysctl.d
/usr/lib/sysctl.d
/lib/sysctl.d
/etc/sysctl.conf

General solution for it:
https://github.com/yast/yast-yast2/pull/1021

**Solution**
Using SysctlConfig class.

**Testing**
Tested manually


teclator:
> Using SysctlConfig class: Handle sysctl entries in different directories and files.



> Using SysctlConfig class: Handle sysctl entries in different directories and files.

Would be nice to have a link to the trello card as well as links to related pull requests, like for example the one https://github.com/yast/yast-yast2/pull/1021 which added SysctlConfig class and the reason of the dependency version update.

Using the storage-ng PR template is recommended:

https://github.com/yast/yast-storage-ng/blob/master/.github/pull_request_template.md

src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -103,20 +101,22 @@ def write_interface_changes(config, old_config)
...
-
-        update_ip_forwarding(sysctl.raw_forward_ipv4, :ipv4)
-        update_ip_forwarding(sysctl.raw_forward_ipv6, :ipv6)
+        sysctl_config = CFA::SysctlConfig.new
+        sysctl_config.load
+        sysctl_config.forward_ipv4 = routing.forward_ipv4
+        sysctl_config.forward_ipv6 = routing.forward_ipv6
+        sysctl_config.save unless sysctl_config.conflict?
+
+        update_ip_forwarding((sysctl_config.forward_ipv4 ? "1" : "0"),
```
teclator: Why were modified the calls to `raw_forward_` methods ? are them not valid anymore?

schubi2: Imo has wished to remove them in the new class. :-)

teclator: Ok, then next time we touch the class we should also get rid of of this documentation that is out of date or never implemented for this new class :stuck_out_tongue_winking_eye: 

https://github.com/yast/yast-yast2/pull/1021/files#diff-4216cbac8e98ed10343a00513e10e917R47

PR #1042, opened 2020-02-26
Fixed setting hostname via installer

mchf:
https://trello.com/c/prc25tfh/1662-sles15-sp2-p5-1164506-y2-network-configuring-hostname-during-install-is-not-preserved-regression-from-15-sp1

## Problem ##

When set hostname in the installer it is not preserved.

## Solution ##

Cover also the other way how user can explicitly set the hostname.

It was tested manually.

- [x] common installation without explicit hostname setup
- [x] common installation with explicit hostname setup via linuxrc
- [x] common installation with explicit hostname setup via YaST UI
- [x] AY installation with first stage only
- [x] AY installation with second stage

### DIFF
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -257,6 +257,18 @@ module Yast
       deep_copy(settings)
     end
 
+    # Stores user's input from hostname field
+    #
+    # @param value [String]
+    # @return [String] stored hostname
+    def store_hostname(value)
+      hostname = Yast::Lan.yast_config.hostname
+      hostname.static = value
+      hostname.installer = value if Stage.initial
+
+      value
+    end
+
     # @param [Hash] settings map of settings to be stored to DNS::
     def StoreSettings(settings)
       settings = deep_copy(settings)
@@ -270,7 +282,8 @@ module Yast
         " ,\n\t"
       )
 
-      DNS.hostname = Ops.get_string(settings, "HOSTNAME", "")
+      store_hostname(settings["HOSTNAME"] || "")
+
       valid_nameservers = NonEmpty(nameservers).each_with_object([]) do |ip_str, all|
         all << IPAddr.new(ip_str) if IP.Check(ip_str)
       end
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Mar  2 21:01:33 UTC 2020 - Michal Filka <mfilka@suse.com>
+
+- bsc#1164506
+  - fixed setting hostname in installer
+- 4.2.60
+
+-------------------------------------------------------------------
 Mon Mar  2 19:12:32 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
 - Do not modify interface name when enslaving it (bsc#1165463)
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -3,6 +3,8 @@ Mon Mar  2 21:01:33 UTC 2020 - Michal Filka <mfilka@suse.com>
 
 - bsc#1164506
   - fixed setting hostname in installer
+- bsc#1164587
+  - fixed setting hostname according to AY profile
 - 4.2.60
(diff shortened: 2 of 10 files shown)


teclator:
> https://trello.com/c/prc25tfh/1662-sles15-sp2-p5-1164506-y2-network-configuring-hostname-during-install-is-not-preserved-regression-from-15-sp1
> ## Problem
> 
> When set hostname in the installer it is not preserved.

And the solution is...? :wink: 
Has been it tested manually?

- https://github.com/yast/yast-storage-ng/blob/master/.github/pull_request_template.md

src/include/network/services/dns.rb
```diff
@@ -257,6 +257,18 @@ def InitSettings
...
     end
 
+    # Stores user's input from hostname field
+    #
+    # @param value [String]
+    # @return [String] stored hostname
+    def store_hostname(value)
+      hostname = Yast::Lan.yast_config.hostname
+      hostname.static = value
+      hostname.installer = value if Stage.initial
```
teclator: Is this workaround really needed? it is not better to remove that save_hostname? check ?

teclator: I guess this fix will not be enough for AY. See https://bugzilla.suse.com/show_bug.cgi?id=1164587

mchf: it is not workaround. It is way how to distinguish what user set explicitly and what is returned as static hostname in the installer otherwise (bcs linuxrc set a default). Hostname can be explicitly set either by linuxrc or here via yast's ui.

teclator: Why not just read the `/etc/hostname`? and write it always in case it is different?

mchf: so instead of
```
if user set hostname in installer
  write it
end
```
you propose
```
if in installer
  try to read /etc/hostname
  if different from what user entered
    write user's input
  end
end
```
And what is the benefit? Btw we currently don't read /etc/hostname directly so it would require to implement this part from scratch otherwise we're not able to distinguish linuxrc's default from what was set by user.

if it is about the ```save_hostname?``` check it can be simply moved from the ```Hostname``` class to the writer in some way. It is not used anywhere else AFAIK.

teclator: I was testing it, and well, it seems to work fine during an installation although as I said, probably will not work for AY and mainly because of that method save_method in the writer that doesn't smell to good to me.

**Note**: linuxrc does not write the /etc/hostname with the default option, only if set explicitly with `Hostname=`. It also permits FQDN (even documented so), if it is the case, we get rid of the hostname because of the characters validation (we are not permitting dots). You can reproduce it with `Hostname=example.com`.

teclator: > so instead of
> 
> ```

For doing this you are asking the hostname itself, if it should be saved, which is in Hostname and not in the writer explicitly and it is more like.
> if user set hostname in installer
>   write it
> end
> ```
> 
> you propose
> 
> ```
> if in installer
>   try to read /etc/hostname
>   if different from what user entered
>     write user's input
>   end
> end
Not really

I propose:

```ruby
read_.hostname
write_hostname if hostname_changed?
```

Which is already proposed here: https://github.com/yast/yast-network/compare/hostname_fixes

The writer just decided how to modify if directly or through hostnamectl depending on the stage.

> ```
> 
> And what is the benefit? Btw we currently don't read /etc/hostname directly so it would require to implement this part from scratch otherwise we're not able to distinguish linuxrc's default from what was set by user.
> 
> if it is about the `save_hostname?` check it can be simply moved from the `Hostname` class to the writer in some way. It is not used anywhere else AFAIK.

BTW, if the issue is already fixed we can leave the cleanup / modifications for later.

mchf: hm, dots are another thing which needs some polishing ... together with aliases. However, validating hostnames with dots is partly done here 65742794babc2df87edc81c09fc177707d16749c already, but i know it do not solve the problem completely. 

I think that AY should work bcs we write network configuration in the first stage. However, I have to test it bcs of other bug.

PR #1044, opened 2020-02-27
Do not crash in inst_lan client if  the network config has not been read (bsc#1101514)

teclator:
## Problem

The `inst_lan` client crashes in **firstboot** because the network has not been read yet.

The network is usually read by the **dhcp** clients, that is, the `inst_setup_dhcp` during an installation or by the `firstboot_dhcp` in firstboot. But, if no previous client read the network config with `Yast::Lan.Read(:cache)`, then, the internals are not initialized for the checks done.

- https://trello.com/c/p0DBYrMD/1665-sles15-p1-1101514-sles15-gmc2-power9-sap-applications-platform-error-handling-post-installation

## Solution

- Modify the check ensuring it not crash in case of no config read.
- Ensure the config is read before checking if some connection config exist in yast_config.
- Do not check if the network is configured when forced the manual configuration

## Tests

- Added unit test to for the `inst_lan` client
- Tested manually in firstboot without autoconfiguration clients.


src/lib/network/clients/inst_lan.rb
```diff
@@ -0,0 +1,92 @@
...
+  private
+
+    # Convenience method that checks whether there is some connection
+    # configuration present in the system
+    #
+    # @return [Boolean] true when there is some connection present in yast
+    #   config; false otherwise
+    def connections_configured?
+      # Ensure we read the current network config
+      Yast::Lan.Read(:cache)
```
jreidinger: NP: inconsistent usage of Yast namespace.

src/lib/network/clients/inst_lan.rb
```diff
@@ -59,20 +59,20 @@ def initialize
     end
 
     def main
+      ret = nil
```
jreidinger: is it really needed?

src/lib/network/clients/inst_lan.rb
```diff
@@ -59,20 +59,20 @@ def initialize
...
-      ret = if network_configured? && !manual_conf_request
-        GetInstArgs.going_back ? :back : :next
-      else
-        LanSequence()
+      if !manual_conf_request
+        log.info("Configured network found: #{network_configured?}")
+        ret = GetInstArgs.going_back ? :back : :next if network_configured?
       end
 
+      ret ||= LanSequence()
```
jreidinger: hmm, not sure if it is more readable.

src/lib/network/clients/inst_lan.rb
```diff
@@ -59,20 +59,20 @@ def initialize
...
 
-      log.info("Configured network found: #{network_configured?}")
-
-      ret = if network_configured? && !manual_conf_request
-        GetInstArgs.going_back ? :back : :next
-      else
-        LanSequence()
+      if !manual_conf_request
+        log.info("Configured network found: #{network_configured?}")
+        ret = GetInstArgs.going_back ? :back : :next if network_configured?
```
jreidinger: should work also with simple
```suggestion
        ret = :auto if network_configured?
```

teclator: So, what about:

```ruby
      ret =
        if manual_conf_request
          LanSequence()
        else
          log.info("Configured network found: #{network_configured?}")
          network_configured? :auto : LanSequence()
        end
```

jreidinger: yes, it is more clear code for me.

PR #1045, opened 2020-03-02
Do not modify interface name when enslaving it (bsc#1165463)

teclator:
## Problem

When enslaving an interface with a physical port id it is renamed adding the description as the name.

- https://bugzilla.suse.com/show_bug.cgi?id=1165463

## Solution

Use a dup of the interface name.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Mar  2 19:12:32 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not modify interface name when enslaving it (bsc#1165463)
+- 4.2.59
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


teclator:
> LGTM
> 
> I'm just thinking if we could somehow avoid a similar bug in the future. Would something like `interface.name.freeze` work?

Well, the interface name can be modified when renaming. Thu, not sure about freezing it.

lslezak:
I see, then OK.

teclator:
> LGTM
> 
> I'm just thinking if we could somehow avoid a similar bug in the future. Would something like `interface.name.freeze` work?

Thinking on it, I think that would be a good idea and at least we should do it in the constructor:

https://github.com/yast/yast-network/blob/master/src/lib/y2network/interface.rb#L77

And when renaming:

https://github.com/yast/yast-network/blob/master/src/lib/y2network/interface.rb#L112

But, will merge as it is by now if you are ok.

PR #1048, opened 2020-03-18
Reload configured interfaces when wriing the config and installing through ssh / vnc

teclator:
## Problem

If the network is configured during installation over ssh or vnc, the interfaces are not reloaded and thus, the changes are not reflected in the inst-sys.

- https://trello.com/c/yjXeyDJA/1711-3-sle15-sp2-p2-1166287-additional-network-devices-and-depending-vlans-are-not-set-online-during-installation
- https://bugzilla.suse.com/show_bug.cgi?id=1166287

## Solution

The reload config still used the Lanitems config which is not modified by YaST anymore. Thus, we have modified the code obtaining the list of configured connections names from `Yast::Lan.yast_config`

## Screenshots

ConfigureVLAN
ConfiguredVLANCheck

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -877,14 +877,14 @@ module Yast
         NetworkService.ReloadOrRestart if Stage.normal || !Linuxrc.usessh
 
       when :remote_installer
-        connections = yast_config&.connections&.map(&:name) || []
+        connection_names = yast_config&.connections&.map(&:name) || []
 
         # last instance handling "special" cases like ssh installation
         # FIXME: most probably not everything will be set properly
         log.info("Running in ssh/vnc installer -> just setting links up")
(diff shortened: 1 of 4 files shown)


src/modules/Lan.rb
```diff
@@ -876,14 +877,14 @@ def activate_network_service
         NetworkService.ReloadOrRestart if Stage.normal || !Linuxrc.usessh
 
       when :remote_installer
-        ifaces = LanItems.getNetworkInterfaces
+        connections = yast_config&.connections&.map(&:name) || []
```
mchf: NP: connections name sounds to be a bit misleading here, I'd keep ifaces or names or something like that

teclator: @mchf updated

PR #1049, opened 2020-03-19
Do not propose a hostname during AY

teclator:
## Problem

AutoYaST crashes when installing a system with dns_section but without a defined hostname

HostnameError

- https://trello.com/c/ZjcRfwqR/1720-important-sle15-sp2-p1-1166953-autoyast-complaining-ie-about-defaulthostname
- https://bugzilla.suse.com/show_bug.cgi?id=1166953

## Solution

- Do not set the static hostname, as it should be read from the system. The value from the profile is set as the installer which should modify the current static.
- Set correctly whether dhcp_hostname should be enable or not


mchf:
lgtm

PR #1052, opened 2020-04-02
S390 devices activation and better chanids handling

imobachgs:
This PR fixes two issues:

* Activation of devices that were already active in S390 systems. It is the network counterpart of https://github.com/yast/yast-s390/pull/81.
* Allow importing `chanids` using the spaces or colons as separators. It is preferred to use colons, to be consistent with the `lszdev` tool, but we need to be backward compatible.

A fix for the documentation will follow.

### DIFF
--- a/src/lib/y2network/autoinst/s390_devices_reader.rb
+++ b/src/lib/y2network/autoinst/s390_devices_reader.rb
@@ -86,6 +86,7 @@ module Y2Network
 
       # Returns the list of channel IDs from a string
       #
+      # @param ids [String] String representing the channel IDs
       # @return [Array<String>]
       def chanids_from(ids)
         ids.to_s.split(CHANIDS_SEPARATOR)
--- a/src/lib/y2network/autoinst/s390_devices_reader.rb
+++ b/src/lib/y2network/autoinst/s390_devices_reader.rb
@@ -66,18 +66,29 @@ module Y2Network
       end
 
       def load_qeth(config, device_section)
-        chanids = device_section.chanids
-        config.read_channel, config.write_channel, config.data_channel = chanids.split(" ")
+        config.read_channel, config.write_channel, config.data_channel =
+          chanids_from(device_section.chanids)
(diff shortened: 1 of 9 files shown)


test/y2network/autoinst_profile/s390_device_section_test.rb
```diff
@@ -62,7 +62,23 @@
...
+    context "using the old syntax for chanids" do
+      let(:hash) do
+        {
+          "type"     => "ctc",
+          "chanids"  => "0.0.0800 0.0.0801",
+          "protocol" => "1"
+        }
+      end
+
+      it "loads properly boot protocol" do
```
teclator, line 78: np:

```suggestion
      it "loads properly the chanids" do
```

src/lib/y2network/autoinst_profile/s390_device_section.rb
```diff
@@ -91,6 +102,25 @@ def init_from_config(config)
...
+        self.chanids = normalized_chanids(hash["chanids"]) if hash["chanids"]
+      end
+
+    private
+
+      # Normalizes the list of channel IDs
+      #
+      # It replaces spaces with colons.
+      #
+      # @return [String]
```
teclator, line 120: np: as it is documented I would also document the `ids` param

src/lib/y2network/autoinst/s390_devices_reader.rb
```diff
@@ -66,18 +66,29 @@ def create_config(device_section)
...
+        config.read_channel, config.write_channel = chanids_from(device_section.chanids)
+      end
+
+      # Separator used for the list of channel IDs
+      CHANIDS_SEPARATOR = ":".freeze
+      private_constant :CHANIDS_SEPARATOR
+
+      # Returns the list of channel IDs from a string
+      #
+      # @return [Array<String>]
```
teclator, line 89: np: as it is documented I would also document the ids parameter

PR #1054, opened 2020-04-15
Removed ISDN leftover hardware methods

teclator:
## Problem

When YaST reads the network config, it tries to find out the kernel package that contains a given module for interfaces without activated drivers. In case of the bugs related, it tries to find out the kernel module for a qeth interface.

The problem is that there no a RPM DB during installation, and thus, there is no 'kernel' package found.

- https://trello.com/c/mUxWuW1N/1317-3-sle-15-sp2-bug-1151187-sporadic-these-packages-to-be-installed-none

## Solution

- Remove the Package.InstallKernel call as it seems to be useless nowadays and which was added mainly for ISDN / DSL configuration which support has been dropped from YaST. (see https://bugzilla.novell.com/show_bug.cgi?id=suse44394)
- Removed also the hardware methods used mainly for configuring ISDN.

## Tests

- Added unit test for NetHwDetection#LoadNetModules
- Tested manually using [yupdate](https://github.com/yast/yast-installation/pull/835)
  - [x] With latest SLE-15-SP2 image in common x86_64 installation
  - [ ] With latest SLE-15-SP2 image in common x390 installation


src/include/network/lan/wizards.rb
```diff
@@ -44,7 +44,6 @@ def initialize_network_lan_wizards(include_target)
       Yast.include include_target, "network/lan/address.rb"
       Yast.include include_target, "network/lan/complex.rb"
       Yast.include include_target, "network/lan/dhcp.rb"
-      Yast.include include_target, "network/lan/hardware.rb"
```
mvidner, line 47: This is wrong, it's a different file.

teclator: Currently it only contains an s390 activation dialog, maybe we should move it to https://github.com/yast/yast-network/blob/master/src/include/network/lan/s390.rb

teclator: BTW, thnx for the catch, fixed.

PR #1056, opened 2020-04-24
Dropped unused methods

mchf:


mchf:
i'm checking if it is somewhere still somehow used ... sometimes i find some small surprise like in this PR.

PR #1057, opened 2020-05-05
Cleanup issue and profile handling

schubi2:
Problem
=========

Each module (storage, firewall,network) has his own AutoYaST Issue and profile handling which is incompatible and doubles the code.
bsc#1171335

Solution
==========

Bring base classes to yast2 package and cleanup the code of the other packages which will use
these common base classes:
yast/yast-yast2#1050
yast/yast-yast2#1055


src/clients/host_auto.rb
```diff
@@ -153,10 +155,16 @@ def check_profile_for_errors(imported_hosts)
       imported_hosts.each do |ip, hosts|
         next unless hosts.any? { |host| host.strip.empty? }
 
-        AutoInstall.issues_list.add(:invalid_value, "host", "names",
+        AutoInstall.issues_list.add(
+          ::Installation::AutoinstIssues::InvalidValue,
+          Y2Network::AutoinstProfile::HostSection.new_from_hashes(
```
imobachgs, line 160: This is not the right place to check for errors (it should be part of Hosts.Import). Please, add a `# TODO` comment here and I will take care on Monday.

PR #1061, opened 2020-05-24
Autoyast based installation does not seem to respect defined udev rules.

teclator:
## Problem

Udev rules are not refreshed after writting

The InterfacesWriter did not refreshed the written udev rules in case of an autoinstallation as we can see here:

https://github.com/yast/yast-network/pull/1059/files#diff-48ec4731f89821c19dc203da64dc00a5L108

In case of the second stage, or, the first stage with the imported network configuration written before the proposal, then, a rename of the interfaces is needed. Thus, we should distinguish between being at the end of the installation (already chrooted) or not.

- https://bugzilla.suse.com/show_bug.cgi?id=1169663

## Solution

The problem with udev rules handling during installation was already fixed in #1059, as well as other issues. Thus, I have backported some of the commits, at least the relevants to SP2, and adapted the code where it was needed. We have to take into account that in SP3, the idea is to write the network configuration during the first stage by default.

## Tests

- Tested manually
- Added some unit tests


mchf:
lgtm

PR #1062, opened 2020-05-28
Drop /usr link in tests directory

imobachgs:
The test does not seem to be useful and it is really annoying when trying to "grep" through the sources.


imobachgs:
Thanks. I also tried to build the package via "rake osc:build" and it worked just fine. Now it is time to merge this change in the rest of branches :smile:

PR #1068, opened 2020-05-28
Drop /usr link in tests directory

imobachgs:
Merge #1062 into `SLE-15-SP2`.

We might need to check whether this bug still applies to SP2:  https://bugzilla.suse.com/show_bug.cgi?id=1136934


teclator:
Good catch, in fact it could be there according to:

https://github.com/yast/yast-network/blob/master/src/lib/y2network/widgets/vlan_interface.rb#L41
https://github.com/yast/yast-network/blob/master/src/lib/y2network/interface_config_builders/vlan.rb#L57

imobachgs:
> LGTM, and verified the bug does not happen in SP2 as we already skip s390 devices. In case it is a s390 activated device, then, the associated linux interface will be shown in the list.

Thanks for checking!

PR #1072, opened 2020-06-10
Avoid Bond Slaves error in s390

joseivanlopez:
## Problem

https://bugzilla.suse.com/show_bug.cgi?id=1172444

In s390, when we are configuring a network card and we go to the Bond Slaves tab, an exception is raised because method `#path` is not defined.

## Solution

Prefix methods with `Yast` namespace.

## Testing

* Added unit tests.

### DIFF
--- a/src/include/network/lan/s390.rb
+++ b/src/include/network/lan/s390.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2019] SUSE LLC
+# Copyright (c) [2019-2020] SUSE LLC
 #
 # All Rights Reserved.
 #
@@ -24,6 +24,7 @@ module Yast
 
     def initialize_network_lan_s390(_include_target)
       Yast.import "FileUtils"
+      Yast.import "Arch"
     end
 
     # Checks if driver was successfully loaded for particular device.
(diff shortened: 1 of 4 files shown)


src/include/network/lan/s390.rb
```diff
@@ -17,13 +17,18 @@
...
+require "yast"
+
 module Yast
   # Functions for accessing and handling s390 specific needs.
   module NetworkLanS390Include
     SYS_DIR = "/sys/class/net".freeze
 
-    def initialize_network_lan_s390(_include_target)
+    def initialize_network_lan_s390(include_target)
+      include_target.class.include Yast
```
jreidinger, line 28: To be honest I do not like this at all, as it can mean more failures in future as YaST collide with global namespaces in some parts like String or FileUtils. I prefer to prefix properly YaST code here with `YaST::`

src/include/network/lan/s390.rb
```diff
@@ -43,7 +44,7 @@ def s390_ReadQethAttribute(devname, attrib)
 
       result = Convert.to_string(
         SCR.Read(
-          path(".target.string"),
+          Yast.path(".target.string"),
```
jreidinger, line 47: Is it enough? What about other calls like SCR or FileUtils?

joseivanlopez:
MU: https://build.suse.de/request/show/220130

PR #1073, opened 2020-06-11
Bring back the installation of 'wireless-tools' when scanning wireless networks

teclator:
## Problem

In order to scan available wireless networks we need the package that provides the `iwlist` command. This dependency was removed in `network-ng` as wicked planned to support the scan directly, but as it is still not the case the user needs to manually install the package in order to perform the wireless networks scan.

- https://bugzilla.suse.com/show_bug.cgi?id=1168479

## Solution

Bring back the installation of the package when missing.

## Tests

- Tested manually
- Unit test added

### DIFF
--- a/src/lib/y2network/widgets/wireless_essid.rb
+++ b/src/lib/y2network/widgets/wireless_essid.rb
@@ -123,7 +123,7 @@ module Y2Network
       IWLIST_PKG = "wireless-tools".freeze
 
       def scan_supported?
-        return true if package_installed?
+        return true if install_needed_packages
 
         Yast::Popup.Error(
           _("The package %s was not installed. It is needed in order to " \
@@ -136,7 +136,7 @@ module Y2Network
       # wlan network (bsc#1112952, bsc#1168479)
       #
       # TODO: drop it when supported by wicked directly
-      def package_installed?
+      def install_needed_packages
         Yast::Stage.initial ||
           Yast::Package.Installed(IWLIST_PKG) ||
           Yast::Package.Install(IWLIST_PKG)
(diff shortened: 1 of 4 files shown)


src/lib/y2network/widgets/wireless_essid.rb
```diff
@@ -110,21 +112,47 @@ def label
...
+      end
+
+      # Require wireless-tools installation in order to be able to scan the
+      # wlan network (bsc#1112952, bsc#1168479)
+      #
+      # TODO: drop it when supported by wicked directly
+      def package_installed?
+        Yast::Stage.initial ||
+          Yast::Package.Installed(IWLIST_PKG) ||
+          Yast::Package.Install(IWLIST_PKG)
```
mchf, line 142: the last part doesn't seem to be answer for ```package_installed?```

mchf: i mean it sounds like check whether the package is installed but in fact it has sideeffect installing the package.

teclator: Yep, I was not sure about the naming, and moving it from the method was an option but do not liked at all. I named firstly `ensure_package_installed`

teclator: we could use `package_installed? || Yast::Package.Install(IWLIST_PKG)` if you prefer

teclator: @mchf I will rename as `install_needed_packages`,  it is fine with you?

mchf: yep

src/lib/y2network/widgets/wireless_essid.rb
```diff
@@ -110,21 +112,47 @@ def label
...
         end
 
-        return unless @update_widget
-
-        @update_widget.update_essid_list(networks)
-        nil
+        networks
       end
 
-    private
```
jreidinger, line 126: why not private?

jreidinger: ah, see above. nevermind

PR #1074, opened 2020-06-11
Fix tests in 390

joseivanlopez:
## Problem

Tests introduced in the a previous PR https://github.com/yast/yast-network/pull/1072 are failing when running in a s390 architecture.

## Solution

Make tests to work for all architectures.

NOTE: Previous PR is not submitted yet. For that reason the version number is not bumped here.

### DIFF
--- a/test/y2network/interface_config_builders/bonding_test.rb
+++ b/test/y2network/interface_config_builders/bonding_test.rb
@@ -55,6 +55,8 @@ describe Y2Network::InterfaceConfigBuilders::Bonding do
       allow(connection_config).to receive(:name).and_return(connection_name)
 
       allow(connection_config).to receive(:find_master).and_return(connection_master)
+
+      allow(Yast::Arch).to receive(:s390).and_return(s390)
     end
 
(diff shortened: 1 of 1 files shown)


test/y2network/interface_config_builders/bonding_test.rb
```diff
@@ -55,6 +55,10 @@
       allow(connection_config).to receive(:name).and_return(connection_name)
 
       allow(connection_config).to receive(:find_master).and_return(connection_master)
+
+      allow(Yast::FileUtils).to receive(:IsDirectory).and_return(true)
```
teclator, line 59: wasn't better to call the s390 method calls?

teclator: maybe we could move that check of layer2 support to a convenience method and mock only it

joseivanlopez: Maybe, but I am not sure if it is worth doing it now. That changes would no be related with the fix about Yast namespace.

PR #1077, opened 2020-06-11
Reduce export

teclator:
## Problem

When the selected network service is NetworkManager, many of the settings are irrelevant when exporting the configuration.

- https://trello.com/c/1RXM7d8u/1851-8-reduce-profile-i-do-not-export-not-relevant-information#comment-5ee1e470b3a5b661db5b2bcd

## Solution

- Do not export any settings by now when NetworkManager is the selected network service.
- Keep the selection of the **network backend** in the `config` object decoupling it from the **NetworkService** module in order to permit having different backends in different configurations (related to https://trello.com/c/87QL0IuF/3606-sles15-sp2-p3-1151841-network-proposal-is-not-a-proposal-it-just-configures-the-network-on-the-fly)

### DIFF
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -89,7 +89,7 @@ module Yast
 
     def config
       # TODO: get it from some config holder
-      Yast::Lan.yast_config
+      @config ||= Yast::Lan.yast_config
     end
 
     def routing_table_widget
--- a/src/include/network/services/routing.rb
+++ b/src/include/network/services/routing.rb
@@ -89,7 +89,7 @@ module Yast
 
     def config
       # TODO: get it from some config holder
-      @config ||= Yast::Lan.yast_config
+      Yast::Lan.yast_config
     end
 
     def routing_table_widget
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -82,7 +82,7 @@ module Yast
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
-        NetworkService.use_networ_kmanager
+        NetworkService.use_network_manager
       else
         log.info("- using wicked")
         log.warn("- NetworkManager requested but not available") if use_network_manager
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -78,9 +78,7 @@ module Yast
 
       log.info("Setting network service according to AY profile")
 
-      use_network_manager = ay_networking_section["managed"]
-      use_network_manager = Lan.UseNetworkManager if use_network_manager.nil?
-
+      use_network_manager = Lan.yast_config&.backend?(:network_manager)
       nm_available = NetworkService.is_backend_available(:network_manager) if use_network_manager
 
       if use_network_manager && nm_available
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -82,13 +82,14 @@ module Yast
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
+        NetworkService.use_networ_kmanager
       else
         log.info("- using wicked")
         log.warn("- NetworkManager requested but not available") if use_network_manager
         Lan.yast_config&.backend = :wicked
+        NetworkService.use_wicked
       end
 
-      Lan.yast_config&.backend&.use!
       NetworkService.EnableDisableNow
     end
 
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -79,9 +79,8 @@ module Yast
       log.info("Setting network service according to AY profile")
 
       use_network_manager = Lan.yast_config&.backend?(:network_manager)
-      nm_available = NetworkService.is_backend_available(:network_manager) if use_network_manager
 
-      if use_network_manager && nm_available
+      if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
 
         NetworkService.use_network_manager
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -82,15 +82,13 @@ module Yast
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
-
-        NetworkService.use_network_manager
       else
         log.info("- using wicked")
         log.warn("- NetworkManager requested but not available") if use_network_manager
-
-        NetworkService.use_wicked
+        Lan.yast_config&.backend = :wicked
       end
 
+      Lan.yast_config&.backend&.use!
       NetworkService.EnableDisableNow
     end
 
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -82,14 +82,13 @@ module Yast
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
-        NetworkService.use_network_manager
       else
         log.info("- using wicked")
         log.warn("- NetworkManager requested but not available") if use_network_manager
         Lan.yast_config&.backend = :wicked
-        NetworkService.use_wicked
       end
 
+      NetworkService.use(Lan.yast_config&.backend&.id)
       NetworkService.EnableDisableNow
     end
 
--- a/src/lib/network/wicked.rb
+++ b/src/lib/network/wicked.rb
@@ -64,6 +64,7 @@ module Yast
     # @param query [String] xpath query. See man wicked for info what is supported there.
     # @return [String] result of the query
     def query_wicked(iface, query)
+      Yast.import "NetworkService"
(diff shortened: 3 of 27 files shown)


src/modules/Lan.rb
```diff
@@ -666,6 +666,8 @@ def Import(settings)
     # we export a 2-level map of typed "devices"
     # @return dumped settings
     def Export
+      return {} if NetworkService.is_network_manager
```
jreidinger, line 669: I think this is wrong. I expect it will return `{ "managed" => true }` so it knows that it is NM.

jreidinger:
and travis is not happy

src/modules/Lan.rb
teclator: Yep, that is what should be expected when wicked is the default, but not otherwise.

src/lib/y2network/autoinst_profile/dns_section.rb
```diff
@@ -103,7 +103,7 @@ def init_from_hashes(hash)
       #
       # @return [Boolean] Result true on success or false otherwise
       def init_from_network(dns, hostname)
-        @dhcp_hostname = hostname.dhcp_hostname
+        @dhcp_hostname = hostname.dhcp_hostname == :any
```
teclator, line 106: This is a fix in the `dns_section` that probably should be backported to SLE-15-SP2

src/lib/network/wicked.rb
```diff
@@ -64,6 +64,7 @@ def parse_hostname(iface)
     # @param query [String] xpath query. See man wicked for info what is supported there.
     # @return [String] result of the query
     def query_wicked(iface, query)
+      Yast.import "NetworkService"
       raise ArgumentError, "A network device has to be specified" if iface.nil? || iface.empty?
       raise "Parsing not supported for network service in use" if !NetworkService.is_wicked
```
mchf, line 69: why not to use ```config&.backend?(:wicked)``` even here when already massively replacing old calls?

src/lib/y2network/widgets/routing_table.rb
```diff
@@ -23,7 +23,7 @@
 require "y2network/interface"
 
 Yast.import "Label"
-Yast.import "NetworkService"
+Yast.import "Lan"
```
mchf, line 26: it seems that import of the Lan module is getting everywhere ... so we probably could think of separating new backend "accessor" from the Lan module

src/modules/Lan.rb
```diff
@@ -736,7 +662,7 @@ def ProposeVirtualized
     def Packages
       pkgs = []
 
-      if NetworkService.is_network_manager
+      if yast_config&.backend?(:network_manager)
```
mchf, line 665: NP: seems that designing aliases like ```config.network_manager?``` or something like that would make code more pretty.

src/modules/Lan.rb
```diff
@@ -894,13 +839,54 @@ def activate_network_service
     # It clears already read configuration.
     def read_config
       system_config = Y2Network::Config.from(:sysconfig)
+      system_config.backend = NetworkService.cached_name
```
mchf, line 842: i think this should be hidden in the factory itself

src/modules/Lan.rb
```diff
@@ -667,18 +601,10 @@ def Import(settings)
...
-        "config"               => NetworkConfig.Export,
-        "interfaces"           => profile.interfaces&.interfaces&.map(&:to_hashes) || [],
-        "ipv6"                 => @ipv6,
-        "routing"              => profile.routing&.to_hashes || {},
-        "managed"              => NetworkService.is_network_manager,
-        "start_immediately"    => autoinst.start_immediately,
-        "keep_install_network" => autoinst.keep_install_network
-      }
+      ay = profile.to_hashes
+      ay["ipv6"] = @ipv6
```
imobachgs, line 605: So at some point, we should move the IPv6 settings into the Config class, right?

src/modules/Lan.rb
```diff
@@ -667,18 +601,10 @@ def Import(settings)
...
+      ay["config"] = NetworkConfig.Export unless ay["managed"]
```
imobachgs, line 606: I would say that these settings are not documented in the AutoYaST guide.

src/include/network/services/routing.rb
```diff
@@ -89,7 +89,7 @@ def RoutingMainDialog
 
     def config
       # TODO: get it from some config holder
-      @config ||= Yast::Lan.yast_config
```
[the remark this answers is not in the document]
teclator: At the end I can revert this change. I wanted to prevent the cache of the config in order to retrieve it from the config folder if added in the future. But will leave it as it is.

src/lib/y2network/backend.rb
```diff
@@ -0,0 +1,100 @@
...
+    # Return whether the backend is available or not
+    #
+    # @return [Boolean]
+    def available?
+      Yast::NetworkService.is_backend_available(id)
+    end
+
+    # Sets the backend to be used as the current network service
+    def use!
+      Yast::NetworkService.public_send("use_#{id}")
```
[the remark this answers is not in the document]
teclator: > I am not sure whether this responsibility/dependency should be here. What about having a `Yast::NetworkService#use` method that receives the backend and sets everything accordingly? In my opinion, this method is just adding some indirection with little benefit.

Let's omit it by now

> 
> About the dependency on `NetworkService`, it is true that the `#available?` method still requires it. We could argue, however, whether `NetworkService` should be responsible for that. But we are not going to rewrite the whole module, right? smile

I tried to avoid the cuplication of code by now, but we could move the check of the package installation to each backend already.

> 
> Finally, why not treating `#available?` and `#use!` as if they were analogous? Well, because at least for me, asking a backend whether it is available or not makes perfect sense. However, I think that the responsibility of setting the backend should be elsewhere (`NetworkService` by now).

src/modules/Lan.rb
teclator: Completely agree, I kept it as it is by now, but should be moved ASAP.

src/modules/Lan.rb
teclator: It is not sysconfig especific, so I preferred to left it out of the reader. Thus, would you prefer to move it?

src/lib/y2network/widgets/routing_table.rb
teclator: The Lan is currently the module which stores the configs. If we add a config manager or something like that, where we could retrieve from the different configs, then, we could get rid of it. But currently it is our central store.

src/lib/network/wicked.rb
teclator: I was in doubt about these query methods, because they depend on the service in use and not strictly in the one configured in the config. THat is why I have not modified here.

If you check, there are still some references to NetworkService, specially when them does not depend on the config at all.

src/modules/Lan.rb
teclator: We could defined them, and even do some meta-programming here, defining the methods dynamically. But I know we are not big fun of it... so I would leave it as it is by now.

@imobachgs, what do you think?

src/lib/network/network_autoyast.rb
```diff
@@ -82,13 +82,14 @@ def set_network_service
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
+        NetworkService.use_networ_kmanager
```
mchf, line 85: seems like a typo

src/modules/Lan.rb
mchf: well it simply raises a question in my head whether we can simply rely on having this separated readers or whether we need even some shared part when creating config storage

src/lib/network/network_autoyast.rb
teclator: Yep, it is, fixing

src/lib/network/network_autoyast.rb
```diff
@@ -82,13 +82,14 @@ def set_network_service
 
       if use_network_manager && Lan.yast_config.backend.available?
         log.info("- using NetworkManager")
+        NetworkService.use_network_manager
```
imobachgs, line 85: If you want, you can have `NetworkService.use` so just write: `NetworkService.use(Last.yast_config&.backend)`.

src/modules/Lan.rb
imobachgs: I am not against but, to be honest, I do not see the benefit.

src/lib/network/network_autoyast.rb
teclator: I left it as it is just to not touch yast2 having to add the method there and having to update the dependency. The benefit is minimal, so I think we could omit adding it by now maintaining the explicit use calls.

teclator: It was needed also when writing the network config. Thus, I have changed my mind adding  `NetworkService#use` as suggested.

https://github.com/yast/yast-yast2/pull/1065

teclator:
@imobachgs @mchf please check https://github.com/yast/yast-network/pull/1077/commits/b9aa7354b0ed0cf7eb8bab5a258448cf21438cd2

PR #1080, opened 2020-06-18
save inst-sys memory by ending the agent process after use (bsc#1172139)

mvidner:
- https://bugzilla.suse.com/show_bug.cgi?id=1172139

[memsample](https://github.com/yast/yast-installation/pull/864) has revealed that `ag_udev_persistent` not only makes NIC names persistent, it also persists as a useless process until the end of the installation, taking up 31MiB (!) of RSS. Fortunately the agent needs no internal state between its 2 calls at install time, so we can simply⁺ terminate it via UnmountAgent after each use. The cost is the slowdown of starting up the process again.

⁺ (In practice it works like slicing up a zombie in half: now you have TWO deadly zombies. Fixed in https://github.com/yast/yast-core/pull/147)

The effect of this PR can be observed in the memsample graphs below, where the green y2start+children line jumps down to hug the y2starts line:

<table>
<tr><td>before</td><td>after</td></tr>
<tr>
<td>

memsample-dash

</td><td>

memsample-noudev-core

</td></tr>
</table>

### DIFF
--- a/src/lib/y2network/udev_rule.rb
+++ b/src/lib/y2network/udev_rule.rb
@@ -160,6 +160,7 @@ module Y2Network
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.rules"), udev_rules.map(&:to_s))
         # Writes changes to the rules file
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
+        Yast::SCR.UnmountAgent(Yast::Path.new(".udev_persistent"))
       end
 
       # Writes drivers specific udev rules to the filesystem
@@ -177,6 +178,7 @@ module Y2Network
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.drivers"), rules_hash)
         # Writes changes to the rules file
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
+        Yast::SCR.UnmountAgent(Yast::Path.new(".udev_persistent"))
       end
 
       # Clears rules cache map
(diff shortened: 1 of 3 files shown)


src/lib/y2network/udev_rule.rb
```diff
@@ -160,6 +160,7 @@ def write_net_rules(udev_rules)
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.rules"), udev_rules.map(&:to_s))
         # Writes changes to the rules file
         Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
+        Yast::SCR.UnmountAgent(Yast::Path.new(".udev_persistent"))
```
lslezak, line 163: What about defining some helper method for wrapping the agent access? Something like

```ruby

def use_udev_agent(&block)
  block.call
ensure
  Yast::SCR.UnmountAgent(Yast::Path.new(".udev_persistent"))
end

...

use_udev_agent do
  Yast::SCR.Write(Yast::Path.new(".udev_persistent.rules"), udev_rules.map(&:to_s))
  # Writes changes to the rules file
  Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
end
```

or we could additionally add some write wrapper

```ruby

def write_udev(&block)
  block.call
ensure
  Yast::SCR.Write(Yast::Path.new(".udev_persistent.nil"), [])
  Yast::SCR.UnmountAgent(Yast::Path.new(".udev_persistent"))
end
```

and simplify the `write_drivers_rules` and `write_net_rules` methods.

mvidner: Hmm, I think that's a case where abstraction complicates things. True, the UnmountAgent calls are repeated, but at 3 occurrences  it is acceptable. Furthermore, the SCR(udev_persistent) calls are all limited to this single source file. The abstraction would pay off if offered as external API, but not here.

lslezak: OK, let's keep the changes as minimal as possible...

PR #1083, opened 2020-06-25
AY schema: added networking setup_before_proposal element

teclator:
## Problem

When the AutoYaST profiles contains the `networking -> setup_before_proposal`option, then, the validation complains.

## Solution

Add it to the schema's networking section.


teclator:
> LGTM. Please, update the [yast2-schema dependency](https://github.com/yast/yast-schema/blob/master/package/yast2-schema.spec#L63) too.

Yep, I was doing that :)

PR #1086, opened 2020-07-09
remove direct access to profile

jreidinger:
trello: https://trello.com/c/kCfh7Ylo/1941-5-drop-references-to-profilecurrent

### DIFF
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -144,7 +144,7 @@ module Yast
     def configure_hosts(write: false)
       log.info("NetworkAutoYast: Hosts configuration")
 
-      unless ay_current_profile.key? "host"
+      if ay_current_profile.empty?
         Host.Write(gui: false)
 
         return true
@@ -210,32 +210,13 @@ module Yast
       instsys_routing.merge(ay_routing)
     end
 
-    # Returns current AY profile in the internal representation
-    #
-    # @return [Hash] hash representing current profile or empty hash
-    def ay_current_profile
-      Yast.import "Profile"
-
-      ay_profile = Profile.current
-
-      return {} if ay_profile.nil? || ay_profile.empty?
-
-      ay_profile
-    end
-
     # Returns networking section of current AY profile
     def ay_networking_section
-      return {} if ay_current_profile["networking"].nil?
-
-      ay_current_profile["networking"]
+      @ay_networking_section || {}
     end
 
-    # Returns global section of current AY profile
-    def ay_general_section
-      return {} if ay_current_profile["general"].nil?
+    attr_writer :ay_networking_section
 
-      ay_current_profile["general"]
-    end
 
     # Returns host section of the current AY profile
     #
@@ -255,10 +236,10 @@ module Yast
     #
     # return <Hash> with hosts configuration
     def ay_host_section
-      return {} if ay_current_profile["host"].nil?
-
-      ay_current_profile["host"]
+      @ay_host_section || {}
     end
+
+    attr_writer :ay_host_section
 
     # Configures given yast submodule according AY configuration
     #
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -168,6 +168,12 @@ module Yast
       ret
     end
 
+    # setter for networking section. Should be done during import.
+    attr_writer :ay_networking_section
+
+    # setter for host section. Should be done during import.
+    attr_writer :ay_host_section
+
   private
 
     # Merges two maps with dns related values.
@@ -215,8 +221,6 @@ module Yast
       @ay_networking_section || {}
     end
 
-    attr_writer :ay_networking_section
-
 
     # Returns host section of the current AY profile
     #
@@ -238,8 +242,6 @@ module Yast
     def ay_host_section
       @ay_host_section || {}
     end
-
-    attr_writer :ay_host_section
 
     # Configures given yast submodule according AY configuration
     #
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -169,9 +169,11 @@ module Yast
     end
 
     # setter for networking section. Should be done during import.
+    # @return [Hash] networking section hash
     attr_writer :ay_networking_section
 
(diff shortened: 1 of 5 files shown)


src/lib/network/network_autoyast.rb
```diff
@@ -144,7 +144,7 @@ def activate_s390_devices(section = nil)
     def configure_hosts(write: false)
       log.info("NetworkAutoYast: Hosts configuration")
 
-      unless ay_current_profile.key? "host"
+      if ay_current_profile.empty?
```
teclator, line 147: This rewrite is strange, originally it was call only if "host" section was present, and now if the profile is empty? at least is what the call and variable name suggest..

src/lib/network/network_autoyast.rb
```diff
@@ -210,32 +210,13 @@ def merge_routing(instsys_routing, ay_routing)
       instsys_routing.merge(ay_routing)
     end
 
-    # Returns current AY profile in the internal representation
-    #
-    # @return [Hash] hash representing current profile or empty hash
-    def ay_current_profile
```
teclator, line 216: But then, there is no ay_current_profile

src/lib/network/network_autoyast.rb
teclator, line 147:
```suggestion
      if ay_host_section.empty?
```

src/lib/network/network_autoyast.rb
```diff
@@ -210,32 +210,13 @@ def merge_routing(instsys_routing, ay_routing)
...
-      return {} if ay_current_profile["networking"].nil?
-
-      ay_current_profile["networking"]
+      @ay_networking_section || {}
     end
 
-    # Returns global section of current AY profile
-    def ay_general_section
-      return {} if ay_current_profile["general"].nil?
+    attr_writer :ay_networking_section
```
teclator, line 218: I guess we plan to use the Y2Network::AutoinstProfile::NetworkingSection object:

https://github.com/yast/yast-network/blob/master/src/lib/y2network/autoinst_profile/networking_section.rb#L37

src/lib/y2network/clients/auto.rb
```diff
@@ -77,6 +77,8 @@ def change
       end
 
       def import(profile)
+        Yast::NetworkAutoYast.instance.ay_networking_section = profile
```
teclator, line 80:
```suggestion
        Yast::NetworkAutoYast.instance.ay_networking_section = 
          AutoinstProfile::NetworkingSection.new_from_hashes(profile)    
```

src/modules/Host.rb
```diff
@@ -159,6 +160,8 @@ def Import(settings)
         set_names(ip, names)
       end
 
+      NetworkAutoYast.instance.ay_host_section = settings
```
teclator, line 163: Similar approach commented below

src/lib/network/network_autoyast.rb
jreidinger: yes, it should just check host section. In fact change should be valid as it do some tricks with hosts config, but if section is empty it should be asme as if it is not there at all, not?

src/lib/network/network_autoyast.rb
jreidinger: yep, it is correct. Previously it was typo.

src/lib/network/network_autoyast.rb
jreidinger: To be honest I am not sure and I do not want to break it more then it is already.

src/lib/y2network/clients/auto.rb
jreidinger: well, does networking section act like Hash? Otherwise I will have to adapt also rest of code.

teclator:
I guess that at some point we could even get rid of the direct check of sections and rely on the imported config. Although there could be exceptions and also host section is independent of networking one, so in general I think that the current approach storing the imported sections in the NetworkAutoyast instance looks ok.

@imobachgs what do you think?

src/lib/network/network_autoyast.rb
teclator: yep, that is what I would expect

src/lib/network/network_autoyast.rb
teclator: Ok, we could do a intermediate step before adapting it completely.

src/lib/y2network/clients/auto.rb
teclator: Not really, and we probably use some defaults already, but then we have to document it very well to not be confused as we could expect an `Y2Network::AutoinstProfile::NetworkingSection` object instead of hash and vice versa.

imobachgs:
> I guess that at some point we could even get rid of the direct check of sections and rely on the imported config. Although there could be exceptions and also host section is independent of networking one, so in general I think that the current approach storing the imported sections in the NetworkAutoyast instance looks ok.
> 
> @imobachgs what do you think?

I agree with your POV, it should be the solution at long-term.

src/lib/network/network_autoyast.rb
```diff
@@ -168,6 +168,12 @@ def keep_net_config?
       ret
     end
 
+    # setter for networking section. Should be done during import.
```
teclator, line 171: Please document the type of the variable.

teclator:
In general looks good to me, although we have to take in mind removed sections that could be cached in the NetworkAutoyast instance and could be a problem if it is done before auto_setup

src/lib/network/network_autoyast.rb
```diff
@@ -169,9 +169,11 @@ def keep_net_config?
     end
 
     # setter for networking section. Should be done during import.
+    # @param [Hash] networking section hash
```
teclator, line 172:
```suggestion
    # @return [Hash] networking section hash
```

src/lib/network/network_autoyast.rb
```diff
@@ -169,9 +169,11 @@ def keep_net_config?
...
     attr_writer :ay_networking_section
 
     # setter for host section. Should be done during import.
+    # @param [Hash] host section hash
```
teclator, line 176:
```suggestion
    # @return [Hash] host section hash
```

PR #1088, opened 2020-07-16
Added a method to only write the yast Y2Network::Config

teclator:
## Problem

Firstboot used to call DNS.Write in order to apply hostname changes but it does not write hostname anymore.

- https://bugzilla.suse.com/show_bug.cgi?id=1173298
- Needed by https://github.com/yast/yast-firstboot/pull/100

## Solution

Added a method to write only **Yast::Lan.yast_config** changes and update the current system_config without touching other components.

## Test

- Added unit test and tested manually

### DIFF
--- a/src/modules/Lan.rb
+++ b/src/modules/Lan.rb
@@ -818,14 +818,14 @@ module Yast
 
     # Reads system configuration
     #
-    # It clears already read configuration.
+    # It resets already read configuration.
     def read_config
       system_config = Y2Network::Config.from(:sysconfig)
       Yast::Lan.add_config(:system, system_config)
       Yast::Lan.add_config(:yast, system_config.copy)
     end
 
-    # Writes current yast config and replace the system config with it
+    # Writes current yast config and replaces the system config with it
     def write_config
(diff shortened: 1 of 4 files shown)


src/modules/Lan.rb
```diff
@@ -818,6 +816,23 @@ def yast_config
...
+    # Reads system configuration
+    #
+    # It clears already read configuration.
+    def read_config
+      system_config = Y2Network::Config.from(:sysconfig)
+      Yast::Lan.add_config(:system, system_config)
+      Yast::Lan.add_config(:yast, system_config.copy)
+    end
+
+    # Writes current yast config and replace the system config with it
```
jreidinger, line 828: replaces

src/modules/Lan.rb
```diff
@@ -818,6 +816,23 @@ def yast_config
       find_config(:yast)
     end
 
+    # Reads system configuration
+    #
+    # It clears already read configuration.
```
jreidinger, line 821: maybe instead of clears I would say that it reset any in memory modifications?

PR #1094, opened 2020-08-11
AutoYaST: Added supplements: autoyast(host,networking,remote) into th…

schubi2:
…e spec file in order to install this packages if the section has been defined in the AY configuration file (bsc#1146494).


lslezak:
I'm not sure if this is correct, PR #1059 actually moved some parts to the first stage, please check...

schubi2:
We have come to the conclusion that this entry does not depend on the first/second stage.

PR #1101, opened 2020-09-10
Do not crash when proposing the bridge configuration for virtualization

teclator:
## Problem

The virtualization config proposal relies on the current network configuration in order to check the current configured interfaces. The proposal crashes because the network config has not been read by any client.

- https://bugzilla.suse.com/show_bug.cgi?id=1176313

## Solution

Force a read of the configuration in case it was not previously initialized.


mchf (approved):
lgtm

PR #1104, opened 2020-09-15
Infers the VLAN_ID from the interface name when not declared explicitly

teclator:
## Problem

**YaST** does not infers the **VLAN_ID** when a **VLAN** interface configuration does not declare it explicitly.

_**Wrong VLAN_ID for enp1s0.110**_

In case of modified the **VLAN** configuration without noting the wrong **ID**, for example adding an alias to `/etc/hosts` through the `hostname` field, then it will write down an incorrect **VLAN_ID** broken your network config.

- https://bugzilla.opensuse.org/show_bug.cgi?id=1176575

## Solution

Infers the **VLAN_ID** from the interface name when it uses the ifcfg-**name.<VLAN_ID>** or ifcfg-**vlan<VLAN_ID>** VLAN interface name schemas.


wfeldt (approved):
lgtm

PR #1108, opened 2020-09-21
Fixed false error in the y2log (bsc#1176653)

lslezak:
- See https://bugzilla.suse.com/show_bug.cgi?id=1176653, there is an error logged in the `y2log` even in a successful installation
- Override the default `OtherCommand` implementation from the `YaST::SCRAgent` (see https://github.com/yast/yast-core/blob/master/agents-perl/lib/YaST/SCRAgent.pm#L83-L89)
- Just to not log an error into the `y2log`
- 4.3.22

### Testing

Run `echo 'OtherCommand("SetRoot")' | ./ag_udev_persistent ; tail -n 1 ~/.y2log` to check that the agent works and logs the message as a warning.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Sep 21 13:15:39 UTC 2020 - Ladislav Slezák <lslezak@suse.cz>
+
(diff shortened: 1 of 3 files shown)


teclator (approved):
LGTM

PR #1113, opened 2020-10-05
Pick the first hostname when there is more than one

teclator:
## Problem 

There are two separate options for configuring the hostname through **DHCP** in sysconfig, one for **IPv4** and one for **IPv6**. 

Both option are global, that is, you can set also per interface which has precedence over the global one but currently **YaST** only set it for **IPv4**.

We did several changes in order to decide which interface is allowed to modify the hostname but the changes done were only for **IPv4**.

- https://bugzilla.suse.com/show_bug.cgi?id=1173915

## Solution

Do not crash when there are more than one hostname, and just pick the first one.
Also do not propose "localhost" as an alias for an interface configured statically.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon Oct  5 09:51:14 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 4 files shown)


mchf (approved):
lgtm

PR #1114, opened 2020-10-05
Modify the logic for reading and writing hostname the config

teclator:
## Problem

The current approach for reading the current hostname and also for writing it could produce unexpected results.

1. The static hostname does not honors the content of `/etc/hostname` but uses the transient hostname obtained by `hostname`
      ```
        $ hostnamectl | grep hostname
         Static hostname: nutello2
         Transient hostname: nutello
      ```
      Screenshot_testing_2020-10-05_12:49:22

2. The logic for knowing if the `hostname` has to be modified or not  is known by the `Y2Network::Hostname` object itself and based on the `Stage` instead of just the content of the static file.

- https://bugzilla.suse.com/show_bug.cgi?id=1173915

## Solution

Modified the logic for writing the config and also for obtaining the transient hostname as it could be modified by other third party tools and not just by wicked.


src/lib/network/network_autoconfiguration.rb
```diff
@@ -91,6 +91,10 @@ def configure_dhcp
       end
 
       activate_changes(dhcp_cards.map(&:name))
+
+      # Force a read of the configuration just for reading the transient
+      # hostname as it could be modified through dhcp since previous read.
+      Lan.read_config
```
mchf, line 97: NP or kind of ~~brain~~storming: in theory it makes more sense to re-read dhcp hostname before displaying it to the user then after activatiing changes. Bcs user could be impressed by YaST UI and kept it open for several hours and the transient host name might change in the meantime ....

teclator: Yep, in fact that is the problem about storing the transient **hostname** when it is set by **dhcp**, it can be modified dynamically. But as currently it is the static one which is shown then it is not a problem.

The only problem could be the proposal of the hostname, but, as we are now not proposing a random hostname at all, now should be ok.

test/y2network/hostname_test.rb
```diff
@@ -98,40 +98,4 @@
...
     end
   end
-
-  describe "#save_hostname?" do
-    context "When used in installer" do
-      before(:each) do
-        allow(Yast::Stage).to receive(:initial).and_return(true)
-      end
-
-      context "without explicitly set hostname" do
```
mchf, line 108: is this logic somehow kept working or not needed anymore?

src/lib/network/network_autoconfiguration.rb
mchf: yast should mostly deal with static hostname mostly. Only case when we need transient hostname is when we use it as a proposal for static hostname. So, we mostly don't need to bother with transient hostname.

test/y2network/hostname_test.rb
teclator: What do you mean exactly?:

Now the static hostname field honors the /etc/hostname content, so, modify it in case it is modified through the dialog.

1. **In case that the hostname is given by linuxrc**

    it is written by the InstallInfCovertor at https://github.com/yast/skelcd-control-leanos/blob/master/control/control.leanos.xml#L514

2. **If it is not given**

    No hostname is proposed at all. In case of dhcp setup, the hostname is modified by wicked, but we will honor the content of the static hostname. So, no hostname will be proposed and at the end of the installation.


Thus, basically what you described at: https://github.com/yast/yast-network/blob/master/doc/hostname.md#installation-proposals

mchf: Ah, i put it to bad line. I meant logic when hostname is set up via linuxrc option. https://github.com/yast/yast-network/pull/1114/files#diff-699f24097595d5e63d94b1c8bbc3ff96L78

src/lib/network/network_autoconfiguration.rb
teclator: Exactly, and if we still want to keep that proposal, then I would do it separately. The transient hostname could be added to the dialog, in order to show the current one, but would do it dynamically instead of storing it as part of the config.

mchf: yes, I agree. So the only point is whether we need to enforce re-reading transient hostname after write ... however, I assume it is cheap so keep it as it is.

mchf (approved):
lgtm

PR #1115, opened 2020-10-07
Modify the hostname when edited through the DNS client direcly (bsc#1177191)

teclator:
## Problem

There is a separate client (`dns`) for configuring the **hostname** and **resolv** settings. But since https://github.com/yast/yast-network/pull/1010 the DNS module does not modify the hostname changes anymore.

- https://bugzilla.suse.com/show_bug.cgi?id=1177191#c2

## Solution

- Added an option for explicitly tell the `Y2Network::Config` object which section we would like to write and call it from the `DNS` client in case the config has been modified since read. 
- Moved the client to the y2network namespace.
- Fixed 'list' action (summary).
- Added unit tests for the client.

# Test

- tested manually

```
suse@vikingo:~$ cat /etc/hostname
teclator
suse@vikingo:~$ sudo yast2 dns edit hostname=vikingo
suse@vikingo:~$ cat /etc/hostname
vikingo

suse@vikingo:~$ sudo yast2 dns edit prueba
Unknown option for command 'edit': prueba
Use 'yast2 dns edit help' for a complete list of available options.

suse@vikingo:~$ sudo yast2 dns edit nameserver1=8.8.8.8
Cannot set nameserver1. Network is managed by NetworkManager.
```

- Added unit tests


src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -36,33 +36,39 @@ module Sysconfig
...
-        write_interfaces(config.interfaces)
-        write_connections(config.connections, old_config)
-        write_dns_settings(config, old_config)
-        write_hostname_settings(config, old_config)
+        write_routes(config.routing.routes) if sections.include?(:routing)
+        write_drivers(config.drivers) if sections.include?(:drivers)
+        write_interfaces(config.interfaces) if sections.include?(:interfaces)
+        write_connections(config.connections, old_config) if sections.include?(:connections)
+        write_dns_settings(config, old_config) if sections.include?(:dns)
+        write_hostname_settings(config, old_config) if sections.include?(:hostname)
```
jreidinger, line 57: well, when I see this I think it deserve a bit of cleaning and unification. So what is idea? Always has same arguments `config, old_config` and method pick what it needs. and match method name with section name.
And then this method can look like 
```
# add note that sections order is important or not?
SECTIONS.each do { |s| send(:"write_#{s}", config, old_config) if sections.include?(s) }
```
this way adding new section write is simple putting it in SECTIONS and then write write_<section> method.

src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -36,33 +36,40 @@ module Sysconfig
...
         log.info "Writing configuration: #{config.inspect}"
         log.info "Old configuration: #{old_config.inspect}"
-        write_ip_forwarding(config.routing) if config.routing
-        write_interface_changes(config, old_config)
-
-        # update /etc/sysconfig/network/routes file
-        file = routes_file_for(nil)
-        file.routes = find_routes_for(nil, config.routing.routes)
-        file.save
+        write_interface_changes(config, old_config) if sections.include?(:interfaces)
```
jreidinger, line 50: just document why it is needed also here beside below

teclator: Checking it seems it is safe enough to be handled by the write_interfaces method, so, moving it.

teclator: In fact, it is somehow a dependency between the interfaces and the configured routes. As it writes or removes the configured interfaces routes I will move to the `write_routing` method.

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Oct  7 11:45:04 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Write hostname changes when modified through the 'dns' client
+  (bs#1177191).
```
jreidinger, line 5: typo bsc

teclator, line 5:
```suggestion
  (bsc#1177191).
```

src/lib/y2network/clients/dns.rb
```diff
@@ -0,0 +1,284 @@
...
+      end
+
+      def log_and_return(&block)
+        # The main ()
+        log.info("----------------------------------------")
+        log.info("Dns module started")
+        ret = block.call
+        log.debug("ret=#{ret}")
+        # Finish
+        log.info("Dns module finished")
```
jreidinger, line 100: well ret is usually next or abort and I find quite annoying to not see it in log. So I would remove that log.debug and instead use `        log.info("Dns module finished(#{ret.inspect})")`

src/lib/y2network/config.rb
```diff
@@ -129,9 +129,9 @@ def initialize(source:, **opts)
     # @param target   [Symbol] Target to write the configuration to (:sysconfig)
     #
     # @see Y2Network::ConfigWriter
-    def write(original: nil, target: nil)
+    def write(original: nil, target: nil, sections: :all)
```
jreidinger, line 132: please document sections parameter

src/lib/y2network/sysconfig/config_writer.rb
```diff
@@ -36,40 +36,50 @@ module Sysconfig
...
 
+      # @return [Array<Symbol>] The different sections handled by the writer
+      SECTIONS = [:routing, :drivers, :interfaces, :connections, :dns, :hostname].freeze
+
       # Writes the configuration into YaST network related modules
       #
       # @param config     [Y2Network::Config] Configuration to write
       # @param old_config [Y2Network::Config] Old configuration
-      def write(config, old_config = nil)
+      def write(config, old_config = nil, sections: :all)
```
jreidinger, line 46: same here, please document sections parameter and ideally reference also SECTIONS constant from it

src/modules/Lan.rb
```diff
@@ -827,9 +827,9 @@ def read_config
     end
 
     # Writes current yast config and replaces the system config with it
-    def write_config
+    def write_config(sections: :all)
```
jreidinger, line 830: and document also here.

test/y2network/clients/dns_test.rb
```diff
@@ -0,0 +1,158 @@
...
+        context "when some of the options to edit are some of the nameservers" do
+          it "sets them" do
+            expect { subject.main }
+              .to change { config.dns.nameservers.map(&:to_s) }
+              .from([]).to(["8.8.8.8"])
+          end
+        end
+      end
+    end
+  end
```
jreidinger, line 157: :+1: for tests

package/yast2-network.changes
```diff
@@ -10437,3 +10437,4 @@ Thu Jun  8 20:46:57 MEST 2000 - dan@suse.cz
 Fri May 19 09:44:08 CEST 2000 - kkaempf@suse.de
 
 - initial version
+
```
teclator: Apparently I removed it by error and do not want to modify previoys changelog entries :P

jreidinger (approved):

PR #1117, opened 2020-10-13
Minor fixes in AY import

mchf:
## Problem ##

1) unexpected nodes in in the networking section (e.g. <comment> node) caused an internal error during import
2) missing ip for statically configured interface caused an internal error
3) missing bootproto caused an internal error

https://bugzilla.suse.com/show_bug.cgi?id=1175360
https://bugzilla.suse.com/show_bug.cgi?id=1175206

## Solution ##

ad 1) implemented simple filtering which excludes unknown nodes from further processing
ad 2) implemented support for missing ip (according to wicked man page it is valid usecase)
ad 3) implemented support for missing bootproto (according to wicked man page it is valid usecase)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,13 @@
 -------------------------------------------------------------------
+Tue Oct 13 12:57:36 UTC 2020 - Michal Filka <mfilka@suse.com>
+
+- bnc#1175360
+  - more robust AY profile parser. Do not crash with internal error
+    on unknown node in interfaces section
+  - support for static configurations without ip address
+- 4.2.83 
+
+-------------------------------------------------------------------
 Wed Oct  7 11:45:04 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
 - Write hostname changes when modified through the 'dns' client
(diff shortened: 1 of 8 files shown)


src/lib/y2network/autoinst_profile/interfaces_section.rb
```diff
@@ -80,7 +80,9 @@ def initialize(*_args)
       #
       # @param hash [Array] see {.new_from_hashes}. In this case it is array of interfaces
       def init_from_hashes(hash)
-        @interfaces = interfaces_from_hash(hash)
+        # we expect list of hashes here, remove everything else
+        h = hash.delete_if { |v| !v.is_a?(Hash) }
```
teclator, line 84: It is really a hash or an array?, it is needed?

teclator (approved):
LGTM

package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Tue Oct 13 12:57:36 UTC 2020 - Michal Filka <mfilka@suse.com>
+
+- bnc#1175360
```
teclator, line 4: what about also mention https://bugzilla.suse.com/show_bug.cgi?id=1175206 ? at least it is also mentioned already in the PR description

mchf: will add ...

teclator:
still LGTM

PR #1121, opened 2020-11-23
Do not show a warn popup when all the slaves are already configured

teclator:
## Problem

Each time that a bond is modified it prompts an annoying message warning the user about slaves configuration changes, even when the bond slaves are configured properly

- https://trello.com/c/wmoYqtac/2184-l3-1178950-changing-bond-details-through-yast-does-not-change-files-on-disk

## Solution

Check whether the bond slaves are configured with (BOOTPROTO='none') instead of (STARTMODE='none')

## Test

- Added a unit test covering the issue.
- Tested manually

### DIFF
--- a/src/lib/y2network/interface_config_builders/bonding.rb
+++ b/src/lib/y2network/interface_config_builders/bonding.rb
@@ -50,10 +50,13 @@ module Y2Network
         connection_config.options
       end
 
-      # Checks if any of given device is already configured and need adaptation for bridge
+      # Returns whether any configuration of the given devices needs to be
+      # adapted in order to be added as a bonding slave
+
       # @param devices [Array<String>] devices to check
-      # @return [Boolean] true if there is device that needs adaptation
-      def already_configured?(devices)
+      # return [Boolean] true if there is a device config that needs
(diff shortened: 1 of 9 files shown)


jreidinger (approved):

src/lib/y2network/interface_config_builders/bonding.rb
```diff
@@ -50,10 +50,13 @@ def bond_options
...
-      # Checks if any of given device is already configured and need adaptation for bridge
+      # Returns whether any configuration of the given devices needs to be
+      # adapted in order to be added as a bonding slave
+
       # @param devices [Array<String>] devices to check
-      # @return [Boolean] true if there is device that needs adaptation
-      def already_configured?(devices)
+      # return [Boolean] true if there is a device config that needs
+      #   to be adaptated; false otherwise
+      def config_need_to_be_adapted?(devices)
```
jreidinger, line 59: well, in general name is fine, I just prefer something like `require_adaptation?`

jreidinger (approved):

jreidinger (approved):

PR #1123, opened 2020-11-24
Fix wireless essid combobox

teclator:
## Problem

Since yast/yast-yast2#1044, when the selected essid is not part of the scanned network list, the list is not populated correctly and it becomes empty.

Original we fixed it with yast/yast-yast2#1113 but unfortunately introduced a new bug (see https://github.com/yast/yast-yast2/pull/1116)

## Solution

Revert the changes in the ComboBox, not introducing the current_items method and fix the update of the widget items just in this repository. Now we store the items list in a variable which is updated with the scanned network list.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Nov 25 22:23:00 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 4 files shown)


teclator:
With this PR https://github.com/yast/yast-yast2/pull/1117, it is not needed to fix anything at all in yast2-network, so, closing it by now.

PR #1124, opened 2020-11-25
Fix hostname hosts update

teclator:
## Problem

The current installation crashes when trying to update the /etc/hosts aliases using the static hostname which is not set.

## Solution

Fix the crash checking if the current hostname is nil or empty in order to delete the /etc/hosts alias.


mchf (approved):
I'd bet we did similar fix several times already :-/

lgtm, thanks

PR #1126, opened 2020-12-09
Clean SCR cached values before write interface config changes.

teclator:
## Problem

When the number of bonding slaved interfaces is decreased, the change is not reflected correctly in the updated file.

- https://bugzilla.suse.com/show_bug.cgi?id=1178950#c29

## Solution

Clean the SCR cached data in order to write the ifcfg-file correctly.

### DIFF
--- a/src/lib/y2network/sysconfig/connection_config_writer.rb
+++ b/src/lib/y2network/sysconfig/connection_config_writer.rb
@@ -42,6 +42,7 @@ module Y2Network
         handler_class = find_handler_class(conn.type)
         return nil if handler_class.nil?
 
+        remove(old_conn) if old_conn
         file.clean
         handler_class.new(file).write(conn)
         file.save
--- a/src/lib/y2network/sysconfig/connection_config_writer.rb
+++ b/src/lib/y2network/sysconfig/connection_config_writer.rb
@@ -42,8 +42,8 @@ module Y2Network
         handler_class = find_handler_class(conn.type)
         return nil if handler_class.nil?
 
-        remove(old_conn) if old_conn
         file.clean
+        remove(old_conn) if old_conn
         handler_class.new(file).write(conn)
         file.save
       end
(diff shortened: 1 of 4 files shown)


src/lib/y2network/sysconfig/interface_file.rb
```diff
@@ -420,8 +420,16 @@ def type
       # to do some clean-up before writing the final values.
       def clean
         @values = self.class.variables.values.each_with_object({}) do |variable, hash|
-          hash[variable.name] = variable.collection? ? {} : nil
+          if variable.collection?
+            clean_collection(variable.name)
+            hash[variable.name] = {}
+            next
```
imobachgs, line 426: The `if-else-end` looks more readable to me in this case.

src/lib/y2network/sysconfig/connection_config_writer.rb
```diff
@@ -42,7 +42,6 @@ def write(conn, old_conn = nil)
         handler_class = find_handler_class(conn.type)
         return nil if handler_class.nil?
 
-        remove(old_conn) if old_conn
```
imobachgs, line 45: Please, check the `remove` method as it seems to do more than just removing a file. Perhaps you might only need to modify that method instead of removing this line.

imobachgs (commented):
I understand the overall approach. However, I am not sure about simply not calling `remove`.

src/lib/y2network/sysconfig/connection_config_writer.rb
teclator: After some discussion we agreed that this removal is not needed as we are caching all the SCR values and the unmanaged will be written again correctly.

If we do not remove the interface then we need to take into account the interface renaming case.

imobachgs:
LGTM. Let's see what Travis thinks about it :)

imobachgs (approved):

PR #1128, opened 2020-12-10
Permit to modify the connection hostname when the static hostname is modified

teclator:
## Problem

We permit to create a hostname to address mapping when we configure a connection static IP address, but if the user modifies the static hostname we are not warned about a broken mapping.

- https://trello.com/c/G1OvvwjV/2104-3-sle15-sp2-1174431-yast-network-yast-does-not-update-ip-aliases-when-the-hostname-is-modified

## Solution

When the static hostname is modified, permit the user to modify the connection hostname when it was mapped to the previous hostname.

HostnameChange


src/include/network/services/dns.rb
```diff
@@ -294,6 +301,26 @@ def StoreSettings(settings)
       nil
     end
 
+    def current_hostname
+      config&.hostname&.static || ""
+    end
+
+    def update_hostname_hosts(hostname)
+      return false if hostname == current_hostname
+      return false if current_hostname.to_s.empty?
```
jreidinger, line 310: why `to_s` is needed? It looks like it cannot be nil

src/include/network/services/dns.rb
```diff
@@ -294,6 +301,26 @@ def StoreSettings(settings)
...
+
+      conn = config.connections.find { |c| c.hostnames.include?(current_hostname) }
+
+      update_hostname = conn && Popup.YesNo(
+        wrap_text(
+          format(_("The interface '%s' static IP address is mapped to the modified hostname.\n\n" \
+                "Would you like to adapt it now?"), conn.name)
```
jreidinger, line 317: I would mention new hostname here to avoid any confusion.
So something like `"The interface '%s' static IP address is mapped to #{current_hostname}. Would you like to adapt it to #{hostname} now?` Also is possible that new hostname is empty?

src/lib/y2network/connection_config/base.rb
```diff
@@ -204,6 +204,23 @@ def static?
         bootproto ? bootproto.static? : true
       end
 
+      # Return the first hostname associated with the primary IP address
+      #
+      # @return [String, nil]
```
jreidinger, line 209: please document when nil can happen. And originally it cannot return nil. So what it was previously? empty string?

src/lib/y2network/connection_config/base.rb
```diff
@@ -204,6 +204,23 @@ def static?
...
+      end
+
+      # Convenience method in order to modify the canonical hostname mapped to
+      # the primary IP address.
+      #
+      # @param name [String] hostnamme mapped to the primary IP address
+      def hostname=(name)
+        short_name = name&.split(".")&.first
+
+        @hostnames = [name, short_name].uniq.compact
```
jreidinger, line 221: hmm, as it is kind of tricky with nils, can you please add some logging here?

src/lib/y2network/connection_config/base.rb
```diff
@@ -204,6 +204,23 @@ def static?
...
+      def hostname
+        hostnames&.first
+      end
+
+      # Convenience method in order to modify the canonical hostname mapped to
+      # the primary IP address.
+      #
+      # @param name [String] hostnamme mapped to the primary IP address
+      def hostname=(name)
+        short_name = name&.split(".")&.first
```
jreidinger, line 219: according to doc name should not be nil. And if it is not nil, then many magic is removed from this method. ( maybe just check it and raise argument error if you want to be super sure. )

src/lib/y2network/dialogs/popup.rb
```diff
@@ -0,0 +1,113 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "cwm/dialog"
+
+module Y2Network
+  module Dialogs
+    # Adapt CWM dialog to allow popup dialogs
+    class Popup < CWM::Dialog
```
jreidinger, line 26: why copy of Popup class is needed? It is not in SLE15 SP2? If so,  then I think it will be nicer to just backport CWM::Popup to yast2 as expected.

src/lib/y2network/dialogs/update_hostname_hosts.rb
```diff
@@ -0,0 +1,64 @@
...
+          connection_hostname_widget
+        )
+      end
+
+      def buttons
+        [ok_button, cancel_button]
+      end
+
+      def ok_button_label
+        _("Modify")
```
jreidinger, line 49: Modify including shortcut is alrady in Label module https://github.com/yast/yast-yast2/blob/66bc64d08c6ff017b88e3cff63fd02d78bbcf7f1/library/general/src/modules/Label.rb#L185

src/lib/y2network/dialogs/update_hostname_hosts.rb
```diff
@@ -0,0 +1,64 @@
...
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm/popup"
+require "y2network/dialogs/popup"
+require "y2network/widgets/update_connection_hostname"
+
+module Y2Network
+  module Dialogs
+    class UpdateHostnameHosts < Popup
```
jreidinger, line 26: class documentation please.

src/lib/y2network/widgets/update_connection_hostname.rb
```diff
@@ -0,0 +1,88 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "cwm"
+
+module Y2Network
+  module Widgets
+    class UpdateConnectionHostname < CWM::CustomWidget
```
jreidinger, line 24: class documentation please.

src/lib/y2network/widgets/update_connection_hostname.rb
```diff
@@ -0,0 +1,88 @@
...
+        @connection = connection
+      end
+
+      def init
+        ip_address_widget.disable
+      end
+
+      def contents
+        VBox(
+          ip_address_widget,
```
jreidinger, line 38: this looks a bit strange as it is always disabled. Why not use simple label widget?

src/lib/y2network/widgets/update_connection_hostname.rb
```diff
@@ -0,0 +1,88 @@
...
+      def label
+        _("&Hostname")
+      end
+
+      def store
+        @connection.hostname = value
+      end
+    end
+  end
+end
```
jreidinger, line 88: In general a lot of new code to maintenance branch and no tests scary me.

src/include/network/services/dns.rb
teclator: The IP address is mapped to the hostname before the change. The new hostname could be empty, but that is not a problem. The user decides if want to remove the hostname completely or not, I thought about adding an explicit 'delete' button, but currently works as the behavior in the address tab, so, looks consistent.

src/lib/y2network/connection_config/base.rb
teclator: Previously was initialized to nil

teclator: https://github.com/yast/yast-network/pull/1128/files#diff-6b7270d3f0aa0539b5197b7ed2502dbcd53c7549dd2d701d3251f4b7c5faf49eL127

src/lib/y2network/dialogs/popup.rb
teclator: You are right, I forgot it was already available and did not checked it.

src/lib/y2network/widgets/update_connection_hostname.rb
teclator: Yep, probably better.

src/lib/y2network/widgets/update_connection_hostname.rb
teclator: We could modify it only in master by now as it is not a bug at all but a user experience improvement as already commented. We can backport it later if really needed or requested.

src/lib/y2network/connection_config/base.rb
```diff
@@ -204,6 +204,25 @@ def static?
...
+      # @return [String, nil] returns the hostname associated with the primary
+      #   IP address or nil
+      def hostname
+        hostnames&.first
+      end
+
+      # Convenience method in order to modify the canonical hostname mapped to
+      # the primary IP address.
+      #
+      # @param name [String, nil] hostnamme mapped to the primary IP address
```
jreidinger, line 218: yardoc complains that here is name and in method hname

teclator:
```suggestion
      # @param hname [String, nil] hostnamme mapped to the primary IP address
```

jreidinger (approved):

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Mon Dec 14 08:27:46 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Allow to modify the hostname mapped to a connection primary IP
+  address when the associated hostname is modified (bsc#1174431)
+- 4.3.3
```
jreidinger, line 6:
```suggestion
- 4.3.33
```

jreidinger (approved):

PR #1129, opened 2020-12-10
Fix bonding slaves sort

teclator:
## Problem

The bonding slaves are read in an unpredictable order and the up and down buttons for sorting them in the interface do not work.

## Solution

Fix slaves reading and sorting buttons.


src/lib/y2network/sysconfig/connection_config_readers/bonding.rb
```diff
@@ -37,7 +37,7 @@ def update_connection_config(conn)
         #
         # @return [Array<String>] bonding slaves defined in the file
         def slaves
-          (file.bonding_slaves || {}).values
+          (file.bonding_slaves || {}).sort.to_h.values
```
jreidinger, line 40: that sort do it via keys? and alphabetical? This can be problem if there are more then 10 bonds, but I expect it cannot happen

teclator: Yep, I was already thinking on that, but If I'm not wrong we limited it to 8 at least in the past. But we could convert  it to integer to be safe.

jreidinger (approved):

PR #1133, opened 2020-12-16
Move additional IP address configuration popup dialog to its own class

teclator:
## Problem

When an IP alias is edited or added to a connection config and the Netmask uses dot-decimal notation, it is converted to the slash notation without the prefix length.

- https://bugzilla.suse.com/show_bug.cgi?id=1174766

## Solution

The bug was fixed for SLE-15-SP2 / Leap 15.2 in #1132 . So, this PR will merge that one but it also moves the dialog to its own class apart becoming its behavior easier to test.

### DIFF
--- a/src/lib/y2network/dialogs/additional_address.rb
+++ b/src/lib/y2network/dialogs/additional_address.rb
@@ -39,11 +39,9 @@ module Y2Network
       end
 
       def contents
-        focus_label = @settings.label.to_s.empty?
-
         VBox(
-          label_widget(focus_label),
-          ip_address_widget(!focus_label),
+          label_widget,
+          ip_address_widget,
           netmask_widget
         )
       end
@@ -65,16 +63,16 @@ module Y2Network
         [ok_button, cancel_button]
       end
 
-      def label_widget(focus)
-        @label_widget ||= IPAddressLabel.new(@name, @settings)
-        @label_widget.focus if focus
-        @label_widget
+      def focus_label?
+        @settings.label.to_s.empty?
       end
 
-      def ip_address_widget(focus)
-        @ip_address_widget ||= Y2Network::Widgets::IPAddress.new(@settings)
-        @ip_address_widget if focus
-        @ip_address_widget
+      def label_widget
+        @label_widget ||= IPAddressLabel.new(@name, @settings, focus: focus_label?)
+      end
+
+      def ip_address_widget
+        @ip_address_widget ||= Y2Network::Widgets::IPAddress.new(@settings, focus: !focus_label?)
       end
 
       def netmask_widget
@@ -84,15 +82,17 @@ module Y2Network
 
     # Widget to modify the label of an additional IP address configuration
     class IPAddressLabel < CWM::InputField
-      def initialize(name, settings)
+      def initialize(name, settings, focus: false)
         textdomain "network"
 
         @name = name
         @settings = settings
+        @focus = focus
       end
 
       def init
         self.value = @settings.label
+        focus if @focus
         Yast::UI.ChangeWidget(Id(widget_id), :ValidChars, Yast::String.CAlnum)
       end
 
--- a/src/lib/y2network/dialogs/additional_address.rb
+++ b/src/lib/y2network/dialogs/additional_address.rb
@@ -1,0 +1,102 @@
+# Copyright (c) [2020] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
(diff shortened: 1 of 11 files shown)


src/lib/y2network/dialogs/additional_address.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def store
+        @settings.label = value
+      end
+
+      def validate
+        return true unless "#{@name}.#{value}" !~ /^[[:alnum:]._:-]{1,15}\z/
+
+        Yast::Popup.Error(_("Label is too long."))
```
jreidinger, line 96: This is confusing as it said only label is too long, but it can be also wrong characters as input is not llimited.

src/lib/y2network/dialogs/additional_address.rb
```diff
@@ -0,0 +1,102 @@
...
+      def label
+        _("&Address Label")
+      end
+
+      def store
+        @settings.label = value
+      end
+
+      def validate
+        return true unless "#{@name}.#{value}" !~ /^[[:alnum:]._:-]{1,15}\z/
```
jreidinger, line 94: I would be positive here so `if ... =~ `

src/lib/y2network/widgets/additional_addresses.rb
```diff
@@ -105,8 +107,7 @@ def init
 
       def refresh_table
         table_items = @settings.aliases.each_with_index.map do |data, i|
-
-          Item(Id(i), data[:label], data[:ip], "/#{data[:prefixlen]}")
+          Item(Id(i), data[:label], data[:ip_address], data[:subnet_prefix])
```
teclator, line 110: It is probably wrong, I have to fix the way the slash is handled

src/lib/y2network/dialogs/additional_address.rb
```diff
@@ -0,0 +1,102 @@
...
+      def initialize(name, settings)
+        textdomain "network"
+
+        @name = name
+        @settings = settings
+      end
+
+      def init
+        self.value = @settings.label
+        focus if @settings.label.to_s.empty?
```
jreidinger, line 82: to be honest I do not like much that focus is handled on this level. What about having focus as constructor parameter and just focus in init when requested in constructor? ( if it is not used top level but in e.g. tab, then it can be handled on that level )

src/lib/y2network/widgets/additional_addresses.rb
```diff
@@ -154,124 +155,16 @@ def handle(event)
...
-          log.warn "IP address #{ip} and mask #{mask} is not valid"
-        end
-        valid_mask
+      # Convenience method to obtain an object with the additionl IP address
+      # data.
+      #
+      # @param index [Integer, nil] the position of the alias to be edited or
+      #   nil in case a new one is wanted
+      # @return [OpenStruct] additional IP address data
+      def settings_struct_for(index)
```
jreidinger, line 166: I think that method name can be without struct so just `settings_for` or even `settings`

src/lib/y2network/interface_config_builder.rb
```diff
@@ -262,27 +262,31 @@ def driver=(value)
...
+    # data
+    #
+    # @param data [IPConfig, nil] Additional IP address configuration
+    # @return [Hash<String>] hash values are `:label` for alias label,
+    #   `:ip_address` for ip address, `:mask` for netmask and `:subnet_prefix` for prefix.
+    def alias_for(data)
+      {
+        label:         data&.label.to_s,
+        ip_address:    data&.address&.address.to_s,
+        subnet_prefix: data&.address&.prefix.to_s,
```
teclator, line 283: Probably better to do as we are expecting an "/" when converting back.

```suggestion
        subnet_prefix: data ? "/#{data.prefix.to_s}" : ""
```

src/lib/y2network/dialogs/additional_address.rb
teclator: Yep, I still have to add the modification of the valid chars when init.

src/lib/y2network/dialogs/additional_address.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    private
+
+      def buttons
+        [ok_button, cancel_button]
+      end
+
+      def label_widget(focus)
+        @label_widget ||= IPAddressLabel.new(@name, @settings)
+        @label_widget.focus if focus
```
jreidinger, line 70: Does it work? I think not as at this time widget is not yet rendered, so focus is ignored.

teclator: Argh, you are right, it is not. Probably tested it before the change.

jreidinger (approved):

jreidinger (approved):

PR #1134, opened 2020-12-21
Propose the hostname to be used when modified

teclator:
## Problem

In the pull request #1128, we added a new dialog in order to modify a connection config hostname when it was mapped to the modified static hostname.

During the sprint review, it was suggested to not show a dialog but to propose the change directly.

## Solution

Do not show a dialog in order to modify the connection config but propose the hostname to be used.

ProposeHostname


src/include/network/services/dns.rb
```diff
@@ -674,5 +684,14 @@ def dhcp?
...
       config.connections.any?(&:dhcp?)
     end
+
+  private
+
+    def propose_hostname_for(conn, hostname)
+      return hostname if hostname.to_s.empty? || hostname.include?(".")
+
+      _, *domain = conn.hostname.split(".")
+      domain.prepend(hostname).join(".")
```
jreidinger, line 694: hmm, why not simple sub? `conn.hostname.sub(/^[^.]+/, hostname)`

jreidinger (approved):
lokks good, just one NP

jreidinger (approved):

PR #1135, opened 2020-12-23
Improve additional addresses prefix length validation

teclator:
## Problem

When the additional IP address is modified using a prefix length without the "/", it is treated as a netmask and an exception is raised.

- https://bugzilla.suse.com/show_bug.cgi?id=1174766#c15

## Solution

Better handling of the additional IP address subnet_prefix when it is given as a prefix length.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Wed Dec 23 09:21:16 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Improve additional addresses validation (bsc#1174766)
+- 4.3.37
+
+-------------------------------------------------------------------
 Mon Dec 21 13:56:55 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
 
 - When the static hostname is modified, propose the hostname to be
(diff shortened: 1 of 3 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Dec 23 09:21:16 UTC 2020 - Knut Anderssen <kanderssen@suse.com>
+
+- Improve additional addresses validation (bsc#1174766)
+- 4.3.37
```
teclator, line 5: In the previous SR, the version was already increased to 4.3.37 by error.

see https://github.com/yast/yast-network/pull/1134/files#diff-a07b34dd719aba232c4133a770816f1c6c41f8a44bf4347910afbd0d943d16acR20

src/lib/y2network/dialogs/additional_address.rb
```diff
@@ -47,18 +47,42 @@ def contents
...
+            )
+          end
+        end
 
         ret
       end
 
     private
 
+      def subnet_prefix_for?(value)
```
jreidinger, line 73: the question mark here looks really strange as it is not question, but normalization

teclator: yep, you are right, I modified the code and it should be removed from current one.

jreidinger (approved):

PR #1137, opened 2020-12-24
Permit to disable the virt bridge config proposal

teclator:
## Problem

The network bridge configuration is the default configuration of SLE or openSUSE  when configured as a KVM or Xen hypervisor, and during an installation there is no way to disable the bridge configuration proposed by YaST.

This proposed configuration was in many cases transparent to the user and was somehow not noticed but since systemd MACAddressPolicy has changed and a bridge does not inherit the MAC address from its ports, it could be a problem to existent DHCP configurations based on mac addresses (see https://bugzilla.suse.com/show_bug.cgi?id=1136600)

- https://bugzilla.suse.com/show_bug.cgi?id=1178603

## Solution

- By now, we would like to permit the user to disable the current bridge config proposal (in a common installation or in AutoYaST). The link will not be shown in the installation summary unless some of the virtualization packages is selected to be installed.

Creating a link to modify the bridge MACAddress policy as suggested in the bug is out of the scope of this PBI (see also the link example here https://github.com/systemd/systemd/pull/12792/files)

## Screenshots

To be honest, I was not very creative, and use the same tests and links we have in the past which were remove by this commi (https://github.com/yast/yast-network/commit/4cc0a6e32be8bfe46314c9913a51c33712724d40)

ProposeBride
NotProposeBridge

### DIFF
--- a/src/lib/y2network/proposal_settings.rb
+++ b/src/lib/y2network/proposal_settings.rb
@@ -20,6 +20,7 @@
 
 require "yast"
 require "y2packager/package"
+require "network/network_autoconfiguration"
 
 module Y2Network
   # Class that stores the proposal settings for network during installation.
@@ -29,6 +30,7 @@ module Y2Network
 
     # @return [Boolean] network service to be used after the installation
     attr_accessor :selected_backend
+    attr_accessor :virt_bridge_proposal
 
     # Constructor
     def initialize
@@ -38,6 +40,7 @@ module Y2Network
       Yast.import "PackagesProposal"
 
       @selected_backend = nil
+      @virt_bridge_proposal = true
     end
 
     def current_backend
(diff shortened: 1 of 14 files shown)


src/lib/y2network/proposal_settings.rb
```diff
@@ -140,6 +147,10 @@ def create_instance
 
   private
 
+    def package_selected?(name)
+      Y2Packager::Resolvable.any?(kind: :package, name: name, status: :selected)
```
lslezak, line 151: What about the upgrade case? What if the package is already installed? Or is this proposal is only for new installations?

teclator: During upgrade the network is not touched at all, this is only for installation & autoinstallation.

imobachgs (approved):

PR #1140, opened 2021-01-14
Add a writer for NetworkManager

imobachgs:
This PR is the first step toward being able to write the network configuration at installation time even if you are using NetworkManager (default option for openSUSE). It implements these changes:

* Add an incomplete configuration writer for NetworkManager.
* Move the common bits to `Y2Network::ConfigWriter`, including collaborators like `Y2Network::Sysconfig::{DNSWriter,HostnameWriter,InterfacesWriter} which live know under `Y2Network::ConfigWriters`. Maybe we should improve the naming.
* Classes to handle files (`InterfaceFile` and `RoutesFile`) live now in the `CFA` module.

Trello: https://trello.com/c/EdjoT1NN/2256-8-write-networkmanager-configuration-during-installation


teclator (approved):
LGTM

PR #1142, opened 2021-01-19
Add unit tests for the save_network client

imobachgs:
The `save_network` client does a lot of stuff at the end of the installation. However, only a small fraction of the code is covered by unit tests. This PR tries to add simple unit tests to, at least, catch obvious errors during development without having to run the whole installation.


teclator (approved):
LGTM

PR #1143, opened 2021-01-20
Save network manager config

imobachgs:
Write NetworkManager configuration at the end of the installation process.

* Live installer: just copy the files under `/etc/NetworkManager/system-connections` as it is already using NetworkManager.
* Normal installation: use the new support for NetworkManager to write the configuration.

Manually tested the regular and the live installer.

Trello: https://trello.com/c/EdjoT1NN/

### DIFF
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -105,13 +105,6 @@ module Yast
         { dir: ::File.join(ETC, "sysctl.d"), file: "70-yast.conf" }
       ]
 
-      # NetworkManager is usually the default in a live installation. Any
-      # configuration applied during the installation should be present in the
-      # target system.
-      if Y2Network::ProposalSettings.instance.network_service == :network_manager
-        copy_recipes << { dir: ::File.join(NETWORK_MANAGER, "system-connections"), file: "*" }
-      end
-
       # just copy files
       copy_recipes.each do |recipe|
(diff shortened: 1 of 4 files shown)


src/lib/network/clients/save_network.rb
```diff
@@ -298,12 +291,32 @@ def configure_lan
...
 
+    # Configures NetworkManager
+    #
+    # When running the live installation, it is just a matter of copying
+    # system-connections to the installed system. In a regular installation,
+    # write the settings in the Yast::Lan.yast_config object.
+    def configure_network_manager
+      return unless Y2Network::ProposalSettings.instance.network_service == :network_manager
+
+      if Yast::Mode.live_installation
```
teclator, line 312: This mode is still valid? I think that we were not able to rely on it anymore and it is detected as normal mode or something like that.

teclator: We probably should document it in the code:

See https://trello.com/c/ueqrCN8I/3630-improve-live-installation-usability-and-behavior

teclator (approved):
LGTM

src/lib/network/clients/save_network.rb
imobachgs: Good point. I have updated the code to rely on `Yast::Lan.system_config.backend` as you suggested offline.

PR #1144, opened 2021-01-20
WIP: Fixed AY's interfaces reader

mchf:
[bnc#1178107](https://bugzilla.suse.com/show_bug.cgi?id=1178107)

## Problem

1) customer uses static configuration without "main" ip address but use aliases instead of that
2) bug in AY's interfaces reader

## Solution

ad 1) extended fix for [bnc#1175360](https://bugzilla.suse.com/show_bug.cgi?id=1175360). This fix introduced support for static configurations withou ip address defined in the profile. However, this fix ignored aliases. This was fixed
ad 2) There was an error in interfaces reader when aliases from profile where read in the interfaces reader but due to bad indexing (to be more precise: lack of a suffix definition) it was storing all ip addresses (aliases and "the main one") in one place. This lead to situation when only last address was stored for the further processing / writing

## TBD

[ ] tests especially for (2)

### DIFF
--- a/src/lib/y2network/autoinst/interfaces_reader.rb
+++ b/src/lib/y2network/autoinst/interfaces_reader.rb
@@ -87,7 +87,7 @@ module Y2Network
         config.name = name_from_section(interface_section)
         config.interface = config.name # in autoyast name and interface is same
         if config.bootproto == BootProtocol::STATIC
-          # we need at leas some ip for the static configuration
+          # we need at least some ip for the static configuration
           if !interface_section.ipaddr && !interfaces_section.aliases
             msg = "Configuration for #{config.name} is invalid #{interface_section.inspect}"
             raise ArgumentError, msg
@@ -129,22 +129,19 @@ module Y2Network
       end
 
       # Loads and intializates interface_section's ipaddr attribute
+      #
+      # @param section [Hash] hash of AY profile's interface section as obtained from parser
+      #
       # @return [ConnectionConfig::IPConfig] created ipaddr object
-      def load_ipaddr(interface_section)
-        if !interface_section.ipaddr.empty?
-          ipaddr = IPAddress.from_string(interface_section.ipaddr)
-        end
+      def load_ipaddr(section)
+        ipaddr = IPAddress.from_string(section.ipaddr) if !section.ipaddr.empty?
+
         # Assign first netmask, as prefixlen has precedence so it will overwrite it
-        ipaddr.netmask = interface_section.netmask if !interface_section.netmask.to_s.empty?
-        if !interface_section.prefixlen.to_s.empty?
-          ipaddr.prefix = interface_section.prefixlen.to_i
-        end
-        if !interface_section.broadcast.empty?
-          broadcast = IPAddress.new(interface_section.broadcast)
-        end
-        if !interface_section.remote_ipaddr.empty?
-          remote = IPAddress.new(interface_section.remote_ipaddr)
-        end
+        ipaddr.netmask = section.netmask if !section.netmask.to_s.empty?
+        ipaddr.prefix = section.prefixlen.to_i if !section.prefixlen.to_s.empty?
+
+        broadcast = IPAddress.new(section.broadcast) if !section.broadcast.empty?
+        remote = IPAddress.new(section.remote_ipaddr) if !section.remote_ipaddr.empty?
 
         ConnectionConfig::IPConfig.new(
           ipaddr, broadcast: broadcast, remote_address: remote
@@ -152,6 +149,9 @@ module Y2Network
       end
 
       # Loads and initializates an IP alias according to given hash with alias details
+      #
+      # @param alias_h [Hash] hash of AY profile's alias section as obtained from parser
+      #
       # @return [ConnectionConfig::IPConfig] alias details
       def load_alias(alias_h, id: nil)
         ipaddr = IPAddress.from_string(alias_h["IPADDR"])
--- a/src/lib/y2network/autoinst/interfaces_reader.rb
+++ b/src/lib/y2network/autoinst/interfaces_reader.rb
@@ -97,8 +97,8 @@ module Y2Network
         end
 
         # handle aliases
-        interface_section.aliases.each_value do |alias_h|
-          config.ip_aliases << load_alias(alias_h)
+        interface_section.aliases.values.each_with_index do |alias_h, index|
+          config.ip_aliases << load_alias(alias_h, id: "_#{index}")
         end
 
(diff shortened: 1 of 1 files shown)


src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
         config.name = name_from_section(interface_section)
         config.interface = config.name # in autoyast name and interface is same
         if config.bootproto == BootProtocol::STATIC
-          if !interface_section.ipaddr
+          # we need at leas some ip for the static configuration
```
teclator, line 90:
```suggestion
          # we need at least some IP for the static configuration
```

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
         config.name = name_from_section(interface_section)
         config.interface = config.name # in autoyast name and interface is same
         if config.bootproto == BootProtocol::STATIC
-          if !interface_section.ipaddr
+          # we need at least some ip for the static configuration
+          if !interface_section.ipaddr && !interfaces_section.aliases
```
teclator, line 91: interface_section attributes are empty strings, arent them? except aliaes that is an empty hash, and interfaces_section looks like a typo

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
...
-          ipaddr = IPAddress.from_string(alias_h["IPADDR"])
-          # Assign first netmask, as prefixlen has precedence so it will overwrite it
-          ipaddr.netmask = alias_h["NETMASK"] if alias_h["NETMASK"]
-          ipaddr.prefix = alias_h["PREFIXLEN"].delete("/").to_i if alias_h["PREFIXLEN"]
-          config.ip_aliases << ConnectionConfig::IPConfig.new(ipaddr, label: alias_h["LABEL"])
+        interface_section.aliases.values.each_with_index do |alias_h, index|
+          config.ip_aliases << load_alias(alias_h, id: "_#{index}")
         end
+
+        # startmode
```
teclator, line 104: isn't it self explanatory?, looks like useless comments

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
...
         end
 
         config
       end
 
+      # Loads and intializates interface_section's ipaddr attribute
+      #
+      # @param section [Hash] hash of AY profile's interface section as obtained from parser
+      #
+      # @return [ConnectionConfig::IPConfig] created ipaddr object
```
teclator, line 135: miss @param documentation

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
...
+        ConnectionConfig::IPConfig.new(
+          ipaddr, broadcast: broadcast, remote_address: remote
+        )
+      end
+
+      # Loads and initializates an IP alias according to given hash with alias details
+      #
+      # @param alias_h [Hash] hash of AY profile's alias section as obtained from parser
+      #
+      # @return [ConnectionConfig::IPConfig] alias details
```
teclator, line 155: miss param documentation

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -88,56 +87,81 @@ def load_generic(config, interface_section)
...
+      def load_alias(alias_h, id: nil)
+        ipaddr = IPAddress.from_string(alias_h["IPADDR"])
```
teclator, line 157: For a moment I thought that the idea of the PR was to check that the alias IPADDR was not empty bu at the end looks like the same code moved to a convenience method, isn't it?

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: no. If you don't put e.g. ipaddr into the profile, you get nil here.

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: well ... it was for me when organizing the work ... if it hurts, i can delete it

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: as in many other methods in the file ;-) will add.

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: dtto

src/lib/y2network/autoinst/interfaces_reader.rb
mchf: no. 

important line is
```
ConnectionConfig::IPConfig.new(ipaddr, id: id, label: alias_h["LABEL"])
```

there is one minor change which makes it work. Before the change it put all aliases under same index in the internal structure (later in ```add_ips``` - if i recall the name correctly - which let t overwriting and we ended always with exactly one ip ... the one under IPADDR or nothing if it was missing)

mvidner:
Please don't forget a test case

src/lib/y2network/autoinst/interfaces_reader.rb
teclator: This is not correct ipaddr is an empty string and aliases will be an empty hash, in fact the second condition is never checked and that is why a different exception is not raised. But this is wrong.

See https://github.com/yast/yast-network/blob/master/src/lib/y2network/autoinst_profile/interface_section.rb#L268

teclator (changes requested):
As we were discussing and @mvidner was already commenting, at least the unit test should be adapted covering the different use cases that at least has failed recently, that is, parsing correctly the different IP addresses provided.

We already suggested to add the tests in a previous PR (https://github.com/yast/yast-network/pull/1144#discussion_r560969166)

src/lib/y2network/autoinst/interfaces_reader.rb
teclator: In fact, already described in this PR https://github.com/yast/yast-network/pull/1035

mchf:
part of PR#1165, closing

PR #1145, opened 2021-01-20
Write NetworkManager config files

teclator:
## Problem

During installation, the network configuration written is wicked specific, when the backend selected to be used by the target system is NetworkManager the configuration is lost.

- https://trello.com/c/EdjoT1NN/2256-8-write-networkmanager-configuration-during-installation

## Solution

- Write a basic NetworkManager configuration from the connection configs that are known. (Ethernet, Wireless (open, WPA-PSK), VLAN)

### DIFF
--- a/src/lib/cfa/nm_connection.rb
+++ b/src/lib/cfa/nm_connection.rb
@@ -28,17 +28,22 @@ module CFA
   #   file.load
   #   puts file.connection["id"]
   class NmConnection < BaseModel
+    KNOWN_SECTIONS = [
+      "bridge", "connection", "ethernet", "ipv4", "ipv6", "vlan", "wifi", "wifi_security"
+    ].freeze
 
-    KNOWN_SECTIONS = ["connection", "ethernet", "ipv4", "ipv6", "wifi", "wifi_security"].freeze
+    # @return [String] file path
+    attr_reader :path
 
     # Constructor
     #
     # @param path [String] File path
     # @param file_handler [.read, .write] Object to read/write the file.
     def initialize(path, file_handler: nil)
+      @path = path
       # FIXME: The Networkmanager lense writes the values surrounded by double
       # quotes which is not valid
-      super(AugeasParser.new("Puppet.lns"), path, file_handler: file_handler)
+      super(AugeasParser.new("Puppet.lns"), @path, file_handler: file_handler)
     end
 
     # Returns the augeas tree for the given section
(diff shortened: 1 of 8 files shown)


src/lib/y2network/network_manager/connection_config_writer.rb
```diff
@@ -31,16 +31,24 @@ class ConnectionConfigWriter
       def write(conn, old_conn = nil)
         return if conn == old_conn
 
-        file = CFA::NmConnection.new(SYSTEM_CONNECTIONS_PATH.join(conn.name).sub_ext(FILE_EXT))
+        path = SYSTEM_CONNECTIONS_PATH.join(conn.name).sub_ext(FILE_EXT)
+        file = CFA::NmConnection.new(path)
         handler_class = find_handler_class(conn.type)
         return nil if handler_class.nil?
 
+        ensure_permissions(path) unless ::File.exist?(path)
```
imobachgs, line 39: You could also pass the mode to the `CFA::NmConnection` constructor if you prefer or even in the `save` method. Anyway, it is OK here.

imobachgs (commented):
In general, it looks good as a first version. My main concern at this point is that we are overwriting the whole file (we are not calling `load`). I think we should not be that aggressive, but we can address that in a different PR.

src/lib/y2network/network_manager/connection_config_writer.rb
teclator: Yep, was not sure about the best place by now, so did it explicitly here by now.

imobachgs:
> In general, it looks good as a first version. My main concern at this point is that we are overwriting the whole file (we are not calling `load`). I think we should not be that aggressive, but we can address that in a different PR.

Anyway, it does not affect the installer use-case, so we can postpone it.

imobachgs (approved):

src/lib/cfa/nm_connection.rb
```diff
@@ -22,28 +22,68 @@
...
+
     # Constructor
     #
     # @param path [String] File path
     # @param file_handler [.read, .write] Object to read/write the file.
     def initialize(path, file_handler: nil)
-      super(AugeasParser.new("Networkmanager.lns"), path, file_handler: file_handler)
+      @path = path
+      # FIXME: The Networkmanager lense writes the values surrounded by double
+      # quotes which is not valid
```
mchf, line 45: was it reported as a bug to the augeas maintainer?

teclator: Not yet, is it in my TODO list. Do you want to take care? that would be nice

PR #1146, opened 2021-01-21
Rename "Sysconfig" backend to "Wicked"

imobachgs:
Until now, we had two different concepts to define how to read/write the configuration: the source (for instance, `sysconfig`) and the backend (`wicked`). It might allow us, in the future, to configure `wicked` using different set of configuration files (`XML` or `sysconfig` files).

However, such distinction is, at best, confusing. So I am proposing to unify those concepts and to rename the `sysconfig` readers/writers to `wicked`. It should fix a problem at installation time that we have detected when testing our recent changes and it should simplify the overall approach.


teclator (approved):
LGTM

PR #1147, opened 2021-01-21
Installation path and routes

teclator:
- Added support for writing device default routes when writing the NetworkManager config (it needs more work for writing all the routes)
- Take into account the installation destination directory when writing the NetworkManager configuration at the end of the installation.

### DIFF
--- a/src/lib/y2network/network_manager/connection_config_writer.rb
+++ b/src/lib/y2network/network_manager/connection_config_writer.rb
@@ -17,8 +17,11 @@
 # To contact SUSE LLC about this file by physical or electronic mail, you may
 # find current contact information at www.suse.com.
 
+require "yast"
 require "cfa/nm_connection"
 require "pathname"
+
+Yast.import "Installation"
 
 module Y2Network
   module NetworkManager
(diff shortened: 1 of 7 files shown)


src/lib/y2network/network_manager/connection_config_writer.rb
```diff
@@ -17,33 +17,51 @@
...
 
-        handler_class.new(file).write(conn)
+        handler_class.new(file).write(conn, routes)
         file.save
       end
 
     private
 
+      def path_for(conn)
+        Yast.import "Installation"
```
imobachgs, line 60: Duplicated (already imported at the beginning).

src/lib/y2network/network_manager/connection_config_writers/base.rb
```diff
@@ -40,23 +40,36 @@ def initialize(file)
...
-          conn.bootproto.dhcp? ? configure_dhcp(conn) : add_ips(conn)
+          conn.bootproto.dhcp? ? configure_dhcp(conn) : configure_ips(conn)
+          configure_routes(conn, routes)
           update_file(conn)
         end
 
       private
 
+        def configure_routes(conn, routes)
+          routes.select { |r| (r.interface&.name == conn.name) && r.is_default? }.each do |route|
```
imobachgs, line 58: I would pass only the relevant routes (from this code, I infer that we are passing all the routes).

teclator: Originally I passed only the routes for the device, but then realized that we probably will need also the global routes so decided to pass all the routes by now, but can move it back to the config writer.

imobachgs (approved):

imobachgs (approved):

PR #1149, opened 2021-01-22
Added support for writing basic NetworkManager configuration

teclator:
## Problem

Until now, the configuration done in yast2-network was lost when switching from 'wicked' to 'NetworkManager' backend. That is specially annoying during installation as all the wireless configuration has to be done again after booting the installed system.

- https://trello.com/c/EdjoT1NN/2256-8-write-networkmanager-configuration-during-installation
- https://bugzilla.suse.com/show_bug.cgi?id=1180716

## Solution

We have implemented support to write the NetworkManager configuration for ethernet and wireless (limited to open or WPA-PSK authentication) devices.

A normal installation uses wicked so, if NetworkManager is selected as the backend for the installed system, YaST2 will "translate" and write the settings to the final system.

In a live-installation, the situation is different: NetworkManager runs during the installation process. In that case, we kept the same approach: copying the NetworkManager configuration files to the target system.

### Pull Requests

* Add a simple NetworkManager writer: https://github.com/yast/yast-network/pull/1140
* Add unit-tests for the `save_network` client: https://github.com/yast/yast-network/pull/1142
* Write NetworkManager configuration at the end of the installation: https://github.com/yast/yast-network/pull/1143
* Write common and wireless NetworkManager connection settings https://github.com/yast/yast-network/pull/1145
* Write connection default route https://github.com/yast/yast-network/pull/1147


imobachgs (approved):
I am approving this PR because it contains a set of PRs that were already reviewed and approved.

PR #1150, opened 2021-01-22
Cache the online status of s390 Group devices in order to speed up the read dialog

teclator:
## Problem

When the network configuration client is run in a z15 machine with many network s390 group device interfaces, the client hangs when it is preparing the devices overview table.

- https://trello.com/c/eDUHOypB/2259-3-sles15-sp3-p2-1180702-network-configuration-does-not-accept-next-in-the-yast-console

## Solution

Cache the online status of the network S390 group devices when the configuration is read. The status is refreshed in case some of the devices are activated through YaST.


src/lib/y2network/dialogs/s390_device_activation.rb
```diff
@@ -88,7 +88,7 @@ def run
           if configured
             interface_name = activator.configured_interface
             builder.name = interface_name
-            add_interface(interface_name)
+            online!(builder)
```
imobachgs, line 91: Please, do not use an exclamation mark unless you have another method with no exclamation (`sub` and `sub!` is OK, for instance).

imobachgs: Actually, `add_interface` and `online!` are private methods that are called only once. And `add_interface` is an one-liner. Why not merging both of them? You could even add an `online:` argument if needed.

teclator: I will rename it to refresh_devices. About the add_interface, I maintained the convenience method because it was already mocked in the unit test, but will merge both and adapt the test.

imobachgs (approved):

PR #1151, opened 2021-01-22
Suggest to modify the VLAN interface name when the VLAN ID is modified

teclator:
## Problem

The VLAN ID and the interface name are configured in different tabs, thus, it is very common to have a misleading interface name when the VLAN ID is modified.

- https://bugzilla.suse.com/show_bug.cgi?id=1174363

## Solution

Suggest a interface name based on the new 'VLAN ID' when it is modified.

## Screenshot

SuggestVlanInterfaceName
ModifiedInterfaceName

### DIFF
--- a/src/lib/y2network/widgets/vlan_id.rb
+++ b/src/lib/y2network/widgets/vlan_id.rb
@@ -42,6 +42,9 @@ module Y2Network
       end
 
       def store
+        return unless modified?
+
+        @config.name = suggested_name if suggest_vlan_name
         @config.vlan_id = value
       end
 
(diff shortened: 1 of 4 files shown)


src/lib/y2network/widgets/vlan_id.rb
```diff
@@ -52,6 +55,25 @@ def minimum
...
+      end
+
+      def suggested_name
+        "vlan#{value}"
+      end
+
+      def suggest_vlan_name
+        Yast::Popup.YesNo(
+          format(
+            _("Would you like to adapt the interface name from '%s' to '%s'?"),
```
mchf, line 72: I'd add some more noise why the change is being suggested. Something like "VLAN with id  20 is being defined. Would you mind to modify device name from vlan0 to vlan20"

mchf (approved):
only one NP which i don't insist on ;-) lgtm

src/lib/y2network/widgets/vlan_id.rb
teclator: Screenshot from 2021-01-25 08-27-58

PR #1153, opened 2021-01-26
Fix minor usability issues

imobachgs:
* When no interfaces are detected, the *Edit* and *Remove* buttons are disabled.
* The *Scan Network* button is disabled when the wireless interface is not
  present (so it is not possible to scan for ESSIDs).
  
Intentionally we are not bumping the version.
  
Related to https://trello.com/c/sWUhr8qn/1713-5-re-think-the-network-ui-approach
See [bsc#1177834](https://bugzilla.suse.com/show_bug.cgi?id=1177834)


teclator (approved):
LGTM

PR #1155, opened 2021-01-27
Fix InterfacesCollection#by_type

imobachgs:
Fix the `InterfacesCollection#by_type` method.

### DIFF
--- a/src/lib/y2network/interfaces_collection.rb
+++ b/src/lib/y2network/interfaces_collection.rb
@@ -83,10 +83,11 @@ module Y2Network
 
     # Returns list of interfaces of given type
     #
-    # @param type [InterfaceType] device type
+    # @param type [InterfaceType,String,Symbol] device type or its short name
     # @return [InterfacesCollection] list of found interfaces
     def by_type(type)
-      InterfacesCollection.new(interfaces.select { |i| i.type == type })
+      short_name = type.is_a?(Y2Network::InterfaceType) ? type.short_name : type.to_s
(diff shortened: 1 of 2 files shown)


teclator (approved):
LGTM

src/lib/y2network/interfaces_collection.rb
```diff
@@ -83,10 +83,11 @@ def by_busid(busid)
 
     # Returns list of interfaces of given type
     #
-    # @param type [InterfaceType] device type
+    # @param type [InterfaceType,String,Symbol] device type or its short name
     # @return [InterfacesCollection] list of found interfaces
     def by_type(type)
-      InterfacesCollection.new(interfaces.select { |i| i.type == type })
+      short_name = type.is_a?(Y2Network::InterfaceType) ? type.short_name : type.to_s
+      InterfacesCollection.new(interfaces.select { |i| i.type.short_name == short_name })
```
mchf, line 90: does it mean that ```==``` operator for InterfaceType is broken or that ```by_type``` just was accepting ```InterfaceType``` as an argument only? In the second case it is not fix of broken method but improvement ;-)

mchf (approved):
one small NP comment but lgtm

src/lib/y2network/interfaces_collection.rb
imobachgs: mmm, good point. I have improved the implementation relying on the `Y2Network::Interfacetype.from_short_name` method.

teclator (approved):
LGTM: even better than before ;)

PR #1156, opened 2021-01-27
Enhance NM writers supporting more authentication modes (WEP, WPA_EAP)

teclator:
## Problem

In https://github.com/yast/yast-network/pull/1149 we added support for writing basic NM wireless configuration but writes for authentication methods like **WEP** or **WPA_EAP** were not implemented.

- https://trello.com/c/QVm1dv53/2264-2-networkmanager-write-wireless-security-settings

## Solution

- Added support for writing WEP config
- Added support for writing WPA_EAP config
- Added widget for EAP TLS client key password
- Some UI enhancement (fixed eap mode selection)

### WPA_EAP config screenshots

|               | Before | After |
| ----------- | --------- | ------- |
| TTLS | TTLS_before | TTLS |
| TLS | Screenshot from 2021-01-27 12-37-09 | TLS |
| PEAP | PEAP_before | PEAP |

## TODO

- I have not tested it, but looks like WPA_EAP configuration using certificates should be broken after installation as we are not copying the files from the inst-sys to the target system.
- The "Details" button when using WPA_EAP with PEAP mode was removed during the network-ng refactorization, thus,there is not way to modify the authentication method or the PEAP version to be used.

### DIFF
--- a/src/lib/y2network/network_manager/connection_config_writers/wireless.rb
+++ b/src/lib/y2network/network_manager/connection_config_writers/wireless.rb
@@ -27,7 +27,6 @@ module Y2Network
       class Wireless < Base
         DEFAULT_MODE = "infrastructure".freeze
         MODE = { "ad-hoc" => "ad-hoc", "master" => "ap", "managed" => "infrastructure" }.freeze
-        SCHEME_PATH = "file://".freeze
 
         # @see Y2Network::ConnectionConfigWriters::Base#update_file
         def update_file(conn)
@@ -37,28 +36,6 @@ module Y2Network
           file.wifi["channel"] = con.channel if conn.channel
 
           write_auth_settings(conn)
-        end
-
-        # Writes autentication settings for WPA-EAP networks
-        #
-        # @param conn [Y2Network::ConnectionConfig::Base] Configuration to write
-        def write_eap_auth_settings(conn)
-          # FIXME: incomplete
-          file.wifi_security["key-mgmt"] = "wpa-eap"
-          section = file.section_for("802-1x")
-
-          section["eap"] = conn.eap_mode
-          section["phase2-auth"] = conn.eap_auth if conn.eap_auth
-          section["password"] = conn.wpa_password if conn.wpa_password
-          section["anonymous-identity"] = conn.wpa_anonymous_identity if conn.eap_mode == "TTLS"
-          section["identity"] = conn.wpa_identity if conn.wpa_identity
-          section["ca-cert"] = File.join(SCHEME_PATH, conn.ca_cert) if conn.ca_cert
-
-          return unless conn.eap_mode == "TLS"
-
-          section["client-cert"] = File.join(SCHEME_PATH, conn.client_cert)
-          section["private-key"] = File.join(SCHEME_PATH, conn.client_key)
-          section["private-key-password"] = File.join(SCHEME_PATH, conn.client_key_password)
         end
 
         # Writes authentication settings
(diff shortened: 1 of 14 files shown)


src/lib/y2network/network_manager/connection_config_writers/wireless.rb
```diff
@@ -58,25 +47,84 @@ def write_eap_auth_settings(_conn)
...
+          auth_mode = conn.auth_mode || :open
+          meth = "write_#{auth_mode}_auth_settings".to_sym
           send(meth, conn) if respond_to?(meth, true)
         end
 
+        # Writes autentication settings for WPA-EAP networks
+        #
+        # @param conn [Y2Network::ConnectionConfig::Base] Configuration to write
+        def write_eap_auth_settings(conn)
+          # FIXME: incomplete
```
imobachgs, line 62: Could you mention what is missing, please?

src/lib/y2network/network_manager/connection_config_writers/wireless.rb
```diff
@@ -58,25 +47,84 @@ def write_eap_auth_settings(_conn)
         #
         # @see #write_eap_auth_settings
         # @see #write_psk_auth_settings
+        # @see #write_open_auth_settings
         # @see #write_shared_auth_settings
         def write_auth_settings(conn)
-          meth = "write_#{conn.auth_mode}_auth_settings".to_sym
+          auth_mode = conn.auth_mode || :open
+          meth = "write_#{auth_mode}_auth_settings".to_sym
```
imobachgs, line 54: NP: Actually, you do not need the `to_sym` call, right?

src/lib/y2network/network_manager/connection_config_writers/wireless.rb
```diff
@@ -58,25 +47,84 @@ def write_eap_auth_settings(_conn)
...
+          file.wifi_security["wep-key-type"] = passphrase_used ? "2" : "1"
+
+          true
+        end
+
+        # Writes autentication settings for WEP networks (open auth)
+        #
+        # @param conn [Y2Network::ConnectionConfig::Base] Configuration to write
+        def write_open_auth_settings(conn)
+          return if (conn.keys || []).compact.all?(&:empty?)
```
imobachgs, line 113: You might be interested in adding a method to the connection class if needed in order to improve readability. Especially given that you use this construction in several places.

imobachgs (changes requested):
It looks like a lot of work given that you need to check a lot of parameter names. In general, it looks good, just some minor comments.

src/lib/y2network/network_manager/connection_config_writers/wireless.rb
teclator: Should be complete now, that is a leftover :)

src/lib/y2network/network_manager/connection_config_writers/wireless.rb
teclator: Yep, I was already thinking on that, so, time to do it ;)

src/lib/y2network/network_manager/connection_config_writers/wireless.rb
```diff
@@ -120,7 +119,7 @@ def write_open_auth_settings(conn)
         #
         # @param conn [Y2Network::ConnectionConfig::Base] Configuration to write
         def write_shared_auth_settings(conn)
-          return if (conn.keys || []).compact.all?(&:empty?)
+          return if !conn.keys?
```
imobachgs, line 122: np: I prefer "unless" but it is just a personal preference.

imobachgs (approved):

PR #1159, opened 2021-02-04
Cache the netcards 'hwinfo' speeding up the access to it (bsc#1180702)

teclator:
## Problem

S390 Group device objects do not cache the hardware info but obtains it querying the Hwinfo class which according to documentation should memoize it. 

But we cannot reset the info forever as we need to reset the info because some device could be activated and we need to refresh the list of interfaces with the associated one.

- https://trello.com/c/eDUHOypB/2259-3-sles15-sp3-p2-1180702-network-configuration-does-not-accept-next-in-the-yast-console

## Solution

- Memoize the hardware wrapper object.
- Reset the hwinfo before obtaining a new interfaces config.

### DIFF
--- a/src/lib/y2network/hwinfo.rb
+++ b/src/lib/y2network/hwinfo.rb
@@ -113,6 +113,8 @@ module Y2Network
       #
       # @return [HardWrapper]
       def hardware_wrapper
+        return @hardware_wrapper if @hardware_wrapper
+
         @hardware_wrapper = HardwareWrapper.new
       end
 
--- a/src/lib/y2network/wicked/interfaces_reader.rb
+++ b/src/lib/y2network/wicked/interfaces_reader.rb
@@ -51,6 +51,7 @@ module Y2Network
       def config
         return @config if @config
 
+        Hwinfo.reset
         find_s390_devices
         find_physical_interfaces
         find_connections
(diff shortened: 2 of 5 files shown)


src/lib/y2network/wicked/interfaces_reader.rb
```diff
@@ -51,6 +51,7 @@ class InterfacesReader
       def config
         return @config if @config
 
+        Hwinfo.reset
```
teclator, line 54: Would we like to ensure that it is read always or not?, we could reset it outside in case it was too much responsibility for the method, but then it could be too much responsibility to others..

src/lib/y2network/hwinfo.rb
```diff
@@ -113,6 +113,8 @@ def hwinfo_from_hardware(name)
       #
       # @return [HardWrapper]
       def hardware_wrapper
+        return @hardware_wrapper if @hardware_wrapper
+
         @hardware_wrapper = HardwareWrapper.new
```
imobachgs, line 118: It might be more concise:

```ruby
@hardware_wrapper ||= HardwareWrapper.new
```

However, it is true that being so "concise" brought us to this situation :-)

imobachgs (approved):

PR #1160, opened 2021-02-04
Network manager reader

imobachgs:
This PR introduces a reader for NetworkManager. In order to accommodate the new code, I needed to do several non-trivial changes.

* `InterfacesReader` does not read connections any more. Now there is a `ConnectionConfigsReader` which takes care of reading the connections (see #1161).
* Add a `CFA::NmConnection.all` method to find all system connection definitions
* Prepare `Y2Network::Wicked::ConfigReader` so it is easy to extract logic to the base class `Y2Network::ConfigReader` (see #1161).
* Read NetworkManager settings. Only ethernet and wireless devices are supported.

## Current problems

There are a few important differences between NetworkManager and our Wicked-based approach. In our current model, we have the concepts of *interfaces* (physical and virtual) and *connections* (configurations). NetworkManager includes the same concepts, but there are some important differences:

* *connections* are not tied to a particular interface. In our current model, *interfaces* and *connections* are related. For instance, the configuration is saved in a `/etc/sysconfig/network/ifcfg-[name].
* *routes* in *NetworkManager* are included as part of the connections. Therefore, they are not related to a particular interface.

## Next steps

~~We could merge this PR because it contains some code reorganization that might be useful. However, there is already an ongoing discussion about [a new approach regarding our UI](https://trello.com/c/sWUhr8qn/1713-5-re-think-the-network-ui-approach). Then, we might discuss whether it makes sense to improve NetworkManager support in the UI or not.~~

We decided to postpone the UI revamp so we will not need the NetworkManager reader yet. I have moved the useful bits to #1161 which, somehow, replaces this PR.


src/lib/cfa/nm_connection.rb
```diff
@@ -31,15 +35,26 @@ class NmConnection < BaseModel
...
     ].freeze
+    SYSTEM_CONNECTIONS_DIR = "/etc/NetworkManager/system-connections".freeze
 
+    class << self
+      # Returns all connection definition files
+      #
+      # @return [Array<NmConnection>]
+      def all
+        directory = File.join(Yast::Installation.destdir, SYSTEM_CONNECTIONS_DIR)
+        files = Dir[File.join(directory, "*.nmconnection")].to_a
```
teclator, line 46: If I'm not wrong NetworkManager reads any ini file under the system-connections directory.

src/lib/cfa/nm_connection.rb
```diff
@@ -31,15 +35,26 @@ class NmConnection < BaseModel
...
+    end
     # Constructor
     #
     # @param path [String] File path
     # @param file_handler [.read, .write] Object to read/write the file.
     def initialize(path, file_handler: nil)
       # FIXME: The Networkmanager lense writes the values surrounded by double
       # quotes which is not valid
-      super(AugeasParser.new("Puppet.lns"), path, file_handler: file_handler)
+      super(AugeasParser.new("Desktop.lns"), path, file_handler: file_handler)
```
teclator, line 57: The puppet lense was ok for writting the config but not for reading, @mchf was taking care of creating an issue in the augeas repository.

src/lib/cfa/nm_connection.rb
```diff
@@ -80,6 +95,20 @@ def add_collection(section, name, values)
       end
     end
 
+    TYPES_MAP = {
+      wifi:     Y2Network::InterfaceType::WIRELESS,
+      ethernet: Y2Network::InterfaceType::ETHERNET
```
teclator, line 100: VLANs are also supported

src/lib/cfa/nm_connection.rb
```diff
@@ -31,15 +35,26 @@ class NmConnection < BaseModel
...
       "bridge", "connection", "ethernet", "ipv4", "ipv6", "vlan", "wifi", "wifi_security"
     ].freeze
+    SYSTEM_CONNECTIONS_DIR = "/etc/NetworkManager/system-connections".freeze
 
+    class << self
+      # Returns all connection definition files
+      #
+      # @return [Array<NmConnection>]
+      def all
+        directory = File.join(Yast::Installation.destdir, SYSTEM_CONNECTIONS_DIR)
```
teclator, line 45: Just wondering if this could be a problem during a live installation and if it would makes sense to use  the Yast::WFM.scr_root as it is defined by the target file https://github.com/yast/yast-yast2/blob/1629c3a520ebd8e8b93f1161f06d9e140d65b9ec/library/system/src/lib/yast2/target_file.rb#L43.

OR even use the `default_file_handler.final_path` method

teclator (approved):
In general looks good to me. 

About the tied interfaces and until we have a UI that is able to cope with profiles, we could read or check the active connection device and use it at least by now.

src/lib/cfa/nm_connection.rb
imobachgs: You are right. Fixed :-)

imobachgs:
We decided to postpone the UI revamp so we will not need the NetworkManager reader yet. However, this PR included some interesting changes to make the readers easier to extend in the future. That code is now in #1161.

src/lib/cfa/nm_connection.rb
teclator: @imobachgs @mchf we were already evaluating the use of desktop instead of the puppet.lns when adding read support. Probably we should do the same change in #1194

PR #1161, opened 2021-02-05
Better readers

imobachgs:
Initially, this code was part of #1160. However, given that we are not going to read NetworkManager settings by now, I have decided to extract only the useful bits (basically some refactoring and clean-up).

* InterfacesReader does not read connections any more. Now there is a ConnectionConfigsReader which takes care of reading the connections.
* Prepare Y2Network::Wicked::ConfigReader so it is easy to extract logic to the base class Y2Network::ConfigReader.


teclator (approved):
LGTM

PR #1163, opened 2021-02-09
Fix the disable of the "Scan Network" button

teclator:
## Problem

When a new configuration for a Wireless card is created associated to an interface which is not present, then the "Scan Network" button should be disabled. 

The problem is that the current check relies on the builder **newly_added** state which is valid for new connection configs but not for checking if a interface is physical or not.

- https://bugzilla.suse.com/show_bug.cgi?id=1177834

## Solution

- Use the hardware information for checking if the associated interface is present or not.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Feb  9 08:18:26 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix check for not present interfaces when deciding if disable
+  the "Scan Network" button during a wireless configuration.
+  (bsc#1177834).
+- 4.3.45
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue Feb  9 08:18:26 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix check for not present interfaces when deciding if disable
```
imobachgs, line 4: "if disabling"

imobachgs (approved):

package/yast2-network.changes
teclator:
```suggestion
- Fix check for not present interfaces when deciding if disabling
```

imobachgs (approved):

PR #1164, opened 2021-02-12
Wireless scanner

imobachgs:
* Reorganize the wireless tab to make it easier to understand.
* Improve the mechanism to select a wireless network.
* Add a mechanism to scan for wireless networks (see `WirelessScanner` and `WirelessCell` classes).

```ruby 
networks = Y2Network::WirelessNetwork.all("wlo1")
networks.map(&:essid) #=> ["MY_WIFI", "OTHER_WIFI"]
```

Choosing a Wireless Network

## Some Screenshots

<details><summary>Wireless Tab</summary>

**Old UI:**
Old Wireless Tab

**New UI:**
New Wireless Tab

</details>

<details><summary>Network Selection</summary>

**Old UI:**
Old Wireless Network Selection

**New UI:**
New Wireless Network Selection

</details>

## References

* Trello: https://trello.com/c/U7NonPdr/2300-improve-wireless-configuration-ux
* [bsc#1177834](https://bugzilla.suse.com/show_bug.cgi?id=1177834)


src/lib/y2network/dialogs/popup.rb
```diff
@@ -52,7 +52,7 @@ def layout
         )
       end
 
-      protected
+    protected
```
teclator, line 55: As there is already a CWM::Popup we should use it, it is already wrong in the branch I created.

https://github.com/yast/yast-yast2/blob/master/library/cwm/src/lib/cwm/popup.rb

imobachgs:
I will create a fresh branch containing only the needed commits.

imobachgs:
> I will create a fresh branch containing only the needed commits.

Just adjusting the base was enough.

test/y2network/wireless_network_test.rb
```diff
@@ -0,0 +1,82 @@
...
+        an_object_having_attributes(
+          essid: "MY_WIFI", mode: "Master", channel: 10, quality: 65
+        ),
+        an_object_having_attributes(
+          essid: "ANOTHER_WIFI", mode: "Master", channel: 3, quality: 42
+        )
+      )
+    end
+
+    it "memoizes the the results" do
```
teclator, line 76:
```suggestion
    it "memoizes the results" do
```

test/y2network/wireless_auth_mode_test.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+  describe ".from_short_name" do
+    it "returns auth mode with the given name" do
+      expect(described_class.from_short_name("none"))
+        .to eq(Y2Network::WirelessAuthMode::NONE)
+    end
+
+    it "returns nil if the given name not found" do
```
teclator, line 40:
```suggestion
    it "returns nil if the given name is not found" do
```

src/lib/y2network/widgets/wireless_networks.rb
```diff
@@ -0,0 +1,79 @@
...
+      def items
+        @networks.map do |network|
+          [
+            network.essid,
+            network.essid,
+            network.mode,
+            network.channel,
+            network.rates.max.to_s,
+            network.quality_percent ? "#{network.quality_percent}%" : "",
+            network.auth_mode&.to_human_string
```
teclator, line 58: I would prefer to have the supported security options instead of the auth mode, but we could do it later.

src/lib/y2network/wireless_cell.rb
```diff
@@ -0,0 +1,62 @@
...
+    # @!attribute [r] mode
+    #   @return [WirelessMode,nil] Wireless mode
+    # @!attribute [r] channel
+    #   @return [Integer] Wireless channel
+    # @!attribute [r] rates
+    #   @return [Array<Bitrate>] Wireles rates
+    # @!attribute [r] quality
+    #   @return [Integer] Signal quality
+    # @!attribute [r] auth_mode
+    #   @return [Array<String>] Security mechanisms
```
teclator, line 38: I would prefer the array of security mechanisms, but by now the authmode is ok.
```suggestion
    #   @return [WirelessAuthMode] Security mechanisms
```

src/lib/y2network/wireless_cell.rb
```diff
@@ -0,0 +1,62 @@
...
+    # @!attribute [r] address
+    #   @return [String,nil] Cell MAC address
+    # @!attribute [r] essid
+    #   @return [String,nil] ESSID
+    # @!attribute [r] mode
+    #   @return [WirelessMode,nil] Wireless mode
+    # @!attribute [r] channel
+    #   @return [Integer] Wireless channel
+    # @!attribute [r] rates
+    #   @return [Array<Bitrate>] Wireles rates
```
teclator, line 34:
```suggestion
    #   @return [Array<Bitrate>] Wireless rates
```

src/lib/y2network/wireless_auth_mode.rb
```diff
@@ -0,0 +1,79 @@
...
+    #
+    # @return [String]
+    def to_human_string
+      _(name)
+    end
+
+    NONE = new(N_("No Encryption"), "none")
+    WEP_OPEN = new(N_("WEP - Open"), "open")
+    WEP_SHARED = new(N_("WEP - Shared Key"), "shared")
+    WPA_PSK = new(N_("WPA-PSK (\"home\")"), "psk")
```
teclator, line 76: Maybe we could even simplify it to **WPA - Home** and **WPA - Enterprise** but it has been defined so for long time.

src/lib/y2network/widgets/wireless_auth.rb
```diff
@@ -54,23 +54,31 @@ def handle(event)
...
+        refresh
+      end
+
     private
 
       def refresh
         case auth_mode_widget.value
-        when "no-encryption" then replace_widget.replace(empty_auth_widget)
-        when "sharedkey", "open" then replace_widget.replace(wep_keys_widget)
+        when "no_encryption" then replace_widget.replace(empty_auth_widget)
```
teclator, line 80: We have to ensure it is read properly from the ifcfg file.

src/lib/y2network/widgets/wireless_networks.rb
imobachgs: Yes, you are right. We need to "rework" the security stuff, but let's do it as part of a new PBI. Somehow, we have been merging several concepts into a single one (authentication mechanism, authentication suites, ciphers, and so on).

src/lib/y2network/wireless_auth_mode.rb
imobachgs: Yes, when we rework that part, we should take the opportunity to adapt them.

src/lib/y2network/widgets/wireless_auth.rb
imobachgs: Yes, I fixed this behavior in commit 8da8620.

src/lib/y2network/widgets/wireless_expert.rb
```diff
@@ -144,7 +144,7 @@ def maximum
       end
 
       def init
-        self.value = @settings.ap_scanmode
+        self.value = @settings.ap_scanmode.to_i
```
teclator, line 147: Well, it should be an integer already, by default it is initialized to 1, but in the InterfaceFile it is declared as a String instead of an Integer. So, I guess we should unify it there.

src/lib/y2network/dialogs/wireless_networks.rb
```diff
@@ -0,0 +1,137 @@
...
+      def handle
+        @block.call
+
+        nil
+      end
+    end
+
+    # This widget displays a list of wireless networks and allows the user to select one
+    #
+    # @example Returning the ESSID of the selected network
```
teclator, line 53: Now it returns a WirelessNetwork, we probably should fix the example

src/lib/y2network/dialogs/wireless_networks.rb
```diff
@@ -0,0 +1,137 @@
...
+      #
+      # @param builder [InterfaceConfigBuilder]
+      def initialize(builder)
+        textdomain "network"
+        @builder = builder
+      end
+
+      # @see CWM::AbstractWidget
+      def title
+        _("Available Wireless Networks")
```
teclator, line 68: Wireless could be redundant in this context, we already removed from the frames, so maybe we could omit it here too, but nothing important to be addressed.

teclator (approved):
Just a minor note, otherwise LGTM.

src/lib/y2network/widgets/wireless_tab.rb
```diff
@@ -36,20 +38,68 @@ def initialize(builder)
       end
 
       def label
-        _("&Wireless Specific")
+        _("&Wireless")
       end
 
       def contents
         VBox(
-          VSpacing(0.5),
```
mchf, line 44: These "0.5" spacing used to have a reason due to different handling in ncurses and qt. No idea whether it is still relevant

imobachgs: Yes, that's true. But the original UI was crowded. Now, with the new layout, we can put some additional space here :smiley:

teclator: Well, not so much if we want to fit well with 80x24 in ncurses. If we use more space then EAP configuration is a problem as it uses the whole interface.

imobachgs: Yes, but it still fits because we removed a frame and a field. In the future, we could reconsider this change.

mchf (approved):
lgtm

PR #1165, opened 2021-02-15
Ay interfaces reader improvements

teclator:
## Problem

The handling of the parsed AutoYAST network interfaces sections is very fragile and prone to errors, specially when the configuration is declared manually.

In case that some of the elements contain a typo it could crash miserably, we should handle it better.

- https://trello.com/c/1jEzPZ0Q/2302-l3-sles15-sp2-1174353-ipv6-installation-failed-new-64-64-ipaddrinvalidaddresserror
  - https://bugzilla.suse.com/show_bug.cgi?id=1174353
- https://trello.com/c/cu1axq4H/2261-sles15-sp2-p2-1178107-l3-autoyast-install-of-sles15-sp2-has-error-internal-error-caller-usr-lib64-ruby-250-ipaddrrb565in-initial 
  - https://bugzilla.suse.com/show_bug.cgi?id=1178107

## Solution

This PR just extends (https://github.com/yast/yast-network/pull/1144)

In the original PR the exception raised is useless as the section attributes are always defined as an empty object.

In this PR we basically will set the IPConfig to nil when the IPADDR is not defined or defined as empty allowing to write somethine like:

```xml
  <interface>
    <device>eth0</device>
    <bootproto>static</bootproto>
    <startmode>auto</bootproto>
  </interface>
</interfaces>
```

It also covers cases when the netmask provided uses the prefix length format:

```xml
  <interface>
    <device>eth1</device>
    <startmode>auto</startmode>
    <bootproto>static</bootproto>
    <ipaddr>192.168.1.100</ipaddr>
    <netmask>16</netmask>
  </interface>
```

It will skip aliases which do not define an IP address and interfaces sections that does not provide a device name.
Last but not least, in case of a typo when defining the startmode it will fallback to the proposed one.

## TODO

There are different scenarios that will not be rescued properly in case that some of the elements contain typos, like a wrong bootprotocol, startmode ... etc. We could pass the AutoYaST issues list as an argument to the AutoYaST config reader populating it will all the errors we found during a read. I will leave it to implement as a separate PR

### DIFF
--- a/src/lib/y2network/autoinst/interfaces_reader.rb
+++ b/src/lib/y2network/autoinst/interfaces_reader.rb
@@ -86,18 +86,13 @@ module Y2Network
         config.bootproto = BootProtocol.from_name(interface_section.bootproto)
         config.name = name_from_section(interface_section)
         config.interface = config.name # in autoyast name and interface is same
-        if config.bootproto == BootProtocol::STATIC
-          # we need at least some ip for the static configuration
-          if !interface_section.ipaddr && !interfaces_section.aliases
-            msg = "Configuration for #{config.name} is invalid #{interface_section.inspect}"
-            raise ArgumentError, msg
-          end
 
-          config.ip = load_ipaddr(interface_section)
-        end
+        config.ip = load_ipaddr(interface_section)
 
         # handle aliases
         interface_section.aliases.values.each_with_index do |alias_h, index|
+          next if alias_h.fetch("IPADDR", "").empty?
+
           config.ip_aliases << load_alias(alias_h, id: "_#{index}")
         end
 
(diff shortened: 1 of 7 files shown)


src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -87,57 +86,91 @@ def load_generic(config, interface_section)
...
+      end
+
+      def prefix_for(value)
+        if value.empty?
+          nil
+        elsif value.start_with?("/")
+          value[1..-1].to_i
+        elsif value.size < 3 # one or two digits can be only prefixlen
+          value.to_i
+        elsif value =~ /^\d{3}$/
```
jreidinger, line 153: can this be merged with previous condition?

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -47,7 +46,10 @@ def config
           log.info "Creating config for interface section: #{interface_section.inspect}"
           config = create_config(interface_section)
           config.propose # propose reasonable defaults for not set attributes
-          load_generic(config, interface_section)
+          unless load_generic(config, interface_section)
+            log.info "Skipping interface as the configuration was wrongly defined"
```
imobachgs, line 50: I think we should warn the user (using autoinstallation issues, for instance).

test/y2network/autoinst/interfaces_reader_test.rb
```diff
@@ -72,12 +80,70 @@
...
+        it "does not set any IPConfig at all" do
+          eth1_config = subject.config.by_name("eth1")
+          expect(eth1_config.ip).to eql(nil)
+        end
+      end
+
+      context "and provides a netmask in prefix length format" do
+        before { eth1["netmask"] = "16" }
+
+        it "initializes correctly the IPConfig" do
```
imobachgs, line 117: np: *initializes IPConfig object properly* (or *IP configuration*).

test/y2network/autoinst/interfaces_reader_test.rb
```diff
@@ -72,12 +80,70 @@
...
+        alias2 = eth0_config.ip_aliases.find { |a| a.label == "TEST3" }
+        expect(alias2.address.to_s).to eql("10.100.0.3/16")
+      end
+
+      context "and some of the aliases do not provide an IPADDR" do
+        before do
+          eth0["aliases"]["alias1"].delete("IPADDR")
+        end
+
+        it "aliases without the IPADDR are skipped" do
```
imobachgs, line 141: "skips those aliases with no IPADDR" (using a 'verb' sounds better).

imobachgs (commented):
I think we should improve the feedback. Otherwise, it looks good.

src/lib/y2network/autoinst/interfaces_reader.rb
teclator: Yep, that is the idea as commented in the TODO but I planned to add it only in master.

test/y2network/autoinst/interfaces_reader_test.rb
teclator:
```suggestion
        it "skips those aliases with noIPADDR" do
```

test/y2network/autoinst/interfaces_reader_test.rb
teclator:
```suggestion
        it "initializes the IPConifg object properly" do
```

imobachgs (approved):

PR #1166, opened 2021-02-19
AY: Add the errors found during the read to the AutoInstall issues list

teclator:
It extends #1165 adding the errors found to the Yast::AutoInstall.issues_list.

## Example

```xml
      <interface>
        <bootproto>stratic</bootproto>
        <startmode>auto</startmode>
        <device>enp1s0</device>
        <ipaddr>192.168.122.10</ipaddr>
        <netmask>24</netmask>
      </interface>
      <interface>
        <bootproto>static</bootproto>
        <startmode>auh</startmode>
        <device>enp1s1</device>
      </interface>
      <interface>
        <bootproto>dhcp</bootproto>
        <startmode>hotplug</startmode>
      </interface>
    </interfaces>
```

AutoYaSTIssuesReportLast

### DIFF
--- a/src/lib/y2network/autoinst/interfaces_reader.rb
+++ b/src/lib/y2network/autoinst/interfaces_reader.rb
@@ -20,8 +20,7 @@
 require "yast"
 require "y2network/dns"
 require "y2network/autoinst/type_detector"
-require "y2storage/autoinst_issues/invalid_value"
-require "y2storage/autoinst_issues/missing_value"
+require "installation/autoinst_issues"
 require "ipaddr"
 
 Yast.import "AutoInstall"
@@ -31,12 +30,15 @@ module Y2Network
     # This class is responsible of importing the AutoYast interfaces section
     class InterfacesReader
       include Yast::Logger
+      include Yast::I18n
 
       # @return [AutoinstProfile::InterfacesSection]
       attr_reader :section
 
       # @param section [AutoinstProfile::InterfacesSection]
       def initialize(section)
+        textdomain "network"
+
         @section = section
       end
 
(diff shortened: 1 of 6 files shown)


src/lib/y2network/autoinst_profile/networking_section.rb
```diff
@@ -34,26 +35,41 @@ module AutoinstProfile
...
+      #  @return [Boolean]
+      # @!attribute keep_install_network
+      #  @return [Boolean]
+      # @!attribute strict_ip_check_timeout
+      #  @return [Boolean]
+      # @!attribute routing
+      #   @return [RoutingSection]
+      # @!attribute dns
+      #   @return [DNSSection]
+      # @!attribute interfaces
```
teclator, line 67: We probably should modify it as it is really a list of `InterfaceSection` and the same could be applied to other attributes but for the time being do not plan to change it.

mchf (approved):
lgtm

src/lib/y2network/autoinst/interfaces_reader.rb
```diff
@@ -20,8 +20,12 @@
 require "yast"
 require "y2network/dns"
 require "y2network/autoinst/type_detector"
+require "y2storage/autoinst_issues/invalid_value"
+require "y2storage/autoinst_issues/missing_value"
```
teclator, line 24: Although the y2storage classes fits better to our needs, @imobachgs considers that is better to use the yast2 classes and try to adapt and unify them instead of reusing the storage ones and I agree with it.

As commented with him, the y2storage ones fits perfectly to our needs and do not wanted to define our own classes by now, but it is better to not add another level of dependency.

imobachgs (approved):

PR #1170, opened 2021-02-24
/etc/hosts record not created automatically in AY installation

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1178260

## Problem ##

If you have static interface configuration and a hostname definition in AY profile it is expected to end up with record like

<ip> <hostname> ...

in /etc/hosts on the installed system

## Solution ##

Had two phases - for first stage only installation where a minor tweak in return value was enough. Second stage fix was a little bit more complicated as it was re-written to  use new network-ng. Also both fixes work only for installation workflows and not when running AY in installed system.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Wed Mar  3 09:45:02 UTC 2021 - Michal Filka <mfilka@suse.com>
+
+- bnc#1178260
+  - fixed writing a record into /etc/hosts for static ips during
+    AY installation
+- 4.2.94
+
+-------------------------------------------------------------------
 Tue Mar  2 11:48:59 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Support both 'bridge_forwarddelay' and 'bridge_forward_delay'.
--- a/src/lib/network/network_autoyast.rb
+++ b/src/lib/network/network_autoyast.rb
@@ -157,6 +157,8 @@ module Yast
       end
       hosts_config = hosts_config.to_h.delete_if { |k, v| k.empty? || v.empty? }
 
+      return false if hosts_config.empty?
+
       configure_submodule(Host, "hosts" => hosts_config, write: write)
     end
 
--- a/src/lib/y2network/clients/auto.rb
+++ b/src/lib/y2network/clients/auto.rb
@@ -23,6 +23,7 @@ Yast.import "Lan"
 Yast.import "Progress"
 Yast.import "Map"
 Yast.import "Host"
+Yast.import "Stage"
 
 module Y2Network
   module Clients
@@ -89,7 +90,7 @@ module Y2Network
 
         Yast::Lan.Import(modified_profile)
 
-        update_etc_hosts
+        update_etc_hosts if Yast::Stage.cont
 
         true
       end
@@ -124,12 +125,9 @@ module Y2Network
         return if !config
 
         static_connections = config.connections.select(&:static?)
-        static_ips = static_connections.map { |c| c.ip.address.address.to_s }
-
-        log.info("Updating /etc/hosts with #{static_ips} #{config.hostname.static}")
-
-        static_ips.each do |sip|
-          Yast::Host.Update(config.hostname.static, config.hostname.static, sip)
+        static_connections.each do |connection|
+          log.info("Updating /etc/hosts with #{connection.ip.address.address} #{config.hostname.static}")
+          connection.hostname = config.hostname.static
         end
 
         nil
--- a/src/lib/y2network/clients/auto.rb
+++ b/src/lib/y2network/clients/auto.rb
@@ -89,7 +89,7 @@ module Y2Network
 
         Yast::Lan.Import(modified_profile)
 
-        update_etc_hosts(profile)
+        update_etc_hosts
 
         true
       end
@@ -128,7 +128,9 @@ module Y2Network
 
         log.info("Updating /etc/hosts with #{static_ips} #{config.hostname.static}")
 
-        static_ips.each { |sip| Yast::Host.Update(config.hostname.static, config.hostname.static, sip) }
+        static_ips.each do |sip|
+          Yast::Host.Update(config.hostname.static, config.hostname.static, sip)
+        end
 
         nil
       end
--- a/src/lib/y2network/clients/auto.rb
+++ b/src/lib/y2network/clients/auto.rb
@@ -22,7 +22,7 @@ require "installation/auto_client"
 Yast.import "Lan"
 Yast.import "Progress"
 Yast.import "Map"
-Yast.import "LanItems"
+Yast.import "Host"
 
 module Y2Network
   module Clients
@@ -89,6 +89,8 @@ module Y2Network
 
         Yast::Lan.Import(modified_profile)
 
+        update_etc_hosts(profile)
+
         true
       end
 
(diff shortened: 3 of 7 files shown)


src/modules/Lan.rb
```diff
@@ -643,10 +643,20 @@ def Import(settings)
...
       # Ensure that the /etc/hosts has been read to no blank out it in case of
       # not defined <host> section (bsc#1058396)
       Host.Read
 
+      # FIXME: hotfix for Host::ResolveHostnameToStaticIPs - the functionality is
+      # not implemented in new config readers/writers which are used in second stage
+      static_connections = config.connections.select { |c| c.static? }
+      static_ips = static_connections.map { |c| c.ip.address.address.to_s }
+
+      log.info("Lan::Import: updating /etc/hosts with static ips #{static_ips} and hostname #{config.hostname.static}")
```
mchf, line 656: this should go away ... it is just check whether they used the patch correctly ;-)

src/modules/Lan.rb
```diff
@@ -643,10 +643,20 @@ def Import(settings)
       end
 
       NetworkConfig.Import(settings["config"] || {})
+
       # Ensure that the /etc/hosts has been read to no blank out it in case of
       # not defined <host> section (bsc#1058396)
       Host.Read
 
+      # FIXME: hotfix for Host::ResolveHostnameToStaticIPs - the functionality is
```
teclator, line 651: In general hhe relation between the Host and connections hostname has to be reevaluated, at least the method description suggested that the method for using the hostname associated to the static IPs was installation specific. 

At the end it is also expected in AutoYaST which makes sense when there is no `<hosts>` section

src/modules/Lan.rb
```diff
@@ -643,10 +643,20 @@ def Import(settings)
...
+      # not implemented in new config readers/writers which are used in second stage
+      static_connections = config.connections.select { |c| c.static? }
```
teclator, line 653: I will put all this logic in a separate method which deserve a unit test.

teclator:
In general LGTM, the different corner cases should be covered by the AutoYaST reader (merge of configs in case of keep installation and so on)

What could be problematic is when the list of hosts is given, in that case the user explicitly decided to not add the static IP alias and we are not overridden it, isn't it?

mchf:
> What could be problematic is when the list of hosts is given, in that case the user explicitly decided to not add the static IP alias and we are not overridden it, isn't it?

well I hardly doubt that it has been ever specified. In my POV explicitly defined thing from the hosts section should win

src/lib/network/network_autoyast.rb
```diff
@@ -157,6 +157,8 @@ def configure_hosts(write: false)
       end
       hosts_config = hosts_config.to_h.delete_if { |k, v| k.empty? || v.empty? }
 
+      return false if hosts_config.empty?
```
teclator, line 160: With this fix, the behavior should be the same than in SLE-12-X that is, should fallback to the ResolveHostname... behavior, other fix than that affecting the second stage will be a real change of behavior.

teclator:
> > What could be problematic is when the list of hosts is given, in that case the user explicitly decided to not add the static IP alias and we are not overridden it, isn't it?
> 
> well I hardly doubt that it has been ever specified. In my POV explicitly defined thing from the hosts section should win

It is not so, because we read the current file in order to not get rid of defaults, so, basically it preserve what is there, in case of <networking> and <host> section, the entries will be added by lan_auto and then not remove by hosts_auto

teclator:
> > > What could be problematic is when the list of hosts is given, in that case the user explicitly decided to not add the static IP alias and we are not overridden it, isn't it?
> > 
> > 
> > well I hardly doubt that it has been ever specified. In my POV explicitly defined thing from the hosts section should win
> 
> It is not so, because we read the current file in order to not get rid of defaults, so, basically it preserve what is there, in case of and section, the entries will be added by lan_auto and then not remove by hosts_auto

After some discussion, apparently the behavior we are trying to introduce was already there ins SLE-11, but checking the code, I have not found references.

In SLE-11, there was a separate client for specifying the hostname, and the domain name, in case that the network configuration was omitted it added the entry to the /etc/hosts but only in a normal installation, never in AutoYaST

That was modified in https://github.com/yast/yast-network/pull/200, which added the call to configure_hosts, but was never part of lan_auto, that is, the behavior was never part of the `Second Stage`.

So, in SLE-12-X, the /etc/hosts entry added automatically based on the configured static IP addresses are `First Stage` specific. Thus, in the bug, the entry added is based on the linuxrc static IP address configuration.

if the AutoYaST installation is done with `ifcfg=*=dhcp`, no entry will be added to /etc/hosts although we specify a different option in the profile.

Of course, it would be nice to propose it in case that there is no hosts section during the second stage, but lan client is not aware of the host section.

That is my main concern about the behavior change. If it is decided that it is preferable to add the host entry always even if there is a declared host section which does not contain it then ok, if not would be preferable that the entry is declared explicitly and have a proper documentation in AutoYaST.

In SLE-15-SP3 it is different because we are doing everything in the first stage and we can check whether the hosts_section is defined or not to propose anything.

package/yast2-network.changes
```diff
@@ -8,6 +16,8 @@ Fri Feb 26 15:23:33 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
     exist.
 - 4.2.92
 
+-------------------------------------------------------------------
+>>>>>>> 6de30074... Updated changelog
```
teclator, line 20: Leftover

mchf: oh %&*!

src/lib/y2network/clients/auto.rb
```diff
@@ -116,6 +118,23 @@ def merge_current_config?
         !!Yast::Lan.autoinst.keep_install_network
       end
 
+      # Whatever is needed in /etc/hosts during AY installation
+      def update_etc_hosts
+        config = Yast::Lan.find_config(:yast)
+        return if !config
+
+        static_connections = config.connections.select(&:static?)
+        static_ips = static_connections.map { |c| c.ip.address.address.to_s }
```
teclator, line 127: Why not assign the hostname to the connection? the writer should take care of updating the hosts entry, or is there some corner case? Isn't possible that otherwise the hostname is overwritten by the specific writer?

https://github.com/yast/yast-network/blob/SLE-15-SP2/src/lib/y2network/sysconfig/connection_config_writers/base.rb#L57

mchf: i don't know ... I'm not that familiar with connections yet. But will give it a try.

src/lib/y2network/clients/auto.rb
```diff
@@ -89,6 +89,8 @@ def import(profile)
 
         Yast::Lan.Import(modified_profile)
 
+        update_etc_hosts
```
teclator, line 92: I thought we agreed on doing it only in case of a Second stage but not as a general approach.

mchf: yeah ... just checking whether you pay attention ... just joking ... I don't know where it disappeared, seems like a copy paste issue ... fixing it.

src/lib/y2network/clients/auto.rb
```diff
@@ -116,6 +119,21 @@ def merge_current_config?
         !!Yast::Lan.autoinst.keep_install_network
       end
 
+      # Whatever is needed in /etc/hosts during AY installation
+      def update_etc_hosts
+        config = Yast::Lan.find_config(:yast)
+        return if !config
+
+        static_connections = config.connections.select(&:static?)
```
teclator, line 127: I would select the ones which are static and which the hostname is empty.

teclator:
Although I would prefer to document it and add an specific entry in the hosts section in case of 2nd Stage, with the latest modifications I think that is ok.

teclator (approved):
LGTM

PR #1172, opened 2021-03-02
Support bridge_forwarddelay and bridge_forward_delay

imobachgs:
In network-ng refactorization, the section attribute was wrongly named (`bridge_forward_delay` instead of `bridge_forwarddelay`). The problem is that new profiles generated since then will have a different attribute than older ones. To fix the problem, we have decided to support both forms. The new one takes precedence.


teclator (approved):
LGTM

PR #1173, opened 2021-03-02
Permit to slave an ethernet card in s390 arch

teclator:
## Problem

When configuring a bond in s390, only **qeth** devices with Layer2 support are considered as valid slave candidates. In case of PCIe devices like the RoCe ones them are recognized as ethernet interfaces and should be offered.

- https://bugzilla.suse.com/show_bug.cgi?id=1182911
- https://trello.com/c/cDgMQdNf

## Solution

- Do not filter out **Ethernet** interfaces when configuring a bond in (s390)
- Do not propose a 1492 MTU size for ethernet devices (LCS ethernet should be detected as LCS connections)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -3,6 +3,7 @@ Wed Mar  3 12:11:39 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Do not filter ethernet cards when configuring a bond in s390
   (bsc#1182911).
+- Only propose a 1492 MTU size when configuring an LCS interface.
 - 4.2.95
 
 -------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


mchf:
Original request (by IBM) was to show only devices which are L2 capable. At the time of implementing it only qeth devices (devices driven by qeth driver) were known to do so. So, original logic was: we want L2 capable device -> check whether it is qeth -> check whether it is qeth with L2 as all other devices were no-go anyway.

However, with this patch every non-qeth device will pass regardless whether it is L2 capable or not.

teclator:
> Original request (by IBM) was to show only devices which are L2 capable. At the time of implementing it only qeth devices (devices driven by qeth driver) were known to do so. So, original logic was: we want L2 capable device -> check whether it is qeth -> check whether it is qeth with L2 as all other devices were no-go anyway.
> 
> However, with this patch every non-qeth device will pass regardless whether it is L2 capable or not.

This patch is a quick one, and in fact it will not, the logic is the same except for ethernet devices. Based on what you comment, I was originally thinking on maintain the original check for qeth interfaces, but as also hsi interfaces could be treated in the same way.. I decided too keep the same logic until we have more knowledge.

mchf (approved):
lgtm

PR #1175, opened 2021-03-04
Merge SLE-15-SP2 into 'master'

teclator:
It merges #1173 into master and skip revert the merge of #1170 as in master the behavior it is quite different


test/y2network/interface_config_builders/bonding_test.rb
```diff
@@ -109,6 +109,8 @@
 
     context "when the architecture is s390" do
       let(:s390) { true }
+      let(:interface1) { Y2Network::Interface.new("iface1", type: Y2Network::InterfaceType::QETH) }
+      let(:interface2) { Y2Network::Interface.new("iface1", type: Y2Network::InterfaceType::QETH) }
```
imobachgs, line 113: np: "iface1" -> "iface2" ?

imobachgs (approved):

test/y2network/interface_config_builders/bonding_test.rb
teclator: It will not affect the test at all, so will fix only in master

PR #1176, opened 2021-03-04
bsc#1182818: it does not consider 'nfsroot' during installation

imobachgs:
Do not consider proposing 'nfsroot' during installation. The reason is that, at that point, it is not possible to know whether the root filesystem will be on a network device or not. Additionally, it triggers the storage probing without having activated multipath devices. See [bsc#1182818](https://bugzilla.suse.com/show_bug.cgi?id=1182818) for further details.


teclator (approved):
LGTM

PR #1180, opened 2021-03-10
Do not apply changes at the end of the first stage.

teclator:
## Problem

At the end of the first stage and when we configure the network according to the profile we should ensure that the configuration is only written but not applied as restarting the service does not makes sense at all at that point.

It was also found that the "default" value for applying the changes immediately when writing the network configuration through AutoYaST was not the same than the one declared in the network_section and that is because the Profile add it automatically:

https://github.com/yast/yast-autoinstallation/blob/master/src/modules/Profile.rb#L250

But it is not the only thing that we could improve. Currently we are initializing the network section values with the defaults, which means that we are not able to identify whether the options were explicitly requested or not.

- https://bugzilla.suse.com/show_bug.cgi?id=1180535

## Solution

- Ensure the configuration is written but not applied at the end of the first stage (configure_lan)
- Do not initialize networking_section values if not declared in the profile
- Modify the start_immediately default option to true as declared in the Profile.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon Mar 14 08:18:11 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Ensure that during an Autoinstallation the network configuration
+  declared in the profile is only written and not applied when it is
+  done at the end of the first stage (bsc#1180535)
+- 4.3.59
+
+-------------------------------------------------------------------
 Fri Mar 12 11:35:19 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
 - Add more space before the "Name Servers and Domain Search List"
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,9 +1,8 @@
 -------------------------------------------------------------------
-Mon Mar 14 08:18:11 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+Mon Mar 15 08:18:11 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
-- Ensure that during an Autoinstallation the network configuration
-  declared in the profile is only written and not applied when it is
-  done at the end of the first stage (bsc#1180535)
+- AutoYaST: configure but not apply the network configuration at
+  the end of the 1st stage (bsc#1180535).
 - 4.3.59
 
 -------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


src/lib/y2network/autoinst/config.rb
```diff
@@ -46,11 +46,13 @@ class Config
       # @option opts [Integer] :ip_check_timetout
       # @option opts [Boolean] :virt_bridge_proposal
       def initialize(opts = {})
-        @before_proposal = opts.fetch(:before_proposal, false)
-        @start_immediately = opts.fetch(:start_immediately, false)
-        @keep_install_network = opts.fetch(:keep_install_network, true)
-        @ip_check_timeout = opts.fetch(:ip_check_timeout, -1)
-        @virt_bridge_proposal = opts.fetch(:virt_bridge_proposal, true)
+        ay_options = opts.delete_if { |_k, v| v.nil? }
```
imobachgs, line 49: `#delete_if` modifies the original hash. Is that what you want? If not, you could use `#reject`.

src/lib/y2network/autoinst/config_reader.rb
```diff
@@ -72,6 +72,16 @@ def config
 
         config
       end
+
+    private
+
+      # Convenience method to check whether the configured network should be
+      # keeped or not (by default returns true)
```
imobachgs, line 79: keeped -> kept

src/modules/Lan.rb
```diff
@@ -800,7 +800,8 @@ def autoinst_settings?(section)
       !section.start_immediately.nil? ||
         !section.keep_install_network.nil? ||
         !section.setup_before_proposal.nil? ||
-        !section.virt_bridge_proposal.nil?
+        !section.virt_bridge_proposal.nil? ||
+        !section.strict_ip_check_timeout.nil?
```
imobachgs, line 804: np: Perhaps it is time to start thinking about applying the De Morgan's law. :-)

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Fri Mar 12 07:18:11 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Ensure that during an Autoinstallation the network configuration
```
imobachgs, line 4: What about something shorter like *AutoYaST: configure but not apply the network configuration at the end of the 1st stage*?

imobachgs (approved):

imobachgs (approved):

PR #1181, opened 2021-03-11
Updated testsuite to address write issue in OBS

mchf:
AFAIK it happens only in OBS.

https://build.opensuse.org/package/live_build_log/openSUSE:Step:15-SP3/yast2-network/standard/x86_64


mchf:
ok, teclator was faster with github.com/yast/yast-network/pull/1179

PR #1182, opened 2021-03-11
LanItems cleanup - part#1

mchf:
follows cleanup started in https://github.com/yast/yast-network/pull/1154

### DIFF
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -166,10 +166,6 @@ module Yast
       Wizard.RestoreHelp(Ops.get_string(@help, "read", ""))
       Lan.AbortFunction = -> { PollAbort() }
       ret = Lan.Read(:cache)
-      # Currently just a smoketest for new config storage -
-      # something what should replace Lan module in the bright future
-      # TODO: find a suitable place for this config storage
-      Y2Network::Config.from(:wicked)
 
       if Lan.HaveXenBridge
         if !Popup.ContinueCancel(
@@ -212,6 +208,8 @@ module Yast
       return true if ret != :abort
 
       return Popup.ConfirmAbort(:painless) if Stage.initial
+
+      return ReallyAbort() if !(Lan::yast_config == Lan::system_config)
 
       true
     end
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -54,7 +54,6 @@ module Yast
       Yast.import "TablePopup"
       Yast.import "CWMTab"
       Yast.import "Stage"
-      Yast.import "LanItems"
       Yast.import "Systemd"
 
       Yast.include include_target, "network/routines.rb"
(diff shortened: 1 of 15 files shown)


src/include/network/lan/complex.rb
```diff
@@ -219,8 +213,6 @@ def input_done?(ret)
 
       return Popup.ConfirmAbort(:painless) if Stage.initial
 
-      return ReallyAbort() if LanItems.GetModified
```
teclator, line 222: isn't there any replacement of this behavior that we could use?

src/lib/y2network/hwinfo.rb
```diff
@@ -68,10 +68,9 @@ class Hwinfo
     class << self
       # Creates a new instance containing hardware information for a given interface
       #
-      # It retrieves the information from two sources:
+      # It retrieves the information from
       #
-      # * hardware (through {Yast::LanItems} for the time being),
-      # * from existing udev rules.
+      # * existing udev rules.
```
teclator, line 73: It should fit now in just one line, isn't it?

src/lib/y2network/interface_config_builders/ctc.rb
```diff
@@ -25,10 +25,24 @@ module InterfaceConfigBuilders
     class Ctc < InterfaceConfigBuilder
       extend Forwardable
 
+      Yast.import "NetworkConfig"
+
       def initialize(config: nil)
         super(type: InterfaceType::CTC, config: config)
       end
 
+      def save
```
teclator, line 34: Could we extend the unit test covering this ?

mchf: yep

src/include/network/lan/complex.rb
mchf: fine tuned a bit

src/lib/y2network/interface_config_builder.rb
```diff
@@ -30,8 +30,6 @@
 module Y2Network
   # Collects data from the UI until we have enough of it to create a
   # {Y2Network::ConnectionConfig::Base} object.
-  #
-  # {Yast::LanItemsClass#Commit Yast::LanItems.Commit(builder)} use it.
```
teclator, line 34: If we decide to not drop LanItems completely by now, at least the Commit method can be already removed, isn't it?

mchf: We're right in the middle ... I have a follow up patch which removes LanItems completely

mchf: For now LanItems::Commit has been partly replaced by direct call to builder.save ... But it also is part of CWM dialog / sequences and it can also be even somewhere else where I haven't found it yet.

teclator: MMm but that is a different Commit:

https://github.com/yast/yast-network/blob/afa07b2ad2c0484d02606328ef55a45cc36e3cfc/src/include/network/lan/complex.rb#L157

teclator (approved):
LGTM

PR #1183, opened 2021-03-11
bsc#1183257: empty interfaces list

imobachgs:
Allow the interfaces list to be empty. As the element is annotated (`config:type`), it is not a problem.

See [bsc#1183257](https://bugzilla.suse.com/show_bug.cgi?id=1183257).


teclator (approved):
LGTM

PR #1184, opened 2021-03-11
Add empty space before name servers frame 

dgdavid:
While working on https://github.com/yast/yast-theme/pull/144, I realize that there is no vertical space between _Modify DNS Configuration_ and "Name Servers and Domain Search List" in the `Hostname/DNS` tab of Network Settings.

Although it is a really small _cosmetic issue_ with which we can live, I fixed it because there enough room in both, TUI and GUI.

| Before | After |
|-|-|
| yast_lan_tui_before | yast_lan_tui_after |
| yast_lan_before | yast_lan_after |

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Fri Mar 12 11:35:19 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Add more space before the "Name Servers and Domain Search List"
+  frame (related to bsc#1183306).
+- 4.3.58
+
+-------------------------------------------------------------------
 Fri Mar 12 09:23:04 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Fix the VLAN interface renaming when suggested by a VLAN ID
--- a/package/yast2-network.spec
+++ b/package/yast2-network.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.3.57
+Version:        4.3.58
 Release:        0
 Summary:        YaST2 - Network Configuration
 License:        GPL-2.0-only
--- a/src/include/network/services/dns.rb
+++ b/src/include/network/services/dns.rb
@@ -190,6 +190,7 @@ module Yast
         ),
         VSpacing(1),
         Left(HBox("MODIFY_RESOLV", HSpacing(1), "PLAIN_POLICY")),
+        VSpacing(1),
         Frame(
           _("Name Servers and Domain Search List"),
           VBox(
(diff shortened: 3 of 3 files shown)


mchf (approved):
lgtm

teclator:
Already approved my @mchf . I guess that he idea is/was to have the options close to each other so looks like in the same block and are related (hostname config first and then DNS config).

dgdavid:
> Already approved my @mchf . I guess that he idea is/was to have the options close to each other so looks like in the same block and are related (hostname config first and then DNS config).
> 
> I would expect that with https://github.com/yast/yast-theme/blob/a330a5f325c20aa582a83ae481eff0dc3eefc47a/theme/SLE/wizard/installation.qss#L419-L429 is enough, but maybe others have a different opinion.

Not, it's not enough :( In fact, I somehow reverted some changes there, see  https://github.com/yast/yast-theme/pull/144/commits/780966dd366c461e23f89796933afa281a3d4741 and the screenshots using both QGroupBox _margin-top_, 10px and 1.5ex

I could play a little bit more with QSS, but honestly I don't think so it worth. Moreover, having mixed techniques to _organize the space around widgets_, VSpacing and QSS looks weird to me.

teclator:
> > Already approved my @mchf . I guess that he idea is/was to have the options close to each other so looks like in the same block and are related (hostname config first and then DNS config).
> > I would expect that with https://github.com/yast/yast-theme/blob/a330a5f325c20aa582a83ae481eff0dc3eefc47a/theme/SLE/wizard/installation.qss#L419-L429 is enough, but maybe others have a different opinion.
> 
> Not, it's not enough :( In fact, I somehow reverted some changes there, see [yast/yast-theme@780966d](https://github.com/yast/yast-theme/commit/780966dd366c461e23f89796933afa281a3d4741) and the screenshots using both QGroupBox _margin-top_, 10px and 1.5ex
> 
> I could play a little bit more with QSS, but honestly I don't think so it worth. Moreover, having mixed techniques to _organize the space around widgets_, VSpacing and a _margin-top_ in QSS, looks weird to me.

It is not a problem at all related to this change, but in the wireless dialog it could be nicer to fix the frame style instead.

dgdavid:
Conclusions after discussing it with @teclator in a video call:

* This isn't about the problem with the _label vertically cut_, just related to it.
* The UI is holding two different groups: _Hostname_ and _DNS Configuration_.
* _Name Servers and Domain Search List_ is actually a **nested frame** of _Modify DNS Configuration_.
* The _Name Servers and Domain Search List_ spacing problem is more evident in GUI than in TUI.
* We should try to wrap it in a `MarginBox` to make it looks like a nested frame.

dgdavid:
@teclator, @mchf 

This is how it looks after last change, using a `MarginBox` instead of adding a `VSpacing`

| Before | After |
|-|-|
| yast_lan_tui_before | yast_lan_tui_after |
| yast_lan_before | yast_lan_after |

Much better now, isn't it?

teclator:
LGTM

PR #1185, opened 2021-03-12
Fix the rename of the vlan interface when suggested

teclator:
## Problem

[Recently](https://github.com/yast/yast-network/pull/1151) we have added a dialog to suggest the user to adapt the **VLAN** interface name in case that the **VLAN ID** was modified, that works fine in case it was newly added, but in case of edited it failed because we need to ensure we modify the old_name too.

- https://trello.com/c/FHg9gXbR/2343-sles15-sp3-p2-1183357-sles-15-sp3-snapshot11-installation-crashes-on-vlan-rename

## Solution

Do not change the builder interface name directly but call rename_interface instead.


mchf (approved):
lgtm

PR #1186, opened 2021-03-17
NM: add support to bridge and bonding configurations

teclator:
## Problem

While in wicked the relationships are declared in the parent device, in NM it is done in the descendants (slaves or ports). Thus, it is not enough with the given configuration to write the settings but also the parent config needs to be provided.

- https://bugzilla.suse.com/show_bug.cgi?id=1181701

## Solution

Give the parent config as an argument in case of a bridge port or bond slave being able to write the relationship correctly.

### DIFF
--- a/src/lib/y2network/network_manager/config_writer.rb
+++ b/src/lib/y2network/network_manager/config_writer.rb
@@ -35,7 +35,11 @@ module Y2Network
       def write_connections(config, _old_config)
         writer = Y2Network::NetworkManager::ConnectionConfigWriter.new
         config.connections.each do |conn|
-          writer.write(conn, nil, routes_for(conn, config.routing.routes)) # FIXME
+          opts = {
+            routes: routes_for(conn, config.routing.routes),
+            parent: conn.find_master(config.connections)
+          }.reject { |_k, v| v.nil? }
+          writer.write(conn, nil, **opts) # FIXME
         end
       end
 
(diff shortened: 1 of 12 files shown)


src/lib/y2network/network_manager/config_writer.rb
```diff
@@ -35,7 +35,11 @@ class ConfigWriter < Y2Network::ConfigWriter
       def write_connections(config, _old_config)
         writer = Y2Network::NetworkManager::ConnectionConfigWriter.new
         config.connections.each do |conn|
-          writer.write(conn, nil, routes_for(conn, config.routing.routes)) # FIXME
+          opts = {
+            routes: routes_for(conn, config.routing.routes),
+            parent: conn.find_master(config.connections)
+          }.reject { |_k, v| v.nil? }
+          writer.write(conn, nil, **opts) # FIXME
```
imobachgs, line 42: What this "FIXME" means?

imobachgs (approved):
Just a minor comment. Otherwise, it looks good.

src/lib/y2network/network_manager/config_writer.rb
teclator: It comes from https://github.com/yast/yast-network/commit/62fac510bb7a651dc685f78afc3762d7460fbb7c#diff-183cc708348de6ae3fa9bf93781e5524decb751c307eae9889d04abfc720350eR38 so I guess we can get rid of it already

teclator: I will remove it in next PR

PR #1187, opened 2021-03-17
AutoYaST: write profile config to NM

teclator:
## Problem

During an **autoinstallation** the network configuration is basically not written to **NetworkManager** because **wicked** is always considered as the target configuration.

During an **autoinstallation**, the translated network configuration has to be written at the very end of the first_stage as in case of done when still in the inst-sys (setup_before proposal or during the confirm dialog) then it should write wicked config.

- https://bugzilla.suse.com/show_bug.cgi?id=1181701

## Solution

- Store whether the backend selected according to **AutoYaST** is **NetworkManager** or not.
- During an autoinstallation and in case that **NetworkManager** is the backend selected to be used, set it at the very end of the installation being able to write wicked config if needed (setup it before or during the proposal dialog).

## Tests

- Adapted unit tests
- Tested manually

## TODO

Detected problems with DNS because the resolv.conf link is missing. The problem happens when the dns system config and the dns yast config looks the same according to the internal state but the fact is that in the target system the system config is not the one from the inst-sys so we probably should reflect that.

Update: NetworkManager does not call netconfig when the connections are configured using an static IP address. Copying the link at the end of the installation is useless as the run config will be empty after the reboot.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Mar 18 12:59:31 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
-AutoYaST: Write NetworkManager configuration according to the
+- AutoYaST: Write NetworkManager configuration according to the
   profile (bsc#1181701)
 - 4.3.62
 
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Mar 18 12:59:31 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+AutoYaST: Write NetworkManager configuration according to the
+  profile (bsc#1181701)
+- 4.3.62
+
+-------------------------------------------------------------------
 Wed Mar 17 18:26:34 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Always provide the layer2 argument when activating a qeth device
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -346,12 +346,7 @@ module Yast
 
     # Sets default network service
     def set_network_service
-      if Mode.autoinst && !Yast::Lan.autoinst.managed.nil?
-        log.info("Setting network service according to AutoYaST preferences")
-      else
-        log.info("Setting network service according to product preferences")
-      end
-
+      log.info("Setting target system network service")
       backend = Y2Network::ProposalSettings.instance.network_service
 
       # NetworkServices caches the selected backend. That is, it assumes the
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -346,14 +346,14 @@ module Yast
 
     # Sets default network service
     def set_network_service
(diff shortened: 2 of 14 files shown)


src/lib/network/clients/save_network.rb
```diff
@@ -346,14 +346,14 @@ def configure_target
 
     # Sets default network service
     def set_network_service
-      if Mode.autoinst
-        NetworkAutoYast.instance.set_network_service
-        return
+      if Mode.autoinst && !Yast::Lan.autoinst.managed.nil?
```
imobachgs, line 349: I would expect these lines to be near the place where the network service is actually set. Is the condition used elsewhere?

src/lib/y2network/proposal_settings.rb
```diff
@@ -123,12 +123,14 @@ def virtual_proposal_required?
...
     def network_service
-      case current_backend
-      when :network_manager
-        network_manager_installed? ? :network_manager : :wicked
-      else
-        :wicked
+      if current_backend == :network_manager
+        return :network_manager if network_manager_installed?
+
+        log.info("NetworkManager is the selected service but it is not installed")
```
imobachgs, line 129: Why not use just a single line?

teclator: Probably will this is enough, I tried to move the ones from the network_service method in the NetworkAutoYast class but in fact this is already enough and were everything is checked.

src/lib/network/clients/save_network.rb
teclator: I agree, get rid of it

src/lib/network/network_autoconfiguration.rb
```diff
@@ -31,16 +31,18 @@ class NetworkAutoconfiguration
...
-    Yast.import "Package"
-    Yast.import "DNS"
-    Yast.import "Arch"
-    Yast.import "Host"
-
     BASH_PATH = Path.new(".target.bash")
 
+    def initialize
+      Yast.import "Lan"
+      Yast.import "NetworkInterfaces"
```
mchf, line 38: ok, when touching it anyway ... I'd bet that at least ```NetworkInterfaces``` is useless here.

src/lib/network/network_autoconfiguration.rb
```diff
@@ -31,16 +31,17 @@ class NetworkAutoconfiguration
...
-    Yast.import "Package"
-    Yast.import "DNS"
-    Yast.import "Arch"
-    Yast.import "Host"
-
     BASH_PATH = Path.new(".target.bash")
 
+    def initialize
+      Yast.import "Lan"
+      Yast.import "NetworkService"
```
mchf, line 38: this one seems to be useless too as the respective code uses new "ProposalSettings" class

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Mar 18 12:59:31 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+AutoYaST: Write NetworkManager configuration according to the
+  profile (bsc#1181701)
```
imobachgs, line 5: Wrong indentation?

imobachgs (approved):

imobachgs (approved):

PR #1188, opened 2021-03-17
Provides always the layer2 argument when activating a qeth device

teclator:
## Problem

There is a checkbox in the s390 qeth activation dialog in order to enable "layer2" support. When the checkbox is checked we add the `layer2=1` argument to the activation call `czhdev -e group_device_id`, but in case of not checked we are using the auto-detection instead of giving the argument `layer2=0` explicitly.

The auto-detection could be fine in zVM but will not work on LPAR.

S390QethActivationDialog

- https://bugzilla.suse.com/show_bug.cgi?id=1183639

## Solution

Add alwasy the argument `layer2` with the option selected to the activation call.

## Tests

- Adapted unit test
- Tested manually

```
y2network/s390_device_activator.rb(configure):80 Activating s390 device: 0.0.0700:0.0.0701:0.0.0702
lib/cheetah.rb(record_commands):160 Executing "/sbin/chzdev qeth 0.0.0700:0.0.0701:0.0.0702 -e layer2\=0".
```

### DIFF
--- a/src/lib/y2network/s390_device_activators/qeth.rb
+++ b/src/lib/y2network/s390_device_activators/qeth.rb
@@ -49,8 +49,7 @@ module Y2Network
         extra_attributes.concat(attributes.split(" ")) if attributes
         # Only set if enable
         extra_attributes << ipa_takeover_attribute if ipa_takeover
-        # Only set if enable
-        extra_attributes << layer2_attribute if layer2
+        extra_attributes << layer2_attribute
         extra_attributes << port_attribute if port_number.to_s != "0"
         extra_attributes
       end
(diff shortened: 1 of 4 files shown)


src/lib/y2network/s390_device_activators/qeth.rb
```diff
@@ -49,8 +49,7 @@ def configure_attributes
         extra_attributes.concat(attributes.split(" ")) if attributes
         # Only set if enable
         extra_attributes << ipa_takeover_attribute if ipa_takeover
-        # Only set if enable
-        extra_attributes << layer2_attribute if layer2
+        extra_attributes << layer2_attribute
```
jreidinger, line 52: I would add here comment with link to bugzilla that attribute always has to be set

jreidinger (approved):

jreidinger (approved):

PR #1189, opened 2021-03-18
Replace double splat argument as could be problematic

teclator:
## Problem

When the PR #1186 was merged the submission to IBS failed because **ruby2.5** and **ruby2.7** differs in the way double splat arguments are handled.

See https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

Basically a test failed because while in **ruby2.7** it expected the method to not receive any argument with `**empty_hash`, in case of **ruby2.5** it passes an empty hash argument.

## Solution

As in **ruby3.0** positional and keyword arguments will be separated and just to avoid possible issues we will use a hash argument instead of a double splat one.

### DIFF
--- a/test/y2network/network_manager/config_writer_test.rb
+++ b/test/y2network/network_manager/config_writer_test.rb
@@ -71,7 +71,7 @@ describe Y2Network::NetworkManager::ConfigWriter do
     end
 
     it "writes connections configuration" do
-      expect(conn_config_writer).to receive(:write).with(eth0_conn, nil, routes: [])
+      expect(conn_config_writer).to receive(:write).with(eth0_conn, nil, routes: [], parent: nil)
       writer.write(config)
     end
   end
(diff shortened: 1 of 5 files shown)


test/y2network/network_manager/config_writer_test.rb
```diff
@@ -71,7 +71,7 @@
     end
 
     it "writes connections configuration" do
-      expect(conn_config_writer).to receive(:write).with(eth0_conn, nil, routes: [])
+      expect(conn_config_writer).to receive(:write).with(eth0_conn, nil, routes: [], parent: nil)
```
dgdavid, line 74: I'm a bit lost here. Why do you still using keyword arguments in the test? Could you please re-check it?

dgdavid (changes requested):

test/y2network/network_manager/config_writer_test.rb
teclator: Well in fact it pass { routes: [], parent: nil } but rubocop will complain if you use brackets ;)

dgdavid: Ah! Sorry! _Quite confusing_ in this case, but if Rubocop is happy then it LGTM too :)

Thanks Knut!

dgdavid (approved):
LTGM

PR #1192, opened 2021-03-29
Use the ESSID to name the NM config file

imobachgs:
Instead of the interface name, use the ESSID to name the `NetworkManager` configuration file. So it will use `MY_WIRELESS.nmconnection` instead of `wlo1.nmconnection`. It keeps the interface name for the rest of the devices.

Bug: [bsc#1183733](https://bugzilla.opensuse.org/show_bug.cgi?id=1183733)
Related to: https://trello.com/c/jrNQOjfN/


dgdavid (approved):
LGTM

PR #1193, opened 2021-03-31
Dropped another small piece of LanItems

mchf:
Another small piece of LanItems is no longer between us.

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -63,6 +63,9 @@ module Yast
           # but also: reset default gw only if DHCP* is used, this branch covers
           #     "No IP address" case, then default gw must stay (#460262)
           # and also: don't delete default GW for usb/pcmcia devices (#307102)
+          #
+          # Is check for hotplug really needed? It contained a typo up to SLE-15-SP4 and
+          # nobody complained - something to think of during next refactoring.
           yast_config = Y2Network::Config.find(:yast)
           if yast_config&.routing&.default_route
             remove_gw = Popup.YesNo(
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -58,7 +58,7 @@ module Yast
       ret = Y2Network::Dialogs::EditInterface.run(builder)
 
       if ret != :back && ret != :abort
-        if builder.boot_protocol.dhcp? && !LanItems.isCurrentHotplug
+        if builder.boot_protocol.dhcp? && !builder.interface.hotplug?
           # fixed bug #73739 - if dhcp is used, dont set default gw statically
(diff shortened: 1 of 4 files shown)


src/lib/y2network/interface.rb
```diff
@@ -132,5 +132,10 @@ def update_udev_rule
           Y2Network::UdevRule.new_bus_id_based_rename(name, hardware.busid, hardware.dev_port)
       end
     end
+
+    # @return [Boolean] true if the interface is hotplug
+    def hotplug?
+      ["usb", "pcmci"].include?(hardware["hotplug"])
```
imobachgs, line 138: `pcmci` or `pcmcia`? I know, the original one was `pcmci`. Just curious.

src/include/network/lan/address.rb
```diff
@@ -55,12 +55,10 @@ def initialize_network_lan_address(include_target)
     # Dialog for setting up IP address
     # @return dialog result
     def AddressDialog(builder:)
-      @builder = builder
-
       ret = Y2Network::Dialogs::EditInterface.run(builder)
 
       if ret != :back && ret != :abort
-        if LanItems.isCurrentDHCP && !LanItems.isCurrentHotplug
+        if builder.boot_protocol.dhcp? && !builder.interface.hardware.hotplug?
```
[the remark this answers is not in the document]
mchf: ups ... obviously error. will fix

src/lib/y2network/interface.rb
mchf: good question.  "pcmci" has been here longer than I work with the code. I've checked source of hwinfo / libhd. And it seems that PCMCIA is correct (as expected anyway)

See https://github.com/openSUSE/hwinfo/blob/4264ba23bc432193d0e1a522909e299cfb35b468/src/hd/hdp.c#L369

wfeldt: It's `pcmcia`. And I really wonder if we should care about it at all or just remove that stuff.

mchf: thanks ... so it obviously was wrong for about 10 years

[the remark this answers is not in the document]
mchf: some modifications/cleanup in related code were done. Hopefully it looks better now.

wfeldt: Which nicely illustrates why we might remove it.

imobachgs: So it remained unused for about 10 years :smile: Actually, we removed another PCMCIA-related feature (in the autoinstallation rules) back in 2016. It never worked for SLE 12. So I agree with @wfeldt here.

imobachgs (approved):

PR #1194, opened 2021-03-31
Write DNS servers to NetworkManager configuration files

imobachgs:
Write DNS settings to NetworkManager configuration files when using static configuration.

~~TODO: augeas will use "quotes" when writing the values, and it will not work. However, if quotes are removed, augeas will crash. Should we use different parsers for reading/writing?~~ Using the *Desktop* lense, as suggested by @teclator, seems tow work.

### DIFF
--- a/src/lib/cfa/nm_connection.rb
+++ b/src/lib/cfa/nm_connection.rb
@@ -21,6 +21,8 @@ require "cfa/base_model"
 require "pathname"
 require "y2network/connection_config/wireless"
 
+Yast.import "WFM"
+
 module CFA
   # Class to handle NetworkManager connection configuration files
   #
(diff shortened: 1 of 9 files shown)


teclator:
> Write DNS settings to NetworkManager configuration files when using static configuration.
> 
> TODO: augeas will use "quotes" when writing the values, and it will not work. However, if quotes are removed, augeas will crash.
> Should we use different parsers for reading/writing?

Probably by now is the only solution, @mchf what is the status of augeas lense fix? you was working on it, any progress?

teclator:
BTW: it only crashes in case of read, isn't it? I thought that we agreed that reading the config was out of the scope by now and only writing the configuration from the installation was the use case so far.

teclator (approved):
LGTM

PR #1195, opened 2021-04-05
Hide the abort button in a network subworkflow during installation

dgdavid:
### Problem

> When user presses Abort button on Network settings page, it works same way as "Back" button and returns to the previous dialog. — https://bugzilla.suse.com/show_bug.cgi?id=1183586

As stated in the [second comment](https://bugzilla.suse.com/show_bug.cgi?id=1183586#c2), this happens

>  because the network **subworkflow** should not use [Abort] (...) It should be possible to **abort only in the main workflow**.

### Solution

To use an installation argument (`hide_abort_button`) to set when the abort button should be hide, since hiding it always during installation will break the installation UI consistence when the client will be used in the main workflow.

### Notes

As it can be seen in the proposed changes, the call to `Yast::Wizard.HideAbortButton` has been placed in _src/include/network/lan/complex.rb_ [`#MainDialog`](https://github.com/yast/yast-network/blob/618c9d5432a59c179b80ca64e7372a0bc9f7cdec/src/include/network/lan/complex.rb#L256) method, since trying to place it somewhere before in the backtrace does not work (it worked for `Yast::Wizard::DisableAbortButton` but not for hiding it, which actually _worked_ partially being hide and displayed again :confused:).

For the record, the full _path_ is 

[_src/lib/network/clients/inst_lan.rb_](https://github.com/yast/yast-network/blob/618c9d5432a59c179b80ca64e7372a0bc9f7cdec/src/lib/network/clients/inst_lan.rb#L67) -> [_src/include/network/lan/wizards.rb_#LanSequences](https://github.com/yast/yast-network/blob/618c9d5432a59c179b80ca64e7372a0bc9f7cdec/src/include/network/lan/wizards.rb#L57) -> [_src/include/network/lan/wizards.rb_#MainSequence](https://github.com/yast/yast-network/blob/618c9d5432a59c179b80ca64e7372a0bc9f7cdec/src/include/network/lan/wizards.rb#L121) -> [_src/include/network/lan/complex.rb_#MainDialog](https://github.com/yast/yast-network/blob/618c9d5432a59c179b80ca64e7372a0bc9f7cdec/src/include/network/lan/complex.rb#L217)

### Tests

* Only tested manually using `yupdate` to apply changes in a SLE 15-SP3 Build 168.1 

### Screenshots

| Patching the installer | The client running in an installation sub-workflow, w/o the abort button |
|-|-|
| Patching the SUSE SLE installer via yupdate | Network inst_lan client without the abort button |

---

Related Trello card (internal link): https://trello.com/c/6M4weND5

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,8 @@
 -------------------------------------------------------------------
 Mon Apr  5 15:21:47 UTC 2021 - David Diaz <dgonzalez@suse.com>
 
-- Do not show the "Abort" button during installation (bsc#1183586).
+- Do not show the "Abort" button when the inst_lan client is called
+  from another installation step (bsc#1183586).
 - 4.4.0
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Apr  5 15:21:47 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Do not show the "Abort" button during installation (bsc#1183586).
+- 4.4.0
+
+-------------------------------------------------------------------
 Mon Mar 29 11:52:08 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Use the ESSID to name the NetworkManager configuration files
--- a/package/yast2-network.spec
+++ b/package/yast2-network.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.3.63
+Version:        4.4.0
 Release:        0
 Summary:        YaST2 - Network Configuration
 License:        GPL-2.0-only
--- a/src/include/network/lan/complex.rb
+++ b/src/include/network/lan/complex.rb
@@ -253,7 +253,7 @@ module Yast
       )
 
       if running_installer
-        Wizard.SetAbortButton(:abort, Label.AbortButton)
+        Wizard.HideAbortButton
       else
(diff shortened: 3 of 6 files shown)


dgdavid (commented):
Should we sent this patch for SLE-15-SP3?

package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.3.63
+Version:        4.4.0
```
dgdavid, line 20: Reviewers, please see https://github.com/yast/yast-network/pull/1195#pullrequestreview-627943908

> Should we sent this patch for SLE-15-SP3?

teclator:
The main problem I see is when it is called from the main workflow:

https://github.com/yast/skelcd-control-leanos/blob/master/control/control.leanos.xml#L565

Currently we are hiding it always during an installation or update.

dgdavid:
Ey @teclator 

Thanks a lot for your chat and hints today afternoon. I have updated the PR to use the new `hide_abort_button` installation argument and created a new one in yast-registration, where the _inst\_client_ is called too.

Please, have a look again when you have time.

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Mon Apr  5 15:21:47 UTC 2021 - David Diaz <dgonzalez@suse.com>
+
+- Do not show the "Abort" button during installation (bsc#1183586).
```
teclator, line 4:
```suggestion
- Do not show the "Abort" button when the inst_lan client is called
  from another installation step (bsc#1183586).
```

teclator (approved):
LGTM

teclator (approved):
LGTM

PR #1198, opened 2021-04-08
Display network configuration in the summary screen for NetworkManager

imobachgs:
Until now, YaST did not show the network configuration details in the summary screen when NetworkManager was the selected backend. Now, that's not the case anymore, and the network configuration is shown always.

* [bsc#1181354](https://bugzilla.suse.com/show_bug.cgi?id=1181354)
* Trello: https://trello.com/c/xM0pDT6t/

Installation summary


teclator (approved):
LGTM

PR #1199, opened 2021-04-09
Write bonding slaves config properly

teclator:
## Problem

When a connection config is written we are adding attributes with empty values which are irrelevant and could be omitted. In case of bonding we are also adding slaves with the _0.0.0.0_ **IPADDR** which is invalid.

- https://trello.com/c/KFilasdD/2381-3-network-ng-clean-up-configuration-to-be-written

## Solution

- Skip empty values when writing.
- Skip the write of the static _0.0.0.0_ **IPADDR**.
- Do not create empty ifroute files.

Written bonding config

### DIFF
--- a/src/lib/y2network/interface_config_builders/bonding.rb
+++ b/src/lib/y2network/interface_config_builders/bonding.rb
@@ -60,7 +60,7 @@ module Y2Network
         devices.any? do |device|
           next false unless yast_config.configured_interface?(device)
 
-          yast_config.connections.by_name(device).bootproto.name != "none"
+          !valid_slave_config?(device)
         end
       end
 
@@ -68,12 +68,13 @@ module Y2Network
       def save
         slaves.each do |slave|
           interface = yast_config.interfaces.by_name(slave)
-          connection = yast_config.connections.by_name(slave)
-          next if connection && connection.bootproto.name == "none"
+          next if valid_slave_config?(slave)
 
+          connection = yast_config.connections.by_name(slave)
           builder = InterfaceConfigBuilder.for(interface.type, config: connection)
           builder.name = interface.name
           builder.configure_as_slave
+          builder.startmode = "hotplug"
           builder.save
         end
 
@@ -117,6 +118,16 @@ module Y2Network
         return false if iface.name == @name
 
         true
+      end
+
+      # Convenience method to check whether the config of an enslaved interface
+      # is valid or not
+      #
+      # @param iface [String] slave name to be validated
+      def valid_slave_config?(iface)
+        conn = yast_config.connections.by_name(iface)
+
+        conn&.bootproto&.name == "none" && conn&.startmode&.name == "hotplug"
       end
     end
   end
--- a/src/lib/y2network/wicked/connection_config_writers/base.rb
+++ b/src/lib/y2network/wicked/connection_config_writers/base.rb
@@ -26,8 +26,11 @@ module Y2Network
       #
       # The derived classes should implement {#update_file} method.
       class Base
+        extend Forwardable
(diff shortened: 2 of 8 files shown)


src/lib/y2network/interface_config_builders/bonding.rb
```diff
@@ -118,6 +119,16 @@ def bondable?(iface)
 
         true
       end
+
+      # Convenience method to check whether the config of an enslaved interface
+      # is valid or not
+      #
+      # @param iface [String] slave name to be validated
+      def valid_slave_config?(iface)
```
imobachgs, line 127: I am not sure whether this logic belongs here. Please, make a note and discuss later, when we have better support for validation/errors reporting.

src/lib/y2network/wicked/connection_config_writers/base.rb
```diff
@@ -40,17 +43,15 @@ def initialize(file)
         #
         # @param conn [Y2Network::ConnectionConfig::Base] Connection to take settings from
         def write(conn)
-          file.bootproto = conn.bootproto&.name
-          file.name = conn.description
-          file.lladdr = conn.lladdress
-          file.startmode = conn.startmode.to_s
-          file.dhclient_set_hostname = dhclient_set_hostname(conn)
+          file.bootproto = value_as_string(conn.bootproto&.name)
```
imobachgs, line 46: np: Perhaps we might think about adding support for such a conversion to "bootproto=" (and the rest of these methods).

imobachgs (commented):
I have the impression that we should improve error handling. I am already working on something about that, so let's move forward with this PR by now.

mchf:
Just a note ... AFAIK this was requested as a feature in the past (to store even unfilled options). There used to be even comment (!) about it in the old code.

src/lib/y2network/wicked/connection_config_writers/base.rb
mchf: +1 we have similar conversions (from string to a <type>) ... so it would be even nice to be consistent

src/lib/y2network/wicked/connection_config_writers/base.rb
```diff
@@ -105,6 +106,13 @@ def add_hostname(conn)
 
           Yast::Host.Update("", conn.hostname, conn.ip.address.address.to_s)
         end
+
+        # @param conn [Y2Network::ConnectionConfig::Base] Connection to take settings from
+        def static_ip?(conn)
+          return false unless conn.bootproto.static?
+
+          conn.ip && conn.ip.address.address.to_s != "0.0.0.0"
```
mchf, line 114: this method does little bit more than described ... it is not only about checking whether it is static ip or not as method name suggests.

teclator: Well, if the bootproto is not static then it reports false as although there is an static IP it is irrelevant, and in case of static IP it should be a valid one, it is valid according to **IPAddr** but not a valid config to be written, valid_static_ip? sounds better to you? what you propose, it is a convenience method so not exposed at all.

imobachgs (approved):
Fine with me. Please, wait for @mchf's feedback before merging. Thanks!

src/lib/y2network/wicked/connection_config_writers/base.rb
mchf: yeah sounds better ... 

code should make things as clear as possible. Doesn't depend if the method is public or not. Especially now when functionality is "hidden" (nested) in a hierarchy of classes.

mchf (approved):
lgtm

PR #1200, opened 2021-04-12
War with LanItems continues, another battle done

mchf:


imobachgs (approved):

src/lib/y2network/virtualization_config.rb
```diff
@@ -72,14 +70,44 @@ def create
...
+        move_routes(builder.name, bridge_builder.name)
       end
 
       true
     end
 
   private
 
+    def yast_config
+      Y2Network::Config.find(:yast)
```
teclator, line 82: Is this really needed?, as the config is an instance variable which is initialized by the constructor.

mchf: right

src/lib/y2network/virtualization_config.rb
```diff
@@ -72,14 +70,38 @@ def create
...
       end
 
       true
     end
 
   private
 
+    # Adds a new interface with the given name
+    def add_device_to_routing(name)
+      return if @config.nil?
```
teclator, line 83: it is an attr_reader, you can use 

```suggestion
      return unless config
```
```suggestion
      return if @config.nil?
```

src/lib/y2network/virtualization_config.rb
```diff
@@ -72,14 +70,38 @@ def create
...
+      return if @config.interfaces.any? { |i| i.name == name }
```
teclator, line 84:
```suggestion
      return if config.interfaces.any? { |i| i.name == name }
```

teclator: And the same with the rest

teclator (approved):
LGTM

PR #1203, opened 2021-04-13
Do not crash when checking if an interface is a hotplug one (bsc#1184623)

teclator:
## Problem

When we try to configure a virtual interface or an interface which is not present it does not have an interface associated and the hotplug check produces a crash.

- https://bugzilla.suse.com/show_bug.cgi?id=1184623

## Solution

- Just do not crash when the interface is not present and return early in case of a hotplug interface.

## Tests

- Tested manually

### DIFF
--- a/src/include/network/lan/address.rb
+++ b/src/include/network/lan/address.rb
@@ -57,31 +57,33 @@ module Yast
     def AddressDialog(builder:)
       ret = Y2Network::Dialogs::EditInterface.run(builder)
 
-      if ret != :back && ret != :abort
-        if builder.boot_protocol.dhcp? && !builder.interface.hotplug?
-          # fixed bug #73739 - if dhcp is used, dont set default gw statically
-          # but also: reset default gw only if DHCP* is used, this branch covers
-          #     "No IP address" case, then default gw must stay (#460262)
-          # and also: don't delete default GW for usb/pcmcia devices (#307102)
-          #
-          # Is check for hotplug really needed? It contained a typo up to SLE-15-SP4 and
-          # nobody complained - something to think of during next refactoring.
-          yast_config = Y2Network::Config.find(:yast)
-          if yast_config&.routing&.default_route
-            remove_gw = Popup.YesNo(
-              _(
-                "A static default route is defined.\n" \
-                "It is suggested to remove the static default route definition \n" \
-                "if one can be obtained also via DHCP.\n" \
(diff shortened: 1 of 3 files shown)


src/include/network/lan/address.rb
```diff
@@ -57,31 +57,33 @@ def initialize_network_lan_address(include_target)
...
-              _(
-                "A static default route is defined.\n" \
-                "It is suggested to remove the static default route definition \n" \
-                "if one can be obtained also via DHCP.\n" \
-                "Do you want to remove the static default route?"
-              )
+      log.info "AddressDialog res: #{ret.inspect}"
+
+      return ret if [:back, :abort].include?(ret)
+      return ret if builder.interface&.hotplug?
```
teclator, line 63: We could even return if !builder.boot_protocol.dhcp?

src/include/network/lan/address.rb
```diff
@@ -57,31 +57,33 @@ def initialize_network_lan_address(include_target)
...
+      if builder.boot_protocol.dhcp?
+        # fixed bug #73739 - if dhcp is used, dont set default gw statically
+        # but also: reset default gw only if DHCP* is used, this branch covers
+        #     "No IP address" case, then default gw must stay (#460262)
+        # and also: don't delete default GW for usb/pcmcia devices (#307102)
+        #
+        # Is check for hotplug really needed? It contained a typo up to SLE-15-SP4 and
+        # nobody complained - something to think of during next refactoring.
+        yast_config = Y2Network::Config.find(:yast)
+        if yast_config&.routing&.default_route
```
imobachgs, line 74: What about merging both conditions?

```ruby
if builder.boot_protocol.dhcp? && yast_config&.routing&.default_route
  # ...
end
```

imobachgs (commented):
The change looks good. I was just considering whether we could simplify it a little bit more...

src/include/network/lan/address.rb
teclator: Yep, was already thinking on moving that

imobachgs: Or even `return ret unless ...`. As it is more readable for you.

imobachgs (approved):
It looks good to me. But perhaps @mchf has additional comments so, please, let's wait for him.

mchf (approved):
lgtm

PR #1204, opened 2021-04-13
Reading errors

imobachgs:
This PR contains a proposal to handle errors when reading the network configuration. The solution could be moved to `yast2.rpm` and used in more places.

**It is still a WIP, but feedback is welcome!**

## The Problem

When reading the network configuration, many things can go wrong. And, usually, YaST is [not able to handle those situations](https://bugzilla.suse.com/show_bug.cgi?id=1181296).

## The Proposed Solution

The idea is to provide a mechanism similar to [AutoinstIssues](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/autoinst_issues.rb) to register and report errors to the user. In this PR, we are handling the case where `BOOTPROTO` contains an invalid name. Additionally, we are falling back to a sensible (not yet) value and asking the user whether we should continue.

<details>
<summary>Screenshot: Invalid network configuration</summary>

invalid network configuration
</details>

<details>
<summary>Code Example: Checking Errors</summary>

```ruby
require "yast"
require "y2network/config_reader"

reader = Y2Network::ConfigReader.for(:wicked)
result = reader.config

puts Y2Network::Errors::Presenter.new(result.errors).to_s
```
</details>

<details>
<summary>Code Example: Adding Errors</summary>

```ruby
require "yast"
require "y2network/errors"

errors = Y2Network::Errors::List.new
errors << Y2Network::Errors::InvalidValue.new("boot protocol", "magic")
errors << Y2Network::Errors::InvalidValue.new("start mode", "random", "auto")
puts Y2Network::Errors::Presenter.new(errors).to_s
```
</details>

## To do

- [ ] Idea: bring `AutoinstIssues` and `Errors` together (using a mixin or even 'inheritance').
- [ ] [AutoinstIssues::List](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/installation/autoinst_issues/list.rb) looks pretty similar to `Errors::List`. Should we merge them? What about the rest?
- [x] Decide about a proper name. *Error* seems related to *exceptions* (*problem*, *issue*, ...)
- [x] Split the `errors.rb` into different files. Move them to `yast2`?
- [ ] Should `MissingValue` and `InvalidValue` be merged into a single class?
- [x] Add a mechanism to identify better the source of the error (the `ifcfg-eth0` file in this case).
- [ ] Do not allow the user to continue when a *fatal* error occurs.
- [ ] Provide a widget to display errors (?)
- [ ] Add tests and documentation.

### DIFF
--- a/src/lib/y2network/config_reader.rb
+++ b/src/lib/y2network/config_reader.rb
@@ -48,9 +48,10 @@ module Y2Network
 
     # Returns the configuration from the given backend
     #
-    # @return [Y2Network::Config] Network configuration
+    # @return [Y2Network::ReadingResult] Network configuration
+    # @raise NotImplementedError
     def config
-      Y2Network::Config.new
+      raise NotImplementedError
     end
   end
 end
--- a/src/lib/y2network/errors.rb
+++ b/src/lib/y2network/errors.rb
@@ -1,0 +1,137 @@
+# encoding: utf-8
+
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "yast"
+require "y2network/wicked/connection_config_reader"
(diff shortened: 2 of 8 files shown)


joseivanlopez:
I like your approach. And yes, it looks too much similar to `AutoinstIssues`. I vote for merging both in a general mechanism provided by *yast2*, e.g., `Yast2::Issues`. And I also agree  that the `Error` term is quite related to exceptions in ruby, so maybe *Issues* would be a better name. Moreover, I guess that some issues do not imply an error, for example, we could have warnings in the list of issues, right?

joseivanlopez:
And I am not sure about providing a widget to display errors. Maybe the presenter is enough. Each module might have its own way of displaying errors: popup, integrated in some part of the UI, etc.

teclator:
I share @joseivanlopez point of view.

imobachgs:
Thanks for the feedback! I have created https://github.com/yast/yast-yast2/pull/1156 to add the basic mechanism to yast2. So, please, have a look when you have time.

imobachgs:
Replaced with https://github.com/yast/yast-yast2/pull/1156 and #1206.

PR #1206, opened 2021-04-16
Handling reading errors (BOOTPROTO and STARTMODE)

imobachgs:
Reports problems with `BOOTMODE` and `STARTMODE` to the user using the new [error reporting mechanism](https://github.com/yast/yast-yast2/pull/1156).

Supersedes #1204.

<details>
<summary>Warn about invalid BOOTPROTO</summary>

network-bootproto-issue
</details>

Bug: [bsc#1181295](https://bugzilla.suse.com/show_bug.cgi?id=1181295)
Trello: https://trello.com/c/3dJHjgeQ/

## Pending

- [x] Adapt the `autoinst` reader
- [x] Adapt unit tests
- [x] Check all calls to read_config

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Thu Apr 22 12:48:27 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
-- Do not crash when the BOOTPROTO or STARTMODE ar missing or
+- Do not crash when the BOOTPROTO or STARTMODE are missing or
   invalid (bsc#1181295).
 - 4.4.6
 
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Apr 22 12:48:27 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not crash when the BOOTPROTO or STARTMODE ar missing or
+  invalid (bsc#1181295).
+- 4.4.6
+
+-------------------------------------------------------------------
 Thu Apr 22 12:22:47 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - bsc#1185106
--- a/src/lib/y2network/reading_result.rb
+++ b/src/lib/y2network/reading_result.rb
@@ -41,13 +41,5 @@ module Y2Network
     def issues?
       issues.any?
     end
-
-    # Generates a new result containing the updated config and the full list of issues
-    #
-    # @param other [ReadingResult] Result to merge
-    # @return [ReadingResult]
-    def merge(other)
-      ReadingResult.new(other.config, issues + other.issues)
-    end
   end
 end
--- a/src/lib/y2network/reading_result.rb
+++ b/src/lib/y2network/reading_result.rb
@@ -1,5 +1,3 @@
-# encoding: utf-8
-
 # Copyright (c) [2021] SUSE LLC
 #
 # All Rights Reserved.
--- a/src/lib/y2network/reading_result.rb
+++ b/src/lib/y2network/reading_result.rb
@@ -1,0 +1,55 @@
+# encoding: utf-8
+
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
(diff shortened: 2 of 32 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Fri Apr 16 12:10:09 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Do not crash when the BOOTPROTO or STARTMODE ar missing or
```
jreidinger, line 4: typo `ar`

src/lib/y2network/reading_result.rb
```diff
@@ -0,0 +1,53 @@
...
+    # @return [Boolean]
+    def issues?
+      issues.any?
+    end
+
+    # Generates a new result containing the updated config and the full list of issues
+    #
+    # @param other [ReadingResult] Result to merge
+    # @return [ReadingResult]
+    def merge(other)
```
jreidinger, line 49: I am still not sure if I get purpose of this method. So it reads updated config, but reports issue for both?

src/lib/y2network/wicked/connection_config_readers/base.rb
```diff
@@ -70,6 +74,33 @@ def connection_config
 
       private
 
+        def find_bootproto
+          bootproto = BootProtocol.from_name(file.bootproto.to_s)
+          return bootproto if bootproto
+
+          fallback = file.ipaddrs.empty? ? BootProtocol::DHCP : BootProtocol::STATIC
+          issue_location = "file:#{file.path}:BOOTPROTO"
```
jreidinger, line 82: I am a bit confused, why is here some string and not instance of Y2Issues::Location? I expect something like `issue_location = Y2Issues::Location.new("file", file.path, "BOOTPROTO")`

src/modules/Lan.rb
```diff
@@ -56,6 +57,7 @@ def main
       Yast.import "NetHwDetection"
       Yast.import "Host"
       Yast.import "IP"
+      Yast.import "HTML"
```
jreidinger, line 60: why it is needed?

src/modules/Lan.rb
```diff
@@ -717,11 +719,17 @@ def yast_config
...
     def read_config
-      system_config = Y2Network::Config.from(:wicked)
+      result = Y2Network::Config.from(:wicked)
+      system_config = result.config
       system_config.backend = NetworkService.cached_name
       Yast::Lan.add_config(:system, system_config)
       Yast::Lan.add_config(:yast, system_config.copy)
+      return true unless result.issues?
+
+      Y2Issues.report(result.issues) == :yes
```
jreidinger, line 732: hmm, I am not sure if there it should not reset back system and yast config? as no basically means that partial configuration is read and replace previous one, not?

src/lib/y2network/reading_result.rb
imobachgs: This method is not needed anymore, so I have just dropped it.

src/lib/y2network/wicked/connection_config_readers/base.rb
imobachgs: The Y2Issues API supports a "string" that [gets parsed](https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/y2issues/issue.rb#L55). However, I have noticed that if you pass the location directly, it will not work (I will fix it).

src/modules/Lan.rb
imobachgs: Just a left-over.

src/modules/Lan.rb
imobachgs: Yes, you are right. I have adapted the method to replace the configuration only when it is accepted by the user.

src/lib/y2network/wicked/connection_config_readers/base.rb
imobachgs: Here is the PR: https://github.com/yast/yast-yast2/pull/1157.

src/lib/y2network/wicked/config_reader.rb
```diff
@@ -214,6 +219,10 @@ def sysctl_config_file
         @sysctl_config_file.load
         @sysctl_config_file
       end
+
+      def issues_list
```
jreidinger, line 223: I am not sure if this does not cause nasty side effects. What e.g. happens if I call `config_reader.read` it reports some issues, User fix it on background and do read again so it calls again `config_reader.read` do I get also old issues list or not?

imobachgs: Actually, I would use a second reader. But it is true that it could be confusing, so I have adapted that code to avoid that side effect.

src/lib/y2network/wicked/config_reader.rb
```diff
@@ -65,8 +66,9 @@ def read
       # Reads the network interfaces
       #
       # @param config [Y2Network::Config] Initial configuration object
+      # @param _issues_list [Y2Issues::List] Issues list
```
jreidinger, line 69: I would document why it is here and maybe mention that in future it can be used if reading interface can also generate some issues.

jreidinger (approved):

PR #1207, opened 2021-04-16
Fixed hosts update when there is no IP address defined

teclator:
## Problem

During an AutoYaST installation yast2-network crashes when there is a connection configured with a **static** bootproto but without an **IP** address defined.

- https://bugzilla.suse.com/show_bug.cgi?id=1184883

## Solution

Do not crash in this case and try to return earlier when it does not make sense to apply changes.

## Test

- Adapted unit tested.
- Tested manually

### DIFF
--- a/src/lib/y2network/clients/auto.rb
+++ b/src/lib/y2network/clients/auto.rb
@@ -122,10 +122,12 @@ module Y2Network
       # Whatever is needed in /etc/hosts during AY installation
       def update_etc_hosts
         config = Yast::Lan.find_config(:yast)
-        return if !config
+        return if [nil, ""].include? config&.hostname&.static
 
-        static_connections = config.connections.select(&:static?)
+        static_connections = config.connections.select { |c| c.static? && c.ip&.address }
         static_connections.each do |connection|
+          next unless connection.hostname.to_s.empty?
+
           log.info("Updating /etc/hosts with" \
                    " #{connection.ip.address.address} #{config.hostname.static}")
           connection.hostname = config.hostname.static
--- a/src/lib/y2network/clients/auto.rb
+++ b/src/lib/y2network/clients/auto.rb
@@ -124,12 +124,11 @@ module Y2Network
         config = Yast::Lan.yast_config
         return if [nil, ""].include?(config&.hostname&.static)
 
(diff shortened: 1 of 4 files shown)


src/lib/y2network/clients/auto.rb
```diff
@@ -122,10 +122,12 @@ def merge_current_config?
       # Whatever is needed in /etc/hosts during AY installation
       def update_etc_hosts
         config = Yast::Lan.find_config(:yast)
-        return if !config
+        return if [nil, ""].include? config&.hostname&.static
 
-        static_connections = config.connections.select(&:static?)
+        static_connections = config.connections.select { |c| c.static? && c.ip&.address }
```
mchf, line 127: in my POV it should be

```&& c.ip```

as ip is the place where nil can get into at https://github.com/yast/yast-network/blob/master/src/lib/y2network/autoinst/interfaces_reader.rb#L128 when no ip address is defined in profile.

btw ```ip``` and ```ip.address``` is non nil up to here as it gets non empty default here https://github.com/yast/yast-network/blob/master/src/lib/y2network/connection_config/base.rb#L91

Comment (teclator) [reply src/lib/y2network/clients/auto.rb:127]: I just tried to avoid any other corner case. That is having **c.ip** without address.

Comment (mchf) [reply src/lib/y2network/clients/auto.rb:127]: > I just tried to avoid any other corner case. That is having **c.ip** without address.

yes ... i understand this part, but main point should be



> in my POV it should be
> 
> `&& c.ip`
> 
> as ip is the place where nil can get into at https://github.com/yast/yast-network/blob/master/src/lib/y2network/autoinst/interfaces_reader.rb#L128 when no ip address is defined in profile.
> 
because

```config.ip = load_ipaddr(interface_section)```

and

```
def load_ipaddr(section)
     return if section.ipaddr.empty?
     <--- snip --->
```

teclator: Take into account merged config, I can modify if you really want it, but will not make any difference at all, the current one is safer.

teclator: Updated

mchf (approved):
seems there were some misunderstanding but in the end I got what I requested, so lgtm ;-)

PR #1210, opened 2021-04-22
Fix hotplug check when there is no hardware info

teclator:
## Problem

When a virtual interface is edited it does not contain hardware information and the hotplug? check crashes miserably.

- https://bugzilla.suse.com/show_bug.cgi?id=1185106

## Solution

Do not crash when there is no hardware information

## Tests

Added unit test for the hotplug? interface method.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Thu Apr 22 09:12:38 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Do not crash when modifying a virtual interface and checking if
-  if it is hotplug (bsc#1184623, bsc#1185106)
+  it is hotplug (bsc#1184623, bsc#1185106)
 - 4.4.5
 
 -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


mchf:
NP: i'm thinking whether it makes sense to initialize the hardware to a kind of empty placeholder to avoid spreading checks like "if hardware" through the code ...

mchf (approved):
lgtm, thanks

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu Apr 22 09:12:38 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Do not crash when modifying a virtual interface and checking if
+  if it is hotplug (bsc#1184623, bsc#1185106)
```
mchf, line 5: if overflow ;-)

mchf: otherwise lgtm

teclator, line 5:
```suggestion
  it is hotplug (bsc#1184623, bsc#1185106)
```

PR #1211, opened 2021-04-22
Dropped setting set_default_route

mchf:
I've decided to remove it completely because:
1) we haven't written this setup since last huge network-ng refactoring
2) we do not have (and never had) support for configuring it via UI
3) no one complained that it haven't worked since last refactoring
4) it worked as "dak magic" anyway (see 2) and was marked as FIXME in the old code
5) this is related to UI only and doesn't affect (auto)installation
6) if it is not explicitly set in ifcfg file then default from /etc/sysconfig/network/dhcp which is provided by wicked is used

### DIFF
--- a/src/lib/y2network/dialogs/add_interface.rb
+++ b/src/lib/y2network/dialogs/add_interface.rb
@@ -51,24 +51,9 @@ module Y2Network
         )
       end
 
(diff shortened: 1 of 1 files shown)


teclator (approved):
LGTM

PR #1212, opened 2021-04-22
Handle invalid values

imobachgs:
Prevent YaST2 from crashing when `BOOTMODE` or `STARTMODE` are invalid. Unlike #1206, it does not use the [Y2Issues mechanism](https://github.com/yast/yast-yast2/pull/1156) because it is not available in `SLE-15-SP3` and backporting all those changes might be risky. So, by now, YaST2 just logs the problem and selects a suitable default value.

Bug: [bsc#1181295](https://bugzilla.suse.com/show_bug.cgi?id=1181295)
Trello: https://trello.com/c/3dJHjgeQ/


mchf (approved):
lgtm

PR #1214, opened 2021-04-26
Refactored dhcliet_set_hostname to use network-ng

mchf:


teclator:
In general look good to me, but what about using the new error reporting class introduced by @imo in #1206

imobachgs (approved):

mchf:
> In general look good to me, but what about using the new error reporting class introduced by @imo in #1206

sorry, but not now.  There are quite some many reports in the```Lan::Read``` which would deserve this, but it is not purpose of this PR.

PR #1215, opened 2021-04-26
Added option to disable network services during installation

teclator:
## Problem

Currently there is no way to disable the default network services during installation. Thus, if the user wants to use a different backend like **systemd-networkd** it will collide with the one enabled during installation.

- https://bugzilla.suse.com/show_bug.cgi?id=1183886

## Solution

- Add a link to the network proposal in order to disable the network services.
- Represented the absense of the network service using its own class. 
- As by now we allowed to write wicked configuration even without a network service we will maintain that behavior.

<details><summary>Screenshots</summary>

Screenshot_testing_2021-04-23_12:21:24
Screenshot_testing_2021-04-23_12:24:38
Screenshot_testing_2021-04-26_10:30:24

</details>


src/lib/network/clients/network_proposal.rb
```diff
@@ -120,23 +123,27 @@ def switch_backend_link
...
+        links.append(network_manager_link) if settings.network_manager_available?
       end
 
-      "<ul><li>#{backend_in_use % current_backend} (#{link_to_another_backend})</li></ul>"
+      "<ul><li>#{backend_in_use % current_backend} (#{links.join(", ")})</li></ul>"
     end
 
     def launch_network_configuration(args)
-      log.info "Launching network configuration"
+      log.info "Launching network configuration NEW"
```
imobachgs, line 146: NEW?

teclator: It is a leftover used for debugging :stuck_out_tongue_closed_eyes:

src/lib/y2network/autoinst_profile/networking_section.rb
```diff
@@ -113,6 +117,7 @@ def self.new_from_network(config)
         return result unless config
 
         result.managed = config.backend?(:network_manager)
+        result.backend = config.backend&.id.to_s if config.backend
```
imobachgs, line 120: We are using a string instead of a symbol, which is OK. But, please, add a test to make sure we are exporting the right value.

src/lib/y2network/none/config_reader.rb
```diff
@@ -0,0 +1,27 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+require "y2network/wicked/config_reader"
+
+module Y2Network
+  module None
+    # This class reads the current configuration from `/etc/sysconfig` files
```
imobachgs, line 23: Please, explain *why* it inherits from `Wicked::ConfigWriter`.

src/lib/y2network/none/config_writer.rb
```diff
@@ -0,0 +1,32 @@
...
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/wicked/config_writer"
+
+module Y2Network
+  module None
+    # By now this class configures Wicked (through sysconfig) according to a
```
imobachgs, line 24: Please, explain *why*.

imobachgs (commented):
In general, it looks good. Just some minor comments about documenting/testing.

imobachgs (approved):

imobachgs (approved):

PR #1216, opened 2021-04-26
Refactor Lan#dhcp_ntp_servers to use network-ng

mchf:


src/modules/Lan.rb
```diff
@@ -702,7 +700,20 @@ def dhcp_ntp_servers
...
+    end
+
+    # Returns hash of NTP servers
+    #
+    # Provides map with NTP servers obtained via any of dhcp aware interfaces
+    #
+    # @return [Hash<String, Array<String>] key is device name, value
+    #                                      is list of ntp servers obtained from the device
+    def ifaces_dhcp_ntp_servers
+      dhcp_ifaces = yast_config.connections.map { |c| c.interface if c.dhcp? }.compact
```
teclator, line 713: np:

```suggestion
      dhcp_ifaces = yast_config.connections.select(&:dhcp?).map(&:interface)
```

teclator (approved):
LGTM

src/modules/Lan.rb
mchf: as customer wishes ;-)

PR #1217, opened 2021-04-27
Kill'em all - R.I.P. LanItems

mchf:
[Trello card](https://trello.com/c/Qu0FpT5C/2263-3-osdistribution-p5-1180085-installed-over-wifi-but-wpasupplicant-ended-up-not-being-installed) which collects related to LanItems removal

## Problem

LanItems has long time been partially replaced by network-ng stuff. However a few pieces were still alive.

## Solution

Cleanup was done in parts. See above trello card for references to other PRs. This PR is just a "big red button" which removes the module LanItems.rb itself. Some research for outter dependencies were done and LanItems is not used much, however something might appear later.


teclator:
Apparently only the [yast-caasp](https://github.com/yast/yast-caasp) repository is still using LanItems. We should adapt it at least in master although probably it is not used anymore...

teclator:
Definitely something to celebrate :tada:  :partying_face: :tada:

mchf:
> Apparently only the [yast-caasp](https://github.com/yast/yast-caasp) repository is still using LanItems. We should adapt it at least in master although probably it is not used anymore...

agree ... at least I was not able to find anything else. And as the LanItems has not been fed for long time, the usage itself cannot be serious. I plan to adapt yast-caasp as a next step.
> 
> Maybe you could add some description adding a link to the Trello Card and a good summary of all the Pull Requests.

ok will try ;-)

teclator (approved):
LGTM

teclator:
Apparently yast-caasp is not the only one still using LanItems.

@mchf see https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:F/yast2-installation/standard/x86_64

So, we need to adapt it

teclator (changes requested):
The dependencies should be dropped before merge this PR.

package/yast2-network.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Tue Apr 27 10:44:05 UTC 2021 - Michal Filka <mfilka@suse.com>
+
+- Dropped LanItems, replaced by network-ng
```
teclator, line 4: Miss bug reference just use this one https://bugzilla.suse.com/show_bug.cgi?id=1185338

```suggestion
- Dropped LanItems, replaced by network-ng (bsc#1185338)
```

teclator (approved):
LGTM

PR #1218, opened 2021-04-28
Omit networks without an essid

teclator:
## Problem

The dialog for selecting a wireless network crashes when there is some hidden network (empty essid).

- https://bugzilla.suse.com/show_bug.cgi?id=1185372

## Solution

Skip networks without an essid from the table list.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
 Wed Apr 28 09:18:27 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
-- Omit hidden networks from the list of wireless network to be
+- Omit hidden networks from the list of wireless networks to be
   selected preventing the dialog to crash (bsc#1185372)
 - 4.3.67
 
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Apr 28 09:18:27 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Omit hidden networks from the list of wireless network to be
+  selected preventing the dialog to crash (bsc#1185372)
+- 4.3.67
+
+-------------------------------------------------------------------
 Thu Apr 22 10:15:42 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
 
 - Do not crash when the BOOTPROTO or STARTMODE ar missing or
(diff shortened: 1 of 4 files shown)


src/lib/y2network/dialogs/wireless_networks.rb
```diff
@@ -121,6 +121,7 @@ def find_networks(cache = true)
           _("Scanning for wireless networks..."), headline: _("Scanning network")
         ) do
           found_networks = Y2Network::WirelessNetwork.all(@builder.interface.name, cache: cache)
+          found_networks.delete_if { |n| n.essid.to_s.empty? }
```
imobachgs, line 124: What about using something like this?

```ruby
found_networks = Y2Network::WirelessNetwork.all(@builder.interface.name, cache: cache)
  .reject { |n| n.essid.to_s.empty? }
```

test/y2network/dialogs/wireless_networks_test.rb
```diff
@@ -74,5 +87,6 @@
         expect(subject.run).to eq(nil)
       end
     end
+
```
imobachgs, line 90: I would omit this blank line.

src/lib/y2network/dialogs/wireless_networks.rb
teclator: fine with it

test/y2network/dialogs/wireless_networks_test.rb
teclator: Yep, also fine

teclator:
And maybe makes sense to return empty in case of nil here https://github.com/yast/yast-network/blob/166c75353147a0c3a58612113a8fa8edca3ac2be/src/lib/y2network/wireless_scanner.rb#L182

mchf:
i believe that this solution works. Only question is whether we want to somehow express that we know there are hidden networks explicitly.

in my POV it is not needed as long as we offer possibility to enter ESSID manually. When someone wants to connect to such network he has to know what to do.

teclator:
Yep, that is basically the main discussion, if we would like to show that in some way or not, I think that for the time being just skipping them is enough.

imobachgs:
> Yep, that is basically the main discussion, if we would like to show that in some way or not, I think that for the time being just skipping them is enough.

Let's keep it simple. After all, you can enter the ESSID manually.

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Apr 28 09:18:27 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Omit hidden networks from the list of wireless network to be
```
mchf, line 4: ```wireless networks```

mchf (approved):
lgtm

package/yast2-network.changes
teclator:
```suggestion
- Omit hidden networks from the list of wireless networks to be
```

PR #1221, opened 2021-05-12
Removed unused methods and imports.

mchf:


src/include/network/services/dns.rb
```diff
@@ -645,10 +645,6 @@ def storeHostnameGlobal(_key, event)
       nil
     end
 
-    def ReallyAbortInst
-      Popup.ConfirmAbort(:incomplete)
```
teclator, line 649: It is used here https://github.com/yast/yast-firstboot/blob/master/src/lib/y2firstboot/clients/hostname.rb#L97

mchf: so it is time to move it there, isn't it?

mchf:
respective change is in TW already

teclator (approved):
With [yast2-firstboot-4.4.2](https://build.opensuse.org/request/show/903243) It looks safer to merge this PR.

PR #1222, opened 2021-05-12
InterfaceFile: Write collections in order

teclator:
## Problem

When a connection has **multiple** IP addresses, the order in which them will be written down to the file is unknown and the primary IP address could be an alias one.

- https://bugzilla.suse.com/show_bug.cgi?id=1185967

## Solution

Ensure that the **InterfaceFile** collection variables are sort when writing them down.

With the fix the collections are written in order:

```
IPADDR='192.168.0.190/24'
IPADDR_1='192.168.1.101/24'
IPADDR_2='192.168.0.181/24'
BOOTPROTO='static'
STARTMODE='auto'
LABEL_1='test'
LABEL_2='te'
```

**Note:** each set is usually written together like:

```
IPADDR='192.168.0.190/24'
BOOTPROTO='static'
STARTMODE='auto'
IPADDR_1='192.168.1.101/24'
LABEL_1='test'
IPADDR_2='192.168.0.181/24'
LABEL_2='te'
```

But for the bug should be enough with this fix.

## Tests

- Tested manually

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed May 12 16:46:00 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Write IP addresses in order preventing an alias to set the
+  primary IP address (bsc#1185967)
+- 4.2.99
+
+-------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


imobachgs (approved):

teclator:
SR: https://build.suse.de/request/show/241221

PR #1223, opened 2021-05-12
Merge 'SLE-15-SP2' into 'SLE-15-SP3'

teclator:
It merges #1222

PR #1224, opened 2021-05-13
Merge SLE-15-SP3 into 'master'

teclator:
It merges #1222 & #1223 into master.

### DIFF
--- a/src/lib/cfa/interface_file.rb
+++ b/src/lib/cfa/interface_file.rb
@@ -608,7 +608,7 @@ module CFA
     # @see #clean_collection
     def write_collection(key, values)
       clean_collection(key)
-      values.each do |suffix, value|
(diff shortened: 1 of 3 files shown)


src/lib/cfa/interface_file.rb
```diff
@@ -608,7 +608,7 @@ def value_as_ipaddr(value)
     # @see #clean_collection
     def write_collection(key, values)
       clean_collection(key)
-      values.each do |suffix, value|
+      values.sort_by { |s, _v| (s == :default) ? "" : s.to_s }.each do |suffix, value|
```
dgdavid, line 611: NP: s/_v/_

dgdavid (approved):
LGTM, just a tiny NP :)

src/lib/cfa/interface_file.rb
teclator: I will leave it as it is as in the other branches it is already written in this way.

PR #1225, opened 2021-05-24
Fix interfaces collection array operations (bsc#1186082)

teclator:
## Problem

We have overridden the **eql?** method in many places but not the **hash** method which is a problem in some Arrays operation where it could give unexpected values depending on the comparison method used.

- https://bugzilla.suse.com/show_bug.cgi?id=1186082

## Solution

- Override the **hash** method in accordance to the **eql?** logic.
- Add / fix some interfaces collection operators.

This will fix the reported error but we should override the **hash** method in other places where we have also made the **eql?** method an alias of the **==** method.

### DIFF
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -95,6 +95,11 @@ module Y2Network
     # https://ruby-doc.org/core-2.3.3/Object.html#method-i-eql-3F
     alias_method :eql?, :==
 
+    # Used by Array or Hash in order to compare equality of elements (bsc#1186082)
+    def hash
+      name.hash
+    end
+
     # Complete configuration of the interface
     #
     # @return [Hash<String, String>] option, value hash map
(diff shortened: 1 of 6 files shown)


imobachgs (approved):

src/lib/y2network/interface.rb
```diff
@@ -95,6 +95,11 @@ def ==(other)
     # https://ruby-doc.org/core-2.3.3/Object.html#method-i-eql-3F
     alias_method :eql?, :==
 
+    # Used by Array or Hash in order to compare equality of elements (bsc#1186082)
+    def hash
+      name.hash
```
joseivanlopez, line 100:
~~~
name = "foo"
interface = Interface.new(name)

interface.hash == name.hash #=> true
~~~

I would use the hash of the class too to make sure we don't have false positives with another kind of objects.

teclator: In general I agree, in this case the **==** method do not contain the class check neither. It would be hard to combine that kinds of objects but it is true that is open to that kind of errors.

teclator: Yep, although it is strange to combine that kind of objects, it is open to that errors and the **==** already contains that check, so will take that into account in next iteration when implemented generally.

PR #1228, opened 2021-05-25
Adopt equatable

teclator:
## Problem

We have overridden the **eql?** method in many places but not the **hash** method which should be in pair. For fixing that @joseivanlopez has provided a new mixin which we could use to avoid possible errors specially if we work with hashes and arrays.

## Solution

Adopt the **Equatable** mixin suggested by Ivan in yast/yast-yast2#1165


src/lib/y2network/hwinfo.rb
```diff
@@ -278,7 +278,11 @@ def used_mac
     # @param other [Hwinfo] Object to compare with
     # @return [Boolean]
     def ==(other)
```
joseivanlopez, line 280: Why isn't equatable used here?

joseivanlopez: I guess we can use: `eql_attr :hwinfo`

teclator: I discarded this case because it uses hwinfo.compact for the comparison, so, we could use `eql_attr :hwinfo` but with
```
def eql_hash
  super.compact
end
```

joseivanlopez: Sure, I would go with that approach.

teclator: Adapted

joseivanlopez (approved):
LGTM

PR #1229, opened 2021-05-25
Allow to set IP forwarding defaults through the control file (bsc#1186280)

teclator:
## Problem

During the installation, we read the current network configuration which includes the **sysctl IP forwarding** settings.
Therefore, if the configuration is modified and written, it will include those settings.

In a common case it happens when we call the **dhcp_setup** client which is very early during installation as you can see in the control file below:

- https://github.com/yast/skelcd-control-Kubic/blob/master/control/control.Kubic.xml#L668

We can adapt the **dhcp setup** writing only the connection changes (ifcfg files) which means that the **/etc/sysctl.d/70-yast.conf** will not be written yet and also it will not be copied to the target system if the network configuration is not modified manually.

That could be fine in most of the cases, and specially in Kubic if it is not expected to touch the network at all but the proposal would be showing that the IP forwarding is off. And apparently it needs to have it enabled by default which means this only change is not enough.

Screenshot_Kubic_2021-05-25_21:54:34

- https://bugzilla.suse.com/show_bug.cgi?id=1186280

## Solution

- The setup_dhcp client will write only what needed (connections).
- The network proposal settings can read and apply the defaults from the control file being written at the end of the installation in case it is needed.

## Example

```xml
    <!-- Example -->
    <network>
        <ipv4_forward>true</ipv4_forward>
    </network>

    <system_roles config:type="list">
      <system_role>
        <id>kubic_admin_role</id>

        <network>
          <ipv6_forward>true</ipv6_forward>
        </network>

```

KubicDefaults

### DIFF
--- a/src/lib/y2network/proposal_settings.rb
+++ b/src/lib/y2network/proposal_settings.rb
@@ -33,6 +33,11 @@ module Y2Network
     # @return [Boolean] network service to be used after the installation
     attr_accessor :selected_backend
     attr_accessor :virt_bridge_proposal
+    attr_accessor :ipv4_forwarding
+    attr_accessor :ipv6_forwarding
+    attr_accessor :defaults_applied
+
+    DEFAULTS = [:ipv4_forwarding, :ipv6_forwarding].freeze
 
     # Constructor
     def initialize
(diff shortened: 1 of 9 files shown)


imobachgs:
I guess the version bump is missing because you plan to release it along with other changes, right?

teclator:
Yep, I'm still working on it, do not plan to merge yet. My fault for not creating it as a draft or with a WIP flag ;)

schubi2:
A simple solution for a not simple problem :-)

src/lib/y2network/proposal_settings.rb
```diff
@@ -164,6 +190,34 @@ def create_instance
...
+
+    # Reads a feature from a given hash and assign it to the corresponding object attribute
+    #
+    # @param feature [String, Symbol] feature name to be read
+    # @param to [String, Symbol] attribute name where to store the feature value
+    # @param source [Hash] from where to read the feature
+    def load_feature(feature, to, source: network_section)
+      value = source[feature.to_s]
+      public_send("#{to}=", value) unless value.nil?
+    end
```
dgdavid, line 210: NP: What about moving the condition to the #load_feature method?

```suggestion
      source.keys.each { |k| load_feature(k, k, source: source) }
    end

    # Reads a feature from a given hash and assign it to the corresponding object attribute
    #
    # @param feature [String, Symbol] feature name to be read
    # @param to [String, Symbol] attribute name where to store the feature value
    # @param source [Hash] from where to read the feature
    def load_feature(feature, to, source: network_section)
      return unless respond_to?("#{to}=")
      
      value = source[feature.to_s]
      public_send("#{to}=", value) unless value.nil?
    end
```

dgdavid (approved):
Code-wise LGTM. But maybe you want to wait for another review from someone with more expertise in the subject.

src/lib/y2network/proposal_settings.rb
teclator: And reading the code maybe makes sense to set the value always. I mean, in case you change the role and it has different values where some of the attributes is not set, then we should not modify it.

teclator:
@dgdavid adapted, see:

- https://github.com/yast/yast-network/pull/1229/commits/b70e2417f93241f272a0d753c7db4b213313528c
- https://github.com/yast/yast-network/pull/1229/commits/cb03266876f4917e31f6c9a479af601ed50e8edb

teclator:
@imobachgs maybe you want to check it too.

dgdavid (approved):

PR #1230, opened 2021-05-26
Use networkmanager lense

mchf:
See [Trello](https://trello.com/c/un8hpdfr/2469-3-ostumbleweed-p3-1185524-changing-to-networkmanager-via-yast-network-throws-an-internal-error) for details

## Problem ##

Installer reported an internal error at the end of installation when writing NetworkManager's configuration to the target.

## Solution

We used to use ```Desktop.lns``` instead of ```NetworkManager.lns``` because of a bug in the later one. As the bug was fixed [here](https://github.com/hercules-team/augeas/pull/723) we can switch to fixed NetworkManager.lns which solves the issue

## TBD
- ~~reference to augeas' pr~~
- version dependency on augeas lenses package with fixed lenses (?)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Mon May 31 17:50:42 UTC 2021 - Michal Filka <mfilka@suse.com>
+
+- bnc#1185524
+  - do not crash at the end of installation when storing wifi
+    configuration for NetworkManager at the target 
+- 4.4.13
+
+-------------------------------------------------------------------
 Fri May 28 13:51:57 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Allow to modify the network IP forwarding configuration defining
--- a/package/yast2-network.spec
+++ b/package/yast2-network.spec
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.4.12
+Version:        4.4.13
 Release:        0
 Summary:        YaST2 - Network Configuration
 License:        GPL-2.0-only
--- a/src/lib/cfa/nm_connection.rb
+++ b/src/lib/cfa/nm_connection.rb
@@ -70,9 +70,7 @@ module CFA
     # @param path [String] File path
     # @param file_handler [.read, .write] Object to read/write the file.
     def initialize(path, file_handler: nil)
-      # FIXME: The Networkmanager lense writes the values surrounded by double
-      # quotes which is not valid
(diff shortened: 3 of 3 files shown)


teclator:
@mchf great!, I think that adding the dependency makes sense.

mchf:
it seems that they do not increase version number for suse patches. I can add dependency for augeas 1.12.0 but it will not distinguish patched version from non-patched

mchf:
@teclator I got confirmation that version number will not be increased on suse side (follows upstream regardless of suse patches). So, what do you prefer now? Adding dependency on augeas 1.12.0 or not adding any dependency or ... ?

teclator:
> @teclator I got confirmation that version number will not be increased on suse side (follows upstream regardless of suse patches). So, what do you prefer now? Adding dependency on augeas 1.12.0 or not adding any dependency or ... ?

I guess that adding the dependency will be useless in this case as we need to ensure we use the NetworkManager lense with your patch.

mchf:
> > @teclator I got confirmation that version number will not be increased on suse side (follows upstream regardless of suse patches). So, what do you prefer now? Adding dependency on augeas 1.12.0 or not adding any dependency or ... ?
> 
> I guess that adding the dependency will be useless in this case as we need to ensure we use the NetworkManager lense with your patch.

I agree. So, I currently see two (bad) possibilities
1) no (version) dependency
2) wait for new augeas upstream release which should potentially contain the patch

I slightly prefer (1) as we cannot know if they will put new augeas into SLE-SPx which will get forked from master one day. And ... we don't know if the patch will get merged to augeas any time soon.

teclator:
> > > @teclator I got confirmation that version number will not be increased on suse side (follows upstream regardless of suse patches). So, what do you prefer now? Adding dependency on augeas 1.12.0 or not adding any dependency or ... ?
> > 
> > 
> > I guess that adding the dependency will be useless in this case as we need to ensure we use the NetworkManager lense with your patch.
> 
> I agree. So, I currently see two (bad) possibilities
> 
>     1. no (version) dependency
> 
>     2. wait for new augeas upstream release which should potentially contain the patch
> 
> 
> I slightly prefer (1) as we cannot know if they will put new augeas into SLE-SPx which will get forked from master one day. And ... we don't know if the patch will get merged to augeas any time soon.

For TW looks ok with (1) once the augeas patch that you already submitted has been accepted. Would be nice to have a dependency but probably SLE-15-SP4 which will be based on master will use the same.. so fine with (1) by now

teclator (approved):
LGTM

PR #1240, opened 2021-06-25
Adapted to the new API of Y2Issues

ancorgs:
Depends on https://github.com/yast/yast-yast2/pull/1178, tests will not pass until that is merged

PR #1242, opened 2021-07-01
Do not crash with internal error on corrupted sysctl file

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1185936

## Problem ##

When any of sysctl files accessed by yast2-network was corrupted, yast2 lan crashed with internal error

## Solution ##

Implemented augeas exception handling together with using new issues reporting mechanism for error report.


src/modules/Lan.rb
```diff
@@ -780,8 +780,12 @@ def read_config(report: true)
     #   parameter is given then all changes will be written
     #
     # @see Y2Network::ConfigWriter
-    def write_config(only: nil)
-      yast_config.write(original: system_config, only: only)
+    def write_config(only: nil, report: true)
+      result = yast_config.write(original: system_config, only: only)
+      if result&.issues? && report
+        return false unless Y2Issues.report(result.issues) == :yes
```
imobachgs, line 786: Beware that this API changed recently and the report method now returns a boolean. See https://github.com/yast/yast-network/pull/1240/files#diff-c536d97bfa2e800b076011ec2976dae8b9c77b83fec29c7a1741554d1f8750c0.

imobachgs (changes requested):
In general, it looks good (the approach is pretty similar to what do while reading the configuration). I wonder whether we should unify *Result classes.

mchf:
> In general, it looks good (the approach is pretty similar to what do while reading the configuration). I wonder whether we should unify *Result classes.

to make it similar to the read case was an intention ;-)  I was thinking of unifying result classes as well ... as these classes are very simple ... and in fact the same.

imobachgs (approved):

PR #1245, opened 2021-07-30
Fix write of device routes (bsc#1188908)

teclator:
## Problem

It was reported [here](https://lists.opensuse.org/archives/list/yast-devel@lists.opensuse.org/thread/S2AZBYIH2QNNB42BVPM5QWJWL6VYB4WH/) that YaST does not supproted the write of routes per interface but that was already supported. 

The problem is that we broke the selection of routes to be written when we adopted Equatable (https://github.com/yast/yast-network/pull/1228)

It happens when we have read the configuration and the interfaces are detected as phisical or virtual ones, but later we compare it with a Y2NetworkInterface, so the classes are different and the comparison fails.

```ruby
Y2Network::Interface.new("eth0") == Y2Network::PhysicalInterface.new("eth0") ## => false

```

- https://bugzilla.opensuse.org/show_bug.cgi?id=1188908

## Solution

By now we will avoid the use of **Equatable** for **Y2NetworkInterface**.

### DIFF
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -83,6 +83,9 @@ module Y2Network
 
     # Determines whether two interfaces are equal
     #
+    # @note although it is preferable to use Yast2::Equatable it uses the class
+    #   hash for comparing objects and it will fail when comparing with
+    #   subclasses objects (bsc#1188908)
     # @param other [Interface] Interface to compare with
     # @return [Boolean]
     def ==(other)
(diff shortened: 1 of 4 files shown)


src/lib/y2network/interface.rb
```diff
@@ -83,7 +81,24 @@ def initialize(name, type: InterfaceType::ETHERNET)
       @renaming_mechanism = :none
     end
 
-    eql_attr :name
+    # Determines whether two interfaces are equal
+    #
+    # @param other [Interface] Interface to compare with
```
jreidinger, line 86: write note here why equatable cannot be used here

jreidinger (approved):

PR #1246, opened 2021-09-06
Read default NTP servers from control file

teclator:
## Problem

During the installation, the default **NTP** servers are inferred from the product name. Depending on it the installation will use a list of **openSUSE** or **SUSE** ntp pool servers.

Currently the product detection is broken and **openSUSE** products are using the **SUSE** servers but there is also no option to define the default ones and the list is currently hard coded.

- https://bugzilla.suse.com/show_bug.cgi?id=1180699
- https://trello.com/c/HtBLfvR0/2280-3-ostumbleweed-p5-1180699-microos-defaults-to-using-the-suse-ntp-pool

## Solution

Read the default servers from the control file in case of defined using the hard coded default ones only in case of missed definition. (see related https://github.com/yast/yast-country/pull/282, https://github.com/yast/skelcd-control-openSUSE/pull/223, https://github.com/yast/yast-installation-control/pull/112)

**Note:** An empty list can be used allowing the installation to not pre-fill the list of **NTP** servers, but it does not influence the RPM selection.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Tue Sep  7 07:41:52 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- Add support for reading the default NTP servers from the control
+  file in case of defined (bsc#1180699)
+- 4.4.23
+
+-------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


imobachgs (approved):

imobachgs (approved):

dgdavid (approved):
LGTM

PR #1247, opened 2021-09-22
Display user defined device name in the overview

mchf:
https://bugzilla.suse.com/show_bug.cgi?id=1190645

## Problem ##

we simply forgot of it during refactoring to network-ng

## Solution ##

Just use what was read from an ifcfg file(s)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Sep 22 08:46:16 UTC 2021 - Michal Filka <mfilka@suse.com>
+
+- bnc#1190645
+  - display user defined device name in the devices overview
+- 4.3.74
+
+-------------------------------------------------------------------
 Tue Jul 20 08:47:30 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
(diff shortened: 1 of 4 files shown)


teclator (commented):
In general LGTM, but would be nice to be covered by a unit test.

teclator (approved):
LGTM

kobliha:
It might be nice to have a test for the future ;) so we do not forget again...

test/y2network/widgets/interfaces_table_test.rb
```diff
@@ -133,5 +133,18 @@
         subject.handle
       end
     end
+
+    context "when the device is named by user" do
```
teclator, line 137: I would expect this context as part of https://github.com/yast/yast-network/blob/77c7669cf136fdccffb2443354d999208431111b/test/y2network/widgets/interfaces_table_test.rb#L121

teclator:
```ruby
context "when the device is configured" do 
  context "and it is named by the user" do
```

teclator (approved):
LGTM

PR #1251, opened 2021-09-23
bsc#1190739: consider <aliases> as case insensitive

imobachgs:
According to [bsc#1190739](https://bugzilla.suse.com/show_bug.cgi?id=1190739), the `<aliases>` section is case sensitive. I guess it is a reminiscence of exporting the content of the `sysconfig` file.

This PR adds a proper `AliasSection` to handle these sections. From now on, the following examples are equivalent:

```xml
<aliases>
  <alias0>
    <IPADDR>192.168.122.1</IPADDR>
    <PREFIXLEN>24</PREFIXLEN>
  </alias0>
</aliases>

<!-- the same than: -->
<aliases>
  <alias0>
    <ipaddr>192.168.122.1</ipaddr>
    <prefixlen>24</prefixlen>
  </alias0>
</aliases>
```

The XML representation of the aliases list is rather weird: instead of being a list, it is a hash where the keys (`alias0`, `alias1`, etc.) are ignored. Actually, this section is [not validated](https://github.com/yast/yast-network/blob/f1c311fc8d6af163ac4f8c92813a98e6a1693909/src/autoyast-rnc/networking.rnc#L70) at all. In order to simplify things,
`InterfaceSection#aliases` is now an array of `AliasSection` objects instead of a hash. However, the XML representation is still the same, but it is easier to reason about it when dealing with the section objects.

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Thu Sep 23 10:01:21 UTC 2021 - Imobach Gonzalez Sosa <igonzalezsosa@suse.com>
+
+- Consider aliases sections as case insensitive (bsc#1190739).
+- 4.2.104
+
+-------------------------------------------------------------------
 Wed Sep 22 08:46:16 UTC 2021 - Michal Filka <mfilka@suse.com>
 
 - bnc#1190645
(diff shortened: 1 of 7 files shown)


teclator (approved):
LGTM

PR #1254, opened 2021-09-27
Fix interfaces table with not configured ifaces (fix for #1247)

teclator:
## Problem

When the interfaces table contains a not configured interface it crashes trying to call the related connection interface. (introduced by #1247)

Reported by openQA https://openqa.opensuse.org/tests/1940439#step/yast2_lan/14

## Solution

Just use the safe navigation operator **(&)**  preventing to call `description` over a **Nil** object.

### DIFF
--- a/test/y2network/widgets/interfaces_table_test.rb
+++ b/test/y2network/widgets/interfaces_table_test.rb
@@ -83,7 +83,7 @@ describe Y2Network::Widgets::InterfacesTable do
       end
     end
 
-    context "ant the device is not configured" do
+    context "and the device is not configured" do
       let(:connections) { Y2Network::ConnectionConfigsCollection.new([]) }
 
       it "shows the hwinfo interface description if present or the interface name if not" do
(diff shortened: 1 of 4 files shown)


test/y2network/widgets/interfaces_table_test.rb
```diff
@@ -61,8 +62,40 @@
...
+          end
+        end
+
+        it "includes its custom device name" do
+          expect(subject.items).to include(a_collection_including(/Custom description/))
+        end
+      end
+    end
+
+    context "ant the device is not configured" do
```
jreidinger, line 86: typo "ant" :ant:

teclator:
```suggestion
    context "and the device is not configured" do
```

jreidinger:
```suggestion
    context "and the device is not configured" do
```

jreidinger (approved):

mchf (approved):
lgtm

PR #1256, opened 2021-09-27
Do not initialize the connection description

teclator:
## Problem

By default we initialize the description to nil, that will affects any new connection as we are initializing the description to an empty string (as we can see here - https://openqa.opensuse.org/tests/1940430#step/yast2_lan_device_settings/41) but it is not the case of read configurations.

Related to #1256 #1247 and https://bugzilla.suse.com/show_bug.cgi?id=1190645
NoDescription

## Solution

Do not initialize the description when a new connection is added.

EmptyVlanName

## Note

No bump version is introduced as #1255 has not been submitted yet, but could add a changelog if needed

### DIFF
--- a/src/lib/y2network/widgets/interfaces_table.rb
+++ b/src/lib/y2network/widgets/interfaces_table.rb
@@ -145,7 +145,7 @@ module Y2Network
       # Returns the connection description if given or the interface friendly name if not
       #
       # @param interface [Interface] Network interface
-      # @param conn [ConnectionConfig::Base] Connection configuration
+      # @param conn [ConnectionConfig::Base, nil] Connection configuration
       # @return [String] Connection description if given or the friendly name for the interface (
       #   description or name) if not
       def description_for(interface, conn)
--- a/src/lib/y2network/widgets/interfaces_table.rb
+++ b/src/lib/y2network/widgets/interfaces_table.rb
@@ -110,7 +110,7 @@ module Y2Network
           # first is (item) ID in table
           interface.name,
           # if user named the connection explicitly, it will have precendence
-          conn&.description || friendly_name(interface),
+          conn&.description.to_s.empty? ? friendly_name(conn, interface) : conn.description
           interface_protocol(conn),
           interface.name,
           note(interface, conn)
--- a/src/lib/y2network/widgets/interfaces_table.rb
+++ b/src/lib/y2network/widgets/interfaces_table.rb
@@ -109,8 +109,7 @@ module Y2Network
         [
           # first is (item) ID in table
           interface.name,
-          # if user named the connection explicitly, it will have precendence
-          conn&.description.to_s.empty? ? friendly_name(conn, interface) : conn.description
+          description_for(interface, conn),
           interface_protocol(conn),
           interface.name,
           note(interface, conn)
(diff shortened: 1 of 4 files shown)


teclator:
@mchf feel free to reopen it if the fix is good enough or create a new one ;)

src/lib/y2network/widgets/interfaces_table.rb
```diff
@@ -110,7 +110,7 @@ def interface_item(interface)
           # first is (item) ID in table
           interface.name,
           # if user named the connection explicitly, it will have precendence
-          conn&.description || friendly_name(interface),
+          conn&.description.to_s.empty? ? friendly_name(conn, interface) : conn.description
```
jreidinger, line 113: well, I think it would be nice to move all that to that friendly_name method and maybe call it display_description or something like this?

teclator: yep, sounds reasonable to me, already tried to move it, in fact that conn parameter is a leftover of the attempt which needs to be fixed :), just was not sure about the method description and passing both parameters so decided to not so far

jreidinger (approved):

src/lib/y2network/widgets/interfaces_table.rb
```diff
@@ -143,11 +142,15 @@ def create_description
...
+      # Returns the connection description if given or the interface friendly name if not
       #
       # @param interface [Interface] Network interface
-      # @return [String] Friendly name for the interface (description or name)
-      def friendly_name(interface)
+      # @param conn [ConnectionConfig::Base] Connection configuration
+      # @return [String] Connection description if given or the friendly name for the interface (
+      #   description or name) if not
+      def description_for(interface, conn)
+        return conn.description unless conn&.description.to_s.empty?
```
mchf, line 152: in theory we can face a problem that someone might use empty name intentionally not to display anything in the overview .... but it sounds a little bit pathologically ;-)

So what about some test(s)?

src/lib/y2network/widgets/interfaces_table.rb
```diff
@@ -143,11 +142,15 @@ def create_description
...
-      # Returns a friendly name for a given interface
+      # Returns the connection description if given or the interface friendly name if not
       #
       # @param interface [Interface] Network interface
-      # @return [String] Friendly name for the interface (description or name)
-      def friendly_name(interface)
+      # @param conn [ConnectionConfig::Base] Connection configuration
+      # @return [String] Connection description if given or the friendly name for the interface (
+      #   description or name) if not
+      def description_for(interface, conn)
```
mchf, line 151: ```interface``` is a bit unnecessary ... we should be able to get it from the connection config. Or vice versa. Positive of such approach would be avoiding theoretical mismatch between interface and connection config.

src/lib/y2network/widgets/interfaces_table.rb
teclator: It is not possible from the config file and also not from the **GUI** after the last changes, but wanted to prevent it was initialized by any builder or step because of a leftover.

The not set should cover the case but I can initialize the connection description to empty if you want :)

src/lib/y2network/widgets/interfaces_table.rb
teclator: Yeah, I already thought about using the connection interface but as the interface was there and in fact the connection was find based on the in the interface name just passed it by parameter. It is not a public API and preferred to be conservative in this case according to last issues :stuck_out_tongue_winking_eye:

imobachgs (approved):

PR #1259, opened 2021-10-05
User NetworkManager lense in SLE-15-SP3 (backported #1230)

teclator:
## Problem

The use of NetworkManager lense was skip because of an error in the lense but as @mchf already provided a fix we should use it in **SLE-15-SP3** instead of the Desktop one as it has other problems. (Already fixed in TW by #1230)

- https://trello.com/c/N782sR6m/2617-osdistribution-p5-1187512-yast-fails-to-convert-wi-fi-connections-to-ones-that-works-in-networkmanager
- https://bugzilla.suse.com/show_bug.cgi?id=1187512

## Solution

Backport #1230 once the SR with the patch for augeas has been already created (https://build.suse.de/request/show/255648)

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Oct  5 09:50:42 UTC 2021 - Michal Filka <mfilka@suse.com>
+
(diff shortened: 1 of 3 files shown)


mchf (approved):
lgtm

PR #1261, opened 2021-10-13
Fix routing

teclator:
## Problem

When the routing client is call yast crashes reading the routing configuration because of a change in the API introduced by (https://github.com/yast/yast-network/pull/1206)

- https://bugzilla.suse.com/show_bug.cgi?id=1191589

## Solution

Adapt the code for obtaining the configuration in case there are no issues reported.


mchf (approved):
"nice" it took so long to catch it.
lgtm

PR #1262, opened 2021-11-03
Do not crash when checking if a virtual interface is connected

teclator:
## Problem

When a configured interface does not have hardware information the connected? method crashes miserably.

- https://bugzilla.suse.com/show_bug.cgi?id=1192270

### DIFF
--- a/src/lib/y2network/interface.rb
+++ b/src/lib/y2network/interface.rb
@@ -75,7 +75,7 @@ module Y2Network
       @renaming_mechanism = :none
     end
 
-    # Whether the interface is connected or not based con hardware information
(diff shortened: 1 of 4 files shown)


src/lib/y2network/interface.rb
```diff
@@ -78,6 +75,22 @@ def initialize(name, type: InterfaceType::ETHERNET)
       @renaming_mechanism = :none
     end
 
+    # Whether the interface is connected or not based con hardware information
```
joseivanlopez, line 78: Typo: con

joseivanlopez (approved):
LGTM

PR #1265, opened 2021-11-04
Use the interface name as the description

teclator:
## Problem

With network-ng the interface sections' device element was marked as
deprecated and the use of name has precedence over it.

But when both elements are given the profile is probably an old one.

- https://bugzilla.suse.com/show_bug.cgi?id=1192270

## Solution

In that case the name will be used as the description (bsc#1192270).

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,14 @@
 -------------------------------------------------------------------
+Thu Nov  4 15:02:24 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- AutoYaST
+  - When the interface section contains the deprecated device
+    element and also the name element, then use the last one as
+    the interface description. (bsc#1192270)
+  - Add the description element to the interface section
+- 4.2.107
+
+-------------------------------------------------------------------
 Wed Nov  3 22:19:45 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - Do not crash when checking if a virtual interface is connected
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,10 +2,10 @@
 Thu Nov  4 15:02:24 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
 
 - AutoYaST
-  - When the interface section contains the deprecated device
-    element and also the name element, then use the last one as
-    the interface description. (bsc#1192270)
-  - Add the description element to the interface section
+  - When the interface section contains the "device" (deprecated)
+    and "name" elements then use the "device" as the "name" and the
+    "name" as the "description". (bsc#1192270)
+  - Add the "description" element to the interface section.
 - 4.2.107
 
 -------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


dgdavid (approved):
LGTM, just a minor comment/preference

src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -260,9 +261,16 @@ def init_from_hashes(hash)
         hash = rename_key(hash, "bridge_forwarddelay", "bridge_forward_delay")
         super(hash)
 
-        if hash["aliases"]
-          self.aliases = hash["aliases"].values.map { |h| AliasSection.new_from_hashes(h) }
+        # When the name and the device attributes are given then the pre network-ng behavior will be
+        # adopted using the name as the description and the device as the name (bsc#1192270).
+        unless (hash.fetch("name", "").empty? || hash.fetch("device", "").empty?)
```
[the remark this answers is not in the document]
teclator: I would need to double check but in case of SLE-15-SP2 I think hash["name"] will be "" if not given and the same with hash["device"] I think it changed in SLE-15-SP3. So, I was a little bit conservative... and tried to make it work in both cases with expected nil or empty values.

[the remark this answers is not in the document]
teclator: BTW would be nice if @imobachgs can check it just to know his opinion about adding the description element in SP2.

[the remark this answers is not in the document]
joseivanlopez: What about this version?

~~~
values = hash.fetch_values("name", "device")

if values.none?(&:nil?) && values.none?(&:empty?)
  ...
end
~~~

package/yast2-network.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Thu Nov  4 15:02:24 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- AutoYaST
+  - When the interface section contains the deprecated device
+    element and also the name element, then use the last one as
```
joseivanlopez, line 6: "then use the device as name and the name as description", right?

package/yast2-network.changes
```diff
@@ -1,3 +1,13 @@
+-------------------------------------------------------------------
+Thu Nov  4 15:02:24 UTC 2021 - Knut Anderssen <kanderssen@suse.com>
+
+- AutoYaST
+  - When the interface section contains the deprecated device
```
imobachgs, line 5: I would use quotes to make the text easier to understand. Moreover, I would rephrase it like *contains the "device" (deprecated) and "name" elements...* or *contains the "device" (deprecated) in addition to the "name" element...".

package/yast2-network.changes
joseivanlopez: BTW, it would be nice to have profile "versions". That makes easier to discover what the profile really means.

imobachgs, line 6: *then use the latter*

imobachgs (changes requested):

package/yast2-network.changes
teclator: > "then use the device as name and the name as description", right?

The **device** is already used as the interface name but the "name" has precedence... I could rephrase it to make it more clear.

imobachgs (approved):

PR #1272, opened 2021-11-10
Fix description for s390 Group Device description

teclator:
## Problem

The `description_for` method added by https://github.com/yast/yast-network/pull/1268 expect two parameters and the first one is an `Interface` or a offline `S390 Group Device`.

## Solution

Fixed the issue and also the documentation, already tested in a s390 machine.

## Screenshot

S390LanOverview


joseivanlopez (approved):
LGTM

PR #1277, opened 2022-01-12
Making the code to fit to inclusive naming initiative

mchf:
- https://trello.com/c/7Hhey5vk/2732-3-featurecould-have-no-slaves-in-yast-anymore

## Problem ##

We're using some terms which doesn't correspond to [inclusive naming](https://inclusivenaming.org/).

## Solution ##

Replace as many occurrences of bad naming practices as possible. This PR is focused for _master_ / _slave_ terms

What was replaced:
* all occurrences in UI
* most occurrences in the source code
* most occurrences in comments (as we (can) generate docu from it)

What was not replaced:
* some occurrences which more / less define an API to other components in the system (sysconfig, NetworkManager). We could not replace these as respective components still use these terms. Even some low level pieces of our code can still use old naming to avoid confusion. This should be replaced when the API becomes obsolete.
* CLI - option names. This PR is already too big. So, fix for CLI deserves its own which will follow this one closely
* AY - we have to open a discussion how to handle backward compatibility. Own PR will follow once it is clear.
* almost all code for wifi was kept untouched. It uses term "master" at several places but without negative meaning in relation to "slaves". If we're asked to solve it too, it will be handled in separate PR.


src/include/network/lan/cmdline.rb
```diff
@@ -206,7 +206,7 @@ def update_builder_from_options!(builder, options)
       case builder.type.short_name
       when "bond"
         # change only if user specify it
-        builder.slaves = options["slaves"].split(" ") if options["slaves"]
+        builder.childs = options["slaves"].split(" ") if options["slaves"]
```
teclator, line 209: I would prefer **members** than **childs** as commented in the feature, but it is just my POV. @mvidner what do you think?

mchf: well ... I've intentionally used grammatically incorrect word "childs". In hope that it won't have some unexpected side meanings. 

At least it will be easy to change by one regexp to something else if not accepted ;-)

teclator: See also http://lkml.iu.edu/hypermail/linux/kernel/2010.0/02186.html, looks like there is some request for using `port` as it is already used for bridges.

teclator: > well ... I've intentionally used grammatically incorrect word "childs". In hope that it won't have some unexpected side meanings.
> 
> At least it will be easy to change by one regexp to something else if not accepted ;-)

I know but think that we have to also update the documentation so would like to at least share it with wicked team and others.

teclator: > well ... I've intentionally used grammatically incorrect word "childs". In hope that it won't have some unexpected side meanings.
> 
> At least it will be easy to change by one regexp to something else if not accepted ;-)

That regexp would rename also the **VLAN** method that were already using **parent / child** terminology so that could be an unexpected side effect change :P but I'm being too nitpicking now :)

src/lib/y2network/presenters/interface_summary.rb
```diff
@@ -64,25 +64,25 @@ def text
...
-          master = connection.find_master(@config.connections)
-          if master
-            master_desc = if master.type.bonding?
-              # TRANSLATORS: text label before device which is master for this device
-              _("Bonding master")
+          parent = connection.find_parent(@config.connections)
+          if parent
+            parent_desc = if parent.type.bonding?
+              # TRANSLATORS: text label before device which is parent for this device
+              _("Bonding parent")
```
teclator, line 80: Why not just...
```suggestion
              _("Bond")
```

src/lib/y2network/widgets/bond_child.rb
```diff
@@ -40,36 +40,36 @@ def initialize(settings)
 
       def contents
         Frame(
-          _("Bond Slaves and Order"),
+          _("Bonded Devices and Order"),
```
teclator, line 43:
```suggestion
          _("Bonding Members and Order"),
```

src/lib/y2network/widgets/bond_childs_tab.rb
```diff
@@ -21,24 +21,24 @@
...
+    class BondChildsTab < CWM::Tab
       def initialize(settings)
         textdomain "network"
 
         @settings = settings
       end
 
       def label
-        _("&Bond Slaves")
+        _("&Bonded devices")
```
teclator, line 37:
```suggestion
        _("&Bonding members")
```

src/include/network/lan/cmdline.rb
mvidner: > See also http://lkml.iu.edu/hypermail/linux/kernel/2010.0/02186.html, looks like there is some request for using `port` as it is already used for bridges.

IMHO it's good to follow the kernel side discussion of the renaming. Here's [version 4 (2020-11) of the above patch set][1] (which was v2, 2020-10), the best part is [5/5 which is the documentation][5].

[5]: https://www.mail-archive.com/linux-kernel@vger.kernel.org/msg2370989.html
[1]: https://www.mail-archive.com/linux-kernel@vger.kernel.org/msg2370988.html

mvidner: The kernel says:

> Major terminology rework done late 2020, to start to remove the use of
> the socially problematic terms "master" and "slave" from the code. The
> "master" device is now referred to as simply the "bond" device and the
> "slave" devices as "ports", but all sysfs, procfs and module options
> have been retained as-is for userspace compatibility...

src/lib/y2network/presenters/interface_summary.rb
mvidner: Kernel : "Bond device"

mchf:
@teclator @mvidner I've modified the patch to use "port(s)" insted of "child(s)". I'll search for minor issues, could you review again?

src/lib/y2network/autoinst_profile/interface_section.rb
```diff
@@ -163,7 +163,7 @@ def self.attributes
       #  @return [String] ???
 
       # @!attribute bonding_slaveX
-      #  @return [String] bonding slave on position X
+      #  @return [String] bonding port on position X
```
mvidner, line 166: So what about AY profiles? Whatever we do, document it.
- commit messages
- PR description
- AY docs? There are some now: https://documentation.suse.com/sles/15-SP3/single-html/SLES-autoyast/#CreateProfile-Network-Interfaces

src/lib/y2network/widgets/bond_port.rb
```diff
@@ -196,15 +196,15 @@ def justify_dev_name(name)
       # device names if at least two devices shared the same physical port id
       # TODO: backend method
       #
-      # @param slaves [Array<String>] bonding slaves
+      # @param ports [Array<String>] devices included in the bonding
       # @return [Hash{String => Array<String>}] of duplicated physical port ids
-      def repeated_physical_port_ids(slaves)
+      def repeated_physical_port_ids(ports)
```
mvidner, line 201: Here we see a previous usage of the term _port_. We should explain in a comment whether ports are the same as slaves or not.

src/lib/y2network/config.rb
```diff
@@ -237,7 +237,7 @@ def configured_interface?(iface_name)
     #   to be modified or deleted if `connection_config` is deleted or renamed
     def connections_to_modify(connection_config)
       result = []
-      bond_bridge = connection_config.find_master(connections)
+      bond_bridge = connection_config.find_parent(connections)
```
mvidner, line 240: Please include the terms _bridge_ and _parent_ in the terms summary.
Maybe some of them are duplicate terms that could be unified, now or later.

src/lib/y2network/network_manager/connection_config_writers/base.rb
```diff
@@ -105,12 +105,12 @@ def configure_dhcp(conn)
...
-          slave_type = "bridge" if parent.type.br?
-          slave_type = "bond" if parent.type.bonding?
-          return unless slave_type
+        def configure_as_port(parent)
+          port_type = "bridge" if parent.type.br?
+          port_type = "bond" if parent.type.bonding?
+          return unless port_type
 
-          file.connection["slave-type"] = slave_type
+          file.connection["slave-type"] = port_type
```
mvidner, line 113:
```suggestion
          # slave-type is NetworkManager API
          file.connection["slave-type"] = port_type
```

I have checked the NM git repo, so far the code has slaves, there is no PR, but there is an Issue
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/779 "changing problematic language in nmcli" which says that around 2022-07 they should have inclusive naming

mvidner (commented):
Thanks!

I can see that this is bigger than I thought, and we cannot do everything at once.
Please list
- the areas that you've changed
- the unchanged ones that you know of, so that we can queue followup tasks

src/lib/y2network/network_manager/connection_config_writers/base.rb
mvidner: Correction: something got merged for NM 1.34.0, apparently only in the D-Bus API, not in config files.
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/985/diffs

src/include/network/lan/cmdline.rb
```diff
@@ -206,7 +206,7 @@ def update_builder_from_options!(builder, options)
       case builder.type.short_name
       when "bond"
         # change only if user specify it
-        builder.slaves = options["slaves"].split(" ") if options["slaves"]
+        builder.ports = options["slaves"].split(" ") if options["slaves"]
```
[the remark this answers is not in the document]
mchf: CLI will be in next PR as this one is huge already.

src/lib/y2network/autoinst_profile/interface_section.rb
mchf: for now I'd keep it. If anything is going to be done in AY profiles it will has to be in separate commit as it will require quite some work with respect to backward compatibility.

src/lib/y2network/network_manager/connection_config_writers/base.rb
mchf: ok, thanks

package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Wed Jan 26 08:59:57 UTC 2022 - Michal Filka <mfilka@suse.com>
+
+- jsc#SLE-22015
+  - make the code to confirm to inclusive naming, most masters and
+    slaves are gone. Kept only where makes sense - interface to
```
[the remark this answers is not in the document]
mchf: ok, somehow done

src/lib/y2network/widgets/bond_port.rb
mchf: Extended method's documentation

src/lib/y2network/config.rb
mchf: bridge and parent are not that related at this place. Here the meaning is bond or bridge. Or vice versa, if the connection has a parent then it is a bond or a bridge device.

src/lib/cfa/interface_file.rb
```diff
@@ -315,10 +315,14 @@ def variable_name(param_name)
     ## BONDING
 
     # @!attribute [r] bonding_master
+    #   TODO: this name doesn't correspond to inclusive naming, however as log as the
```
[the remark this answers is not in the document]
mchf: I'm not aware of it. As main task was to get bad words mainly from UI, I hardly doubt about it.

mvidner:
Dictionaries for the **bond** word:
- https://en.wiktionary.org/wiki/bond#Noun
- https://slovnik.seznam.cz/preklad/anglicky_cesky/bond

src/lib/y2network/widgets/bond_port.rb
```diff
@@ -196,15 +196,23 @@ def justify_dev_name(name)
...
+      def repeated_physical_port_ids(b_ports)
         physical_port_ids = {}
 
-        slaves.each do |slave|
-          if physical_port_id?(slave)
-            p = physical_port_ids[physical_port_id(slave)] ||= []
-            p << slave
+        b_ports.each do |b_port|
+          if physical_port_id?(b_port)
+            p_port = physical_port_id(b_port)
```
mvidner: @mchf I have changed this method and the following one to make the code+comments clearer

mvidner (changes requested):
OK, thanks!

src/lib/y2network/widgets/interfaces_table.rb
```diff
@@ -91,8 +91,8 @@ def note(interface, conn)
 
         return "" unless conn
 
-        master = conn.find_master(config.connections)
-        return format(_("enslaved in %s"), master.name) if master
+        parent = conn.find_parent(config.connections)
+        return format(_("included in %s"), parent.name) if parent
```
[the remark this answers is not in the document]
mchf: yes

package/yast2-network.changes
```diff
@@ -1,3 +1,17 @@
+-------------------------------------------------------------------
+Wed Jan 26 08:59:57 UTC 2022 - Michal Filka <mfilka@suse.com>
+
+- jsc#SLE-22015
+  - make the code to confirm to inclusive naming, most masters and
```
mvidner, line 5:
```suggestion
  - make the code conform to inclusive naming, most masters and
```

mvidner (approved):

PR #1280, opened 2022-02-01
None slaves work in cli animore

mchf:
Follow up of https://github.com/yast/yast-network/pull/1277

## Problem ##

CLI used naming according to old standards for bonding.

## Solution ##

- options were renamed in a same way like in GUI
- the change is backward incompatible. CLI is not much used (according to fact that some bugs were present in CLI for years), so investing into backward compatibility sounds useless. (Other option is coexistence of bond_ports and slaves options and marking the later as "deprecated" which is going to be removed ... if someone remembers it in the future ;-))

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+Tue Feb  1 06:28:14 UTC 2022 - Michal Filka <mfilka@suse.com>
+
+- jsc#SLE-22015
+  - renamed CLI options to conform inclusive naming
+  - the change is backward incompatible
+- 4.4.36
+
+-------------------------------------------------------------------
 Wed Jan 26 08:59:57 UTC 2022 - Michal Filka <mfilka@suse.com>
 
 - jsc#SLE-22015
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,8 @@
 Tue Feb  1 06:28:14 UTC 2022 - Michal Filka <mfilka@suse.com>
 
 - jsc#SLE-22015
-  - renamed CLI options to conform inclusive naming
+  - renamed slaves CLI option to conform to inclusive naming:
+    yast lan add name=bond0 bond_ports=eth0 eth1 bootproto=dhcp
   - the change is backward incompatible
 - 4.4.36
 
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -4,7 +4,7 @@ Tue Feb  1 06:28:14 UTC 2022 - Michal Filka <mfilka@suse.com>
 - jsc#SLE-22015
   - renamed slaves CLI option to conform to inclusive naming:
     yast lan add name=bond0 bond_ports=eth0 eth1 bootproto=dhcp
-  - the change is backward incompatible
+  - the "slaves=" option stays for backward compatibility
 - 4.4.36
 
 -------------------------------------------------------------------
--- a/src/clients/lan.rb
+++ b/src/clients/lan.rb
@@ -90,7 +90,7 @@ module Yast
             "example" => [
               "yast lan add name=vlan50 ethdevice=eth0 bootproto=dhcp",
               "yast lan add name=br0 bridge_ports=eth0 eth1 bootproot=dhcp",
-              "yast lan add name=bond0 slaves=eth0 eth1 bootproto=dhcp",
+              "yast lan add name=bond0 bond_ports=eth0 eth1 bootproto=dhcp",
               "yast lan add name=dummy0 type=dummy ip=10.0.0.100"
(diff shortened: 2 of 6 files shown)


suntorytimed:
IMHO we should document the changed parameters in the Release Notes and let the documentation team know that their documentation needs to be adjusted.

mchf:
@suntorytimed So, you're fine with renaming the parameter, just mention it in the release notes (and docu)

suntorytimed:
@mchf yes, but I have also asked PM for their input on this, as they might have a different opinion.

kobliha:
Nitpicking: "animore" :)?

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Tue Feb  1 06:28:14 UTC 2022 - Michal Filka <mfilka@suse.com>
+
+- jsc#SLE-22015
+  - renamed CLI options to conform inclusive naming
```
mvidner, line 5:
```suggestion
  - renamed CLI options to conform to inclusive naming:
    yast lan add name=bond0 bond_ports=eth0 eth1 bootproto=dhcp
```

mvidner (approved):
OK, but the proposed changelog is not very helpful, so I'm suggesting an example to improve it

package/yast2-network.changes
mchf: @mvidner Updated partly according to the proposal, partly using own creativity.

suntorytimed:
@mchf the feedback from PM is, that we should keep the old option in place as a deprecated parameter in addition to the bond_ports parameter.

src/clients/lan.rb
```diff
@@ -160,6 +160,11 @@ def main
             "help" => _("Bond Ports"),
             "type" => "string"
           },
+          "slaves"   => {
+            # Commandline option help
+            "help" => _("Bond Ports (obsolete, use bond_ports instead)"),
```
mvidner, line 165: This will confuse the hell out of translators, because they don't see the code.
Add a comment please

src/include/network/lan/cmdline.rb
```diff
@@ -172,7 +178,7 @@ def DeleteHandler(options)
     # @param options [Hash{String => String}] action options
     # @return [String] infered device type; an empty string if not infered
     def infered_type(options)
-      return "bond" if options.include? "slaves"
+      return "bond" if options.include? "bond_ports"
```
mvidner, line 181:
```suggestion
      return "bond" if options.include? "slaves"
      return "bond" if options.include? "bond_ports"
```

src/include/network/lan/cmdline.rb
```diff
@@ -206,7 +212,7 @@ def update_builder_from_options!(builder, options)
       case builder.type.short_name
       when "bond"
         # change only if user specify it
-        builder.ports = options["slaves"].split(" ") if options["slaves"]
+        builder.ports = options["bond_ports"].split(" ") if options["bond_ports"]
```
mvidner, line 215:
```suggestion
        builder.ports = options["slaves"].split(" ") if options["slaves"]
        builder.ports = options["bond_ports"].split(" ") if options["bond_ports"]
```

mvidner (changes requested):
I think these pieces of old behavior still need to be preserved

src/include/network/lan/cmdline.rb
mchf: Hopefully not. That's reason I've added https://github.com/yast/yast-network/pull/1280/commits/7a5a3278932d6ce097aaaf8cfa7b5e9fb55c6fd9#diff-cd51e63dfa3bcb1671b7efb41b3f76ead7339528b230c594ae3882029620ee75R112

mvidner: Yes, I see now.

May I point out that my way is less code and easier to understand and review? :-)

package/yast2-network.changes
```diff
@@ -1,3 +1,12 @@
+-------------------------------------------------------------------
+Tue Feb  1 06:28:14 UTC 2022 - Michal Filka <mfilka@suse.com>
+
+- jsc#SLE-22015
+  - renamed slaves CLI option to conform to inclusive naming:
+    yast lan add name=bond0 bond_ports=eth0 eth1 bootproto=dhcp
+  - the change is backward incompatible
```
mvidner, line 7:
```suggestion
  - the "slaves=" option stays for backward compatibility
```

src/include/network/lan/cmdline.rb
mchf: yep. But my way makes behavior cleaner when both `slaves` and `bond_ports` are present simultaneously

mvidner (approved):
thanks!

PR #1281, opened 2022-02-04
Bring back overview help (bsc#1195155)

teclator:
## Problem

Apparently the help text for the **Overview** tab was removed when network-ng and the new widgets were introduced (see https://github.com/yast/yast-network/commit/9312e52420d993272badb4de41f91ed78cdb33ef)

The removed help was:

Previous help

- https://bugzilla.suse.com/show_bug.cgi?id=1195155

## Solution

Added help per widget

HelpPerWidget

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Fri Feb  4 15:14:10 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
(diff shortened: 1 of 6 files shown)


dgdavid (approved):
LGTM, but maybe we should add those changes in SP3 too. And perhaps first, in order to follow the [maintenance workflow](https://yastgithubio.readthedocs.io/en/latest/maintenance-branches/#maintenance-work-flow) What do you think?

PR #1282, opened 2022-02-09
Do not setup DHCP in case of configured iBFT devices

teclator:

PR #1284, opened 2022-02-22
Fixed active device detection (bsc#1196276)

teclator:
The problem was introduced by https://github.com/yast/yast-network/pull/1282/files#diff-5316412e8bda05f275f2710e425296ad554b326f59d3d320da81b3bdbbdbd56cR52

- https://bugzilla.suse.com/show_bug.cgi?id=1196276

Fixed loop for active device detection which should go next instead of return false.


joseivanlopez (approved):
LGTM

PR #1285, opened 2022-02-24
Do not consider iBFT configuration for proposing DHCP setup

teclator:
## Problem

1. In #1282 we skipped **DHCP setup** completely when there was some active interface configured by **iBFT**. 
 
    This change is probably too much at this moment and we should skip only the **iBFT** interfaces as candidates to **setup DHCP** but continue configuring any not configured interface in case there is no **connection config (ifcfg-file)** present in the system.

2. In #1283 we applied also the **same logic** when deciding if the **network settings** dialog needs to be run or not. 

    This behavior should be maintained as for example in case that all the interfaces present in the system are **iBFT** configured then the dialog will be run when it is not needed. (Like in https://bugzilla.suse.com/show_bug.cgi?id=1194711)

- https://bugzilla.suse.com/show_bug.cgi?id=1194911#1284 

## Solution

1. **DHCP Setup Client**
    - It is skipped if there is some network interface **UP** and configured through a **ifcfg file**.
    - If it is not skipped then try to setup with **DHCP** not configured interfaces **(iBFT configured interfaces are considered as configured).**
2. The **network settings** dialog is only shown if there is no active and configured interface (including iBFT configured interfaces).

## Tests

Tested manually and added unit tests.

## Screenshots

**Scenario with 1 interface configured with iBFT**

| Description | Screenshot |
| --------------- | --------------- |
| iPXE eth0 configured | Screenshot_testing_2022-02-24_12:57:13 |
| First screen after language selection | Screenshot_testing_2022-02-24_12:54:45 |
| Clicking on Network Configuration | Screenshot_testing_2022-02-24_12:55:00 |

**Scenario with 1 interface configured with iBFT and 1 not configured**

| Description | Screenshot |
| --------------- | --------------- |
| iPXE eth0 configured | Screenshot_testing_2022-02-24_12:57:13 |
| Before yast.ssh is called | Screenshot_testing_2022-02-24_13:05:47 |
| After yast.ssh is called | Screenshot_testing_2022-02-24_13:07:35 |
| First screen after language selection | Screenshot_testing_2022-02-24_12:54:45 |
| Clicking on Network Configuration | Screenshot_testing_2022-02-24_13:07:44 |

### DIFF
--- a/src/lib/network/network_autoconfiguration.rb
+++ b/src/lib/network/network_autoconfiguration.rb
@@ -47,25 +47,14 @@ module Yast
     # system
     #
     # returns [Boolean] true when at least one interface is active
-    def any_iface_active?
-      Yast::Lan.Read(:cache)
+    def any_iface_active?(ibft_included: false)
+      Lan::Read(:cache)
+
       config.interfaces.any? do |interface|
         next false unless active_config?(interface.name)
+        return true if ibft_included && ibft_interfaces.include?(interface.name)
 
         config.connections.by_name(interface.name)
-      end
-    end
-
-    # Checks if any of available interfaces is active and configured by iBFT or with a connection
-    # file present in the system
-    #
-    # returns [Boolean] true when at least one interface is active and configured
-    def network_configured?
-      Yast::Lan.Read(:cache)
-      config.interfaces.any? do |interface|
-        next false unless active_config?(interface.name)
-
-        config.connections.by_name(interface.name) || ibft_interfaces.include?(interface.name)
       end
     end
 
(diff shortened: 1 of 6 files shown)


src/lib/network/network_autoconfiguration.rb
```diff
@@ -43,10 +43,24 @@ def initialize
...
     #
     # returns [Boolean] true when at least one interface is active
     def any_iface_active?
+      Yast::Lan.Read(:cache)
+      config.interfaces.any? do |interface|
+        next false unless active_config?(interface.name)
+
+        config.connections.by_name(interface.name)
+      end
+    end
```
ancorgs, line 57: I see quite some code duplication between this and `#network_configured?`

teclator: Yep, I preferred to duplicate the logic by now, but could try to reorganize / extract some common logic.

mtomaschewski:
I've reopened https://bugzilla.suse.com/show_bug.cgi?id=1194911 see:
- https://bugzilla.suse.com/show_bug.cgi?id=1194911#c42

PR #1286, opened 2022-02-25
Revert last PR skipping DCHP setup completely in case of iBFT configured

teclator:
After some discussions, apparently #1282 was good enough and #1285 was wrong and has been requested to be reverted.

https://bugzilla.suse.com/show_bug.cgi?id=1194911#c49

Basically, if **iBFT** is used it should provide all require settings and configuring another interface with DHCP could be a problem with the reverse path.

Therefore, this PR reverts #1285 .


ancorgs (approved):
LGTM

PR #1287, opened 2022-03-03
Add qeth and hsi connection writers (bsc#1196582)

teclator:
## Problem

When yast2-network **NetworkManager** writers were [added](https://github.com/yast/yast-network/pull/1149) we omitted s390 devices like **qeth**, **lcs**, **ctc** and **hsi** which are supported by **wicked** writers.

This lack of support has been reported by:

- https://bugzilla.suse.com/show_bug.cgi?id=1196582

## Solution

Added **qeth** and **hsi** connection writers which are currently the most important ones for s390 according to the bug report.

### DIFF
--- a/src/lib/y2network/network_manager/connection_config_writers/hsi.rb
+++ b/src/lib/y2network/network_manager/connection_config_writers/hsi.rb
@@ -1,0 +1,31 @@
+# Copyright (c) [2021] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE LLC about this file by physical or electronic mail, you may
+# find current contact information at www.suse.com.
+
+require "y2network/network_manager/connection_config_writers/qeth"
+
+module Y2Network
+  module NetworkManager
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Hsi
+      # object to the underlying system.
+      class Hsi < Qeth
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


src/lib/y2network/network_manager/connection_config_writers/hsi.rb
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) [2021] SUSE LLC
```
ancorgs, line 1: Wrong year in copyright in this file and in several others.

src/lib/y2network/network_manager/connection_config_writers/hsi.rb
```diff
@@ -0,0 +1,31 @@
...
+# find current contact information at www.suse.com.
+
+require "y2network/network_manager/connection_config_writers/qeth"
+
+module Y2Network
+  module NetworkManager
+    module ConnectionConfigWriters
+      # This class is responsible for writing the information from a ConnectionConfig::Hsi
+      # object to the underlying system.
+      class Hsi < Qeth
```
ancorgs, line 27: An empty class is something that certainly looks weird.

ancorgs: Seems to be a result of this, which couples the name of the file to its format, which I'm not sure is 100% correct.
https://github.com/yast/yast-network/blob/166c75353147a0c3a58612113a8fa8edca3ac2be/src/lib/y2network/network_manager/connection_config_writer.rb#L70

teclator: We could move this specific options to a **mixin** to be included by **Qeth** and **Hsi** classes which are written in the same way. But that fix nothing and we already have the same in wicked writers. So, for a bug report looks weird to change the current approach.

teclator: The same approach were already used by several wicked writers which does not different in options from an **ethernet** connection like:

- https://github.com/yast/yast-network/blob/166c75353147a0c3a58612113a8fa8edca3ac2be/src/lib/y2network/wicked/connection_config_writers/hsi.rb
- https://github.com/yast/yast-network/blob/166c75353147a0c3a58612113a8fa8edca3ac2be/src/lib/y2network/wicked/connection_config_writers/qeth.rb
- https://github.com/yast/yast-network/blob/166c75353147a0c3a58612113a8fa8edca3ac2be/src/lib/y2network/wicked/connection_config_writers/lcs.rb

teclator: Modifying the way the connection writers are selected is probably too much for a bug report, so it is something completely discarded by now.

ancorgs (approved):
LGTM

PR #1288, opened 2022-03-09
Write NetworkManager s390 settings to ethernet section

teclator:
## Problem

The NetworkManager connection's **s390 specific options** are writing in the **connection section** instead of the **ethernet** one which does not break the load of the connection settings but ignore the **ethernet s390 specific options**.

- https://bugzilla.suse.com/show_bug.cgi?id=1196582

## Solution 

Write the **s390 specific options** in the ethernet section.


ancorgs (approved):
LGTM

PR #1290, opened 2022-05-04
Disable loading the Ruby bindings (bsc#1197120)

mvidner:
## References

- See https://github.com/yast/yast-core/pull/155 for a full description
- https://bugzilla.suse.com/show_bug.cgi?id=1197120
- https://trello.com/c/PMo8iWEx/2946-3-tw-p3-1197120-what-to-do-about-libjemalloc

### DIFF
--- a/src/servers_non_y2/ag_udev_persistent
+++ b/src/servers_non_y2/ag_udev_persistent
@@ -20,6 +20,11 @@
 # find current contact information at www.suse.com.
 
 package ag_udev_persistent;
+
+# None of the bindings are needed here.
+# And dlopening Ruby would make jemalloc complain about a failed TLS allocation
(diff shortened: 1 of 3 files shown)


src/servers_non_y2/ag_udev_persistent
```diff
@@ -20,6 +20,7 @@
 # find current contact information at www.suse.com.
 
 package ag_udev_persistent;
+BEGIN { $ENV{Y2DISABLELANGUAGEPLUGINS} = "1"; }
```
jreidinger, line 23: NP: Comment in code explaining WHY it is needed, would be nice.

mvidner: @jreidinger good point, thanks, added

PR #1301, opened 2022-07-21
Added a FCoE connections generator (bsc#1199554)

teclator:
## Problem

When installing a server with FCoE on a target LUN it gets stuck in subsequent reboots due to a wrong network configuration.

- https://bugzilla.suse.com/show_bug.cgi?id=1199554
- https://trello.com/c/vA1WOdhV

## Solution

Added a class to generate the needed configuration for a **FCoE** device being aware of it during installation.

## Testing

- *Added a new unit test*
- *Tested manually* (pending due to problems with server)

### DIFF
--- a/test/y2network/fcoe_conn_generator_test.rb
+++ b/test/y2network/fcoe_conn_generator_test.rb
@@ -1,0 +1,131 @@
+#!/usr/bin/env rspec
+
+# Copyright (c) [2022] SUSE LLC
+#
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of version 2 of the GNU General Public License as published
+# by the Free Software Foundation.
+#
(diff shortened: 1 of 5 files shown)


test/y2network/fcoe_conn_generator_test.rb
```diff
@@ -0,0 +1,131 @@
...
+      before do
+        generator.update_connections_for(device)
+      end
+
+      it "sets the connection STARTMODE to 'nfsroot'" do
+        expect(conn.startmode.to_s).to eql("nfsroot")
+      end
+
+      it "sets the connection BOOTPROTO to 'static'" do
+        expect(conn.bootproto&.name).to eql("static")
```
dgdavid, line 103: NP: I think `&.` is no needed here since `conn.bootproto` mustn't be nil at this point.

dgdavid (approved):
LGTM, thank you!

PR #1304, opened 2022-08-05
[WIP] Check STIG security policy

imobachgs:
This PR is just a PoC to use security policy validation for network settings.

See https://github.com/yast/yast-security/pull/128


imobachgs:
This PR is not needed anymore. The check is executed in yast/yast-security#128.

PR #1305, opened 2022-09-01
Disable backends

teclator:
## Problem

During installation, if apparently the selected network service has not been modified (wicked is default backend) the service is not enabled explicitly in the target system which could end with it disabled.

- https://bugzilla.suse.com/show_bug.cgi?id=1202479

## Solution

During installation, force an enablement of the selected network service even when if has not been modified and also ensure not selected backend is disabled in order to prevent more than one end running after the installation. 

It depends on (https://github.com/yast/yast-yast2/pull/1269)

## Testing

Tested manually

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -4,7 +4,7 @@ Thu Sep  1 08:30:47 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
 - At the end of the installation, force an enablement of the
   selected network service even when the selected one has not been
   modified and ensure other backends are disabled (bsc#1202479)
-- 4.4.52
+- 4.4.50
 
 -------------------------------------------------------------------
 Thu Jul 21 08:38:40 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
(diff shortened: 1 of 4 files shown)


package/yast2-network.spec
```diff
@@ -17,7 +17,7 @@
 
 
 Name:           yast2-network
-Version:        4.4.49
+Version:        4.4.50
```
mchf, line 20: version doesn't match to the changelog one

teclator: Ups :P

package/yast2-network.changes
```diff
@@ -1,3 +1,11 @@
+-------------------------------------------------------------------
+Thu Sep  1 08:30:47 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
+- At the end of the installation, force an enablement of the
+  selected network service even when the selected one has not been
+  modified and ensure other backends are disabled (bsc#1202479)
+- 4.4.52
```
teclator, line 7:
```suggestion
- 4.4.50
```

mchf (approved):
lgtm

PR #1306, opened 2022-09-02
During installation force to enable selected backend and disable not selected one (master)

teclator:
It merges (#1305) into master and needs (https://github.com/yast/yast-yast2/pull/1270)

### DIFF
--- a/src/lib/network/clients/save_network.rb
+++ b/src/lib/network/clients/save_network.rb
@@ -355,6 +355,10 @@ module Yast
       # at all specially in a live installation where NM is the backend by
       # default. For detecting changes we should reset the cache first.
       NetworkService.reset!
+      # Ensure not selected backend is disabled in order to not end with two network backends
+      # running at the same time. (bsc#1202479)
+      NetworkService.send(:disable_service, :wicked)
+      NetworkService.send(:disable_service, :network_manager)
       case backend
       when :network_manager
         log.info("- using NetworkManager")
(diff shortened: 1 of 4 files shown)


src/lib/network/clients/save_network.rb
```diff
@@ -355,6 +355,10 @@ def set_network_service
       # at all specially in a live installation where NM is the backend by
       # default. For detecting changes we should reset the cache first.
       NetworkService.reset!
+      # Ensure not selected backend is disabled in order to not end with two network backends
+      # running at the same time. (bsc#1202479)
+      NetworkService.send(:disable_service, :wicked)
+      NetworkService.send(:disable_service, :network_manager)
```
jreidinger, line 361: well, I kind of expect that it will ensure the EnableDisableNow method. I do not like this change that basically move responsibility for disabling services to caller instead on NetworkService method itself.

teclator: The method will take care of enabling and disabling the selected **backend** and the current **backend** but as we have seen in the bug, in case of installation, the selected and current **backend** could be the same which in that case it does not touch any service at all but at installation time we need to ensure the other backends are disabled too. We probably could think in a better way but by now I think it is not a responsability for the **EnablingDisablingNow** mehtod

teclator: I'm not saying we cannot move the responsibility there but not see doing it for this specific bug.

jreidinger (approved):

teclator:
Tried to rerun the jobs which are still failing because the container does not contain the yast2 dependency because of a build problem (thanks @lslezak for taking a look!)

lslezak:
I fixed the CI images in OBS (added `FileProvides: /usr/bin/gdb-add-index gdb` to the project config), the image now builds fine.
I have restarted the tests, now everything is green. :heavy_check_mark:

teclator:
Thanks, merging it!!

PR #1307, opened 2022-09-12
Activate s390 devices before anything else (bsc#1199746)

teclator:
## Problem

In **AutoYaST** it is not possible to write udev rules for more than one s390 interface.

- https://bugzilla.suse.com/show_bug.cgi?id=1199746

## Solution

In order to write custom udev rules for s390 devices then the devices should be activated before the configuration is read or otherwise them will be ignored.

## Testing

- *Tested manually*


dgdavid (commented):
Do not forget bumping version and updating changelog

teclator:
Probably missed to commit and push it, so, fixed.

dgdavid (approved):
LGTM

PR #1309, opened 2022-09-29
Fix no route gw nm

teclator:
## Problem

When no route gateway is defined the NetworkManager routing writer crashes.

- [bsc#1203866](https://bugzilla.suse.com/show_activity.cgi?id=1203866)

## Solution

When writing the NetworkManager configuration ignore a route when there is no route gateway.

### DIFF
--- a/src/lib/y2network/network_manager/connection_config_writers/base.rb
+++ b/src/lib/y2network/network_manager/connection_config_writers/base.rb
@@ -64,7 +64,7 @@ module Y2Network
         #
         # @param routes [<Array<Y2Network::Route>] routes associated with the connection
         def configure_routes(routes)
-          routes.select(&:default?).each { |r| configure_gateway(r) }
+          routes.select(&:default?).each { |r| configure_gateway(r) if r.gateway }
         end
 
         # @param route [Y2Network::Route] route to be written
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu Sep 29 09:15:47 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
+
+- Fixed issue when writing the NetworkManager config without a
+  gateway (bsc#1203866)
+- 4.3.86
+
+-------------------------------------------------------------------
 Thu Jul 21 08:38:40 UTC 2022 - Knut Anderssen <kanderssen@suse.com>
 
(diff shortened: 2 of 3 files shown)


src/lib/y2network/network_manager/connection_config_writers/base.rb
```diff
@@ -64,7 +64,7 @@ def write(conn, opts = {})
         #
         # @param routes [<Array<Y2Network::Route>] routes associated with the connection
         def configure_routes(routes)
-          routes.select(&:default?).each { |r| configure_gateway(r) }
+          routes.select(&:default?).each { |r| configure_gateway(r) if r.gateway }
```
mchf, line 67: Well ... it solves the issue by not raising the internal error. But question is if we want to get into this situation. From logs 
```2022-09-28 05:19:41 <1> install(3712) [Ruby] dialogs/route.rb(initialize):38 route dialog with route: #<Y2Network::Route:0x00005574bc318518 @to=:default, @interface=nil, @gateway=nil, @options=""> and devices ...``` 
so it seems that default route was created without an ip. Does it make sense to allow it?

teclator: it is allowed by the interface and also by wicked which permits for example a default just with device, so, once that is allowed in routing config and interface then yes, it makes sense but not when it is translated from wicked to NM which is the current issue.

mchf: Yeah device route(s) are fine, but even ```@interface``` is nil here so it still looks weird to me. But i didn't check code in details.

teclator: if there is no interface then there should be probably not associated to the connection related to the interface... as that route do not makes any sense we could add a validation in the route dialog to prevent that.

mchf: yep ... that is exactly what i tried to point out. So, what to do now?

teclator: > Yeah device route(s) are fine, but even `@interface` is nil here so it still looks weird to me. But i didn't check code in details.

Confirmed this situation is not crashing, so the issue really happen when there was a specific device declared in the route.

mchf: ok then

mchf (approved):
lgtm

PR #1310, opened 2022-09-29
NM: Do not write a route gateway when it is not defined (SLE-15-SP4)

teclator:
It merges #1309 into SLE-15-SP4


imobachgs (approved):
This code was already approved. I would have written something like:

```ruby
routes.select { |r| r.default? && r.gateway }.each { |r| configure_gateway(r) }
```

So all the logic to filter the routes is in the same place. Anyway,the current fix works, so let's keep the same code that it was already merge in SLE-15-SP3.

PR #1312, opened 2022-10-10
Ay needed packages

teclator:
## Problem

We are only checking needed packages for the selected backend when it is NetworkManager but not in case of wicked or netconfig which in case of AutoYaST now could be needed as otherwise no package will be installed since sysconfig-netconfig is not required anymore.

- [bsc#1201435](https://bugzilla.suse.com/show_bug.cgi?id=1201435)
- 
## Solution

Return needed packages for the selected backend in case them are not installed.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/src/lib/y2network/backend.rb
+++ b/src/lib/y2network/backend.rb
@@ -27,8 +27,6 @@ module Y2Network
     include Yast::I18n
     include Yast2::Equatable
 
-    PACKAGES = [].freeze
-
     # @return [Symbol] backend id
     attr_reader :id
 
@@ -90,8 +88,11 @@ module Y2Network
       Yast::NetworkService.is_backend_available(id)
     end
 
+    # Return a list of the required packages for this backend
+    #
+    # @return [Array<String>] list of packages required by the backend
     def packages
-      self.class.const_get(:PACKAGES)
+      []
     end
   end
 end
--- a/src/lib/y2network/backend.rb
+++ b/src/lib/y2network/backend.rb
@@ -27,6 +27,8 @@ module Y2Network
     include Yast::I18n
     include Yast2::Equatable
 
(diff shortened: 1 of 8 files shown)


imobachgs:
Related to https://github.com/yast/yast-autoinstallation/pull/853.

src/lib/y2network/backend.rb
```diff
@@ -87,5 +89,9 @@ def self.by_id(id)
     def available?
       Yast::NetworkService.is_backend_available(id)
     end
+
+    def packages
+      self.class.const_get(:PACKAGES)
```
imobachgs, line 94: np: to be honest, it looks a little bit over-engineered. Just defining a method that returns the list should be enough. Additionally, if I wanted to create a new derived class, I would need to know that I need to define a constant inside the class to define the list of packages (it is a kind of indirection).

Anyway, that's just an opinion, feel free to keep it as it is because I know you already have tested the code and I do not want to invalidate those tests :-)

imobachgs (approved):

src/lib/y2network/backend.rb
teclator: > np: to be honest, it looks a little bit over-engineered. Just defining a method that returns the list should be enough. Additionally, if I wanted to create a new derived class, I would need to know that I need to define a constant inside the class to define the list of packages (it is a kind of indirection).

Just keep it simple.

> 
> Anyway, that's just an opinion, feel free to keep it as it is because I know you already have tested the code and I do not want to invalidate those tests :-)

imobachgs (approved):

lkocman:
perhaps related issue https://bugzilla.suse.com/show_bug.cgi?id=1203934

PR #1315, opened 2022-12-28
Do not crash when the NETMASK or PREFIX are invalid

teclator:
## Problem

YaST crashes reporting an exception when there is some interface configuration with an invalid NETMASK or PREFIX.

- https://trello.com/c/b6cnNm2A/3252-sles15-sp4-p2-1206551-l3-invalid-netmask-breaks-yast-and-network-setting-is-not-displayed

## Solution

Instead of crashing YaST will report the issues found allowing the user to stop or to continue with the broken configuration.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

Screenshot from 2022-12-29 07-43-09


src/lib/y2network/wicked/connection_config_readers/base.rb
```diff
@@ -158,14 +159,37 @@ def all_ips
...
-        def build_ip(ip, prefix, netmask)
+        # @param id      [String] Hash key for the IP Address
+        def build_ip(ip, id)
           ipaddr = ip.clone
           return ipaddr if ip.prefix?
 
-          ipaddr.netmask = netmask if netmask
-          ipaddr.prefix = prefix if prefix
+          begin
+            netmask = file.netmasks[id]
```
imobachgs, line 168: np: what about moving those begin/rescue blocks to smaller methods? This method has more than 20 lines and most of the code is about registering an issue.

src/lib/y2network/wicked/connection_config_readers/base.rb
```diff
@@ -158,14 +159,37 @@ def all_ips
...
+              netmask, fallback: nil, location: issue_location
+            )
+            issues_list << issue
+          end
+
+          prefix = file.prefixlens[id]
+          return ipaddr unless prefix
+
+          begin
+            address = ipaddr.address.clone
```
imobachgs, line 182: I do not get why you are cloning the address here. Actually, it looks like you are just using the new `address` variable in the next line, but nothing else.

teclator: I do not wanted to modify the address at all but only take advantage of the already provided validations as we are not modifying it in other circumstances like when a netmask is assigned.

teclator: I was also thinking in doing it overriding the `prefix=` method similar to what we do for the netmask.

https://github.com/yast/yast-network/blob/master/src/lib/y2network/ip_address.rb#L87

imobachgs: The intention is still not clear to me. A comment would help. Otherwise, if the only line you expect to fail is the `address`, you might put only that line into the begin/rescue block.

teclator: Well, I do not want the prefix to be assigned in case it is invalid.

imobachgs (approved):

PR #1317, opened 2023-01-03
[WIP] Fix network overview for NetworkManager users

clanig:
Please allow me to propose a solution to make navigating the network section more flawlessly.

## Problem

The user is displayed a message when entering the tab `Overview` in the `Network Settings` of YaST and by acknowledging that message, forced to switch to tab `Global Options`. Apparently this is done to prevent a user from accessing controls that can't be used. This solution might be acceptable in most cases, although it is not possible to check the list of network interfaces anymore, even without the intention to change anything.

However, in case of the ncurses UI it breaks the option to use a combination of `TAB`, `Arrow Keys`, `Enter` to navigate instead of `Alt+<KEY>` shortcuts. (See bug report.)

https://bugzilla.suse.com/show_bug.cgi?id=1205715

## Solution

This PR would disable the buttons `Add`, `Edit`, `Delete` on the tab `Overview` as well as the interfaces table if NetworkManager is used for configuration.
~It will not force the user to `Global Options` anymore when acknowledging the warning. Instead it will inform about the network setup method option in `Global Options`. As a result, the user is still able to see the properties of the network interfaces on the `Overview` page.~
The popup message is not shown anymore. Instead, in the box which usually displays the interface properties, an information is displayed that Network Manager is managing the interfaces.
In the help of the `Overview` page, an information about fixing this by selecting `Wicked` is given.

### Reason for hiding all interfaces:
It looks that the interfaces displayed might differ from reality if NM is used. C.f. https://bugzilla.suse.com/show_bug.cgi?id=1205715#c7
From my perspective it makes sense to handle this separately if required. As a result, these are hidden as well with the current code.

## Testing

The code was tested manually on openSUSE Tumbleweed.

## Request for comments

1. Please give some comments on the proposed change. 
2. This would be my first contribution to the project - if I missed a side effect, please let me know.
3. I changed the ~warning~ message to only tell that NetworkManager is used. If `:network_manager` as backend means that the network could also be disabled, please let me know as well. To me it appears that we know that the network is not disabled but just NetworkManager is used if that is the value.

## Screenshots

Screenshot from 2023-01-12 09-52-29

Screenshot from 2023-01-12 09-53-15

Screenshot from 2023-01-12 09-55-25

Screenshot from 2023-01-12 09-55-39

### DIFF
--- a/src/lib/y2network/presenters/interface_summary.rb
+++ b/src/lib/y2network/presenters/interface_summary.rb
@@ -87,7 +87,11 @@ module Y2Network
         end
 
         if hardware.nil? || !hardware.exists?
-          rich << "<b>(" << _("No hardware information") << ")</b><br>"
+          if @config.backend?(:network_manager)
+            rich << "<b>" << _("Network interfaces are managed by Network Manager.") << "</b><br>"
+          else
+            rich << "<b>(" << _("No hardware information") << ")</b><br>"
+          end
         else
           rich << "<b>(" << _("Not connected") << ")</b><br>" if !hardware.link
           rich << "<b>MAC : </b>" << hardware.mac << "<br>" if hardware.mac
@@ -104,9 +108,11 @@ module Y2Network
             rich << Yast::HTML.Bold(dev_name) << "<br>"
           end
 
-          rich << "<p>"
-          rich << _("The device is not configured. Press <b>Edit</b>\nto configure.\n")
-          rich << "</p>"
+          if !@config.backend?(:network_manager)
+            rich << "<p>"
+            rich << _("The device is not configured. Press <b>Edit</b>\nto configure.\n")
+            rich << "</p>"
+          end
         end
         rich
       end
--- a/src/lib/y2network/widgets/add_interface.rb
+++ b/src/lib/y2network/widgets/add_interface.rb
@@ -30,6 +30,10 @@ module Y2Network
         textdomain "network"
       end
 
+      def init
+        disable if Yast::Lan.yast_config&.backend?(:network_manager)
+      end
+
       def label
         Yast::Label.AddButton
       end
--- a/src/lib/y2network/widgets/delete_interface.rb
+++ b/src/lib/y2network/widgets/delete_interface.rb
@@ -41,6 +41,7 @@ module Y2Network
       # @see CWM::AbstractWidget#init
       def init
         disable unless @table.value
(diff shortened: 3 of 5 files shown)


ancorgs:
> Instead of just disabling the whole tab, in case Network Manager is active, there should be a bold-faced message IN the tab content: "Controlled by Network Manager"

The only thing I miss in your "screenshot" is the part that indicates you can switch to Wicked to get those controls enabled by using the "Global Options" tab.

shundhammer:
> The only thing I miss in your "screenshot" is the part that indicates you can switch to Wicked to get those controls enabled by using the "Global Options" tab.

That's intentional: It would be something to go to the help text. Static texts in dialogs should be minimalistic, not a replacement for the help text.

Also it should be self-explanatory how to do that: It's in the first tab, after all.

clanig:
Thanks for your comments. I have updated the code and the description according to your suggestions.

Due to possible inconsistencies, currently it makes the most sense to me not displaying any interfaces.

shundhammer:
The table still makes sense, even if the network interfaces are managed by Network Manager (which is the default for most home users).

teclator:
> The table still makes sense, even if the network interfaces are managed by **NetworkManager** (which is the default for most home users).

Well, we have to discuss about it because we are showing basically the ifcfg files in `/etc/sysconfig/network/` which could be a different configuration than the **NetworkManager** one...

mchf:
From my POV after first quick reading:
- i dislike(d) that switching to another tab too, but
- in some cases it might be useful to have some content here, so
- i don't see empty overview as a proper solution as I can imagine some bug reports about empty overview coming right after merging it. I'd prefer either (slightly easier solution) removing all irrelevant tabs (running NM? -> we cannot help you with Overview tab -> we will not make it available for you). We already use this concept of "on demand tabs" sometimes in y2-network (e.g. for s390) so we can extend it. Or even better is to fill everything with proper data - do not read config from wicked / netconfig when using NM but read it from NM directly and display it r/o to the user.

clanig:
I agree that the best option would be reading the correct values from Network Manager. I will take a look and check the feasibility.

clanig:
I have not forgotten this PR. I still intend to work on it soon but I am currently a bit busy.

jreidinger:
well, no activity for more then three years. So lets close it. Feel free to reopen when you update it.
Thanks

PR #1318, opened 2023-01-13
Copy only the selected backend configuration to the target system (bsc#1206723)

teclator:
## Problem

Until now, the **wicked** or **sysconfig network configuration** was always copying to the target system during a normal installation. 

Recent changes in the Basesystem filesystem packages removed the /etc/sysconfig/network hierarchy from it not being used by default anymore which raises an exception when YaST tries to merge the **dhcp** and **config** files from the instsys.

- https://trello.com/c/SEM7dN4G
- https://bugzilla.suse.com/show_bug.cgi?id=1206723

## Solution

Copy only the configuration of the selected backend to the target system.

## Testing

- *Added a new unit test*
- *Tested manually a MicroOS installation with*
  - https://build.opensuse.org/package/revisions/Base:System/filesystem (Revision 233)
  - https://build.opensuse.org/package/show/openSUSE:Factory/sysconfig (v0.0.90)
  - https://build.opensuse.org/package/show/openSUSE:Factory/wicked (v0.6.71)


src/lib/y2network/config_copier.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    # Convenvenience method for copying a list of files into the target system.
+    # It takes care of creating the target directory but only if some file
+    # needs to be copied
+    #
+    # @param files [Array<String>] list of short filenames to be copied
+    # @param path [String] path where the files resides and where will be
+    # copied in the target system
+    # @return [Boolean] whether some file was copied
+    def copy_files_to_target(files, path)
```
imobachgs, line 68: The arguments look confusing to me.

* `files`: it is a list of glob patterns (not exactly names). So you can do `ifcfg-*`, right?
* `path`: the name looks too generic to me (a *path* can be a file or a directory, but we expect it to be just a directory). 

Additionally, I would swap the arguments, so you can write:

```ruby
copy_files_to_target("/etc/NetworkManager/system-connections") # use '*' by default.
copy_files_to_target("/etc", ["hosts"])
```

Having said that, I wonder whether supporting *directories* and *files* would make sense:

```ruby
copy_to_target("/etc/hosts")
copy_to_target("/etc/NetworkManager/system-connections", include: ["*.nmconnections"])
```

src/lib/y2network/network_manager/config_copier.rb
```diff
@@ -0,0 +1,34 @@
...
+require "y2network/config_copier"
+
+module Y2Network
+  module NetworkManager
+    # This class copies Wicked specific configuration to the target system
+    class ConfigCopier < Y2Network::ConfigCopier
+      NETWORK_MANAGER = "/etc/NetworkManager".freeze
+
+      def copy
+        copy_files_to_target(["*"], File.join(NETWORK_MANAGER, "system-connections"))
```
imobachgs, line 30: np: At the end, each subclass implements its own `#copy` method. And that's fine. But what about having a method that returns the list of things to copy, and the parent class `#copy` takes care of that? In the current implementation, I feel we are using the inheritance just to share a helper method (`#copy_files_to_target`).

imobachgs (commented):
Nothing critical, just a few comments about the overall approach because I feel it is influenced by the old code (which is neither good or bad).

teclator:
> Nothing critical, just a few comments about the overall approach because I feel it is influenced by the old code (which is neither good or bad).

Yes, in fact it is an part of the code was almost moved as it is.

src/lib/y2network/network_manager/config_copier.rb
teclator: Yep, that was the main reason and using inheritance for that is unnecessary, so fine with the helper method.

imobachgs (approved):

imobachgs (approved):

mchf:
Well ... already merged, but as i was asked for an opinion ... I've tested that, read through and it LGTM

PR #1319, opened 2023-01-19
Do not setup DHCP in case of firmware configured devices

teclator:
## Problem

Similar to #1282 but **YaST** should also skip **DHCP** setup in case that there is some interface configured by firmware **(iBFT, NBFT or Redfish)**.

- https://trello.com/c/eFJXEeep/3274-3-urgent-top-priority-connect-to-nbft-disks
- https://jira.suse.com/browse/PED-967
- https://jira.suse.com/browse/PED-3138

## Solution

When there is some active interface configured by **firmware** **yast2-network** will also skip the **DHCP** setup completely as it was doing when there was some active interface with an ifcfg file.

In case that `wicked firmware` command is not available it will be logged by **Cheetah**, the same happens if the **ibft** extension is not available and for us it will means that there are no firmware configured devices at all.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Wed Jan 25 13:52:26 UTC 2023 - Knut Alejandro Anderssen González <kanderssen@suse.com>
 
 - During installation, do not configure DHCP if there is some
-  active interface configured by firmware (jsc#PED-3138).
+  active interface configured by firmware (jsc#PED-967).
 - 4.5.15
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Wed Jan 25 13:52:26 UTC 2023 - Knut Alejandro Anderssen González <kanderssen@suse.com>
+
+- During installation, do not configure DHCP if there is some
+  active interface configured by firmware (jsc#PED-3138).
+- 4.5.15
+
+-------------------------------------------------------------------
 Tue Jan 24 09:37:43 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
 
 - Fix the return of packages needed by the selected backend when
--- a/src/lib/network/wicked.rb
+++ b/src/lib/network/wicked.rb
@@ -92,7 +92,7 @@ module Yast
     # @return [Array <String>] array of interface names
     def firmware_interfaces
       output = Yast::Execute.stdout.locally!(WICKED_PATH, "firmware", "interfaces")
-      interfaces = output.gsub(/^(ibft|nbft|redfish)\s+/, "").split(/\s+/)
+      interfaces = output.gsub(/^\w+\s+/, "").split(/\s+/)
       (ibft_interfaces + interfaces).uniq
     end
   end
--- a/src/lib/network/wicked.rb
+++ b/src/lib/network/wicked.rb
@@ -84,7 +84,7 @@ module Yast
     #
     # @return [Array <String>] array of interface names
     def ibft_interfaces
-      Yast::Execute.stdout.locally!(IBFT_CMD, "-l").gsub("\n", " ").split(" ").uniq
+      Yast::Execute.stdout.locally!(IBFT_CMD, "-l").gsub("\n", " ").split(/\s+/).uniq
     end
 
     # Returns an array of interface names which are configured via firmware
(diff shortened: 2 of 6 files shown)


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Wed Jan 25 13:52:26 UTC 2023 - Knut Alejandro Anderssen González <kanderssen@suse.com>
+
+- During installation, do not configure DHCP if there is some
+  active interface configured by firmware (jsc#PED-3138).
```
ancorgs, line 5: Shouldn't we use the number of the Epic (PED-967)?

teclator: I was not sure, just checked linuxrc and wicked PRs... so modifying it ..

src/lib/network/wicked.rb
```diff
@@ -85,5 +86,14 @@ def query_wicked(iface, query)
     def ibft_interfaces
       Yast::Execute.stdout.locally!(IBFT_CMD, "-l").gsub("\n", " ").split(" ").uniq
     end
+
+    # Returns an array of interface names which are configured via firmware
+    #
+    # @return [Array <String>] array of interface names
+    def firmware_interfaces
+      output = Yast::Execute.stdout.locally!(WICKED_PATH, "firmware", "interfaces")
+      interfaces = output.gsub(/^(ibft|nbft|redfish)/, "").gsub("\n", " ").split(" ")
```
ancorgs, line 95: Wouldn't be better to split by `/\s/` just in case they start to use tabs?

test/wicked_test.rb
```diff
@@ -77,4 +77,29 @@ class DummyNetwork
       expect(subject.ibft_interfaces).to eql(["eth0.42", "eth0", "eth1"])
     end
   end
+
+  describe "#firmware_interfaces" do
+    let(:stdout) { instance_double("Yast::Execute") }
+    let(:output) { "ibft    eth1 eth1.10\nibft    eth2 ibft\nnbft    eth3\nredfish usb0 usb0.42" }
```
ancorgs, line 83: Although I believe the regexp is fully correct, I would like to see an example with "ibft" or "nbft" as part of the name of the device. Just as a reminder if someone changes the regexp.

```
ibft    eth1 eth1.10
nbft    eth3
nbft    nbft0 nbft0.40
```

src/lib/network/wicked.rb
ancorgs: Thinking it twice I see another problem (not serious but...). What if tomorrow they invent a new firmware technology that is not ibft, nbft or redfish?

Maybe we should rely more in the format (ignore the fist column and catch the rest) than in the content (check for the technology names).

src/lib/network/wicked.rb
```diff
@@ -83,7 +84,16 @@ def query_wicked(iface, query)
...
-      Yast::Execute.stdout.locally!(IBFT_CMD, "-l").gsub("\n", " ").split(" ").uniq
+      Yast::Execute.stdout.locally!(IBFT_CMD, "-l").gsub("\n", " ").split(/\s+/).uniq
+    end
+
+    # Returns an array of interface names which are configured via firmware
+    #
+    # @return [Array <String>] array of interface names
+    def firmware_interfaces
+      output = Yast::Execute.stdout.locally!(WICKED_PATH, "firmware", "interfaces")
+      interfaces = output.gsub(/^(ibft|nbft|redfish)\s+/, "").gsub("\n", " ").split(/\s+/)
```
ancorgs, line 95: I don't think `gsub("\n", " ")` is needed since `\s` includes `\n`.

teclator: In fact it is not :+1:

ancorgs: Did you see my previous command about parsing based on the format (columns) instead of strings (technology names). https://github.com/yast/yast-network/pull/1319/files#r1086825042

Maybe it deserves a comment in the code (FIXME, TODO, whatever).

teclator: So do you prefer something like

`
 interfaces = output.gsub(/^\w+\s+/, "").split(/\s+/)
`

teclator: BTW we probably would like to modify it in the future as it would be nice to now which interfaces are configured by which technology to show it in the overview summary.

ancorgs (approved):
LGTM

PR #1321, opened 2023-01-24
Fix packages needed by the selected backend in AutoYaST

teclator:
## Problem

The packages needed by a **backend** selected explicitly in the **AutoYaST** profiles are not installed during an **AutoInstallation**.

- https://bugzilla.suse.com/show_bug.cgi?id=1207221

## Solution

Check the autoinst config when evaluating the packages needed to ensure we do not use instsys backend.

## Testing

- *Added a new unit test*
- *Tested manually*
  - With **managed = true** it configures **NetworkManager**
  - With **managed = false** it configures **wicked**
  - With **backend = wicked** it configures **wicked**
  - With **backend = network_manager**  it configures **NetworkManager**
  - With no option given it configures **NetworkManager** as it is the default in the SLE-Micro5.4 control file.

### DIFF
--- a/test/lan_test.rb
+++ b/test/lan_test.rb
@@ -106,7 +106,7 @@ describe "LanClass" do
         Y2Network::ProposalSettings.create_instance
       end
 
-      it "checks the AutoYAST selected backend" do
+      it "checks the AutoYaST selected backend" do
         expect(Yast::Package).to receive(:Installed).with("NetworkManager", target: :autoinst)
         Yast::Lan.Packages
       end


package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Tue Jan 24 09:37:43 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix the return of packages neede by the selected backend when
```
imobachgs, line 4: typoe: needed

src/modules/Lan.rb
```diff
@@ -651,12 +651,12 @@ def ProposeVirtualized
...
-      backend = yast_config&.backend
-
-      return pkgs unless backend
+      backend = Mode.autoinst ? autoinst.selected_backend : yast_config&.backend
+      return [] unless backend
 
-      backend.packages.each { |p| pkgs << p if !Package.Installed(p) }
+      pkgs = []
+      target = Mode.autoinst ? :autoinst : :system
+      backend.packages.each { |p| pkgs << p if !Package.Installed(p, target: target) }
```
imobachgs, line 659: What about a more idiomatic way? And we could use a better name for the variable.

```ruby
# ok, the name sucks, but it is more descriptive
non_installed_pkgs = backend.packages.reject { |p| Package.Installed(p, target: target) }
```

imobachgs (approved):

PR #1325, opened 2023-03-01
Fix proposal settings test

teclator:
## Problem

The packages is not building for s390x and i586 architectures.

- https://bugzilla.suse.com/show_bug.cgi?id=1208796

## Solution

Fixed a unit test mocking which causes the tests to fail depending on how the tests are sorted.

## Testing

- *Fixed current new unit test for the proposal settings*


shundhammer (approved):
LGTM

PR #1327, opened 2023-05-02
Firmware configured interface (bsc#1209589)

teclator:
## Problem

When an interface is configured by firmware there is no information in the description allowing also to configure it through the UI which could be problematic therefore the buttons should be disabled in order to prevent the user to shoot himself in the foot.

- [*bsc=1209589*](https://bugzilla.suse.com/show_bug.cgi?id=1209589)
- [*PBI*](https://trello.com/c/2AmgYNN6)

## Solution

1. The description of the selected interface will show information about the extension used for configuring it by firmware. 
2. The edit button will be disabled in case of configured by firmware
3. As a plus task the delete button will be disabled in case of not configured or configured by firmware interface.

## Testing

- *Added a new unit test*
- *Tested manually*

## Screenshots

Screenshot from 2023-05-04 09-30-20

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -2,7 +2,7 @@
 Thu May  4 11:21:13 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
 
 - Display information about firmware configured interfaces and
-  prevent the forbid to edit them (bsc#1209589)
+  forbid editing them (bsc#1209589)
 - 4.5.18
 
 -------------------------------------------------------------------
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Thu May  4 11:21:13 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Display information about firmware configured interfaces and
+  prevent the forbid to edit them (bsc#1209589)
+- 4.5.18
+
+-------------------------------------------------------------------
 Wed Mar  1 14:52:46 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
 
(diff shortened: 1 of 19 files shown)


src/lib/network/wicked.rb
```diff
@@ -84,16 +84,37 @@ def query_wicked(iface, query)
...
     # @return [Array <String>] array of interface names
     def firmware_interfaces
-      output = Yast::Execute.stdout.locally!(WICKED_PATH, "firmware", "interfaces")
-      interfaces = output.gsub(/^\w+\s+/, "").split(/\s+/)
+      interfaces = firmware_interfaces_by_extension.values.flatten
       (ibft_interfaces + interfaces).uniq
     end
+
+    # Returns a hash with each firmware extension as the key and the specific extension
+    # configured interfaces as the value
```
imobachgs, line 99: np: An example might help to understand this.

imobachgs (approved):

package/yast2-network.changes
```diff
@@ -1,3 +1,10 @@
+-------------------------------------------------------------------
+Thu May  4 11:21:13 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Display information about firmware configured interfaces and
+  prevent the forbid to edit them (bsc#1209589)
```
imobachgs, line 5: The message looks confusing to me (the second part). What do you mean?

imobachgs (changes requested):

package/yast2-network.changes
teclator: even to me xxDDD

teclator:
```suggestion
  forbid to edit them (bsc#1209589)
```

imobachgs (approved):

PR #1329, opened 2023-05-08
Summary fix when no interfaces are present in the system (bsc#1211161)

teclator:
## Problem

When there are no interfaces at all the description of the selected one is still set which ends with a crash error introduced by #1327 

*Short description of the original problem.*

- [*bsc#1211161*](https://bugzilla.suse.com/show_bug.cgi?id=1211161)
- *[openQA link](https://openqa.opensuse.org/tests/3271807/modules/network_configuration/steps/2)*

## Solution

Do not set the description when there are no interfaces present on the system and only show the edit link when there is an interface.

## Testing

- *Added unit test*
- *Tested manually*

## Screenshots

| Before #1327  | with the provided fix |
| --- | --- |
| Screenshot from 2023-05-08 11-55-28 | No description interfaces table |

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,11 @@
 -------------------------------------------------------------------
+Mon May  8 09:07:07 UTC 2023 - Knut Anderssen <kanderssen@suse.com>
+
+- Fix summary crash when there is no interface available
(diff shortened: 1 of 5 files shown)


joseivanlopez (approved):
LGTM

PR #1331, opened 2023-05-10
Do not set EAP auth when EAP mode is TLS (bsc#1211026)

teclator:
## Problem

YaST writes the WIRELESS_EAP_AUTH attribute as "mschapv2" when the EAP mode is TLS which is wrong.

- *[bsc#1211026](https://bugzilla.suse.com/show_bug.cgi?id=1211026)*

## Solution

Do no set the WIRELESS_EAP_AUTH attribute when EAP mode is TLS.

## Testing

- *Added a new unit test*


dgdavid (approved):
How many acronyms! :exploding_head: 

LGTM, thanks!

PR #1336, opened 2023-06-14
Installation crashes when writing VLAN configuration into NetworkManager

mchf:
## Problem

Installation crashed when installing MicroOS (5.3) with configured vlan.

This bug is not MicroOS specific. First crucial prerequisite for the bug to happen was that target was configured to use NetworkManager as network service. The issue happened in test where user connected to installation source via vlan device. The configuration of the device itself is not important. What caused trouble was vlan id. Say you have device eth0.1006 where 1006 is vlan id. At the end of installation when network configuration is being stored installer crashed when writing NetworkManager's configuration via augeas to the target.

- [bsc#1211431](https://bugzilla.suse.com/show_bug.cgi?id=1211431)

## Solution

Problem was caused by storing vlan id in wrong format. YaST passed integer where String was expected

## Testing

- Tested manually
- Tested by the security team

### DIFF
--- a/src/lib/y2network/network_manager/connection_config_writers/vlan.rb
+++ b/src/lib/y2network/network_manager/connection_config_writers/vlan.rb
@@ -28,7 +28,7 @@ module Y2Network
         # @see Y2Network::ConnectionConfigWriters::Base#update_file
         # @param conn [Y2Network::ConnectionConfig::Vlan] Configuration to write
         def update_file(conn)
-          file.vlan["id"] = conn.vlan_id
+          file.vlan["id"] = conn.vlan_id.to_s
           file.vlan["parent"] = conn.parent_device
           file.vlan["type"] = "vlan"
         end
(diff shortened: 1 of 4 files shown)


src/lib/y2network/network_manager/connection_config_writers/vlan.rb
```diff
@@ -28,7 +28,7 @@ class Vlan < Base
         # @see Y2Network::ConnectionConfigWriters::Base#update_file
         # @param conn [Y2Network::ConnectionConfig::Vlan] Configuration to write
         def update_file(conn)
-          file.vlan["id"] = conn.vlan_id
+          file.vlan["id"] = conn.vlan_id.to_s
```
teclator, line 31: Would be nice to add a unit test for VLAN connection which currently is missing ;)

teclator, line 31: I'm wondering if we could prevent this in the future as we have some definition of the method and we already know it is an integer variable...

teclator: https://github.com/yast/yast-network/blob/master/src/lib/cfa/interface_file.rb#L313

teclator (approved):
LGTM

PR #1339, opened 2023-07-04
Fix typo when writing a wireless channel SLE-15-SP3

teclator:
## Problem

There is a typo which breaks the wireless network manager writer when an specific wireless channel is specified.

- https://bugzilla.suse.com/show_bug.cgi?id=1212976

## Solution

Fixed typo

## Testing

- *Added a new unit test*


imobachgs (approved):
Ouch 😳. LGTM.

PR #1346, opened 2023-11-16
Fix report of AutoYaST strict IP check

teclator:
## Problem

There is no way to disable the strict IP check in AutoYaST in SLE-12-SP5 and an error is always reported although by default it is just 1 second report.

- [bsc#1216859](https://bugzilla.suse.com/show_bug.cgi?id=1216859#c8)

## Solution

Do not show an error report if the timeout is less than 0.

### DIFF
--- a/src/clients/lan_auto.rb
+++ b/src/clients/lan_auto.rb
@@ -97,22 +97,14 @@ module Yast
         Builtins.y2error("Writing lan config failed") if !result
         @ret &&= result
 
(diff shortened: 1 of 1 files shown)


waynechen55:
Any plan to merge it ?  From my perspective, if it is safe to have this fix, then merging it is better.

shundhammer (approved):
LGTM

PR #1347, opened 2023-12-13
Fix drivers read

teclator:
## Problem

It was reported that yast2 lan does not show the proper kernel modules and that is mainly because yast2-network read only the modules of the first driver read by hwinfo.

- https://bugzilla.suse.com/show_bug.cgi?id=1217652

## Solution

Read all the modules from listed by hwinfo.

## Testing

- *Added a new unit test*
- *Tested manually*


shundhammer (approved):
LGTM

PR #1359, opened 2024-03-12
PoC: Guard secret attributes against leaking to the logs V1.0

shundhammer:
# Proof of Concept [_superseded_]

## Problem

Secret attributes of the wifi connection object might leak to the logs, e.g. if methods like `inspect()` are used.

## Fix

Use a custom `inspect()` method that clones the original object and sanitizes all fields that should not be logged verbatim (replacing each one with a special string `<sanitized>`).

## Related PR

**Superseded** by PR #1360 which uses `attr_secret` from `YaST2::SecretAttributes`.

### DIFF
--- a/src/lib/y2network/connection_config/wireless.rb
+++ b/src/lib/y2network/connection_config/wireless.rb
@@ -78,6 +78,7 @@ module Y2Network
       attr_accessor :client_key
       # @return [String] client private key password
       attr_accessor :client_key_password
+      alias_method :super_inspect, :inspect
 
       def initialize
         super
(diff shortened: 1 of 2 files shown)


joseivanlopez:
I think `SecretAttributes` mixin could be useful here, see https://github.com/yast/yast-yast2/blob/master/library/general/src/lib/yast2/secret_attributes.rb.

src/lib/y2network/connection_config/wireless.rb
```diff
@@ -113,6 +114,28 @@ def mode=(wireless_mode)
...
         !(keys || []).compact.all?(&:empty?)
       end
+
+      # Return a clone of this object with all sensitive fields (passwords etc.) obscured.
+      # Use this for log output to avoid leaking passwords.
+      def sanitized
+        san = dup
+        san.wpa_psk = sanitized_field(san.wpa_psk)
+        san.wpa_password = sanitized_field(san.wpa_password)
+        san.client_key_password = sanitized_field(san.client_key_password)
```
teclator, line 124: Consider adding also **_wep_keys_** to the list of attributes to be concealed.

shundhammer: yup, done in the new one #1360.

shundhammer:
Superseded by PR #1360 which uses `secret_attr` from `YaST::SecretAttributes`.

PR #1368, opened 2024-11-06
Assign the global default route to an specific connection when possible

teclator:
## Problem

- https://bugzilla.suse.com/show_bug.cgi?id=1232531

The default gateway is lost after installation if the gateway is configured manually and it’s not network interface specific. It’s the wicked to NM conversion that is not implemented in this case.

## Solution

When the default route does not belongs to any specific interface it will be associated to the connection which has an ip address belonging to the same network than the default route gateway.

## Testing

- *Added a new unit test*
- *Tested manually*

### DIFF
--- a/src/lib/y2network/network_manager/config_writer.rb
+++ b/src/lib/y2network/network_manager/config_writer.rb
@@ -91,14 +91,16 @@ module Y2Network
 
         explicit = routes.select { |r| r.interface&.name == conn.name }
 
-        return explicit if !conn.static_valid_ip?
+        return explicit unless conn.static_valid_ip?
+
+        conn_network = (IPAddr.new conn.ip.address.to_s).to_range
 
         # select the routes without an specific interface and which gateway belongs to the
         # same network
         global = routes.select do |r|
           next if r.interface || !r.default? || !r.gateway
 
-          (IPAddr.new conn.ip.address.to_s).to_range.include?(IPAddr.new(r.gateway.to_s))
+          conn_network.include?(IPAddr.new(r.gateway.to_s))
         end
 
         explicit + global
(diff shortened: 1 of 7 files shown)


src/lib/y2network/network_manager/config_writer.rb
```diff
@@ -68,7 +87,21 @@ def write_dns(config, old_config, _issues_list)
...
+        explicit = routes.select { |r| r.interface&.name == conn.name }
+
+        return explicit if !conn.static_valid_ip?
+
+        # select the routes without an specific interface and which gateway belongs to the
+        # same network
+        global = routes.select do |r|
+          next if r.interface || !r.default? || !r.gateway
+
+          (IPAddr.new conn.ip.address.to_s).to_range.include?(IPAddr.new(r.gateway.to_s))
```
imobachgs, line 101: You could move `IPAddr.new conn.ip.address.to_s` outside of the iteration (and calculated it once).

src/lib/y2network/network_manager/config_writer.rb
```diff
@@ -68,7 +87,21 @@ def write_dns(config, old_config, _issues_list)
       # @param routes [Array<Route>] List of routes to search in
       # @return [Array<Route>] List of routes for the given connection
       def routes_for(conn, routes)
-        routes.select { |r| r.interface&.name == conn.name }
+        return routes.reject(&:interface) if conn.nil?
+
+        explicit = routes.select { |r| r.interface&.name == conn.name }
+
+        return explicit if !conn.static_valid_ip?
```
imobachgs, line 94: Why not use `unless`?

imobachgs (commented):
Just two minor comments. Otherwise, it looks good.

src/lib/y2network/network_manager/config_writer.rb
teclator: no strong reason, so, just switch to unless

imobachgs (approved):

PR #1372, opened 2025-02-12
Added dialog for changing the bond ports naming schema to BusID (bsc#1233653) - SLE-15-SP6

teclator:
## Problem

When a user does some change with the module writing down the configuration, YaST calls  `udevadm trigger --subsystem-match=net --action=add` at some point. 

If there is some bond port using the MAC address as the renaming schema (udev rule) the interface could be wrongly renamed by the udev-generator as it uses the *current MAC* of the NIC instead of the permanent one (the backup of the bond uses the same MAC of the active one).

- https://bugzilla.suse.com/show_bug.cgi?id=1233653

## Solution

In order to prevent the user to shoot himself in the foot a dialog has been implemented to warn the user allowing him to change the renaming mechanism of all the affected bond ports just accepting it.

## Testing

- *Tested manually*

## Screenshots

image

### DIFF
--- a/src/lib/y2network/dialogs/bonding_fix.rb
+++ b/src/lib/y2network/dialogs/bonding_fix.rb
@@ -1,4 +1,4 @@
-# Copyright (c) [2021] SUSE LLC
+# Copyright (c) [2025] SUSE LLC
 #
 # All Rights Reserved.
 #
--- a/src/lib/y2network/dialogs/bonding_fix.rb
+++ b/src/lib/y2network/dialogs/bonding_fix.rb
@@ -141,7 +141,7 @@ module Y2Network
         invalid_bonds.each do |bond|
           bond.ports.each do |port|
             interface = iface_for(port)
-            interface.rename(interface.name, :bus_id)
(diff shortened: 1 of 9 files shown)


src/lib/y2network/dialogs/bonding_fix.rb
```diff
@@ -0,0 +1,179 @@
+# Copyright (c) [2021] SUSE LLC
```
imobachgs, line 1:
```suggestion
# Copyright (c) [2025] SUSE LLC
```

imobachgs (approved):

imobachgs (changes requested):
Oh, I forgot about the changes file. Please, add an entry.

imobachgs (approved):

teclator:
Introduced another change to prevent the user to modify a bond using a port with a mac address udev rule. If only the renaming schema need to be modified it will be reported so, in other case it will be part of the config adaptation needed by the port.

Screenshot From 2025-02-27 10-58-20
Screenshot From 2025-02-27 10-59-00
Screenshot From 2025-02-27 10-59-41

imobachgs (approved):

PR #1375, opened 2025-10-08
Fix new ruby

jreidinger:
## Problem

With security fixed ruby write nil to user resets also password.

- obs https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Staging:H/yast2-network/standard/x86_64
- bzilla https://bugzilla.suse.com/show_bug.cgi?id=1251273

## Solution

Reorder calling nil after password is read.


mchf (approved):
LGTM

PR #1376, opened 2026-03-10
Dropped requirement on update-desktop-files

mchf:
Removed all references to update-desktop-files from the package's spec file. Bumped version

### DIFF
--- a/package/yast2-network.changes
+++ b/package/yast2-network.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Tue Mar 10 09:00:28 UTC 2026 - Michal Filka <mfilka@suse.com>
+
+- jsc#PED-14507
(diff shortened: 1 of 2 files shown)


mchf:
Script used for the modification:
https://paste.opensuse.org/pastes/2887651d437b

lslezak (approved):
LGTM.

But first please adapt the yast2-devtools to not call the update of the desktop files during RPM build: https://github.com/yast/yast-devtools/blob/master/build-tools/rpm/macros.yast#L146-L148. That would cause all packages to fail.

Merge this PR after the updated yast2-devtools package is accepted in Factory.

(Actually the [%yast_desktop_files](https://github.com/yast/yast-devtools/blob/master/build-tools/rpm/macros.yast#L113-L126) macro can be removed completely, it is not used directly by any YaST package.)

